PR #22, opened 2026-01-22
sigrok - logic analyzer driver

mangelajo:
## Release Notes

* **New Features**
  * Added Sigrok driver for digital and analog signal capture and analysis.
  * Supports multiple output formats: VCD, CSV, BITS, ASCII, BINARY, and SRZIP.
  * Device scanning and channel mapping capabilities.
  * Real-time protocol decoder support with configurable decoders and output formats.
  * Streaming capture functionality for continuous data acquisition.

* **Documentation**
  * Added comprehensive Sigrok driver documentation with configuration and usage examples.

* **Infrastructure**
  * Updated CI workflows and container environment to include Sigrok CLI tooling.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/client.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/client.py
@@ -1,0 +1,37 @@
+from dataclasses import dataclass
+
+from .common import CaptureConfig, CaptureResult
+from jumpstarter.client import DriverClient
+
+
+@dataclass(kw_only=True)
+class SigrokClient(DriverClient):
+    """Client methods for the Sigrok driver."""
+
+    def scan(self) -> str:
+        return self.call("scan")
+
+    def capture(self, config: CaptureConfig | dict) -> CaptureResult:
+        return CaptureResult.model_validate(self.call("capture", config))
+
+    def capture_stream(self, config: CaptureConfig | dict):
+        """Stream capture data from sigrok-cli.
+
+        Args:
+            config: CaptureConfig or dict with capture parameters
+
+        Yields:
+            bytes: Chunks of captured data
+        """
+        for chunk in self.streamingcall("capture_stream", config):
+            yield chunk
+
+    def get_driver_info(self) -> dict:
+        return self.call("get_driver_info")
+
+    def get_channel_map(self) -> dict:
+        return self.call("get_channel_map")
+
+    def list_output_formats(self) -> list[str]:
+        return self.call("list_output_formats")
+
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from typing import Any
+from typing import Any, Iterator
 
 from pydantic import BaseModel, Field
 
@@ -69,12 +69,12 @@ class CaptureResult(BaseModel):
         from base64 import b64decode
         return b64decode(self.data_b64)
 
-    def decode(self) -> list[Sample] | dict[str, list[int]] | str:
+    def decode(self) -> Iterator[Sample] | dict[str, list[int]] | str:
         """Parse captured data based on output format.
 
         Returns:
-            - CSV format: list[Sample] with timing and all values per sample
-            - VCD format: list[Sample] with timing and only changed values
+            - CSV format: Iterator[Sample] yielding samples with timing and all values per sample
+            - VCD format: Iterator[Sample] yielding samples with timing and only changed values
             - Bits format: dict[str, list[int]] with channel→bit sequences
             - ASCII format: str with ASCII art visualization
             - Other formats: raises NotImplementedError (use .data for raw bytes)
@@ -85,11 +85,11 @@ class CaptureResult(BaseModel):
         if self.output_format == OutputFormat.CSV:
             from .csv import parse_csv
             samples_data = parse_csv(self.data, self.sample_rate)
-            return [Sample.model_validate(s) for s in samples_data]
+            return (Sample.model_validate(s) for s in samples_data)
         elif self.output_format == OutputFormat.VCD:
             from .vcd import parse_vcd
             samples_data = parse_vcd(self.data, self.sample_rate)
-            return [Sample.model_validate(s) for s in samples_data]
+            return (Sample.model_validate(s) for s in samples_data)
         elif self.output_format == OutputFormat.BITS:
             return self._parse_bits()
         elif self.output_format == OutputFormat.ASCII:
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -43,8 +43,9 @@ class CaptureConfig(BaseModel):
     triggers: dict[str, str] | None = Field(default=None, description="e.g., {'D0': 'rising'}")
     channels: list[str] | None = Field(default=None, description="override default channels by name")
     output_format: str = Field(
-        default="srzip",
-        description="csv, srzip, vcd, binary, bits, ascii",
+        default=OutputFormat.VCD,
+        description="Output format (default: vcd - efficient change-based format with timing). "
+        "Options: vcd, csv, srzip, binary, bits, ascii",
     )
     decoders: list[DecoderConfig] | None = Field(default=None, description="real-time protocol decoding")
 
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -1,0 +1,49 @@
+from __future__ import annotations
+
+from typing import Any
+
+from pydantic import BaseModel, Field
+
+
+class DecoderConfig(BaseModel):
+    """Protocol decoder configuration (real-time during capture)."""
+
+    name: str
+    channels: dict[str, str] | None = None
+    options: dict[str, Any] | None = None
+    annotations: list[str] | None = None
+    stack: list["DecoderConfig"] | None = None
+
+
+class CaptureConfig(BaseModel):
+    sample_rate: str = Field(default="1M", description="e.g., 8MHz, 1M, 24000000")
+    samples: int | None = Field(default=None, description="number of samples; None for continuous")
+    pretrigger: int | None = Field(default=None, description="samples before trigger")
+    triggers: dict[str, str] | None = Field(default=None, description="e.g., {'D0': 'rising'}")
+    channels: list[str] | None = Field(default=None, description="override default channels by name")
+    output_format: str = Field(
+        default="srzip",
+        description="csv, srzip, vcd, binary, bits, ascii",
+    )
+    decoders: list[DecoderConfig] | None = Field(default=None, description="real-time protocol decoding")
+
+
+class CaptureResult(BaseModel):
+    """Result from a capture operation.
+
+    Note: data is base64-encoded for reliable JSON transport. Client methods
+    automatically decode it to bytes for you.
+    """
+    data_b64: str  # Base64-encoded binary data
+    output_format: str
+    sample_rate: str
+    channel_map: dict[str, str]
+    triggers: dict[str, str] | None = None
+    decoders: list[DecoderConfig] | None = None
+
+    @property
+    def data(self) -> bytes:
+        """Get the captured data as bytes (auto-decodes from base64)."""
+        from base64 import b64decode
+        return b64decode(self.data_b64)
+
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -105,6 +105,23 @@ class CaptureResult(BaseModel):
     triggers: dict[str, str] | None = None
     decoders: list[DecoderConfig] | None = None
 
+    def __str__(self) -> str:
+        """Format CaptureResult with truncated data_b64 field."""
+        data_len = len(self.data_b64)
+        if data_len <= 50:
+            data_preview = self.data_b64
+        else:
+            # Show first 50 and last 50 chars with ellipsis
+            data_preview = f"{self.data_b64[:25]}...{self.data_b64[-25:]} ({data_len} chars)"
+
+        return (
+            f"CaptureResult(output_format='{self.output_format}', "
+            f"sample_rate='{self.sample_rate}', "
+            f"data_size={len(self.data)} bytes, "
+            f"channels={len(self.channel_map)}, "
+            f"data_b64='{data_preview}')"
+        )
+
     @property
     def data(self) -> bytes:
         """Get the captured data as bytes (auto-decodes from base64)."""
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -79,6 +79,11 @@ class CaptureResult(BaseModel):
             - ASCII format: str with ASCII art visualization
             - Other formats: raises NotImplementedError (use .data for raw bytes)
 
+        Note:
+            Channel names in the output depend on how the data was captured:
+            - If captured with channel mapping, sigrok-cli outputs mapped names (vcc, cs, etc.)
+            - If captured without mapping, outputs device names (D0, D1, etc.)
+
         Raises:
             NotImplementedError: For binary/srzip formats (use .data property)
         """
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -100,27 +100,35 @@ class CaptureResult(BaseModel):
             )
 
     def _parse_bits(self) -> dict[str, list[int]]:
-        """Parse bits format to dict of channel→bit sequences."""
+        """Parse bits format to dict of channel→bit sequences.
+
+        Sigrok-cli bits format: "D0:10001\\nD1:01110\\n..."
+        Each line has format "channel_name:bits"
+
+        Note: For large sample counts, sigrok-cli wraps bits across multiple
+        lines with repeated channel names. We accumulate all occurrences.
+        """
         text = self.data.decode("utf-8")
         lines = [line.strip() for line in text.strip().split("\n") if line.strip()]
 
-        # bits format is just columns of 0/1
-        # TODO: Need to determine channel mapping from somewhere
-        # For now, return as generic numbered channels
         result: dict[str, list[int]] = {}
 
         for line in lines:
-            # Each line might be space/comma separated bits
-            bits = [int(b) for b in line if b in "01"]
-            if not result:
-                # Initialize channels
-                for i, bit in enumerate(bits):
-                    result[f"CH{i}"] = [bit]
-            else:
-                # Append to existing channels
-                for i, bit in enumerate(bits):
-                    if f"CH{i}" in result:
-                        result[f"CH{i}"].append(bit)
+            # Bits format: "D0:10001" or "A0:10001"
+            if ":" in line:
+                channel_device_name, bits_str = line.split(":", 1)
+                channel_device_name = channel_device_name.strip()
+
+                # Map device name (D0) to user-friendly name (vcc) if available
+                channel_name = self.channel_map.get(channel_device_name, channel_device_name)
+
+                # Parse bits from this line
+                bits = [int(b) for b in bits_str if b in "01"]
+
+                # Accumulate bits for this channel (may appear on multiple lines)
+                if channel_name not in result:
+                    result[channel_name] = []
+                result[channel_name].extend(bits)
 
         return result
 
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -22,8 +22,50 @@ class OutputFormat:
 class Sample(BaseModel):
     """A single sample with timing information."""
     sample: int  # Sample index
-    time_ns: int  # Time in nanoseconds
+    time: float  # Time in seconds (full precision)
     values: dict[str, int | float]  # Channel values (digital: 0/1, analog: voltage)
+
+    def __str__(self) -> str:
+        """Format sample with clean time display using appropriate unit (fs/ps/ns/μs/ms/s)."""
+        time_str = self._format_time(self.time)
+        return f"Sample(sample={self.sample}, time={time_str}, values={self.values})"
+
+    @staticmethod
+    def _format_time(time_s: float) -> str:
+        """Format time in seconds to the most appropriate unit.
+
+        Args:
+            time_s: Time in seconds
+
+        Returns:
+            Formatted string like "1.5ns", "2.3μs", "1.5ms", "2s"
+        """
+        # Special case for zero
+        if time_s == 0:
+            return "0s"
+
+        abs_time = abs(time_s)
+
+        # Define units in descending order (seconds to femtoseconds)
+        units = [
+            (1.0, "s"),
+            (1e-3, "ms"),
+            (1e-6, "μs"),
+            (1e-9, "ns"),
+            (1e-12, "ps"),
+            (1e-15, "fs"),
+        ]
+
+        # Find the most appropriate unit
+        for scale, unit in units:
+            if abs_time >= scale or scale == 1e-15:  # Use fs as minimum
+                value = time_s / scale
+                # Format with up to 6 significant digits, remove trailing zeros
+                formatted = f"{value:.6g}"
+                return f"{formatted}{unit}"
+
+        # Fallback (should never reach here)
+        return f"{time_s:.6g}s"
 
 
 class DecoderConfig(BaseModel):
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -5,6 +5,27 @@ from typing import Any
 from pydantic import BaseModel, Field
 
 
+class OutputFormat:
+    """Constants for sigrok output formats."""
+    CSV = "csv"
+    BITS = "bits"
+    ASCII = "ascii"
+    BINARY = "binary"
+    SRZIP = "srzip"
+    VCD = "vcd"
+
+    @classmethod
+    def all(cls) -> list[str]:
+        return [cls.CSV, cls.BITS, cls.ASCII, cls.BINARY, cls.SRZIP, cls.VCD]
+
+
+class Sample(BaseModel):
+    """A single sample with timing information."""
+    sample: int  # Sample index
+    time_ns: int  # Time in nanoseconds
+    values: dict[str, int | float]  # Channel values (digital: 0/1, analog: voltage)
+
+
 class DecoderConfig(BaseModel):
     """Protocol decoder configuration (real-time during capture)."""
 
@@ -46,4 +67,60 @@ class CaptureResult(BaseModel):
         """Get the captured data as bytes (auto-decodes from base64)."""
         from base64 import b64decode
         return b64decode(self.data_b64)
+
+    def decode(self) -> list[Sample] | dict[str, list[int]] | str:
+        """Parse captured data based on output format.
+
+        Returns:
+            - CSV format: list[Sample] with timing and all values per sample
+            - VCD format: list[Sample] with timing and only changed values
+            - Bits format: dict[str, list[int]] with channel→bit sequences
+            - ASCII format: str with ASCII art visualization
+            - Other formats: raises NotImplementedError (use .data for raw bytes)
+
+        Raises:
+            NotImplementedError: For binary/srzip formats (use .data property)
+        """
+        if self.output_format == OutputFormat.CSV:
+            from .csv import parse_csv
+            samples_data = parse_csv(self.data, self.sample_rate)
+            return [Sample.model_validate(s) for s in samples_data]
+        elif self.output_format == OutputFormat.VCD:
+            from .vcd import parse_vcd
+            samples_data = parse_vcd(self.data, self.sample_rate)
+            return [Sample.model_validate(s) for s in samples_data]
+        elif self.output_format == OutputFormat.BITS:
+            return self._parse_bits()
+        elif self.output_format == OutputFormat.ASCII:
+            return self.data.decode("utf-8")
+        else:
+            raise NotImplementedError(
+                f"Parsing not implemented for {self.output_format} format. "
+                f"Use .data property to get raw bytes."
+            )
+
+    def _parse_bits(self) -> dict[str, list[int]]:
+        """Parse bits format to dict of channel→bit sequences."""
+        text = self.data.decode("utf-8")
+        lines = [line.strip() for line in text.strip().split("\n") if line.strip()]
+
+        # bits format is just columns of 0/1
+        # TODO: Need to determine channel mapping from somewhere
+        # For now, return as generic numbered channels
+        result: dict[str, list[int]] = {}
+
+        for line in lines:
+            # Each line might be space/comma separated bits
+            bits = [int(b) for b in line if b in "01"]
+            if not result:
+                # Initialize channels
+                for i, bit in enumerate(bits):
+                    result[f"CH{i}"] = [bit]
+            else:
+                # Append to existing channels
+                for i, bit in enumerate(bits):
+                    if f"CH{i}" in result:
+                        result[f"CH{i}"].append(bit)
+
+        return result
 
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import re
 from enum import Enum
+
 from pydantic import BaseModel, Field, field_validator
 
 
@@ -79,7 +80,7 @@ class DecoderConfig(BaseModel):
 
     @field_validator("options")
     @classmethod
-    def validate_options(cls, v: dict[str, str | int | float | bool] | None) -> dict[str, str | int | float | bool] | None:
+    def validate_options(cls, v: dict[str, str | int | float | bool] | None) -> dict[str, str | int | float | bool] | None:  # noqa: E501
         if v is None:
             return v
         for key, value in v.items():
--- a/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
+++ b/python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
@@ -1,11 +1,11 @@
 from __future__ import annotations
 
-from typing import Any, Iterator
+import re
+from enum import Enum
+from pydantic import BaseModel, Field, field_validator
 
-from pydantic import BaseModel, Field
 
-
-class OutputFormat:
+class OutputFormat(str, Enum):
     """Constants for sigrok output formats."""
     CSV = "csv"
     BITS = "bits"
@@ -16,7 +16,7 @@ class OutputFormat:
 
     @classmethod
     def all(cls) -> list[str]:
-        return [cls.CSV, cls.BITS, cls.ASCII, cls.BINARY, cls.SRZIP, cls.VCD]
+        return [member.value for member in cls]
 
 
 class Sample(BaseModel):
@@ -26,7 +26,7 @@ class Sample(BaseModel):
     values: dict[str, int | float]  # Channel values (digital: 0/1, analog: voltage)
 
     def __str__(self) -> str:
-        """Format sample with clean time display using appropriate unit (fs/ps/ns/μs/ms/s)."""
+        """Format sample with clean time display using appropriate unit (fs/ps/ns/us/ms/s)."""
         time_str = self._format_time(self.time)
         return f"Sample(sample={self.sample}, time={time_str}, values={self.values})"
 
@@ -38,7 +38,7 @@ class Sample(BaseModel):
             time_s: Time in seconds
 
         Returns:
-            Formatted string like "1.5ns", "2.3μs", "1.5ms", "2s"
+            Formatted string like "1.5ns", "2.3us", "1.5ms", "2s"
         """
         # Special case for zero
         if time_s == 0:
@@ -50,7 +50,7 @@ class Sample(BaseModel):
         units = [
             (1.0, "s"),
             (1e-3, "ms"),
-            (1e-6, "μs"),
+            (1e-6, "us"),
             (1e-9, "ns"),
             (1e-12, "ps"),
             (1e-15, "fs"),
@@ -73,9 +73,24 @@ class DecoderConfig(BaseModel):
 
     name: str
     channels: dict[str, str] | None = None
-    options: dict[str, Any] | None = None
+    options: dict[str, str | int | float | bool] | None = None
     annotations: list[str] | None = None
     stack: list["DecoderConfig"] | None = None
+
+    @field_validator("options")
+    @classmethod
+    def validate_options(cls, v: dict[str, str | int | float | bool] | None) -> dict[str, str | int | float | bool] | None:
+        if v is None:
+            return v
+        for key, value in v.items():
+            str_key = str(key)
+            str_value = str(value)
+            if ":" in str_key or ":" in str_value:
+                raise ValueError(
+                    f"Decoder option key/value must not contain ':' (got key={str_key!r}, value={str_value!r}). "
+                    "Colons are used as sigrok-cli delimiters and could cause option injection."
+                )
+        return v
 
 
 class CaptureConfig(BaseModel):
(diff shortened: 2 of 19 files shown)


mangelajo:
@ambient-code please rebase

mangelajo:
@kirkbrauer @raballew this is long, but independent of almost everything else. It is extensively tested with the sigrok cli dummy driver. I have also tested it with hardware.

It would be a cool addition to 0.9.0

I am happy if the docs and the proposed api makes sense to you.

python/packages/jumpstarter-driver-sigrok/pyproject.toml
```diff
@@ -0,0 +1,42 @@
...
+
+[tool.hatch.build.hooks.pin_jumpstarter]
+name = "pin_jumpstarter"
+
+[dependency-groups]
+dev = [
+    "pytest-cov>=6.0.0",
+    "pytest>=8.3.3",
+    "pytest-asyncio>=0.24.0",
+]
```
raballew, line 45: [HIGH] Missing driver entry points registration.

The `pyproject.toml` has no `[project.entry-points."jumpstarter.drivers"]` section. Every other driver package in this repo registers its driver class via this entry point, and without it the Jumpstarter framework won't be able to discover the `Sigrok` driver at runtime.

Suggested fix: add an entry-points section like:
```toml
[project.entry-points."jumpstarter.drivers"]
Sigrok = "jumpstarter_driver_sigrok.driver:Sigrok"
```

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+
+        if not self.channels:
+            return []
+
+        # Build channel map: device_name=user_name
+        channel_map = ",".join(f"{dev}={user}" for dev, user in self.channels.items())
+        return ["-C", channel_map] if channel_map else []
+
+    def _config_args(self, cfg: CaptureConfig, *, continuous: bool = False) -> list[str]:
+        parts = [f"samplerate={cfg.sample_rate}"]
```
raballew, line 186: [HIGH] Potential command option injection via unsanitized `sample_rate`.

`CaptureConfig.sample_rate` is a free-form `str` that gets interpolated directly into sigrok-cli's `-c` argument without validation. A malicious client could send something like `sample_rate="1M,capture_file=/etc/shadow"` and inject arbitrary sigrok-cli configuration options.

Suggested fix: validate `sample_rate` against a strict regex (e.g., `r"^\d+(\.\d+)?\s*(k|M|G)?(Hz)?$"`) via a Pydantic `field_validator`, or parse it into a numeric Hz value server-side.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+    def _build_capture_command(self, cfg: CaptureConfig) -> tuple[list[str], Path, TemporaryDirectory]:
+        tmpdir = TemporaryDirectory()
+        outfile = Path(tmpdir.name) / f"capture.{cfg.output_format}"
+
+        cmd: list[str] = self._base_driver_args()
+        cmd += self._channel_args(cfg.channels)
+        cmd += self._config_args(cfg)
+        cmd += self._trigger_args(cfg)
+        cmd += self._decoder_args(cfg)
+        cmd += ["-O", cfg.output_format, "-o", str(outfile)]
```
raballew: [HIGH] No validation on `output_format` enables path traversal and option injection.

`CaptureConfig.output_format` is a free-form `str` used without validation. Path traversal like `output_format="x/../../../etc/cron.d/payload"` would resolve outside the temp directory. Format-specific options like `"csv:header=false"` can also be injected. `OutputFormat` is a plain class with string constants rather than a `StrEnum`, so Pydantic does not enforce valid values at construction.

Suggested fix: convert `OutputFormat` to `StrEnum` (Python 3.11+) and change `CaptureConfig.output_format` type to `OutputFormat`, or add a Pydantic `field_validator` that checks `v in OutputFormat.all()`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd.py
```diff
@@ -0,0 +1,266 @@
...
+def _parse_timescale(line: str) -> float:
+    """Parse timescale line and return multiplier to convert to seconds."""
+    parts = line.split()
+    if len(parts) >= 3:
+        value = parts[1]
+        unit = parts[2]
+        # Convert to seconds multiplier
+        unit_multipliers = {"s": 1.0, "ms": 1e-3, "us": 1e-6, "ns": 1e-9, "ps": 1e-12}
+        return float(value) * unit_multipliers.get(unit, 1.0)
+    return 1.0
```
raballew: [HIGH] `_parse_timescale` fallback to `1.0` seconds silently corrupts timestamps.

When `_parse_timescale` cannot parse the timescale line (fewer than 3 tokens), it returns `1.0` (seconds), overwriting the caller's default of `1e-9` (nanoseconds). This means timestamps would be off by ~1e9x. Also, the `unit_multipliers` dict omits `"fs"` (femtoseconds), which is a valid VCD timescale unit, causing another silent fallback to `1.0`.

Suggested fix: add `"fs": 1e-15` to `unit_multipliers`. Change the fallback return to `1e-9` or raise `ValueError` on an unparseable timescale.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+            return {
+                "data_b64": b64encode(data).decode("ascii"),
+                "output_format": cfg.output_format,
+                "sample_rate": cfg.sample_rate,
+                "channel_map": self.channels,
+                "triggers": cfg.triggers,
+                "decoders": [d.model_dump() for d in cfg.decoders] if cfg.decoders else None,
+            }
+        finally:
+            tmpdir.cleanup()
```
raballew: [MEDIUM] `TemporaryDirectory` leaked when `_build_capture_command` raises.

The `TemporaryDirectory` is created inside `_build_capture_command` (line 134) but cleaned up by the caller's `finally` block (line 98). If `_channel_args`, `_trigger_args`, or `_decoder_args` raises `ValueError`, the tempdir is created but never returned, so `tmpdir.cleanup()` in the finally block is never reached.

Suggested fix: create the `TemporaryDirectory` in `capture()` before calling `_build_capture_command` and pass the path in, or wrap the `_build_capture_command` body in try/except that calls `tmpdir.cleanup()` on failure.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv_test.py
```diff
@@ -0,0 +1,133 @@
...
+@pytest.fixture
+def demo_client(demo_driver_instance):
+    """Create a client for the demo Sigrok driver."""
+    with serve(demo_driver_instance) as client:
+        yield client
+
+
+@pytest.mark.skipif(which("sigrok-cli") is None, reason="sigrok-cli not installed")
+def test_csv_format_basic(demo_client: SigrokClient):
+    """Test CSV format capture with demo driver."""
```
raballew, line 133: [MEDIUM] CSV parser has zero unit tests; all tests require sigrok-cli.

All 4 tests in `csv_test.py` are gated by `@pytest.mark.skipif(which("sigrok-cli") is None, ...)`. The `parse_csv()` function and its helpers are pure Python with no external dependencies, yet have no unit tests. Meanwhile `vcd_test.py` correctly provides pure unit tests with constructed test data.

Suggested fix: add unit tests constructing CSV byte data inline and calling `parse_csv()` directly. Cover: digital-only, analog-only, mixed channels, sample rate parsing variants, and empty input.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,199 @@
...
+                # Parse bits from this line
+                bits = [int(b) for b in bits_str if b in "01"]
+
+                # Accumulate bits for this channel (may appear on multiple lines)
+                if channel_name not in result:
+                    result[channel_name] = []
+                result[channel_name].extend(bits)
+
+        return result
+
```
raballew, line 225: [MEDIUM] No test file for `common.py`; multiple public methods are untested.

No `common_test.py` exists. Untested public methods include: `Sample._format_time()` (6 unit branches), `CaptureResult.__str__()` (truncation logic), `CaptureResult.data` (base64 decoding), `CaptureResult._parse_bits()` (bits format parsing), and `OutputFormat.all()`.

Suggested fix: create `common_test.py` with unit tests for each method. For `_format_time`, test each unit boundary. For `_parse_bits`, construct bits-format byte data and verify channel mapping.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd_test.py
```diff
@@ -0,0 +1,217 @@
...
+            data_b64=b64encode(vcd_content.encode("utf-8")).decode("ascii"),
+            output_format=OutputFormat.VCD,
+            sample_rate="1MHz",
+            channel_map={},
+        )
+
+        samples = list(result.decode())
+        assert len(samples) >= 1
+        # First sample at time 0
+        assert samples[0].time == 0.0
```
raballew, line 136: [MEDIUM] `test_vcd_parser_timescale_variations` has tautological assertions.

The test defines 5 test cases with `expected_time_ns` values but never asserts the second sample's time against them. The only assertions are `len(samples) >= 1` and `samples[0].time == 0.0`, both trivially true for every case. The `_multiplier` and `expected_time_ns` parameters are computed but unused.

Suggested fix: assert `samples[1].time` against the expected time in seconds for each case using `pytest.approx` for float comparison.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+    @export
+    def capture(self, config: CaptureConfig | dict) -> dict:
+        """One-shot capture; returns dict with base64-encoded binary data."""
+        self._ensure_executable()
+        cfg = CaptureConfig.model_validate(config)
+        cmd, outfile, tmpdir = self._build_capture_command(cfg)
+
+        try:
+            self.logger.debug("Running sigrok-cli: %s", " ".join(cmd))
+            subprocess.run(cmd, check=True)
```
raballew: [MEDIUM] No subprocess timeout on `capture()` and `scan()`.

Both `subprocess.run` calls have no `timeout` parameter. If sigrok-cli hangs (e.g., waiting for a trigger that never fires), these calls block indefinitely with no recovery path.

Suggested fix: add a `timeout` parameter to both `subprocess.run` calls. Consider making timeout configurable with sensible defaults (e.g., 300s for capture, 30s for scan).

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd.py
```diff
@@ -0,0 +1,266 @@
...
+    # Parse value changes if present on the same line
+    if len(parts) > 1:
+        values_str = parts[1]
+        _parse_vcd_value_changes(values_str, channel_map, current_values)
+
+    # Return sample data if we have values
+    if current_values:
+        return {"time": current_time_s, "values": current_values}
+
+    return None
```
raballew: [MEDIUM] Dead code: `_parse_vcd_timestamp_line` is never called.

`_parse_vcd_timestamp_line` (33 lines) is defined but never referenced anywhere. The refactored `_parse_vcd_body` uses `_parse_timestamp` and `_parse_vcd_value_changes` directly.

Suggested fix: delete `_parse_vcd_timestamp_line` (lines 117-149).

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd.py
```diff
@@ -0,0 +1,266 @@
...
+    i = 0
+    while i < len(values_str):
+        char = values_str[i]
+
+        # Single bit change (e.g., "1!", "0abc" for multi-char identifiers)
+        if char in "01xzXZ":
+            symbol, new_i = _extract_symbol(values_str, i + 1)
+            if symbol in channel_map:
+                channel = channel_map[symbol]
+                current_values[channel] = 1 if char == "1" else 0
```
raballew, line 143: [MEDIUM] VCD parser silently maps x/z states to integer 0.

Single-bit x/z values are mapped to 0 via `1 if char == "1" else 0`. Binary vectors containing x/z fail `int(binary_value, 2)` and the except handler also returns 0. A signal transitioning from 1 to x (unknown/high-impedance) is silently reported as transitioning to 0, which could be misleading.

Suggested fix: use `None` for x/z states and update `Sample.values` type to `dict[str, int | float | None]`, or at least log a warning.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,199 @@
...
+            NotImplementedError: For binary/srzip formats (use .data property)
+        """
+        if self.output_format == OutputFormat.CSV:
+            from .csv import parse_csv
+            samples_data = parse_csv(self.data, self.sample_rate)
+            return (Sample.model_validate(s) for s in samples_data)
+        elif self.output_format == OutputFormat.VCD:
+            from .vcd import parse_vcd
+            samples_data = parse_vcd(self.data, self.sample_rate)
+            return (Sample.model_validate(s) for s in samples_data)
```
raballew: For CSV and VCD, decode() returns a generator expression rather than a list. This means len(result.decode()) raises TypeError, and iterating the result a second time silently yields nothing.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
```diff
@@ -0,0 +1,495 @@
...
+
+from .common import CaptureConfig, CaptureResult, OutputFormat
+from .driver import Sigrok
+from jumpstarter.common.utils import serve
+
+# Skip all integration tests if sigrok-cli is not available
+pytestmark = pytest.mark.skipif(
+    which("sigrok-cli") is None,
+    reason="sigrok-cli not found in PATH"
+)
```
raballew: [MEDIUM] Module-level `pytestmark` skips metadata tests that don't need sigrok-cli.

The module-level `pytestmark` skips ALL tests in this file, including `test_get_driver_info`, `test_get_channel_map`, and `test_list_output_formats`, which only call metadata methods and never invoke sigrok-cli. These tests should run in CI even without sigrok-cli installed.

Suggested fix: remove the module-level `pytestmark`. Apply the skip marker only to individual tests that actually invoke sigrok-cli (the ones that call `scan()` or `capture()`).

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,199 @@
...
+        if time_s == 0:
+            return "0s"
+
+        abs_time = abs(time_s)
+
+        # Define units in descending order (seconds to femtoseconds)
+        units = [
+            (1.0, "s"),
+            (1e-3, "ms"),
+            (1e-6, "μs"),
```
raballew: [MEDIUM] Non-ASCII characters violate project rules.

Lines 29, 42, and 53 use the Greek mu character (U+03BC) for microseconds where the textual alternative `"us"` exists (and is already used in `vcd.py`). Lines 137, 170 use right arrow (U+2192) where `->` is the textual alternative. `driver_test.py:338` also uses right arrow. Per project rules: "prefer textual representation instead of icon."

Suggested fix: replace mu with `"us"` and right arrow with `"->"`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+        for decoder in self._flatten_decoders(cfg.decoders):
+            pin_map = self._resolve_decoder_channels(decoder)
+            segments = [decoder.name]
+
+            for pin_name, channel_name in pin_map.items():
+                segments.append(f"{pin_name}={self._resolve_channel(channel_name)}")
+
+            if decoder.options:
+                for key, value in decoder.options.items():
+                    segments.append(f"{key}={value}")
```
raballew, line 229: [MEDIUM] Decoder option injection via `Any`-typed values.

`DecoderConfig.options` is typed as `dict[str, Any]` and values are interpolated into sigrok-cli `-P` arguments without sanitization. Values containing `:` could inject additional protocol decoder segments.

Suggested fix: restrict the type to `dict[str, str | int | float | bool]` and validate that neither keys nor values contain the `:` separator.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+        """One-shot capture; returns dict with base64-encoded binary data."""
+        self._ensure_executable()
+        cfg = CaptureConfig.model_validate(config)
+        cmd, outfile, tmpdir = self._build_capture_command(cfg)
+
+        try:
+            self.logger.debug("Running sigrok-cli: %s", " ".join(cmd))
+            subprocess.run(cmd, check=True)
+
+            data = outfile.read_bytes()
```
raballew, line 89: [LOW] Unbounded memory consumption from capture file read.

`outfile.read_bytes()` reads the entire capture file into memory with no size limit, then base64-encodes it (roughly doubling memory use). A very large capture could exhaust available memory.

Suggested fix: check file size before reading and enforce a maximum (e.g., 100 MB).

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv.py
```diff
@@ -0,0 +1,141 @@
...
+    Args:
+        channel_names: List of channel names
+        row: List of value strings
+
+    Returns:
+        Dict mapping channel name to parsed value
+    """
+    values = {}
+
+    for channel, value in zip(channel_names, row, strict=True):
```
raballew, line 128: [LOW] `_parse_csv_row` crashes on column count mismatch.

`zip(channel_names, row, strict=True)` raises `ValueError` if the column count doesn't match the channel count, which can happen with truncated captures.

Suggested fix: remove `strict=True` or use `itertools.zip_longest` to handle mismatches gracefully.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,199 @@
...
+    decoders: list[DecoderConfig] | None = None
+
+    def __str__(self) -> str:
+        """Format CaptureResult with truncated data_b64 field."""
+        data_len = len(self.data_b64)
+        if data_len <= 50:
+            data_preview = self.data_b64
+        else:
+            # Show first 50 and last 50 chars with ellipsis
+            data_preview = f"{self.data_b64[:25]}...{self.data_b64[-25:]} ({data_len} chars)"
```
raballew: [LOW] Misleading comment says "50 chars" but code uses 25.

The comment says "Show first 50 and last 50 chars with ellipsis" but the code actually slices `[:25]` and `[-25:]`.

Suggested fix: remove the comment entirely; the f-string is self-explanatory.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+        if hasattr(super(), "__post_init__"):
+            super().__post_init__()
+
+    def _ensure_executable(self):
+        """Ensure sigrok-cli is available."""
+        if self.executable is None:
+            raise FileNotFoundError(
+                "sigrok-cli executable not found in PATH. "
+                "Please install sigrok-cli to use this driver."
+            )
```
raballew, line 44: [LOW] No test for `_ensure_executable` error path.

`_ensure_executable()` raises `FileNotFoundError` when `self.executable is None`, but no test verifies this.

Suggested fix: add a test like `Sigrok(driver="demo", executable=None)` then call `scan()` and assert `pytest.raises(FileNotFoundError)`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv_test.py
```diff
@@ -0,0 +1,133 @@
...
+    # Decode the CSV data
+    samples = list(result.decode())
+    assert isinstance(samples, list)
+    assert len(samples) > 0
+
+    # Verify timing progresses correctly
+    for sample in samples:
+        assert isinstance(sample.time, float)
+        # Verify timing progresses (1/100kHz = 0.00001s per sample)
+        assert sample.time == sample.sample * 0.00001
```
raballew: [LOW] Float equality without tolerance in assertion.

`assert sample.time == sample.sample * 0.00001` uses exact float equality, which can fail due to floating-point precision issues.

Suggested fix: change to `assert sample.time == pytest.approx(sample.sample * 0.00001, rel=1e-6, abs=1e-12)` or use a mathematically correct representation.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv_test.py
```diff
@@ -0,0 +1,133 @@
...
+            "D4": "clk",
+            "D5": "sda",
+            "D6": "scl",
+            "D7": "gnd",
+        },
+    )
+
+
+@pytest.fixture
+def demo_client(demo_driver_instance):
```
raballew: [LOW] Duplicate test fixtures across test modules.

`demo_driver_instance` and `demo_client` fixtures are duplicated nearly identically in `csv_test.py` and `driver_test.py`. Other driver packages in this repo use `conftest.py` for shared fixtures.

Suggested fix: create a `conftest.py` with shared fixtures and remove the duplicates.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,199 @@
...
+            (1e-9, "ns"),
+            (1e-12, "ps"),
+            (1e-15, "fs"),
+        ]
+
+        # Find the most appropriate unit
+        for scale, unit in units:
+            if abs_time >= scale or scale == 1e-15:  # Use fs as minimum
+                value = time_s / scale
+                # Format with up to 6 significant digits, remove trailing zeros
```
raballew, line 64: [LOW] Unnecessary comments throughout driver code.

Multiple comments just restate what the code already says: inline field comments on the `Sample` model, step-by-step narration in `_format_time`, section separator banners in `driver.py`, and the magic number comment for default sample count. Per project rules, comments are acceptable only as a last resort when the code cannot be made self-explanatory.

Suggested fix: remove the narration comments. Use `Field(description=...)` for model fields if metadata is needed. Extract magic numbers to named constants. Remove section separator banners.

raballew (changes requested):

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
mangelajo: I disagree, comments are important to humans :) as long as those match the code.

mangelajo:
@ambient-code please handle comments.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv_test.py
mangelajo: I agree the CSV generated by sigrok is quite under control, is not just any CSV, we know how sigrok generates it.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
mangelajo: I agree, we can try to alleviate this if we see it happening.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv_test.py
kirkbrauer: @ambient-code please implement this and the missing `common.py` test cases to ensure we have full coverage for this feature which incorporates rather complex implementation details on the driver side that must be tested.

kirkbrauer (approved):
I think this looks OK as a first attempt, I asked the @ambient-code to add some additional tests, but I think this is good to merge assuming it has been tested.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
raballew: Disagree.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
raballew: Failure modes must be tested.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
raballew: Would it be possible to leave capture running overnight?

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd.py
raballew: Curious - Is this tested? Why did you choose logging it over None?

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd_test.py
raballew: Couldn't it also use a >0 comparison instead or is this all because of floating point precision?

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv_test.py
raballew: But this does not mean that Sigrok never changes. I think we should test this behaviour.

raballew (changes requested):
Better, but I need clarification on a few things.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd.py
```diff
@@ -0,0 +1,238 @@
...
+        value_end += 1
+
+    # Extract symbol
+    symbol, next_pos = _extract_symbol(values_str, value_end)
+
+    if symbol in channel_map:
+        try:
+            return int(binary_value, 2), symbol, next_pos
+        except ValueError:
+            return 0, symbol, next_pos
```
raballew, line 203: [HIGH] The binary value parser maps the entire multi-bit value to `0` when any single bit is `x` or `z`. For example, `b10x10` hits `ValueError` on `int(binary_value, 2)` and the except handler returns `0` for the whole value. A 32-bit bus value with one unknown bit silently becomes zero with no warning.

Suggested fix: replace `x`/`z` characters with `'0'` before calling `int(binary_value, 2)`, and log a warning about the substitution.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,225 @@
...
+                return f"{formatted}{unit}"
+
+        # Fallback (should never reach here)
+        return f"{time_s:.6g}s"
+
+
+class DecoderConfig(BaseModel):
+    """Protocol decoder configuration (real-time during capture)."""
+
+    name: str
```
raballew, line 75: [HIGH] `DecoderConfig.name` has no validation. While `options` rejects colons (lines 81-94), the name is passed directly to the `-P` flag (`segments = [decoder.name]` at driver.py:222) and the `-A` flag (driver.py:234). A name like `"uart:rx=D0"` would inject into the colon-delimited argument.

Suggested fix: add a `field_validator` for `name` with a pattern like `^[a-zA-Z0-9_-]+$` to reject colons, spaces, and special characters.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/client.py
```diff
@@ -0,0 +1,37 @@
...
+class SigrokClient(DriverClient):
+    """Client methods for the Sigrok driver."""
+
+    def scan(self) -> str:
+        return self.call("scan")
+
+    def capture(self, config: CaptureConfig | dict) -> CaptureResult:
+        return CaptureResult.model_validate(self.call("capture", config))
+
+    def capture_stream(self, config: CaptureConfig | dict):
```
raballew, line 17: [HIGH] `get_driver_info` and `get_channel_map` use bare `dict` return types without type parameters, and `capture_stream` has no return type annotation at all.

Suggested fix: use `dict[str, str]` for `get_channel_map`, parameterize the dict for `get_driver_info`, and annotate `capture_stream` with `-> Generator[bytes, None, None]`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
```diff
@@ -0,0 +1,294 @@
...
+    def scan(self) -> str:
+        """List devices for the configured driver."""
+        self._ensure_executable()
+        assert self.executable is not None
+        cmd = [self.executable, "--driver", self.driver, "--scan"]
+        result = subprocess.run(cmd, capture_output=True, text=True, check=True, timeout=self.timeout)
+        return result.stdout
+
+    @export
+    def get_driver_info(self) -> dict:
```
raballew, line 62: Use a `TypedDict` or dedicated result type for `capture()`. Add `-> AsyncGenerator[bytes, None]` for `capture_stream`. Parameterize the dict return for `get_driver_info`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv.py
```diff
@@ -0,0 +1,141 @@
...
+        try:
+            if "." in value or "e" in value.lower():
+                values[channel] = float(value)
+            else:
+                values[channel] = int(value)
+        except ValueError:
+            # Keep as string if not a number
+            values[channel] = value
+
+    return values
```
raballew, line 140: Change the annotation to `-> dict[str, int | float | str]`, or remove the string fallback by raising or logging instead.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
```diff
@@ -0,0 +1,469 @@
...
+    assert len(samples) > 0
+
+    # Check first sample for analog values
+    first_sample = samples[0]
+    assert isinstance(first_sample, Sample)
+    assert len(first_sample.values) > 0
+
+    # Analog values should be floats (voltages)
+    for _channel, value in first_sample.values.items():
+        assert isinstance(value, (int, float))
```
raballew, line 469: Many driver tests skip when `sigrok-cli` is not installed. Especially the command-building helpers, channel resolution, and decoder logic (`_flatten_decoders`, `_resolve_decoder_channels`) have zero dedicated unit tests and no mocked alternatives.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
```diff
@@ -0,0 +1,469 @@
...
+
+    # Verify we still get a proper CaptureResult model
+    assert isinstance(result, CaptureResult)
+    assert result.data
+    assert isinstance(result.data, bytes)
+    assert len(result.data) > 0
+    assert result.output_format == "srzip"
+
+
+@pytest.mark.skip(reason="sigrok-cli demo driver doesn't support streaming to stdout (-o -)")
```
raballew, line 173: `test_capture_stream_with_demo` is unconditionally skipped with `@pytest.mark.skip` and no mocked version exists.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv.py
```diff
@@ -0,0 +1,141 @@
...
+    # Parse the CSV data
+    reader = csv.reader(data_lines)
+
+    # First row is types (logic, V DC, etc.) - use for channel name inference
+    types_row = next(reader)
+
+    # Get channel names from types
+    channel_names = _infer_channel_names(types_row)
+
+    # Parse and yield data rows one by one
```
raballew, line 42: I dont like those comments.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
```diff
@@ -0,0 +1,469 @@
...
+
+
+def test_ensure_executable_raises_when_none():
+    """Test that _ensure_executable raises FileNotFoundError when executable is None."""
+    driver = Sigrok(driver="demo", executable=None)
+    with pytest.raises(FileNotFoundError, match="sigrok-cli executable not found"):
+        driver.scan()
+
+
+@pytest.mark.skipif(which("sigrok-cli") is None, reason="sigrok-cli not installed")
```
raballew, line 17: The driver passes `timeout=self.timeout` to `subprocess.run()`, but no test verifies that timeout enforcement actually works. There is no test that patches `subprocess.run` to raise `TimeoutExpired`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,225 @@
...
+            return v
+        for key, value in v.items():
+            str_key = str(key)
+            str_value = str(value)
+            if ":" in str_key or ":" in str_value:
+                raise ValueError(
+                    f"Decoder option key/value must not contain ':' (got key={str_key!r}, value={str_value!r}). "
+                    "Colons are used as sigrok-cli delimiters and could cause option injection."
+                )
+        return v
```
raballew, line 94: `validate_options` (rejects colons for injection prevention) and `validate_sample_rate` (enforces format) have zero test coverage.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,225 @@
...
+
+        # Fallback (should never reach here)
+        return f"{time_s:.6g}s"
+
+
+class DecoderConfig(BaseModel):
+    """Protocol decoder configuration (real-time during capture)."""
+
+    name: str
+    channels: dict[str, str] | None = None
```
raballew, line 76: A channel key containing `:` would inject into the `-P` argument. Add a `field_validator` for `channels` keys rejecting colons and other delimiter characters.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
```diff
@@ -0,0 +1,225 @@
...
+        default=OutputFormat.VCD,
+        description="Output format (default: vcd - efficient change-based format with timing). "
+        "Options: vcd, csv, srzip, binary, bits, ascii",
+    )
+    decoders: list[DecoderConfig] | None = Field(default=None, description="real-time protocol decoding")
+
+    @field_validator("sample_rate")
+    @classmethod
+    def validate_sample_rate(cls, v: str) -> str:
+        if not re.match(r"^\d+(\.\d+)?\s*(k|M|G)?(Hz)?$", v):
```
raballew, line 113: remove `\s*` from the regex pattern.

raballew (dismissed):

kirkbrauer:
@mangelajo This PR has been pending for quite a while, I think we should decide if we want to merge it or close it for later.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver.py
mangelajo: Pushing back on this one. `capture()` returns a dict that is immediately deserialized into `CaptureResult` (a Pydantic model already defined in `common.py`) on the client side. `CaptureResult` already fully types these fields. Adding a TypedDict would be redundant since the dict doesn't survive the protobuf wire serialization (`struct_pb2.Value`) anyway — types are erased at the wire boundary.

For `capture_stream`, adding an `AsyncGenerator[bytes, None]` annotation is fine and will be included.
For `get_driver_info`, will parameterize the return type to `dict[str, Any]`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
mangelajo: Pushing back. sigrok-cli's internal config parser accepts spaces between the number and unit prefix (e.g., `1 MHz` is valid as a `-c samplerate=1 MHz` argument value — it's a single string, no shell splitting). Removing `\s*` would reject valid input. No tests use spaces so this is unlikely to matter in practice, but tightening it is technically incorrect.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv.py
mangelajo: Pushing back. These comments explain sigrok-specific CSV format quirks (analog preview lines like `A0: -10.0000 V DC`, type header rows instead of column names, channel naming conventions). These are non-obvious domain details about sigrok's non-standard CSV output that aren't self-documenting from the code alone. A developer unfamiliar with sigrok's CSV format would genuinely need this context. This aligns with the earlier discussion where mangelajo noted 'comments are important to humans as long as those match the code.'

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv.py
mangelajo: Pushing back on changing the behavior. The `str` fallback is intentional defensive coding for protocol decoder annotation rows in sigrok CSV output, which contain text strings that are genuinely not numeric. Raising would crash the parser on valid sigrok output.

However, the type annotation is wrong — it says `dict[str, int | float]` but the function can return `dict[str, int | float | str]`. Will fix the annotation to match the actual behavior.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
mangelajo: The skip reason is valid — sigrok-cli's demo driver genuinely doesn't support streaming to stdout (`-o -`). This is a known sigrok-cli limitation, not a test issue.

A mocked unit test for the streaming plumbing could be added alongside this, but it would test different things (Python asyncio chunk reading, not sigrok-cli integration). Will add a mocked test for the streaming path separately, keeping this integration test skipped.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
mangelajo: Fixed in 335726fe. Added `validate_name` field_validator with pattern `^[a-zA-Z][a-zA-Z0-9_-]*$` — rejects colons, spaces, and special characters. All valid sigrok decoder names (uart, spi, i2c, sdcard_spi, onewire-link, etc.) match this pattern. Tests added in `TestValidateDecoderName`.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/vcd.py
mangelajo: Fixed in 335726fe. The binary value parser now:
1. Substitutes x/z bits with 0 (preserving other bits) instead of zeroing the entire value
2. Logs a warning with the channel name and original binary value
3. Falls back to 0 only if the substituted value still can't be parsed

For example, `b10x10` now produces `0b10010` (18) with a warning, instead of silently producing 0.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/client.py
mangelajo: Fixed in 335726fe. Updated type annotations:
- `get_driver_info() -> dict[str, Any]`
- `get_channel_map() -> dict[str, str]`
- `capture_stream() -> Generator[bytes, None, None]`

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/csv.py
mangelajo: Fixed the annotation in 335726fe. Changed `_parse_csv_row` return type to `dict[str, int | float | str]` to match actual behavior. The str fallback is intentional (protocol decoder annotation rows), as explained in the earlier pushback comment.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
mangelajo: Fixed in 335726fe. Added 24 unit tests for all command-building helpers that run without sigrok-cli:
- `TestBaseDriverArgs` (3 tests)
- `TestChannelArgs` (4 tests)
- `TestConfigArgs` (4 tests)
- `TestTriggerArgs` (2 tests)
- `TestDecoderArgs` (4 tests)
- `TestFlattenDecoders` (2 tests)
- `TestResolveChannel` (5 tests)
- `TestBuildCaptureCommand` (1 test)
- `TestBuildStreamCommand` (1 test)

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/driver_test.py
mangelajo: Fixed in 335726fe. Added `TestTimeoutEnforcement` with 2 tests that patch `subprocess.run` to raise `subprocess.TimeoutExpired` and verify both `scan()` and `capture()` propagate the exception.

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
mangelajo: Fixed in 335726fe. Added test coverage:
- `TestValidateSampleRate` (11 tests): valid formats (number, kHz, MHz, GHz, decimal, with space), invalid inputs (injection, letters-only, empty, semicolon)
- `TestValidateOptions` (6 tests): valid options, None, colon in key/value rejected, bool/float values

python/packages/jumpstarter-driver-sigrok/jumpstarter_driver_sigrok/common.py
mangelajo: Fixed in 335726fe. Added `validate_channels` field_validator that rejects colons in both keys and values. Tests added in `TestValidateDecoderChannels` (4 tests).

PR #137, opened 2026-01-23
operator: fix authentication spec api

mangelajo:
Authentication was duplicated in two places:
* spec.controller.authentication
* spec.authentication

The original intent of the design was using just spec.authentication for the overall auth
config, but it got duplicated and implemented on spec.controller

This fixes the problem before we release this in 0.8.0

## Summary by CodeRabbit

* **Breaking Changes**
  * Authentication configuration has been relocated from controller-level to top-level specification. Update existing Jumpstarter configurations to reference the new top-level authentication path instead of controller-scoped settings.

* **Bug Fixes**
  * Improved script compatibility for non-Linux environments.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>


mangelajo:
the pytest matrix does not detect the skips, and requires to run it seems, meeh https://github.com/orgs/community/discussions/44490 we may see ourselves running them all...

mangelajo:
@evakhoni can you have an eye on this one?  thanks thanks! :D

evakhoni (approved):
@mangelajo LGTM
the only thing i can think of, is if anybody already using the operator with the old spec, it will probably be just discarded silently by the operator, and defaults will be applied, without any warning for the user. but I guess at this point there are no one besides our devs so probably not an issue.

mangelajo:
> @mangelajo LGTM the only thing i can think of, is if anybody already using the operator with the old spec, it will probably be just discarded silently by the operator, and defaults will be applied, without any warning for the user. but I guess at this point there are no one besides our devs so probably not an issue.

yep, I think I am the only user of that so far %)

and I have one instance configured like that, so I am ready to switch it when necessary :D 
but yes :)

PR #213, opened 2026-02-11
bump fls to 0.2.0

bennyz:
## Summary by CodeRabbit

* **Chores**
  * Updated build-time dependency version.


mangelajo (approved):

mangelajo:
We should probably force container builds with the dockerfile is modified. it happens in e2e and it seems to be fine, but the process is slightly different

mangelajo:
> We should probably force container builds with the dockerfile is modified. it happens in e2e and it seems to be fine, but the process is slightly different

ah wait!, we have the build-pr-images :D it now works

bennyz:
> > We should probably force container builds with the dockerfile is modified. it happens in e2e and it seems to be fine, but the process is slightly different
> 
> ah wait!, we have the build-pr-images :D it now works

i should have used it to test my OCI PRs 😅

PR #266, opened 2026-02-25
SSHWrapperClient.stream_async should be a asynccontextmanager

NickCao:
And the stream method is provided by DriverClient base class

## Summary by CodeRabbit

* **Refactor**
  * Improved asynchronous stream handling in the SSH driver component through enhanced context management, resulting in better resource lifecycle control and increased operational robustness.


mangelajo (approved):
oops thanks!

Issue #57, opened 2025-07-08
j: better exception handling


mangelajo:
Capture most common exceptions and present them on a nice way to the user.

bkhizgiy:
implemented in: https://github.com/jumpstarter-dev/jumpstarter/pull/339 and https://github.com/jumpstarter-dev/jumpstarter/pull/328

PR #328, opened 2026-03-17
Improve j commands exception handling

bkhizgiy:
Updated exception handling for J commands to show cleaner, actionable errors for common failures: timeouts, TLS/certificate issues, DNS/connection problems, file and permission errors, bad JSON, and user cancellations. It also now maps key gRPC errors (DEADLINE_EXCEEDED, UNAVAILABLE, UNAUTHENTICATED/PERMISSION_DENIED, INVALID_ARGUMENT) to friendlier messages.

Also made sure the re-auth wrapper uses the same fallback mapping and improved jmp login fetch handling so invalid JSON/content-type responses are reported clearly.

Related to: https://github.com/jumpstarter-dev/jumpstarter/issues/57

## Summary by CodeRabbit

* **New Features**
  * Improved error messaging: timeouts, TLS cert issues, DNS, file/JSON errors, gRPC auth/argument problems and interrupts now map to clear, actionable CLI messages.
  * Stricter login validation: tighter checks on login endpoints, auth-config payloads, argument parsing, and conflicting insecure-flag handling.

* **Refactor**
  * Centralized exception translation for consistent error handling across sync/async flows.

* **Tests**
  * Added comprehensive tests for exception mappings, reauthentication flows, gRPC cases, and login validation.

### DIFF
--- a/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
+++ b/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
@@ -224,7 +224,9 @@ def handle_exceptions_with_reauthentication(login_func):
                 _handle_exception_group_with_reauth(eg, login_func)
             except (ConnectionError, JumpstarterException, click.ClickException) as e:
                 _handle_single_exception_with_reauth(e, login_func)
-            except Exception:
+            except Exception as e:
+                if common_exc := _map_common_exception(e):
+                    raise common_exc from None
                 raise
 
         return wrapped
--- a/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
+++ b/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/exceptions.py
@@ -1,4 +1,5 @@
 import asyncio
+import json
 import socket
 import ssl
 import types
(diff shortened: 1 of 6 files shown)


bkhizgiy:
There are many cases that could be covered, and I’m not sure how many of them are actually needed vs. overkill, so feel free to suggest adding anything missing or removing anything unnecessary.

mangelajo:
Nice improvement!

mangelajo (approved):

raballew (changes requested):
The timeout type tuple in `_map_runtime_exception` lists `TimeoutError`, `asyncio.TimeoutError`, `socket.timeout`, and `FutureTimeoutError` -- but since Python 3.11 (your minimum) these are all the same object. Just `TimeoutError` is enough.

`handle_exceptions_with_reauthentication` is missing a `KeyboardInterrupt` handler. The other two decorators both catch it and show "Cancelled by user." cleanly, but this one only has `except Exception` which won't catch `KeyboardInterrupt` (it inherits from `BaseException`), so the user gets a raw traceback instead.

The login tests (`test_login_cli_shows_timeout_message`, `test_login_cli_shows_certificate_message`) monkeypatch `fetch_auth_config` to raise `click.ClickException` directly, so they never exercise the exception mapping logic in `fetch_auth_config`. If that mapping code were removed, the tests would still pass.

bkhizgiy:
@raballew thanks for the feedback, I"ll post it as a follow-up PR.

mangelajo:
oops, thank you @raballew and @bkhizgiy

PR #333, opened 2026-03-17
feat: consolidate TLS flags into single --insecure option

### LINKED ISSUE #318: Inconsistent TLS/insecure flag naming in jmp login
`jmp login` has three separate flags for disabling TLS/security, each with a different naming pattern:

| Flag | Purpose |
|------|---------|
| `--insecure-tls-config` | Disable endpoint TLS verification |
| `--insecure-login-tls` | Skip TLS for login endpoint |
| `--insecure-login-http` | Use HTTP instead of HTTPS for login |

The naming is inconsistent (`insecure-tls-config` vs `insecure-login-tls` vs `insecure-login-http`) and the distinction between these flags is not immediately clear.

raballew:
## Summary

- Rename `--insecure-tls-config` / `--insecure-tls` to `--insecure` across all CLI commands
- Remove `--insecure-login-tls` and `--insecure-login-http` from `jmp login` -- `--insecure` now covers the entire connection chain
- If `--insecure` is passed and no URL scheme is provided for login, HTTP is used; otherwise HTTPS with TLS verification disabled
- Consistent flag naming across `jmp login`, `jmp-admin create client/exporter`, `jmp-admin import client/exporter`

Closes https://github.com/jumpstarter-dev/jumpstarter/issues/318

## Test plan

- [x] `--insecure` flag accepted and defaults to false
- [x] `jmp login --help` shows only `--insecure`, no `--insecure-tls-*` or `--insecure-login-*` flags
- [x] `jmp-admin create client --help` shows `--insecure`
- [x] `jmp-admin create exporter --help` shows `--insecure`
- [x] `jmp-admin import client --help` shows `--insecure`
- [x] `jmp-admin import exporter --help` shows `--insecure`
- [x] Insecure confirmation prompt updated
- [x] All 23 existing tests pass

## Summary by CodeRabbit

* **Documentation**
  * Updated docs and examples to use the unified --insecure flag for authentication, login, and distributed setup; examples reflect non-interactive login usage.

* **Refactor**
  * Consolidated TLS/HTTP toggles into a single --insecure option across CLI flows and commands; login flow simplified to use it.

* **New Features**
  * Added compact CLI output modes ("name" and "path").

* **Tests**
  * Updated and added tests covering the --insecure flag and non-interactive login.

* **Chores**
  * Added local tooling entries to .gitignore.


mangelajo:
Insecure could mean a few things: I would suggest the behavior is instead:
Rename --insecure-tls-config / --insecure-tls to --insecure-tls across all CLI commands
Remove --insecure-login-tls and --insecure-login-http from jmp login -- --insecure-tls , what is the purpose of insecure-login-http? if the user passes http it must know what he's doing, but we should probably print a warning, and ask the user unless it's a non interactive session

I think it's better to be explicit about what is insecure, tls in this case. But I would like to hear @bennyz @kirkbrauer @bkhizgiy

kirkbrauer:
@mangelajo I do agree that some of these flags are unnecessarily long, especially if they are required to be used often. However, having the arguments be longer is good to discourage use, but that's a bit premature here since it's still required as part of some of our setup instructions.

If we are able to eliminate the need for many of these flags, then combining them might be fine as the user is just accepting risk whenever they use the flag such as `-k` in cURL. I think your proposal makes sense to unify under `--insecure-tls` (maybe we can add `-k` as well?). However, I agree that since we're not primarily a HTTP client, it is true that `--insecure` is ambiguous since we also have an auth mechanism.

mangelajo:
I agree with @kirkbrauer 

I would propose:
 --insecure-tls + -k covering them all

Then

If --insecure-tls is passed and no URL scheme is provided for login, HTTPS + insecure TLS is used; otherwise http:// schema needs to be explicitly specified.

WDYT?

mangelajo (changes requested):
see main thread comments

python/docs/source/getting-started/guides/setup-distributed-mode.md
```diff
@@ -7,7 +7,7 @@ controller service, configuring drivers, and running the exporter.
 The jumpstarter-controller endpoints are secured by TLS. However, in release 0.7.x,
 the certificates are self-signed and rotated on every restart. This means the client
 will not be able to verify the server certificate. To bypass this, you should use the
-`--insecure-tls-config` flag when creating clients and exporters. This issue will be
+`--insecure-tls` flag when creating clients and exporters. This issue will be
 resolved in the next release. See [issue #72](https://github.com/jumpstarter-dev/jumpstarter/issues/72)
```
raballew: @mangelajo I think we should remove the issue reference here.

python/packages/jumpstarter-cli/jumpstarter_cli/login_test.py
```diff
@@ -53,7 +54,7 @@ def test_validate_login_endpoint_url_rejects_unsupported_scheme() -> None:
 
 
 def test_validate_login_endpoint_url_rejects_http_without_explicit_opt_in() -> None:
-    with pytest.raises(click.ClickException, match="Use --insecure-login-http"):
+    with pytest.raises(click.ClickException, match="Use --insecure-tls"):
```
mangelajo, line 57: this is a bit weird and not 100% exact, but I think the overall experience is much better.

python/packages/jumpstarter-driver-flashers/README.md
```diff
@@ -147,7 +147,7 @@ Options:
   --force-exporter-http          Force use of exporter HTTP
   --force-flash-bundle TEXT      Force use of a specific flasher OCI bundle
   --cacert FILE                  CA certificate to use for HTTPS
-  --insecure-tls                 Skip TLS certificate verification
+  --insecure                     Disable TLS verification and allow insecure connections
```
mangelajo: The flashers code isn't changed and is --insecure-tls

https://github.com/jumpstarter-dev/jumpstarter/blob/main/python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py#L1486

mangelajo (changes requested):
just one nit with the flashers README.md change

python/packages/jumpstarter-driver-flashers/README.md
raballew: created https://github.com/jumpstarter-dev/jumpstarter/issues/428 to track other drivers

mangelajo:
```suggestion
  --insecure-tls -k              Disable TLS verification and allow insecure connections
```

mangelajo: I mean this :D

mangelajo (dismissed):
see the suggestion :D

mangelajo (approved):

mangelajo:
@raballew the simplified login fails now for some reason:
https://github.com/jumpstarter-dev/jumpstarter/actions/runs/24136369402/job/70425509887?pr=333#step:6:4008

all subsequent failures seem to be related to the failed login.

raballew (commented):
The e2e failure is because `LOGIN_ENDPOINT` is a plain HTTP server (`login.${BASEDOMAIN}:8086`), but with the flag consolidation, `--insecure-tls` only disables certificate verification — it still defaults to `https://` when no scheme is given.

The fix is to explicitly specify the `http://` scheme in the test:

```diff
-  run jmp login test-client-oidc@${LOGIN_ENDPOINT} --insecure-tls --nointeractive \
+  run jmp login test-client-oidc@http://${LOGIN_ENDPOINT} --insecure-tls --nointeractive \
```

PR #343, opened 2026-03-19
feat: show expires at and remaining columns in lease listing

### LINKED ISSUE #32: Show lease expiration time by default
Currently `jmp get leases` shows the lease start time and duration by default, and I don't think there's a way to change that. It would be more useful to show the lease expiration time (by default) because in most cases I want to know until when (or how much longer) I can use the system, without having to do the math myself. Reproduced on 0.7.5.dev13+gde656df25.

raballew:
## Summary
- Replace `BEGIN TIME` and `DURATION` columns with `EXPIRES AT` and `REMAINING` in `jmp get leases` output
- `EXPIRES AT` shows the calculated expiration timestamp
- `REMAINING` shows human-readable time left (e.g. `2h 15m`) or `expired`

Fixes #32

## Test plan
- [x] Unit tests: column headers include EXPIRES AT and REMAINING (`test_rich_add_columns_has_expires_at_and_remaining`)
- [x] Unit tests: expiration computed from effective_end_time, effective_begin_time+duration, begin_time+duration (`test_compute_expires_at_*`)
- [x] Unit tests: remaining time formatting including expired state (`test_format_remaining_*`)
- [x] Unit tests: table rendering with and without timing data (`test_rich_add_rows_*`)
- [x] E2E BATS test: verify `jmp get leases` output contains EXPIRES AT and REMAINING columns

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/client/grpc.py
+++ b/python/packages/jumpstarter/jumpstarter/client/grpc.py
@@ -218,10 +218,11 @@ class Lease(BaseModel):
         return None
 
     @staticmethod
-    def _format_remaining(expires_at):
+    def _format_remaining(expires_at, *, now=None):
         if expires_at is None:
             return ""
-        now = datetime.now(tz=expires_at.tzinfo)
+        if now is None:
+            now = datetime.now(tz=expires_at.tzinfo)
         remaining = expires_at - now
         if remaining.total_seconds() <= 0:
             return "expired"
--- a/python/packages/jumpstarter/jumpstarter/client/grpc.py
+++ b/python/packages/jumpstarter/jumpstarter/client/grpc.py
@@ -225,10 +225,9 @@ class Lease(BaseModel):
         remaining = expires_at - now
         if remaining.total_seconds() <= 0:
             return "expired"
-        total_seconds = int(remaining.total_seconds())
-        days, remainder = divmod(total_seconds, 86400)
-        hours, remainder = divmod(remainder, 3600)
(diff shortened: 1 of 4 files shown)


python/packages/jumpstarter/jumpstarter/client/grpc.py
```diff
@@ -203,27 +203,51 @@ def from_protobuf(cls, data: client_pb2.Lease) -> Lease:
...
+        if self.effective_end_time:
+            return self.effective_end_time
+        if self.effective_begin_time and self.duration:
+            return self.effective_begin_time + self.duration
+        if self.begin_time and self.duration:
+            return self.begin_time + self.duration
+        return None
+
+    @staticmethod
+    def _format_remaining(expires_at):
```
kirkbrauer, line 221: We should be able to eliminate the days and seconds calculations by relying on the `timedelta` `.days` and `.seconds` fields directly. We might also want to look into a library such as `humanize` to generalize this formatting across the codebase as we implement more scenarios like this.

kirkbrauer (approved):
This looks good! Just a few optimizations to make and I think we're good to merge.

python/packages/jumpstarter/jumpstarter/client/grpc.py
mangelajo: @raballew ^

raballew:
@kirkbrauer @mangelajo please take another look

mangelajo (approved):
looks good :)

PR #385, opened 2026-03-25
fix(operator): exclude internal DNS names from certs when using external issuer

### LINKED ISSUE #384: operator: using external certissuer attempts to create internal dns domains and fails
from @david-baylibre:
I installed the operator from the repo dist/install.yaml, this is great, devs don't have to deal with Helm and permissions.

I can now have annotations on my ingresses.
However, I'm getting errors when enabling cert-manager with Let'sEncrypt, it tries to get a cert for the local domain:
```
 describe cert jumpstarter-controller-tls
  Warning  Failed     36s   cert-manager-certificates-issuing          The certificate request has failed to complete and will be retried: Failed to wait for order resource "jumpstarter-controller-tls-1-1511209746" to become ready: order is in "errored" state: Failed to create Order: 400 urn:ietf:params:acme:error:rejectedIdentifier: Invalid identifiers requested :: Cannot issue for "jumpstarter-controller": Domain name needs at least one dot (and 3 more problems. Refer to sub-problems for more information.); subproblems:
           urn:ietf:params:acme:error:malformed: [dns: jumpstarter-controller] Invalid identifiers requested :: Domain name needs at least one dot
           urn:ietf:params:acme:error:malformed: [dns: jumpstarter-controller.jumpstarter-dev01] Invalid identifiers requested :: Domain name does not end with a valid public suffix (TLD)
           urn:ietf:params:acme:error:malformed: [dns: jumpstarter-controller.jumpstarter-dev01.svc] Invalid identifiers requested :: Domain name does not end with a valid public suffix (TLD)
           urn:ietf:params:acme:error:malformed: [dns: jumpstarter-controller.jumpstarter-dev01.svc.cluster.local] Invalid identifiers requested :: Domain name does not end with a valid public suffix (TLD)
```
On the ingress side, it does not try to get a cert because secretName is missing (unlike login which gets a cert). Adding tls:certSecret: does not help.
If I had secretName, I do get a cert from LetsEncrypt but back to the original problem: tls: first record does not look like a TLS handshake in Jumpstarter.

Should certs be obtained by ingresses (through secretName and annotations), then be used by Jumpstarter only, not ingresses which are supposed to do SSL passthrough?

mangelajo:
## Summary

- When an external cert-manager issuer (e.g. Let's Encrypt) is configured via `spec.certManager.server.issuerRef`, the operator no longer includes internal Kubernetes service DNS names (like `jumpstarter-controller`, `*.svc.cluster.local`) in Certificate resources. These caused ACME validation failures because such domains are not publicly resolvable.
- Internal K8s service DNS names are still included when using self-signed CA mode (the default), which has no trouble issuing certs for any domain.
- Added unit tests for `collectControllerDNSNames`, `collectRouterDNSNames`, and `isExternalIssuer`, plus e2e test assertions verifying external issuer certificates contain only public DNS names.

Fixes #384

## Test plan

- [x] Unit tests pass (`make test` in operator directory)
- [x] New unit tests cover both `includeInternalNames=true` (self-signed) and `includeInternalNames=false` (external issuer) paths
- [x] E2e tests updated to assert controller/router certs do NOT contain internal DNS names when external issuer is used
- [ ] Manual validation: deploy with an ACME ClusterIssuer and verify Certificate resources only contain public DNS names

### DIFF
--- a/controller/deploy/operator/internal/controller/jumpstarter/certificates_test.go
+++ b/controller/deploy/operator/internal/controller/jumpstarter/certificates_test.go
@@ -1,0 +1,298 @@
+/*
+Copyright 2025. The Jumpstarter Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jumpstarter
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1alpha1 "github.com/jumpstarter-dev/jumpstarter-controller/deploy/operator/api/v1alpha1"
+)
+
+var _ = Describe("isExternalIssuer", func() {
+	It("should return false when Server is nil", func() {
+		js := &operatorv1alpha1.Jumpstarter{
+			Spec: operatorv1alpha1.JumpstarterSpec{
+				CertManager: operatorv1alpha1.CertManagerConfig{
+					Enabled: true,
+				},
+			},
+		}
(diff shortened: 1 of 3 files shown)


controller/deploy/operator/internal/controller/jumpstarter/certificates_test.go
```diff
@@ -0,0 +1,298 @@
+/*
+Copyright 2025. The Jumpstarter Authors.
```
bennyz, line 2: 2026, but I think dates are not really necessary 😅

bennyz (approved):

bkhizgiy (approved):

PR #392, opened 2026-03-28
feat: add gPTP driver for IEEE 802.1AS / PTP time synchronization

vtz:
## Summary

- Adds `jumpstarter-driver-gptp` wrapping linuxptp (`ptp4l`/`phc2sys`) to provide IEEE 802.1AS / PTP precision time synchronization for automotive Ethernet testing
- Includes `Gptp` driver (real hardware), `MockGptp` driver (testing), `GptpClient` with `wait_for_sync` helper and Click CLI
- Pydantic models for PTP status, offsets, sync events, port stats, and parent info with validated state transitions
- Comprehensive test suite: unit tests, end-to-end mocked tests via `serve()`, and stateful tests enforcing PTP state machine transitions
- Documentation integrated into the jumpstarter.dev docs site with full API reference, configuration examples, and troubleshooting guide

### Driver Architecture
- **`Gptp` driver**: Manages `ptp4l` and `phc2sys` subprocess lifecycle, parses real-time log output for sync status, offset measurements, port state changes, and clock quality metrics
- **`MockGptp` driver**: Pluggable backend for testing without PTP hardware — supports a `StatefulPtp4l` backend that enforces valid PTP state machine transitions
- **`GptpClient`**: Async client with `wait_for_sync(threshold_ns, timeout)` convenience method and full Click CLI (`j gptp status`, `j gptp start`, `j gptp offset`, etc.)

### Models (`common.py`)
- `GptpStatus`, `GptpOffset`, `GptpSyncEvent`, `GptpPortStats`, `GptpParentInfo`
- `PortState` and `ServoState` enums with `VALID_PORT_TRANSITIONS` map

### Test Coverage
- **Unit**: ptp4l log parsing, config generation, HW timestamping detection, Pydantic model validation
- **E2E Mocked**: Full driver↔client lifecycle over gRPC via `serve()`, error paths, CLI invocation
- **Stateful**: `StatefulPtp4l` mock enforcing PTP state transitions, illegal transition rejection, operation ordering

## Test plan
- [x] `make pkg-test-jumpstarter-driver-gptp` passes all tests
- [x] `make lint` passes with no new errors
- [ ] Manual: verify docs render correctly on jumpstarter.dev after merge

### DIFF
--- a/python/packages/jumpstarter-driver-gptp/README.md
+++ b/python/packages/jumpstarter-driver-gptp/README.md
@@ -273,3 +273,4 @@ $ which ptp4l
 - Check physical layer: `ethtool <interface>` should show link up
 - Review ptp4l logs in the exporter output
 - Ensure both ends use the same domain number and transport
+
--- a/python/packages/jumpstarter-driver-gptp/README.md
+++ b/python/packages/jumpstarter-driver-gptp/README.md
@@ -123,7 +123,7 @@ export:
 
 PTP ports transition through these states:
 
-```
+```text
 INITIALIZING → LISTENING → SLAVE (synchronized to master)
                          → MASTER (elected as grandmaster)
                          → PASSIVE (backup, not active)
@@ -145,8 +145,14 @@ The clock servo tracks synchronization quality:
 ```{eval-rst}
 .. autoclass:: jumpstarter_driver_gptp.client.GptpClient()
     :members: start, stop, status, get_offset, get_port_stats,
-              get_clock_identity, get_parent_info, set_priority1,
               is_synchronized, wait_for_sync, monitor
+```
+
+```{note}
+`get_clock_identity()`, `get_parent_info()`, and `set_priority1()` are
+available on the ``MockGptp`` driver for testing. On the real ``Gptp`` driver
+they raise ``NotImplementedError`` until ptp4l UDS management socket
+integration is added.
 ```
 
 ## Examples
@@ -179,19 +185,6 @@ with serve(Gptp(interface="eth0")) as gptp:
     gptp.stop()
 ```
 
-### Force master role
-
-```python
-with serve(Gptp(interface="eth0", role="auto")) as gptp:
-    gptp.start()
-    gptp.wait_for_sync()
-
-    # Override BMCA: become grandmaster
-    gptp.set_priority1(0)
-    status = gptp.status()
-    assert status.port_state.value == "MASTER"
-```
-
 ### Using MockGptp in tests
 
 ```python
@@ -211,12 +204,12 @@ def test_my_application():
 When used inside `jmp shell`, the driver provides these commands:
 
 ```console
-$ j gptp start              # Start PTP synchronization
-$ j gptp stop               # Stop PTP synchronization
-$ j gptp status             # Show sync status
-$ j gptp offset             # Show current clock offset
-$ j gptp monitor -n 20      # Monitor 20 sync events
-$ j gptp set-priority 0     # Force grandmaster role
+j gptp start              # Start PTP synchronization
+j gptp stop               # Stop PTP synchronization
+j gptp status             # Show sync status
+j gptp offset             # Show current clock offset
+j gptp monitor -n 20      # Monitor 20 sync events
+j gptp set-priority 0     # Force grandmaster role
 ```
 
 ## Hardware Requirements
--- a/python/packages/jumpstarter-driver-gptp/README.md
+++ b/python/packages/jumpstarter-driver-gptp/README.md
@@ -160,7 +160,11 @@ integration is added.
 ### Basic lifecycle
 
 ```python
-with serve(Gptp(interface="eth0")) as gptp:
+from jumpstarter.common.utils import env
+
+with env() as client:
+    gptp = client.gptp
+
     gptp.start()
 
     # Wait for synchronization (up to 30 seconds)
@@ -176,7 +180,11 @@ with serve(Gptp(interface="eth0")) as gptp:
 ### Monitoring sync events
 
 ```python
-with serve(Gptp(interface="eth0")) as gptp:
+from jumpstarter.common.utils import env
+
+with env() as client:
+    gptp = client.gptp
+
     gptp.start()
     for event in gptp.monitor():
         print(f"[{event.event_type}] offset={event.offset_ns:.0f}ns state={event.port_state}")
--- a/python/packages/jumpstarter-driver-gptp/README.md
+++ b/python/packages/jumpstarter-driver-gptp/README.md
@@ -1,0 +1,282 @@
+# gPTP driver
+
+`jumpstarter-driver-gptp` provides IEEE 802.1AS (gPTP) and IEEE 1588 (PTPv2)
+time synchronization for Jumpstarter. It manages the
+[linuxptp](https://linuxptp.nwtime.org/) stack (`ptp4l` and `phc2sys`) as
+supervised subprocesses, enabling precise clock synchronization between an
+exporter host and a target device over automotive Ethernet or standard IP networks.
+
+gPTP is the foundation of Time-Sensitive Networking (TSN), required for
+applications like sensor fusion, ADAS, and synchronized diagnostics in
+automotive ECU testing.
+
+## Installation
+
+```{code-block} console
+:substitutions:
+$ pip3 install --extra-index-url {{index_url}} jumpstarter-driver-gptp
+```
+
+**System requirements** (on the exporter host):
+
+```console
+# Debian/Ubuntu
+$ sudo apt install linuxptp ethtool
+
+# Fedora/RHEL
+$ sudo dnf install linuxptp ethtool
+```
+
+## Configuration
+
+### gPTP Slave (802.1AS, automotive Ethernet)
+
+The most common automotive scenario — synchronize to an external grandmaster:
+
+```yaml
+export:
+  gptp:
+    type: jumpstarter_driver_gptp.driver.Gptp
+    config:
+      interface: eth0
+      domain: 0
+      profile: gptp
+      transport: L2
+      role: slave
+      sync_system_clock: true
+```
+
+### IEEE 1588 UDP mode
+
+For networks that use standard PTPv2 over UDP:
+
+```yaml
+export:
+  ptp:
+    type: jumpstarter_driver_gptp.driver.Gptp
+    config:
+      interface: eth0
+      domain: 0
+      profile: default
+      transport: UDPv4
+      role: auto
+```
+
+### gPTP Grandmaster
+
+Force this host to act as the PTP grandmaster:
+
+```yaml
+export:
+  gptp:
+    type: jumpstarter_driver_gptp.driver.Gptp
+    config:
+      interface: eth0
+      profile: gptp
+      role: master
+      sync_system_clock: false
+```
+
+### Combined with SOME/IP
+
+gPTP provides the time base for other automotive Ethernet protocols:
+
+```yaml
+export:
+  gptp:
+    type: jumpstarter_driver_gptp.driver.Gptp
+    config:
+      interface: eth0
+      profile: gptp
+      role: auto
+  someip:
+    type: jumpstarter_driver_someip.driver.SomeIp
+    config:
+      host: 192.168.1.100
+```
+
+### Config parameters
+
+| Parameter          | Description                                          | Type       | Required | Default  |
+| ------------------ | ---------------------------------------------------- | ---------- | -------- | -------- |
+| interface          | Network interface for PTP (e.g. `eth0`, `enp3s0`)    | str        | yes      |          |
+| domain             | PTP domain number (0-127)                            | int        | no       | 0        |
+| profile            | `"gptp"` (IEEE 802.1AS) or `"default"` (IEEE 1588)  | str        | no       | `"gptp"` |
+| transport          | `"L2"`, `"UDPv4"`, or `"UDPv6"`                     | str        | no       | `"L2"`   |
+| role               | `"master"`, `"slave"`, or `"auto"` (BMCA election)  | str        | no       | `"auto"` |
+| sync_system_clock  | Run `phc2sys` to sync CLOCK_REALTIME to PHC          | bool       | no       | true     |
+| ptp4l_extra_args   | Additional ptp4l command-line arguments              | list[str]  | no       | []       |
+
+## PTP Standards Reference
+
+### IEEE 802.1AS (gPTP) vs IEEE 1588 (PTPv2)
+
+| Feature           | 802.1AS (gPTP)               | IEEE 1588 (PTPv2)             |
+| ----------------- | ---------------------------- | ----------------------------- |
+| Transport         | Layer 2 only                 | L2, UDPv4, UDPv6             |
+| Timestamping      | Hardware required             | HW or software               |
+| Accuracy          | Sub-microsecond              | Sub-microsecond to ms         |
+| Use case          | Automotive, industrial TSN   | General purpose               |
+| Profile setting   | `profile: gptp`              | `profile: default`            |
+
+### Port State Machine
+
+PTP ports transition through these states:
+
+```
+INITIALIZING → LISTENING → SLAVE (synchronized to master)
+                         → MASTER (elected as grandmaster)
+                         → PASSIVE (backup, not active)
+                         → FAULTY (error detected)
+```
+
+### Servo States
+
+The clock servo tracks synchronization quality:
+
+- **s0** (unlocked): Initial state, no sync
+- **s1** (calibrating): Frequency adjustment in progress
+- **s2** (locked): Fully synchronized, offset stable
+
+## API Reference
+
+### GptpClient
+
+```{eval-rst}
+.. autoclass:: jumpstarter_driver_gptp.client.GptpClient()
+    :members: start, stop, status, get_offset, get_port_stats,
+              get_clock_identity, get_parent_info, set_priority1,
+              is_synchronized, wait_for_sync, monitor
+```
+
+## Examples
+
+### Basic lifecycle
+
+```python
+with serve(Gptp(interface="eth0")) as gptp:
+    gptp.start()
+
+    # Wait for synchronization (up to 30 seconds)
+    if gptp.wait_for_sync(timeout=30.0):
+        offset = gptp.get_offset()
+        print(f"Synchronized! Offset: {offset.offset_from_master_ns:.0f} ns")
+    else:
+        print("Sync timeout")
+
+    gptp.stop()
+```
+
+### Monitoring sync events
+
+```python
+with serve(Gptp(interface="eth0")) as gptp:
+    gptp.start()
+    for event in gptp.monitor():
+        print(f"[{event.event_type}] offset={event.offset_ns:.0f}ns state={event.port_state}")
+        if event.event_type == "fault":
+            break
+    gptp.stop()
+```
+
+### Force master role
+
+```python
+with serve(Gptp(interface="eth0", role="auto")) as gptp:
+    gptp.start()
+    gptp.wait_for_sync()
+
+    # Override BMCA: become grandmaster
+    gptp.set_priority1(0)
+    status = gptp.status()
+    assert status.port_state.value == "MASTER"
+```
+
+### Using MockGptp in tests
+
+```python
+from jumpstarter_driver_gptp.driver import MockGptp
+from jumpstarter.common.utils import serve
+
+def test_my_application():
+    with serve(MockGptp()) as gptp:
+        gptp.start()
+        assert gptp.is_synchronized()
+        assert abs(gptp.get_offset().offset_from_master_ns) < 1000
+        gptp.stop()
+```
+
+## CLI Commands
+
+When used inside `jmp shell`, the driver provides these commands:
+
+```console
+$ j gptp start              # Start PTP synchronization
+$ j gptp stop               # Stop PTP synchronization
+$ j gptp status             # Show sync status
+$ j gptp offset             # Show current clock offset
+$ j gptp monitor -n 20      # Monitor 20 sync events
+$ j gptp set-priority 0     # Force grandmaster role
+```
+
+## Hardware Requirements
+
+### PTP-capable NICs
+
+For sub-microsecond accuracy, the network interface must support hardware
+timestamping. Common PTP-capable NICs:
+
+- Intel i210, i225, i226 (automotive-grade variants available)
+- Intel X710, XL710, E810
+- Broadcom BCM5719, BCM5720
+- TI AM65x / Jacinto 7 (embedded automotive)
+- NXP S32G (automotive gateway)
+
+### Verifying hardware timestamping
+
+```console
+$ ethtool -T eth0
+# Look for:
+#   hardware-transmit
+#   hardware-receive
+#   hardware-raw-clock
+```
+
+If hardware timestamping is not available, the driver automatically falls back
+to software timestamping (`-S` flag) with a warning. Software timestamping
+provides millisecond-level accuracy, sufficient for development but not for
+production TSN.
+
+## Troubleshooting
+
+### Permission denied
+
+`ptp4l` requires `CAP_NET_RAW` (or root) for Layer 2 transport and hardware
+timestamping:
+
+```console
+$ sudo setcap cap_net_raw+ep $(which ptp4l)
+# or run the exporter as root
+```
+
+### No hardware timestamping
+
+If you see "falling back to software timestamping":
+1. Check NIC support: `ethtool -T <interface>`
+2. Verify the NIC driver is loaded: `lsmod | grep <driver>`
+3. Some virtualized NICs (virtio, veth) only support software timestamping
+
+### ptp4l not found
+
+Ensure linuxptp is installed:
+```console
+$ which ptp4l
+/usr/sbin/ptp4l
+```
+
+### No sync achieved
+
+- Verify the link partner (DUT) is running a gPTP stack
+- Check physical layer: `ethtool <interface>` should show link up
+- Review ptp4l logs in the exporter output
+- Ensure both ends use the same domain number and transport
--- a/python/packages/jumpstarter-driver-gptp/README.md
+++ b/python/packages/jumpstarter-driver-gptp/README.md
@@ -1,10 +1,11 @@
 # gPTP driver
 
-`jumpstarter-driver-gptp` provides IEEE 802.1AS (gPTP) and IEEE 1588 (PTPv2)
-time synchronization for Jumpstarter. It manages the
+`jumpstarter-driver-gptp` provides IEEE 802.1AS-2020 (gPTP) and IEEE 1588-2019
+(PTPv2.1) time synchronization for Jumpstarter. It manages the
 [linuxptp](https://linuxptp.nwtime.org/) stack (`ptp4l` and `phc2sys`) as
 supervised subprocesses, enabling precise clock synchronization between an
-exporter host and a target device over automotive Ethernet or standard IP networks.
+exporter host and a target device over automotive Ethernet or standard IP
+networks. Parameter ranges and defaults follow the respective standards.
 
 gPTP is the foundation of Time-Sensitive Networking (TSN), required for
 applications like sensor fusion, ADAS, and synchronized diagnostics in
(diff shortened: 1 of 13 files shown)


python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+        lines.append("follow_up_info\t\t1")
+        lines.append("gmCapable\t\t1")
+
+    if role == "slave":
+        lines.append("slaveOnly\t\t1")
+    elif role == "master":
+        lines.append("priority1\t\t0")
+        lines.append("priority2\t\t0")
+
+    lines.append(f"\n[{interface}]")
```
raballew, line 128: The `interface` value gets interpolated straight into an INI section header with no validation. A config value like `eth0]\nmalicious_key\tvalue\n[fake` would break out of the section and inject arbitrary ptp4l configuration.

Please add a regex check in `__post_init__` to restrict `interface` to valid Linux interface names, something like `re.fullmatch(r'[a-zA-Z0-9][a-zA-Z0-9._-]{0,14}', self.interface)`. Same value also flows into `ethtool` and `ptp4l` subprocess args (lines 162, 236).

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+                self.interface,
+            )
+
+        cmd = [
+            "ptp4l",
+            "-f", self._config_file.name,
+            "-i", self.interface,
+            ts_flag,
+            "-m",
+            *self.ptp4l_extra_args,
```
raballew: These extra args are splatted directly into the `ptp4l` exec call with zero filtering, and ptp4l runs with `CAP_NET_RAW`. An exporter config could pass something like `["-f", "/etc/shadow"]` to override the config file, or `["--uds_address", "/tmp/evil"]` to redirect the management socket.

Please add a denylist validation in `__post_init__` that rejects at least `-f`, `--config`, `-i`, `--interface`, `--uds_address`, and `--log_file`. A comment explaining that `ptp4l_extra_args` is trusted-operator-only input would also be good.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+
+        config_content = _generate_ptp4l_config(
+            interface=self.interface,
+            domain=self.domain,
+            profile=self.profile,
+            transport=self.transport,
+            role=self.role,
+        )
+
+        self._config_file = tempfile.NamedTemporaryFile(
```
raballew: This temp file is created with `delete=False` but only cleaned up in `stop()` (line 297). If the process crashes or `start()` raises after this point (e.g. ptp4l exits immediately at line 232), the temp file leaks. The default permissions also depend on umask, which could leave the config world-readable.

Consider wrapping the rest of `start()` in a try/except that calls cleanup on failure. Also worth setting `mode=0o600` on the file descriptor after creation via `os.fchmod`.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+            *cmd,
+            stdout=asyncio.subprocess.PIPE,
+            stderr=asyncio.subprocess.STDOUT,
+        )
+
+        self._port_state = "INITIALIZING"
+        self._servo_state = "s0"
+        self._priority1 = 128
+        self._stats = {}
+        self._reader_task = asyncio.get_event_loop().create_task(
```
raballew: `asyncio.get_event_loop()` is deprecated since Python 3.10 and will emit a DeprecationWarning if there's no running loop. Since this code runs inside an async method, just use `asyncio.create_task(...)` directly.

Also, if `_read_ptp4l_output` raises an unexpected exception, it'll be silently swallowed (nobody awaits this task or checks `task.exception()`). Consider adding a `task.add_done_callback(...)` that logs unhandled exceptions.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+                stderr=asyncio.subprocess.STDOUT,
+            )
+
+    @export
+    async def stop(self) -> None:
+        """Stop PTP synchronization."""
+        self._require_started()
+
+        if self._reader_task is not None:
+            self._reader_task.cancel()
```
raballew, line 328: The teardown order here is inverted: the reader task (which reads from ptp4l's stdout) gets cancelled *before* ptp4l is terminated. This means the reader might miss the final output, or worse, the cancel could race with ptp4l's shutdown.

The safer order is: (1) terminate ptp4l, (2) wait for it with timeout, (3) kill if needed + `await proc.wait()`, (4) *then* cancel the reader task, (5) terminate phc2sys similarly, (6) clean up the config file.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+            except asyncio.CancelledError:
+                pass
+            self._reader_task = None
+
+        if self._phc2sys_proc is not None:
+            self._phc2sys_proc.terminate()
+            try:
+                await asyncio.wait_for(self._phc2sys_proc.wait(), timeout=5.0)
+            except asyncio.TimeoutError:
+                self._phc2sys_proc.kill()
```
raballew, line 340: After `kill()` there's no `await self._phc2sys_proc.wait()`, so the process becomes a zombie until the event loop collects it (if ever). Same issue with ptp4l's kill path a few lines below (around line 293). These are privileged processes (`CAP_NET_RAW`), so zombie cleanup matters.

Please add `await proc.wait()` after each `kill()` call.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+                f"Invalid role: {self.role!r}. Must be one of {_VALID_ROLES}"
+            )
+
+    @classmethod
+    def client(cls) -> str:
+        return "jumpstarter_driver_gptp.client.GptpClient"
+
+    def _supports_hw_timestamping(self) -> bool:
+        try:
+            result = subprocess.run(
```
raballew: This is a blocking `subprocess.run()` call inside what will be called from an async context (via `start()` -> `_supports_hw_timestamping()`). It blocks the entire event loop while ethtool runs.

Please replace with `asyncio.create_subprocess_exec` + `await asyncio.wait_for(proc.communicate(), timeout=10.0)` or similar.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+                f"ptp4l exited immediately with code {self._ptp4l_proc.returncode}"
+            )
+
+        if self.sync_system_clock and hw_ts:
+            phc2sys_cmd = ["phc2sys", "-a", "-rr", "-m"]
+            self.logger.info("Starting phc2sys: %s", " ".join(phc2sys_cmd))
+            self._phc2sys_proc = await asyncio.create_subprocess_exec(
+                *phc2sys_cmd,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.STDOUT,
```
raballew: phc2sys's stdout is piped (`asyncio.subprocess.PIPE`) but nobody ever reads from it. If phc2sys writes enough to fill the pipe buffer (typically 64KB on Linux), it'll block and hang. This is a classic deadlock pattern with subprocess pipes.

Since the phc2sys output isn't used, change this to `stdout=asyncio.subprocess.DEVNULL`.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+            sync_count=self._stats.get("sync_count", 0),
+            followup_count=self._stats.get("followup_count", 0),
+            pdelay_req_count=self._stats.get("pdelay_req_count", 0),
+            pdelay_resp_count=self._stats.get("pdelay_resp_count", 0),
+            announce_count=self._stats.get("announce_count", 0),
+        )
+
+    @export
+    @validate_call(validate_return=True)
+    def get_clock_identity(self) -> str:
```
raballew, line 521: This always returns an empty string. The mock version (line 480) returns a real-looking `"aa:bb:cc:ff:fe:dd:ee:ff"`, so tests pass while the real driver returns nothing useful. Same issue with `get_parent_info()` (line 374) returning empty defaults, and `set_priority1()` (line 386) writing to memory without actually communicating to ptp4l.

These stubs are exported as production API and documented in the README. Once published on PyPI, they become part of the semver contract. Please either remove these from `@export`/client/CLI/README, or raise `NotImplementedError("requires ptp4l UDS integration")` so callers know they're not functional yet.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+    @export
+    async def read(self) -> AsyncGenerator[GptpSyncEvent, None]:
+        """Stream periodic sync status updates.
+
+        Yields a GptpSyncEvent approximately once per second with current
+        offset, delay, and state information.
+        """
+        self._require_started()
+        prev_state = self._port_state
+        for _ in range(100):
```
raballew: Hardcoded `range(100)` means the monitor stream silently stops after 100 events with no indication to the client. For a real PTP sync monitor this could mean losing visibility mid-test.

Consider `while True` with a way to break (e.g. checking if the process is still running), or at least make the limit configurable.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/client.py
```diff
@@ -0,0 +1,197 @@
...
+        :param poll_interval: Polling interval in seconds
+        :returns: True if synchronized, False if timeout expired
+        :rtype: bool
+        """
+        deadline = time.monotonic() + timeout
+        while time.monotonic() < deadline:
+            try:
+                if self.is_synchronized():
+                    return True
+            except Exception:
```
raballew: Bare `except Exception: pass` here swallows everything, including unexpected errors like `PermissionError`, `MemoryError`, `KeyboardInterrupt` (well, that one is BaseException). If `is_synchronized()` is failing for a reason other than "not ready yet", this loop will silently retry until timeout with no indication of what's wrong.

Please narrow this to catch only the expected exceptions (like `RuntimeError` from the "not started" check, or the gRPC `DriverError`).

python/packages/jumpstarter-driver-gptp/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+requires-python = ">=3.11"
+dependencies = [
+    "anyio>=4.10.0",
+    "jumpstarter",
+    "click",
+]
+
+[project.entry-points."jumpstarter.drivers"]
+Gptp = "jumpstarter_driver_gptp.driver:Gptp"
+MockGptp = "jumpstarter_driver_gptp.driver:MockGptp"
```
raballew: Registering `MockGptp` as a production entry point means it shows up in driver discovery alongside real drivers. No other Jumpstarter driver registers its mocks this way. If someone typos a config or an automation tool picks up available drivers, they could end up running the mock in production and get fake PTP data with no indication that sync isn't real.

Please remove this line. The mock is importable directly for tests; it doesn't need an entry point.

python/packages/jumpstarter-driver-gptp/pyproject.toml
```diff
@@ -0,0 +1,47 @@
...
+[tool.hatch.metadata.hooks.vcs.urls]
+Homepage = "https://jumpstarter.dev"
+source_archive = "https://github.com/jumpstarter-dev/repo/archive/{commit_hash}.zip"
+
+[tool.pytest.ini_options]
+addopts = "--cov --cov-report=html --cov-report=xml"
+log_cli = true
+log_cli_level = "INFO"
+testpaths = ["jumpstarter_driver_gptp"]
+asyncio_mode = "auto"
```
raballew, line 33: `asyncio_mode = "auto"` is set here but `pytest-asyncio` isn't listed in the dev dependencies (lines 42-45). This likely works because it's pulled in transitively from the workspace, but it's fragile and will break if someone runs this package's tests in isolation. Also the project uses anyio instead of asyncio. Please replace asyncio with anyio and fix imports.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,591 @@
...
+        )
+
+        await asyncio.sleep(0.5)
+        if self._ptp4l_proc.returncode is not None:
+            raise RuntimeError(
+                f"ptp4l exited immediately with code {self._ptp4l_proc.returncode}"
+            )
+
+        if self.sync_system_clock and hw_ts:
+            phc2sys_cmd = ["phc2sys", "-a", "-rr", "-m"]
```
raballew: The `-a` flag on phc2sys makes it discover and sync *all* PHC-to-system-clock pairs on the host, not just the one for the managed interface. On a multi-NIC box this could have unintended side effects.

Also, `start_new_session=True` should be added to both `create_subprocess_exec` calls (here and the ptp4l one at line 225) so that the child processes don't share the driver's process group. Without this, a stray signal to the driver's process group could leave orphaned privileged ptp4l/phc2sys processes.

raballew (changes requested):

python/packages/jumpstarter-driver-gptp/pyproject.toml
kirkbrauer: Just an additional note, the `jumpstarter-driver-power` does register its `MockPower` driver in this way, but that is because the power driver is really just a stub that is supposed to be implemented for a specific piece of hardware. I would concur with @raballew that we should remove this and not register it similar to how the `jumpstarter-driver-opendal` has a mock driver, but doesn't register it.

https://github.com/jumpstarter-dev/jumpstarter/blob/b21e591274e323f83ff599cd115307b2818a5458/python/packages/jumpstarter-driver-power/pyproject.toml#L15

raballew:
@vtz please update the comments in case you are ready for another review.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Added `_INTERFACE_RE = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9._-]{0,14}$")` and validation in `__post_init__` that rejects any interface name not matching valid Linux interface names. Added tests for injection attempts, overly-long names, and valid names like `eth0`, `enp3s0`, `ens0f0.100`, `br-lan`.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Added `_DENIED_PTP4L_ARGS` frozenset and `_validate_extra_args()` that rejects `-f`, `--config`, `-i`, `--interface`, `--uds_address`, and `--log_file` (including `--config=value` form). Validation runs in `__post_init__`. Added tests covering denylist rejection and allowed args.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. The entire `start()` body is now wrapped in `try/except` that calls `await self._cleanup()` on any failure. Temp file creation uses `tempfile.mkstemp()` + `os.fchmod(fd, 0o600)` for explicit permissions. The `_cleanup()` method handles partial state (processes, reader task, config file) safely.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Replaced `asyncio.get_event_loop().create_task(...)` with `asyncio.create_task(...)`. Added `self._reader_task.add_done_callback(self._on_reader_done)` that logs unhandled exceptions from the background reader.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Teardown order in `_cleanup()` is now: (1) terminate ptp4l + wait, (2) cancel reader task, (3) terminate phc2sys + wait, (4) remove config file. Both `stop()` and the `start()` failure path use this same `_cleanup()` method.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Added `await proc.wait()` after every `kill()` call for both ptp4l and phc2sys in the `_cleanup()` method.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Replaced `phc2sys -a -rr -m` with `phc2sys -s <interface> -c CLOCK_REALTIME -w -m` to scope synchronization to the managed interface only. Also added `start_new_session=True` to both `ptp4l` and `phc2sys` subprocess calls so child processes get their own process group.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Refactored `_supports_hw_timestamping()` to be an `async` method using `asyncio.create_subprocess_exec` and `asyncio.wait_for(proc.communicate(), timeout=10.0)` instead of the blocking `subprocess.run()`. This avoids blocking the event loop while ethtool runs.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Changed phc2sys stdout/stderr to `asyncio.subprocess.DEVNULL` since its output is not consumed. This prevents the classic pipe-buffer deadlock.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. `get_clock_identity()`, `get_parent_info()`, and `set_priority1()` now raise `NotImplementedError` with explanatory messages. These require PTP management TLV integration via the UDS socket, which is not yet implemented. The mock driver still provides full implementations for testing.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Replaced `for _ in range(100)` with `while True` for indefinite streaming. Also added a check for `_ptp4l_proc.returncode` so the stream terminates gracefully if ptp4l exits.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/client.py
vtz: Fixed in f2d7fa56. Narrowed the `except Exception` in `wait_for_sync()` to `except RuntimeError` to only catch the expected "not started" errors while letting other exceptions propagate properly.

python/packages/jumpstarter-driver-gptp/pyproject.toml
vtz: Fixed in f2d7fa56. Removed `MockGptp` from the production entry points in pyproject.toml. Good point from @kirkbrauer — unlike `MockPower`, `MockGptp` is purely a test fixture backed by a stateful fake (`StatefulPtp4l`), not a reusable emulation driver, so it does not belong in production discovery.

python/packages/jumpstarter-driver-gptp/pyproject.toml
vtz: Fixed in f2d7fa56. Added `pytest-asyncio>=0.24.0` as an explicit dev dependency and kept `asyncio_mode = "auto"` in `[tool.pytest.ini_options]`, consistent with other driver packages in the workspace.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Fixed in f2d7fa56. Replaced `phc2sys -a -rr -m` with `phc2sys -s <interface> -c CLOCK_REALTIME -w -m` to scope synchronization to the managed interface only. Also added `start_new_session=True` for better process group management.

vtz:
@raballew All 14 review comments have been addressed and replied to individually. Here's a summary:

**Security (3 items):**
- Interface name validation with regex + length check
- `ptp4l_extra_args` denylist for dangerous flags (`-f`, `--config`, `-i`, etc.)
- Secure temp file creation with `mkstemp` + `fchmod 0o600`, cleanup in `try/except`

**Async correctness (2 items):**
- `_supports_hw_timestamping()` converted to async using `asyncio.create_subprocess_exec`
- Replaced deprecated `asyncio.get_event_loop().create_task` with `asyncio.create_task` + done callback

**Process lifecycle (3 items):**
- Corrected teardown order: ptp4l → reader task → phc2sys → config cleanup
- Added `await proc.wait()` after every `kill()` to prevent zombies
- `start()` wrapped in `try/except` calling `_cleanup()` on failure

**phc2sys (2 items):**
- Scoped to managed interface: `-s <interface> -c CLOCK_REALTIME -w -m` (replaced `-a`)
- stdout/stderr → `DEVNULL` to prevent pipe-buffer deadlock

**API & streaming (2 items):**
- Stubs (`get_clock_identity`, `get_parent_info`, `set_priority1`) raise `NotImplementedError`
- `read()` uses `while True` instead of `range(100)`

**Client & config (2 items):**
- `wait_for_sync()` narrowed to `except RuntimeError`
- Removed `MockGptp` from entry points; added `pytest-asyncio` to dev deps

Ready for re-review when you have time.

python/packages/jumpstarter-driver-gptp/README.md
```diff
@@ -0,0 +1,276 @@
...
+integration is added.
+```
+
+## Examples
+
+### Basic lifecycle
+
+```python
+with serve(Gptp(interface="eth0")) as gptp:
+    gptp.start()
```
mangelajo: users will/should never use the serve() helper, that's internal to our tests.

They can use the env() helper to get a client., then do client.gptp.start(), etc..

python/packages/jumpstarter-driver-gptp/README.md
```diff
@@ -0,0 +1,276 @@
...
+    else:
+        print("Sync timeout")
+
+    gptp.stop()
+```
+
+### Monitoring sync events
+
+```python
+with serve(Gptp(interface="eth0")) as gptp:
```
mangelajo: same

python/packages/jumpstarter-driver-gptp/README.md
```diff
@@ -0,0 +1,276 @@
...
+```
+
+### Using MockGptp in tests
+
+```python
+from jumpstarter_driver_gptp.driver import MockGptp
+from jumpstarter.common.utils import serve
+
+def test_my_application():
+    with serve(MockGptp()) as gptp:
```
mangelajo, line 210: same

mangelajo (changes requested):
left some comments about the readme.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/client.py
```diff
@@ -0,0 +1,228 @@
...
+
+        Returns:
+            True if synchronized.
+
+        Raises:
+            RuntimeError: If ptp4l is not started.
+        """
+        return self.call("is_synchronized")
+
+    def wait_for_sync(self, timeout: float = 30.0, poll_interval: float = 1.0) -> bool:
```
raballew: [HIGH] `wait_for_sync` is missing a `threshold_ns` parameter. The current signature is `wait_for_sync(self, timeout, poll_interval)` and it only checks `is_synchronized()` (SLAVE + servo state). There is no offset-based gating. Consider adding a `threshold_ns: float | None = None` parameter that, when provided, additionally checks `abs(self.get_offset().offset_from_master_ns) < threshold_ns` before returning True.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/client.py
```diff
@@ -0,0 +1,228 @@
...
+                    break
+
+        @base.command(name="set-priority")
+        @click.argument("priority", type=int)
+        def set_priority(priority):
+            """Set clock priority1 for BMCA."""
+            self.set_priority1(priority)
+            click.echo(f"Priority1 set to {priority}")
+
+        return base
```
raballew: [MEDIUM] The CLI exposes start, stop, status, offset, monitor, and set-priority commands, but `port-stats` and `parent-info` are missing. The client already has `get_port_stats()` and `get_parent_info()` methods, so adding corresponding `@base.command(name="port-stats")` and `@base.command(name="parent-info")` commands should be straightforward.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,841 @@
...
+            )
+        if self.transport not in _VALID_TRANSPORTS:
+            raise ValueError(
+                f"Invalid transport: {self.transport!r}. Must be one of {_VALID_TRANSPORTS}"
+            )
+        if self.role not in _VALID_ROLES:
+            raise ValueError(
+                f"Invalid role: {self.role!r}. Must be one of {_VALID_ROLES}"
+            )
+        _validate_extra_args(self.ptp4l_extra_args)
```
raballew, line 223: I do not have access to those IEEE standards, but it think a lot of those extra args have defined ranges and require input validation. i can not judge about them. @vtz can you confirm that you properly validate inputs according to the standard? Also it would be interesting to know which version of the standard you are implement against. 2022 or 2025?

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,841 @@
...
+        """
+        proc = self._ptp4l_proc
+        if proc is None or proc.stdout is None:
+            return
+        while True:
+            raw = await proc.stdout.readline()
+            if not raw:
+                self._ptp4l_proc = None
+                self._synchronized_invalidate()
+                break
```
raballew, line 291: [MEDIUM] `_read_ptp4l_output` sets `self._ptp4l_proc = None` on EOF, and `_require_started()` checks that same attribute then accesses `.returncode` on it. While CPython's GIL and single-threaded async loop prevent actual races today, the pattern is fragile. Capturing `proc = self._ptp4l_proc` into a local variable at the top of `_require_started` and working with that reference would make the code more resilient to future refactoring.

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
```diff
@@ -0,0 +1,841 @@
...
+            self.logger.info("Starting ptp4l: %s", " ".join(cmd))
+            self._ptp4l_proc = await asyncio.create_subprocess_exec(
+                *cmd,
+                stdout=asyncio.subprocess.PIPE,
+                stderr=asyncio.subprocess.STDOUT,
+                start_new_session=True,
+            )
+
+            self._port_state = "INITIALIZING"
+            self._servo_state = "s0"
```
raballew, line 408: [LOW] When ptp4l is not installed, the current code lets `FileNotFoundError` propagate through a generic `except Exception` block, producing a raw traceback. Catching `FileNotFoundError` specifically and re-raising as a `RuntimeError` with a helpful message (e.g., "ptp4l not found, install linuxptp") would make the failure much easier to diagnose.

raballew (approved):

mangelajo:
@ambient-code can you handle comments and rebase?

python/packages/jumpstarter-driver-gptp/jumpstarter_driver_gptp/driver.py
vtz: Addressed in 659a173c. The driver targets **IEEE 802.1AS-2020** (gPTP) and **IEEE 1588-2019** (PTPv2.1) via the linuxptp reference implementation.

Input validation now enforced per the standards:
- **domain**: 0-127 (IEEE 1588-2019 §7.1; IEEE 802.1AS-2020 always uses domain 0)
- **priority1**: 0-255 (IEEE 1588-2019 §7.6.2.2) — validated in `set_priority1()` on both real and mock drivers
- **interface**: regex-validated (Linux IFNAMSIZ constraint)
- **profile/transport/role**: enum-validated against the supported values
- **ptp4l_extra_args**: denylist blocks safety-critical flags

The remaining ptp4l parameters (e.g. `logSyncInterval`, `logAnnounceInterval`) are passed through `ptp4l_extra_args` and validated by ptp4l itself against its own range checks. We deliberately do not duplicate ptp4l's internal validation for these, since they are version-dependent and ptp4l reports clear errors for out-of-range values.

Test default values reference the standards: `clock_class=248` (slave-only, IEEE 1588-2019 Table 5), `clock_accuracy=0xFE` (unknown, Table 6), `priority1=128` (default per §7.6.2.2).

vtz:
@mangelajo All three comments have been addressed — README examples now use `env()` with `client.gptp` attribute access instead of `serve()` (only the MockGptp unit test example still uses `serve()`, which is appropriate for tests).

All review comments from @raballew and @kirkbrauer have also been addressed and replied to individually. The branch has been rebased onto latest `main`. Ready for re-review.

raballew:
@vtz could you rebase this branch? those changes lgtm - lets see what @kirkbrauer and @mangelajo say.

python/packages/jumpstarter-driver-gptp/examples/exporter.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: jumpstarter.dev/v1alpha1
+kind: ExporterConfig
+endpoint: grpc://localhost:8082
+drivers:
```
raballew: `drivers:` should be `export:` — that is the actual field name in `ExporterConfigV1Alpha1`. With `drivers:`, Pydantic silently drops the block and the exporter starts with no child drivers, so `j gptp` never appears in the shell. Also missing a required `metadata:` block (e.g. `metadata:\n  name: gptp-example`).

raballew (changes requested):
The example uses `drivers:` but the schema field is `export:` (see `ExporterConfigV1Alpha1` in `jumpstarter/config/exporter.py`). Pydantic silently ignores unknown fields, so the exporter starts with an empty composite — no drivers are loaded and `j gptp` never appears in the shell. All other driver examples in the repo use `export:`. Please change `drivers:` to `export:` here, and also add the required `metadata.name` field which the schema mandates.

raballew:
I was able to test the mock setup but have not support for gptp on my hardware @mangelajo @vtz do could you validate that it actually works for you?

PR #397, opened 2026-03-31
fix: clean up listenQueues entry on listener disconnect

### LINKED ISSUE #362: Possible memory leak in listenQueues on listener disconnect
## Description

I think this might be a memory leak.

At `controller/internal/service/controller_service.go:442`, `Listen()` creates an entry in the `listenQueues` sync.Map via `LoadOrStore`. When the listener disconnects (ctx.Done at line 445 or stream.Send failure at line 449), the function returns without removing the entry.

There is no `listenQueues.Delete()` call anywhere in the codebase, so channels accumulate in the map over time as leases are created and disconnected.

## Suggested Fix

- Add `defer s.listenQueues.Delete(leaseName)` after the `LoadOrStore` call

raballew:
## Summary
- The `listenQueues` sync.Map in the controller never cleaned up entries when a listener disconnected, causing unbounded memory growth over time as orphaned channels accumulated
- Fix: add `defer s.cleanupListenQueue(leaseName)` after `LoadOrStore` in `Listen()` so the entry is removed when the function returns for any reason (context cancellation, stream send failure, etc.)

## Test plan
- 5 new tests: cleanup on context cancellation, cleanup on stream send failure, new listener works after cleanup, Dial doesn't panic after cleanup, messages flow correctly through a fresh queue
- All existing controller unit tests pass with no regressions

Closes #362

### DIFF
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -440,6 +440,7 @@ func (s *ControllerService) Listen(req *pb.ListenRequest, stream pb.ControllerSe
 	}
 
 	queue, _ := s.listenQueues.LoadOrStore(leaseName, make(chan *pb.ListenResponse, 8))
+	defer s.cleanupListenQueue(leaseName)
 	for {
 		select {
 		case <-ctx.Done():
@@ -450,6 +451,10 @@ func (s *ControllerService) Listen(req *pb.ListenRequest, stream pb.ControllerSe
 			}
 		}
 	}
+}
+
+func (s *ControllerService) cleanupListenQueue(leaseName string) {
+	s.listenQueues.Delete(leaseName)
 }
 
 // Status is a stream of status updates for the exporter.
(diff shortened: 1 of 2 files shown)


mangelajo:
Hmm, I thought this had already been fixed, let me read :) thanks @raballew !!

controller/internal/service/controller_service.go
```diff
@@ -440,6 +440,7 @@ func (s *ControllerService) Listen(req *pb.ListenRequest, stream pb.ControllerSe
 	}
 
 	queue, _ := s.listenQueues.LoadOrStore(leaseName, make(chan *pb.ListenResponse, 8))
+	defer s.cleanupListenQueue(leaseName)
```
mangelajo, line 443: I believe you could have simply defer s.listenQueues.Delete(leaseName) :) 

but that's fine :D

mangelajo (approved):

PR #399, opened 2026-03-31
fix: wire keepalive configuration fields into gRPC server options

### LINKED ISSUE #363: Keepalive configuration fields parsed but never wired
## Description

Five keepalive configuration fields at `controller/internal/config/types.go:53-81` are parsed from configuration but never passed to `keepalive.ServerParameters` at `config/grpc.go:12-31`. Operators configuring these parameters see no effect.

The affected fields are: `Timeout`, `IntervalTime`, `MaxConnectionIdle`, `MaxConnectionAge`, `MaxConnectionAgeGrace`. Only `MinTime` and `PermitWithoutStream` are actually used (in the `EnforcementPolicy`).

## Suggested Fix

Either wire the fields into a `keepalive.ServerParameters` and pass it as an additional `grpc.ServerOption`, or remove the unused fields from the struct.

raballew:
## Summary
- The five keepalive config fields (Timeout, IntervalTime, MaxConnectionIdle, MaxConnectionAge, MaxConnectionAgeGrace) were parsed from configuration but never wired into `keepalive.ServerParameters`, making them silently ignored
- `LoadGrpcConfiguration` now returns `[]grpc.ServerOption` and constructs a `ServerParameters` option when any keepalive field is non-zero
- Propagates the slice type change through all callers (controller service, router service, cmd entrypoints)

## Test plan
- 8 new tests covering: timeout/interval wiring, connection lifetime fields, invalid value errors for all 5 fields, empty fields produce zero-value params, backward compatibility
- All existing config tests pass with no regressions
- Full `go build ./...` clean

Closes #363

### DIFF
--- a/controller/internal/config/grpc.go
+++ b/controller/internal/config/grpc.go
@@ -7,25 +7,62 @@ import (
 	"google.golang.org/grpc/keepalive"
 )
 
-// LoadGrpcConfiguration loads the gRPC server configuration from the parsed Config struct.
-// It creates a gRPC server option with keepalive enforcement policy configured.
-func LoadGrpcConfiguration(config Grpc) (grpc.ServerOption, error) {
+func LoadGrpcConfiguration(config Grpc) ([]grpc.ServerOption, error) {
 	ka := config.Keepalive
 
-	// Parse MinTime with default of 1s
 	minTime, err := ParseDuration(ka.MinTime)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse keepalive minTime: %w", err)
 	}
 	if minTime == 0 {
-		minTime = 1e9 // 1 second default
+		minTime = 1e9
 	}
 
-	// Create the keepalive enforcement policy
 	policy := keepalive.EnforcementPolicy{
 		MinTime:             minTime,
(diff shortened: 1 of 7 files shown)


controller/internal/config/grpc.go
```diff
@@ -7,25 +7,62 @@ import (
...
 	ka := config.Keepalive
 
-	// Parse MinTime with default of 1s
 	minTime, err := ParseDuration(ka.MinTime)
 	if err != nil {
 		return nil, fmt.Errorf("failed to parse keepalive minTime: %w", err)
 	}
 	if minTime == 0 {
-		minTime = 1e9 // 1 second default
+		minTime = 1e9
```
mangelajo, line 18: I would keep the comment here, it's not obvious that 1e9 means s second in this context. May be there is a constant we can rely on.

mangelajo:
@coderabbitai can you verify if this patch could potentially change behavior of any server using the default settings?

mangelajo (changes requested):
just one nit.

controller/internal/config/grpc.go
mangelajo: this is a nit...

mangelajo (approved):

PR #406, opened 2026-04-06
ci(e2e): run operator e2e on main push, drop helm matrix

mangelajo:
- Trigger workflow on push to main so merges are covered.
- Run e2e with METHOD=operator only; remove helm matrix leg and ARM/helm exclude.

Made-with: Cursor

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -2,6 +2,9 @@ name: End-to-end tests
 
 on:
   workflow_dispatch:
+  push:
+    branches:
+      - main
   pull_request:
   merge_group:
 
@@ -21,7 +24,7 @@ jobs:
       - uses: dorny/paths-filter@v3
         id: filter
         with:
-          base: ${{ github.base_ref || github.event.merge_group.base_ref || 'main' }}
+          base: ${{ github.base_ref || github.event.merge_group.base_ref || github.ref }}
           filters: |
             e2e:
               - 'controller/**'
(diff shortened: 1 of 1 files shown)


mangelajo:
some tests marked as missing, but just because we need to change the project config due to the new matrix.

.github/workflows/e2e.yaml
```diff
@@ -2,6 +2,9 @@ name: End-to-end tests
 
 on:
   workflow_dispatch:
+  push:
```
raballew, line 5: [AI-generated, human reviewed]

With the new `push` trigger, the `e2e-tests` job condition at line 38 may skip tests on push-to-main:

```yaml
if: needs.changes.outputs.should_run == 'true' || github.event_name == 'workflow_dispatch'
```

For `push` events, `github.base_ref` is empty, so `dorny/paths-filter` falls back to the `base` parameter (`'main'`). On squash merges or force-pushes, the push payload's `before` SHA may not be a direct ancestor, causing the filter to produce unexpected results. Since the `if` condition doesn't include `github.event_name == 'push'`, e2e tests could be silently skipped on push to main.

raballew (commented):
The PR correctly:

- Adds `push` trigger on `main` branch for e2e tests
- Removes `helm` from the matrix `method` dimension
- Removes the ARM/helm `exclude` rule
- Hardcodes `METHOD: operator` instead of using `${{ matrix.method }}`

One medium-severity concern identified below regarding the interaction between the new `push` trigger and the existing `dorny/paths-filter` gate.

raballew (dismissed):

.github/workflows/e2e.yaml
mangelajo: oh right, I forgot we had that filter, checking, thanks! :)

mangelajo:
@raballew Thanks for the detailed review.

**paths-filter on `push`:** Updated the `base` input to fall back to `github.ref` instead of the literal `'main'`. That matches [dorny/paths-filter’s long-lived-branch behavior](https://github.com/dorny/paths-filter#supported-workflows): when the workflow is triggered by a push and `base` is the same branch ref as the push, changes are detected against the previous commit on that branch—avoiding the weak default we had when `github.base_ref` was empty.

**Job `if`:** We’re **not** adding a blanket `push` bypass; downstream jobs still use `needs.changes.outputs.should_run == 'true' || github.event_name == 'workflow_dispatch'` and rely on the corrected filter (option A).

Pushed as a single amended commit on `ci/e2e-operator-main-push` (fixup squashed). If anything still looks off, happy to adjust.

.github/workflows/e2e.yaml
mangelajo: Ok, now the filtering base is fixed, and we always run the main E2E tests on all pushes (not the compat ones), as we just want an indication of any patch introducing flakiness in our codebase over the time.

PR #422, opened 2026-04-07
feat: shell completion for jmp, jmp-admin, and j

### LINKED ISSUE #35: RFE: terminal autocomplete for jumpstarter

raballew:
## Summary
- Auto-source shell completions for `jmp`, `jmp-admin`, and `j` when entering `jmp shell` (bash, zsh, fish)
- Bake `j` subcommand names into shell init scripts at startup for instant TAB completion (no gRPC on every keypress)
- Add `j completion {bash,zsh,fish}` subcommand with fast-path dispatch that avoids the full async stack and catches `SystemExit` cleanly
- Add shared `make_completion_command` factory in `jumpstarter-cli-common` reused across all three CLIs

Closes #35

## Test plan
- [x] 12 unit tests for `_generate_shell_init` covering bash/zsh/fish with and without j_commands, profiles, and unknown shells
- [x] 6 unit tests for `j completion` covering bash/zsh/fish script generation, error cases, and SystemExit handling
- [x] Manual: `jmp shell` then TAB-complete `jmp`, `jmp-admin`, and `j` subcommands in bash/zsh/fish
- [x] Manual: `j completion bash | source /dev/stdin` outside jmp shell

### DIFF
--- a/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/completion.py
+++ b/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/completion.py
@@ -11,6 +11,8 @@ def make_completion_command(cli_group_factory: Callable[[], click.Command], prog
         """Generate shell completion script."""
         cli_group = cli_group_factory()
         comp_cls = get_completion_class(shell)
+        if comp_cls is None:
+            raise click.ClickException(f"Unsupported shell: {shell}")
         comp = comp_cls(cli_group, {}, prog_name, complete_var)
         click.echo(comp.source())
 
--- a/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/completion.py
+++ b/python/packages/jumpstarter-cli-common/jumpstarter_cli_common/completion.py
@@ -1,0 +1,17 @@
+from typing import Callable
+
+import click
+from click.shell_completion import get_completion_class
+
+
+def make_completion_command(cli_group_factory: Callable[[], click.Command], prog_name: str, complete_var: str):
+    @click.command("completion")
+    @click.argument("shell", type=click.Choice(["bash", "zsh", "fish"]))
+    def completion(shell: str):
+        """Generate shell completion script."""
+        cli_group = cli_group_factory()
+        comp_cls = get_completion_class(shell)
+        comp = comp_cls(cli_group, {}, prog_name, complete_var)
+        click.echo(comp.source())
+
+    return completion
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
@@ -42,7 +42,7 @@ _TOKEN_REFRESH_THRESHOLD_SECONDS = 120
 
 
 
-def _run_shell_only(lease, config, command, path: str) -> int:
+def _run_shell_only(lease, config, command, path: str, j_commands: list[str] | None = None) -> int:
     """Run just the shell command without log streaming."""
     allow = config.drivers.allow if config is not None else getattr(lease, "allow", [])
     unsafe = config.drivers.unsafe if config is not None else getattr(lease, "unsafe", False)
@@ -59,6 +59,7 @@ def _run_shell_only(lease, config, command, path: str) -> int:
         lease=lease,
         insecure=insecure,
         passphrase=passphrase,
+        j_commands=j_commands,
     )
 
 
@@ -324,8 +325,18 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
                                 warning_text = monitor.status_message[len(HOOK_WARNING_PREFIX) :]
                                 click.echo(click.style(f"Warning: {warning_text}", fg="yellow", bold=True))
 
-                            # Run the shell command
-                            exit_code = await anyio.to_thread.run_sync(_run_shell_only, lease, config, command, path)
+                            # Extract j command names for static shell completion
+                            j_commands = None
+                            try:
+                                cli_group = client.cli()
+                                if hasattr(cli_group, "list_commands"):
+                                    j_commands = cli_group.list_commands(None)
+                            except Exception:
+                                pass
+
+                            exit_code = await anyio.to_thread.run_sync(
+                                _run_shell_only, lease, config, command, path, j_commands
+                            )
 
                             # Shell has exited. For auto-created leases (release=True), call
                             # EndSession to trigger afterLease hook while keeping log stream
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
@@ -331,8 +331,8 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
                                 cli_group = client.cli()
                                 if hasattr(cli_group, "list_commands"):
                                     j_commands = cli_group.list_commands(None)
-                            except Exception:
-                                pass
+                            except Exception as e:
+                                logger.debug("Failed to extract j commands for completion: %s", e)
 
                             exit_code = await anyio.to_thread.run_sync(
                                 _run_shell_only, lease, config, command, path, j_commands
--- a/python/packages/jumpstarter/jumpstarter/common/utils.py
+++ b/python/packages/jumpstarter/jumpstarter/common/utils.py
@@ -1,4 +1,5 @@
 import os
+import re
 import signal
 import sys
 import tempfile
@@ -85,7 +86,17 @@ def _run_process(
     return process.wait()
 
 
+_SAFE_COMMAND_NAME = re.compile(r"^[a-zA-Z0-9_-]+$")
+
+
+def _validate_j_commands(j_commands: list[str] | None) -> list[str] | None:
+    if j_commands is None:
+        return None
+    return [cmd for cmd in j_commands if _SAFE_COMMAND_NAME.match(cmd)]
+
+
 def _generate_shell_init(shell_name: str, use_profiles: bool, j_commands: list[str] | None = None) -> str:
+    j_commands = _validate_j_commands(j_commands)
     if shell_name.endswith("bash"):
         lines = []
         if use_profiles:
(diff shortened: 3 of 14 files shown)


mangelajo:
Something is weird at my side:
```
➜  python git:(039-shell-completion) ✗ export PATH=$(pwd)/.venv/bin:$PATH
➜  python git:(039-shell-completion) ✗ jmp shell -l board-type=qc8650
[04/08/2026 16:07:58] INFO     [jumpstarter.client.lease] Acquiring lease 019d6d6b-da70-76c2-9235-d789b272009b for selector board-type=qc8650 for duration 0:30:00
                      INFO     Waiting for ready connection at /var/folders/j_/9v47_5750_s85_5j4tybsx200000gn/T/jumpstarter-dtb075vs/socket
[04/08/2026 16:07:59] INFO     [jumpstarter_cli.shell] Waiting for beforeLease hook to complete...
(eval):39: command not found: compdef
(eval):39: command not found: compdef
(eval):39: command not found: compdef
majopela@majopela-mac python % exit
[04/08/2026 16:08:03] INFO     [jumpstarter.client.lease] Releasing Lease 019d6d6b-da70-76c2-9235-d789b272009b

**Comment by mangelajo** [comment]
this is MacOS / zsh

**Comment by mangelajo** [comment]
```
➜  python git:(039-shell-completion) ✗ jmp shell -l target=j784s4evm,enabled=true
[04/10/2026 12:24:59] INFO     [jumpstarter.client.lease] Acquiring lease 019d76ec-6e10-707c-98aa-a3d6922e3327 for selector target=j784s4evm,enabled=true for duration 0:30:00
                      INFO     Waiting for ready connection at /var/folders/j_/9v47_5750_s85_5j4tybsx200000gn/T/jumpstarter-3z7n8ze5/socket
[04/10/2026 12:25:01] INFO     [jumpstarter_cli.shell] Waiting for beforeLease hook to complete...
majopela@majopela-mac python % j
Usage: j [OPTIONS] COMMAND [ARGS]...

  Generic composite device

Options:
  --log-level [DEBUG|INFO|WARNING|ERROR|CRITICAL]
                                  Set the log level
  --help                          Show this message and exit.

Commands:
  power    SNMP power control commands
  serial   Serial port client
  ssh      Run SSH command with arguments
  storage  Software-defined flasher interface
  tcp      Generic Network Connection
  tmt      Run TMT command with arguments
  vnc      Open a VNC session and block until the user closes it.
majopela@majopela-mac python % j
power    serial   ssh      storage  tcp      tmt      vnc

majopela@majopela-mac python % j storage

```

I found 2 problems:

* [CRITICAL] The nice prompt is lost
* [LOW]  Anything beyond first level of j command is not autocompleted (but could be analyzed on a follow up PR)

For jmp it seems to work well.

**Comment by raballew** [comment]
@ambient-code why is CI failing?

**Comment by mangelajo** [comment]
Now completion for bare jmp does not work, but the prompt is back

**Comment by raballew** [comment]
```
⬢ [toolbx]~/code/jumpstarter/python% autoload -Uz compinit && compinit
⬢ [toolbx]~/code/jumpstarter/python% export PATH="$(uv run which jmp | xargs dirname):$PATH" 
Installed 4 packages in 2ms
⬢ [toolbx]~/code/jumpstarter/python% eval "$(uv run jmp completion zsh)"
Installed 4 packages in 2ms
⬢ [toolbx]~/code/jumpstarter/python% jmp c    jmp c
completion  -- Generate shell completion script.
config      -- Manage local configurations
create      -- Create a resource
```

File: python/packages/jumpstarter/jumpstarter/common/utils.py [hunk 1]
```diff
@@ -127,44 +267,28 @@ def launch_shell(
...
+        if shell_name.endswith("fish"):
+            return _launch_fish(shell, init_file, common_env, context, lease)
+        else:
+            return _run_process([shell], common_env, lease)
+    finally:
+        if init_file:
+            try:
+                os.unlink(init_file.name)
+            except OSError:
+                pass
```
evakhoni: tbh this whole file doesn't look pretty. I was itching to refactor it a bit while back, as it was already a bit of an if else jungle in combination with partially similar special case treatments for the different shells, and in this aspect your bot made it somewhat worse. i mean it's not a deal breaker in any sense, but for readability / ease to maintain of the codebase, imo worth giving it a bit of thought

python/packages/jumpstarter/jumpstarter/common/utils.py
evakhoni, line 1: TBH, this whole file doesn't look pretty. I was itching to refactor it a bit while back, as it was already a bit of an if else jungle in combination with partially overlapping special case treatments for the different shells, and in this aspect your bot just made it further away from being any elegant code :) 
not a deal breaker in any sense, but i think it worth to optimize it a bit.

python/packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -95,29 +250,14 @@ def launch_shell(
...
-
     shell = os.environ.get("SHELL", "bash")
     shell_name = os.path.basename(shell)
-
     common_env = os.environ | {
         JUMPSTARTER_HOST: host,
         JMP_DRIVERS_ALLOW: "UNSAFE" if unsafe else ",".join(allow),
-        "_JMP_SUPPRESS_DRIVER_WARNINGS": "1",  # Already warned during client initialization
+        "_JMP_SUPPRESS_DRIVER_WARNINGS": "1",
     }
```
evakhoni: while I'm not a big fan of a file packed full of comments, but completely stripping everything including the docstrings and previous comments made by other devs is a strange decision 🤷

python/packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -127,44 +267,28 @@ def launch_shell(
...
-        cmd = [shell]
-        if not use_profiles:
-            cmd.append("--no-rcs")
-        cmd.extend(["-o", "inc_append_history", "-o", "share_history"])
-        return _run_process(cmd, env, lease)
+    init_file = None
+    if init_content:
+        init_file = tempfile.NamedTemporaryFile(mode="w", suffix=".sh", delete=False)
+        init_file.write(init_content)
+        init_file.close()
```
evakhoni, line 341: this one is inconsistent in a way it writes it here for fish but inside _launch_xxx for the rest of the shells.
it's also a duplicate code that could be made generic for all the 3 shells by using mkdtemp tmpdir for all of them, probably in a separate function.

python/packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -84,6 +88,157 @@ def _run_process(
     return process.wait()
 
 
+_SAFE_COMMAND_NAME = re.compile(r"^[a-zA-Z0-9_-]+$")
+
+
+def _validate_j_commands(j_commands: list[str] | None) -> list[str] | None:
+    if j_commands is None:
+        return None
+    return [cmd for cmd in j_commands if _SAFE_COMMAND_NAME.match(cmd)]
```
evakhoni, line 142: that's an interesting one. so basically it tries to protect from maliciously set resource name in ExporterConfig? not sure how real is the attack surface there. I tried creating one locally and attempted to autocomplete on unpatched version using `autoload -Uz compinit && compinit && source <(_J_COMPLETE=zsh_source j)` and it just escaped the special characters for me by default. and even if it's not, it seems it would not execute it by itself anyway but rather just autocomplete it and only if the user proceeds with the malformed command then execution is possible, which is unlikely unless the user have absolutely no idea what he's doing :)

also it's a one-liner anyway, and never reused in code (test cases aside, even though I doubt there's any meaningful complexity to test here anyway) , so it's probably better to inline in _generate_shell_init.

python/packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -84,6 +88,157 @@ def _run_process(
...
+
+
+def _resolve_cli_paths() -> tuple[str, str, str]:
+    jmp = shutil.which("jmp") or "jmp"
+    jmp_admin = shutil.which("jmp-admin") or "jmp-admin"
+    j = shutil.which("j") or "j"
+    return jmp, jmp_admin, j
```
evakhoni, line 149: same goes for here. doesn't warrant a separate function IMO.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -324,8 +325,18 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
...
-                            # Run the shell command
-                            exit_code = await anyio.to_thread.run_sync(_run_shell_only, lease, config, command, path)
+                            # Extract j command names for static shell completion
+                            j_commands = None
+                            try:
+                                cli_group = client.cli()
+                                if hasattr(cli_group, "list_commands"):
+                                    j_commands = cli_group.list_commands(None)
+                            except Exception as e:
+                                logger.debug("Failed to extract j commands for completion: %s", e)
```
evakhoni, line 335: this one is a bit of concern, as it seems like it generate only a shallow command list, rather than a tree, so only the `j commands` will get completion, but `j commands subcommands` will get nothing, and they are the useful ones IMO. this is also inconsistent with the fallback that calls `j completion <shell>` which from my tests will autocomplete the full tree.

a better solution if we want a proper autocompletion caching, would be building the full tree here recursively. 

another option to explore, is making j completion to generate and return a full tree inside its j completion command, rather than a simple completion script that includes `response=("${(@f)$(env COMP_WORDS="${words[*]}" COMP_CWORD=$((CURRENT-1)) _J_COMPLETE=zsh_complete j)}")` as it does currently, so then by loading j completion once at jmp shell start (rather than passing a shallow list as we do now) we're all set and covered. this option may be more complex though. 

yet another option would be to implement a proper completion tree with either of the options above in a separate PR as it seems like an effort on its own, and for now relying on the build-in autocompletion in j which is slow because of dynamic fetching, but it works nevertheless and works well.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/completion.py
```diff
@@ -0,0 +1,19 @@
...
+    @click.command("completion")
+    @click.argument("shell", type=click.Choice(["bash", "zsh", "fish"]))
+    def completion(shell: str):
+        """Generate shell completion script."""
+        cli_group = cli_group_factory()
+        comp_cls = get_completion_class(shell)
+        if comp_cls is None:
+            raise click.ClickException(f"Unsupported shell: {shell}")
+        comp = comp_cls(cli_group, {}, prog_name, complete_var)
+        click.echo(comp.source())
```
evakhoni, line 17: I recall this one, I also mentioned it in https://github.com/jumpstarter-dev/jumpstarter/issues/35 as evaluate the usefulness of `jmp complete` subcommand.. however upon further examination, the standard way for Click/Typer seems to be: 
`_JMP_COMPLETE=<zsh/bash/fish>_source jmp` which already exists without the implementation. the `jmp completion <shell>` format seems to be used by the Cobra library in tools written in go such as kubectl, and there are other patterns with other libraries and languages.

so in this case, we're just adding another command exposing the same completion script for which a command already exists, and it's not the native way for python. there is still "some" benefit of familiarity, as we're working closely with k8s/openshift where the cli is built in Cobra so users may be familiar with it more, but that's a thin one.

python/packages/jumpstarter/jumpstarter/common/utils.py
```diff
@@ -84,6 +88,157 @@ def _run_process(
...
+
+def _generate_shell_init(shell_name: str, use_profiles: bool, j_commands: list[str] | None = None) -> str:
+    j_commands = _validate_j_commands(j_commands)
+    jmp, jmp_admin, j = _resolve_cli_paths()
+    if shell_name.endswith("bash"):
+        lines = []
+        if use_profiles:
+            lines.append('[ -f ~/.bashrc ] && source ~/.bashrc')
+        lines.append(f'eval "$({jmp} completion bash 2>/dev/null)"')
+        lines.append(f'eval "$({jmp_admin} completion bash 2>/dev/null)"')
```
evakhoni, line 160: so we're basically adding jmp completion into jmp shell itself here. IMHO not sure how useful it is considering that most jmp commands were triggered prior to that in order to reach jmp shell in the first place, which means that we probably really want jmp completion integrated into the user's shell rc file, one way could be adding it there via jumpstarter install.sh script. I'm not saying this one is redundant, but I definitely think that a system-wide one is missing.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
evakhoni: personally I would suggest the later, just sticking with build-in completion for this PR, and follow up in a separate one just handling the caching properly. **if you want, this is something that I can do instead**, as I already have some ideas of how to implement it properly, and I"m also thinking about expanding it to dynamic completion for specific subcommands, at least for jmp, where we would fetch a discoverable values (this part is completely absent from both jmp as well as j at the moment)

evakhoni (commented):
@raballew I've seen that you submitted some code changes related to my last review, but I can't tell which of them address which of my concerns. ptal at my comments if you can mark those you already addressed. thanks!

PR #423, opened 2026-04-07
Add Jumpstarter Enhancement Proposal (JEP) process and docs

kirkbrauer:
## Summary

Introduces the **Jumpstarter Enhancement Proposal (JEP)** process: an index and template under the Sphinx docs (`python/docs/source/internal/jeps/`), JEP-0000 process text, and the Renode JEP-0010. JEPs are submitted and reviewed **via pull request** (no issue template).

Also adds **Cursor** and **Claude** integration: `.cursor/rules/jep-process.mdc` (single source of truth), `.claude/rules/jep-process.md` symlink, and the `propose-jep` skill.

## Latest follow-up (this push)

- Renamed `.cursor/rules/jep-adr-process.mdc` → **`jep-process.mdc`** and updated `.claude` symlink + references in the skill and JEP README.
- **Sphinx** `make html` verified locally with `uv sync --group docs`.

## Reviewers

Please **re-review** on the current head when you can; prior `CHANGES_REQUESTED` reviews were on an older commit. Resolve conversation threads that are already addressed so the PR can merge.

PR #533, opened 2026-04-09
Add Renode emulator driver for embedded target simulation

vtz:
## Summary

- Introduce `jumpstarter-driver-renode`, a composite driver that enables Renode-based virtual hardware targets in Jumpstarter
- Users can define any Renode-supported platform (STM32, S32K, Nucleo H753ZI, etc.) via exporter YAML configuration without modifying driver code
- Includes `RenodeMonitor` (async telnet client), `RenodePower` (process lifecycle), `RenodeFlasher` (firmware loading via `sysbus LoadELF`/`LoadBinary`), and `RenodeClient` (composite client with CLI extension)
- Adds ADR-0001 documenting architectural decisions: control interface (telnet monitor), UART exposure (PTY terminals), configuration model (managed mode with extra_commands), and firmware loading strategy

## Design Decisions

Key architectural decisions are documented in `python/docs/source/contributing/adr/0001-renode-integration.md`:

| Decision | Choice | Rationale |
|---|---|---|
| Control Interface | Telnet Monitor | Lowest-common-denominator, mirrors QEMU's QMP pattern, uses `anyio.connect_tcp` |
| UART Exposure | PTY Terminals | Consistent with QEMU driver's PySerial integration |
| Configuration Model | Managed Mode | Driver constructs monitor commands from YAML params; `extra_commands` for customization |
| Firmware Loading | Deferred to Flash | Aligns with QEMU's `flash()` then `on()` semantic |

## Files Changed

- `python/packages/jumpstarter-driver-renode/` — Full driver package (monitor, driver, client, tests, examples)
- `python/docs/source/contributing/adr/` — ADR index and ADR-0001 for Renode integration
- `python/docs/source/reference/package-apis/drivers/renode.md` — Docs symlink to driver README

## Test Plan

- [ ] Unit tests for `RenodeMonitor` (connect, execute, close, timeout, error handling)
- [ ] Unit tests for `RenodePower` (on/off, firmware loading, UART configuration)
- [ ] Unit tests for `RenodeFlasher` (flash, format detection, not-running error)
- [ ] Unit tests for `Renode` config validation (missing platform, defaults, UART config)
- [ ] Client and CLI tests (property access, monitor command forwarding)
- [ ] E2E test with `serve()` (requires Renode installed, marked with `skipif`)
- [ ] CI: Add Renode installation step to `python-tests.yaml` (follow-up)


python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,229 @@
...
+        cmdline = [
+            renode_bin,
+            "--disable-xwt",
+            "--plain",
+            "--port",
+            str(port),
+        ]
+
+        self.logger.info("starting Renode: %s", " ".join(cmdline))
+        self._process = Popen(cmdline, stdin=PIPE, stdout=PIPE, stderr=PIPE)
```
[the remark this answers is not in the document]
vtz: Fixed in fd842df2. Replaced `PIPE` with `DEVNULL` for stdin/stdout/stderr.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,229 @@
...
+        for cmd in self.parent.extra_commands:
+            await self._monitor.execute(cmd)
+
+        if self.parent._firmware_path:
+            load_cmd = self.parent._load_command or "sysbus LoadELF"
+            await self._monitor.execute(
+                f'{load_cmd} @"{self.parent._firmware_path}"'
+            )
+
+        await self._monitor.execute("start")
```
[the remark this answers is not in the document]
vtz: Fixed in fd842df2. Wrapped monitor setup in try-except, calling `off()` on failure to clean up the subprocess.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,229 @@
...
+            except Exception:
+                pass
+            await self._monitor.disconnect()
+            self._monitor = None
+
+        self._process.terminate()
+        try:
+            self._process.wait(timeout=5)
+        except TimeoutExpired:
+            self._process.kill()
```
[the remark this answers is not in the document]
vtz: Fixed in fd842df2. Using `anyio.to_thread.run_sync` for the blocking `wait()` in `off()`. Note: `close()` remains synchronous as it's called during driver teardown, not from an async context.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
```diff
@@ -0,0 +1,129 @@
...
+    async def execute(self, command: str) -> str:
+        """Send a command and return the response text (excluding the prompt)."""
+        if self._stream is None:
+            raise RuntimeError("not connected to Renode monitor")
+
+        logger.debug("monitor> %s", command)
+        await self._stream.send(f"{command}\n".encode())
+        response = await self._read_until_prompt()
+        logger.debug("monitor< %s", response.strip())
+        return response
```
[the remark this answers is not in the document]
vtz: Fixed in fd842df2. `execute()` now inspects the response for known error markers and raises `RenodeMonitorError`. Updated test accordingly.

vtz:
@kirkbrauer ADR-0001 has been restructured to align with the JEP template format from #423 (metadata table, Abstract, Motivation, Rejected Alternatives, Prior Art, Implementation History, References, and license footer). The DD-N design decisions and Consequences sections were already in the right shape.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,555 @@
...
+        assert driver._load_command == "sysbus LoadBinary"
+
+    @pytest.mark.anyio
+    async def test_dump_not_implemented(self):
+        """dump() raises NotImplementedError."""
+        driver = _make_driver()
+        flasher: RenodeFlasher = driver.children["flasher"]
+
+        with pytest.raises(NotImplementedError, match="not supported"):
+            await flasher.dump("/dev/null")
```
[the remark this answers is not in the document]
vtz: Fixed in e19be083. Replaced the broken `AsyncMock(return_value=iter(...))` pattern with idiomatic `__aiter__`/`__anext__` using `StopAsyncIteration`. The old pattern silently yielded 0 chunks (with `RuntimeWarning: coroutine was never awaited`); the new one properly iterates. All 32 tests pass with 0 warnings now.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
...
+        assert driver.uart == "sysbus.usart2"
+        assert driver.machine_name == "my-machine"
+        assert driver.monitor_port == 9999
+        assert driver.extra_commands == ["command1", "command2"]
+
+    def test_renode_pty_path(self):
+        """PTY path is inside the temp directory."""
+        driver = _make_driver()
+        assert driver._pty.endswith("/pty")
+        assert driver._tmp_dir.name in driver._pty
```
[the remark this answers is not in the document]
vtz: Fixed in 51a01cf9. Using `Path(driver._pty) == Path(driver._tmp_dir.name) / "pty"` to mirror the driver implementation.

mangelajo:
@vtz can we convert the ADR into a JEP which we are trying to push at the community level?, there is no community decission to use ADRs, and that may get confusing to keep both in repo.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
...
+# 5.5 E2E test with serve()
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.skipif(
+    shutil.which("renode") is None,
+    reason="Renode not installed",
+)
+@pytest.mark.xfail(
+    platform.system() == "Darwin" and os.getenv("GITHUB_ACTIONS") == "true",
```
mangelajo: why?, I think this is just copying from other places.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
...
+        assert driver.get_machine_name() == "test-mcu"
+
+
+# ---------------------------------------------------------------------------
+# 5.5 E2E test with serve()
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.skipif(
+    shutil.which("renode") is None,
```
mangelajo, line 575: can we modify the .github workflows for python testing to install renode so this is rested in CI?

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+
+        cmd = load_command or "sysbus LoadELF"
+        self.parent._firmware_path = firmware_path
+        self.parent._load_command = cmd
+
+        if hasattr(self.parent.children["power"], "_process"):
+            monitor = self.parent.children["power"]._monitor
+            if monitor is not None:
+                await monitor.execute(f'{cmd} @"{firmware_path}"')
+                await monitor.execute("machine Reset")
```
raballew, line 95: [HIGH] The `load_command` parameter in `flash()` is client-supplied and interpolated directly into a monitor command without validation. A remote client can inject arbitrary monitor commands by passing something like `load_command='logFile @"/tmp/exfil"\nsysbus LoadELF'`. The injected value is also persisted in `self.parent._load_command` and replayed during subsequent `power.on()` calls.

Suggested fix: validate `load_command` against an allowlist of known Renode load commands (e.g., `sysbus LoadELF`, `sysbus LoadBinary`, `sysbus LoadSymbolsFrom`) and raise `ValueError` if it doesn't match.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+        """Return the Renode machine name."""
+        return self.machine_name
+
+    @export
+    async def monitor_cmd(self, command: str) -> str:
+        """Send an arbitrary command to the Renode monitor."""
+        power: RenodePower = self.children["power"]
+        if power._monitor is None:
+            raise RuntimeError("Renode is not running")
+        return await power._monitor.execute(command)
```
raballew, line 284: [HIGH] `monitor_cmd` forwards arbitrary strings to the Renode telnet monitor without any filtering. The Renode monitor supports commands that interact with the host filesystem (`logFile`, `include`, `CreateFileTerminal`, `LoadPlatformDescription`), so any authenticated Jumpstarter client can use these to read or write files on the host. For comparison, the existing QEMU driver does not expose an unrestricted QMP passthrough.

Suggested fix: implement an allowlist of permitted monitor commands, or gate this export behind an explicit `allow_raw_monitor: true` configuration flag that defaults to `false`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
```diff
@@ -0,0 +1,139 @@
...
+        with fail_after(timeout):
+            while True:
+                try:
+                    self._stream = await connect_tcp(host, port)
+                    self._buffer = b""
+                    await self._read_until_prompt()
+                    logger.info("connected to Renode monitor at %s:%d", host, port)
+                    return
+                except OSError:
+                    await sleep(0.5)
```
raballew, line 53: [HIGH] TCP socket streams leak on retry iterations in `connect()`. When `connect_tcp` succeeds but `_read_until_prompt()` subsequently fails with an `OSError`, the except clause catches it and the loop retries. On the next iteration, `self._stream = await connect_tcp(...)` overwrites the reference to the previous stream without closing it. This is a realistic scenario during Renode startup when the telnet port accepts connections before the monitor is ready.

Suggested fix: close the existing stream before retrying:
```python
except OSError:
    if self._stream is not None:
        await self._stream.aclose()
        self._stream = None
    await sleep(0.5)
```

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
```diff
@@ -0,0 +1,139 @@
...
+        logger.debug("monitor> %s", command)
+        await self._stream.send(f"{command}\n".encode())
+        response = await self._read_until_prompt()
+        logger.debug("monitor< %s", response.strip())
+
+        stripped = response.strip()
+        if stripped and any(stripped.startswith(m) for m in self._ERROR_MARKERS):
+            raise RenodeMonitorError(stripped)
+
+        return response
```
raballew: [HIGH] `execute()` has no timeout and can block an async task indefinitely. It calls `_read_until_prompt()` which runs an unbounded `while True` loop. If Renode stops sending data but keeps the TCP connection open (process hang, simulation deadlock), `self._stream.receive(4096)` blocks forever. Unlike `connect()` which wraps in `fail_after(timeout)`, `execute()` has no cancellation scope.

Suggested fix: wrap the read in a configurable timeout:
```python
async def execute(self, command: str, timeout: float = 30) -> str:
    ...
    with fail_after(timeout):
        response = await self._read_until_prompt()
    ...
```

python/packages/jumpstarter-driver-renode/pyproject.toml
```diff
@@ -0,0 +1,59 @@
...
+]
+requires-python = ">=3.11"
+dependencies = [
+    "jumpstarter",
+    "jumpstarter-driver-composite",
+    "jumpstarter-driver-network",
+    "jumpstarter-driver-opendal",
+    "jumpstarter-driver-power",
+    "jumpstarter-driver-pyserial",
+]
```
raballew, line 18: [HIGH] Missing workspace source entry in the top-level `python/pyproject.toml`. Every driver package in the workspace has an explicit entry in `[tool.uv.sources]` (e.g., `jumpstarter-driver-qemu = { workspace = true }`), but the new `jumpstarter-driver-renode` package has no such entry. Without it, other workspace packages listing `jumpstarter-driver-renode` as a dependency won't resolve to the local workspace copy during development, causing build failures since the package doesn't exist on PyPI yet.

Suggested fix: add `jumpstarter-driver-renode = { workspace = true }` to `[tool.uv.sources]` in `python/pyproject.toml`, alphabetically after `jumpstarter-driver-qemu`.

python/packages/jumpstarter-driver-renode/pyproject.toml
```diff
@@ -0,0 +1,59 @@
+[project]
+name = "jumpstarter-driver-renode"
+dynamic = ["version", "urls"]
+description = "Renode emulation framework driver for Jumpstarter"
```
raballew, line 4: [HIGH] The `jumpstarter-all` meta-package lists every driver as a dependency, but the new renode driver is missing from `python/packages/jumpstarter-all/pyproject.toml`. Users installing `jumpstarter-all` won't get the renode driver, which breaks the expected "install everything" contract.

Suggested fix: add `"jumpstarter-driver-renode"` to the `dependencies` list in `python/packages/jumpstarter-all/pyproject.toml`, after `jumpstarter-driver-qemu`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+        loading during power-on. If already running, loads the firmware
+        and resets the machine.
+        """
+        firmware_path = self.parent._tmp_dir.name + "/firmware"
+        async with await FileWriteStream.from_path(firmware_path) as stream:
+            async with self.resource(source) as res:
+                async for chunk in res:
+                    await stream.send(chunk)
+
+        cmd = load_command or "sysbus LoadELF"
```
raballew: [MEDIUM] `flash()` always defaults to `sysbus LoadELF` regardless of input file type. A user passing a raw binary firmware without specifying `load_command` will get a `RenodeMonitorError` because `LoadELF` will fail on non-ELF content.

Suggested fix: detect file format from ELF magic bytes (`\x7fELF`) at the start of the written firmware file and default to `sysbus LoadELF` when magic bytes match, `sysbus LoadBinary` otherwise.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
```diff
@@ -0,0 +1,139 @@
...
+        stripped = line.strip()
+        if not stripped:
+            return False
+        if stripped.startswith(MACHINE_PROMPT_PREFIX) and stripped.endswith(
+            MACHINE_PROMPT_SUFFIX
+        ):
+            inner = stripped[1:-1]
+            if inner and b")" not in inner:
+                return True
+        return False
```
raballew: [MEDIUM] The prompt detector `_is_prompt` matches any line fitting `(text-without-closing-paren)`, which also matches legitimate output like `(enabled)`, `(default)`, or `(none)`. A false positive causes early return from `_read_until_prompt`, truncating the response and corrupting subsequent command-response pairing.

Suggested fix: constrain matching to known prompt patterns. Track the expected machine name and validate against it, rather than accepting any parenthesized word.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
```diff
@@ -0,0 +1,139 @@
...
+            raise RuntimeError("not connected to Renode monitor")
+
+        logger.debug("monitor> %s", command)
+        await self._stream.send(f"{command}\n".encode())
+        response = await self._read_until_prompt()
+        logger.debug("monitor< %s", response.strip())
+
+        stripped = response.strip()
+        if stripped and any(stripped.startswith(m) for m in self._ERROR_MARKERS):
+            raise RenodeMonitorError(stripped)
```
raballew: [MEDIUM] Error detection only checks the start of the entire response text via `stripped.startswith(m)`. In a multi-line response where informational output comes before an error, the error marker is buried after the first line and `startswith` doesn't match. The error is silently returned as a normal string.

Suggested fix: check each line of the response individually:
```python
for line in stripped.splitlines():
    if any(line.startswith(m) for m in self._ERROR_MARKERS):
        raise RenodeMonitorError(stripped)
```

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
```diff
@@ -0,0 +1,139 @@
...
+
+        Raises RenodeMonitorError if the response indicates a command failure.
+        """
+        if self._stream is None:
+            raise RuntimeError("not connected to Renode monitor")
+
+        logger.debug("monitor> %s", command)
+        await self._stream.send(f"{command}\n".encode())
+        response = await self._read_until_prompt()
+        logger.debug("monitor< %s", response.strip())
```
raballew: [MEDIUM] `execute()` does not reject newline characters in the command string. Since the command is sent with an appended newline over telnet, embedded newlines cause each line to be interpreted as a separate Renode command. The method only reads the response to the first command's prompt, leaving the remaining responses in the buffer. This is the root mechanism enabling the command injection issues.

Suggested fix: reject newline characters as defense-in-depth:
```python
if "\n" in command or "\r" in command:
    raise ValueError("monitor commands must not contain newline characters")
```

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+        if self._process is not None:
+            if self._monitor is not None:
+                self._monitor = None
+            self._process.terminate()
+            try:
+                self._process.wait(timeout=5)
+            except TimeoutExpired:
+                self._process.kill()
+            self._process = None
+
```
raballew, line 207: [MEDIUM] `close()` sets `self._monitor = None` without calling `disconnect()` on the monitor, leaking the underlying `SocketStream`. By contrast, `off()` properly calls `await self._monitor.execute("quit")` and `await self._monitor.disconnect()`. The two methods implement the same cleanup logic independently, with `close()` missing critical steps.

Suggested fix: close the underlying socket before dropping the reference, or restructure so shared cleanup lives in a single helper that both methods call.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+        cmd = load_command or "sysbus LoadELF"
+        self.parent._firmware_path = firmware_path
+        self.parent._load_command = cmd
+
+        if hasattr(self.parent.children["power"], "_process"):
+            monitor = self.parent.children["power"]._monitor
+            if monitor is not None:
+                await monitor.execute(f'{cmd} @"{firmware_path}"')
+                await monitor.execute("machine Reset")
+                self.logger.info("firmware hot-loaded and machine reset")
```
raballew, line 96: [MEDIUM] `RenodeFlasher` navigates `self.parent.children["power"]._process` and `self.parent.children["power"]._monitor`, traversing two levels of indirection to access private attributes of a sibling driver. The same violation occurs in `Renode.monitor_cmd()` at lines 242-245.

Suggested fix: add a public method to `RenodePower` (e.g., `send_monitor_command(command)`) that encapsulates the monitor interaction, and call that instead of reaching into private state.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+        """
+        firmware_path = self.parent._tmp_dir.name + "/firmware"
+        async with await FileWriteStream.from_path(firmware_path) as stream:
+            async with self.resource(source) as res:
+                async for chunk in res:
+                    await stream.send(chunk)
+
+        cmd = load_command or "sysbus LoadELF"
+        self.parent._firmware_path = firmware_path
+        self.parent._load_command = cmd
```
raballew, line 89: [MEDIUM] `RenodeFlasher.flash()` writes to `self.parent._firmware_path` and `self.parent._load_command`, while `RenodePower.on()` reads them. This communication between siblings through shared mutable parent state is implicit with no interface contract, no type annotation, and no structural enforcement of the ordering dependency.

Suggested fix: make the firmware loading concern explicit. Have `RenodePower.on()` accept firmware parameters, or have `Renode` expose a method that coordinates the interaction.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
+from __future__ import annotations
```
raballew, line 1: [MEDIUM] Several test gaps found:

1. No test for process cleanup when `power.on()` fails mid-startup. `driver.py:135-137` has an explicit error-cleanup path (`except Exception: await self.off(); raise`) that prevents orphaned Renode processes, but those lines are not exercised. Add a test that mocks `_monitor.execute` to raise on a mid-sequence command and assert `mock_process.terminate()` was called.

2. Missing tests for monitor error paths: (a) `monitor.py:32` uses `fail_after(timeout)` which raises `TimeoutError`, but no test covers this; (b) `monitor.py:96` raises `ConnectionError` when `stream.receive()` returns empty bytes, also uncovered. Add a test where `connect_tcp` always raises `OSError` and verify `TimeoutError`, and a test where the stream returns `b""` to verify `ConnectionError`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
...
+        assert power._process is None
+
+    @pytest.mark.anyio
+    async def test_power_off_idempotent(self):
+        """Second off() call logs warning and does nothing."""
+        driver = _make_driver()
+        power: RenodePower = driver.children["power"]
+        power._process = None
+
+        await power.off()
```
raballew, line 367: [MEDIUM] `test_power_on_idempotent` and `test_power_off_idempotent` contain zero assertions. They only verify "doesn't crash", not "does nothing." They would pass even if `on()` silently spawned a second process or `off()` leaked resources.

Suggested fix: patch `Popen`/`RenodeMonitor` and assert they were not called for the idempotent case. For `test_power_off_idempotent`, assert `_process` remains `None`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+    @export
+    async def flash(self, source, load_command: str | None = None):
+        """Flash firmware to the simulated MCU.
+
+        If the simulation is not yet running, stores the firmware for
+        loading during power-on. If already running, loads the firmware
+        and resets the machine.
+        """
+        firmware_path = self.parent._tmp_dir.name + "/firmware"
+        async with await FileWriteStream.from_path(firmware_path) as stream:
```
raballew, line 79: [LOW] When the simulation is not running, `flash()` silently stores firmware for deferred loading during `power.on()` instead of raising an error. This is arguably better UX and consistent with the QEMU driver pattern, but worth noting the design choice.

Suggested fix: if this is the intended design, consider documenting it more explicitly.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+            async with self.resource(source) as res:
+                async for chunk in res:
+                    await stream.send(chunk)
+
+        cmd = load_command or "sysbus LoadELF"
+        self.parent._firmware_path = firmware_path
+        self.parent._load_command = cmd
+
+        if hasattr(self.parent.children["power"], "_process"):
```
raballew, line 91: [LOW] The `hasattr` guard `hasattr(self.parent.children["power"], "_process")` is always `True` because `RenodePower._process` is declared as `field(init=False, default=None)`, so the attribute always exists. This is dead code; the real guard is the nested `if monitor is not None` on line 64.

Suggested fix: replace `hasattr(...)` with `power._process is not None and power._monitor is not None`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+                pass
+            await self._monitor.disconnect()
+            self._monitor = None
+
+        self._process.terminate()
+        try:
+            await to_thread.run_sync(self._process.wait, 5)
+        except TimeoutExpired:
+            self._process.kill()
+        self._process = None
```
raballew, line 189: [LOW] `off()` leaves inconsistent state if `terminate()` raises on an already-exited process. `ProcessLookupError` propagates without executing `self._process = None`, leaving the driver in a state where subsequent `off()` calls fail and `on()` short-circuits.

Suggested fix: wrap the terminate/wait/kill sequence in try/finally to always set `self._process = None`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+from .monitor import RenodeMonitor
+from jumpstarter.driver import Driver, export
+
+logger = logging.getLogger(__name__)
+
+
+def _find_free_port() -> int:
+    with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
+        s.bind(("127.0.0.1", 0))
+        return s.getsockname()[1]
```
raballew, line 47: [LOW] TOCTOU race in `_find_free_port()`: the ephemeral port is discovered by binding and immediately releasing a socket. A local process could race to bind the same port before Renode does. The QEMU driver avoids this by using Unix domain sockets.

Suggested fix: switch to a Unix domain socket for the monitor connection if Renode supports it, or use `--port 0` if Renode supports reporting its actual bound port.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,245 @@
...
+                load_cmd = self.parent._load_command or "sysbus LoadELF"
+                await self._monitor.execute(
+                    f'{load_cmd} @"{self.parent._firmware_path}"'
+                )
+
+            await self._monitor.execute("start")
+            self.logger.info("Renode simulation started")
+        except Exception:
+            await self.off()
+            raise
```
raballew, line 167: [LOW] `RenodePower.on()` handles multiple responsibilities in ~53 lines: idempotency check, binary location, port allocation, command-line construction, process spawn, monitor connection, machine creation, platform loading, UART setup, extra commands, firmware loading, simulation start, and error cleanup.

Suggested fix: extract the monitor setup sequence (lines 110-133) into a private method like `_configure_simulation()` to improve readability.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
...
+
+
+@pytest.fixture
+def anyio_backend():
+    return "asyncio"
+
+
+# ---------------------------------------------------------------------------
+# 5.1 RenodeMonitor unit tests
+# ---------------------------------------------------------------------------
```
raballew, line 28: [LOW] Six decorative separator comments reference internal plan task numbers (5.1-5.6) and restate what the immediately-following class name already says (e.g., `# 5.1 RenodeMonitor unit tests` followed by `class TestRenodeMonitor:`). These add noise without information.

Suggested fix: remove all six separator blocks. The class and function names already convey the grouping.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
```diff
@@ -0,0 +1,562 @@
+from __future__ import annotations
+
```
raballew, line 2: [LOW] Several minor paths lack test coverage: (1) `_find_renode()` error path when binary not on PATH; (2) `RenodePower.read()` raising `NotImplementedError`; (3) `Renode.monitor_cmd()` success path; (4) `close()` timeout-kill branch; (5) E2E test only exercises start/stop, not flash or console interaction.

Suggested fix: add targeted tests for each: (1) mock `shutil.which` returning `None`; (2) parallel to `test_dump_not_implemented`; (3) mock monitor and verify forwarding; (4) mock `wait` raising `TimeoutExpired`; (5) extend E2E with flash and `monitor_cmd` when Renode is available.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
raballew: Not 100% sure on this one. This might be a false positive but I am keeping it for your reference.

raballew (dismissed):

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
vtz: You're right — this was copied from the QEMU test pattern. Removed in c311ab6c.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver_test.py
vtz: Done in c311ab6c. Added Renode installation steps for both Linux (deb) and macOS (brew) to the python-tests workflow.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
vtz: Fixed in c311ab6c. `load_command` is now validated against an allowlist of known Renode load commands (`sysbus LoadELF`, `sysbus LoadBinary`, `sysbus LoadSymbolsFrom`). Invalid values raise `ValueError`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
vtz: Fixed in c311ab6c. `monitor_cmd` is now gated behind `allow_raw_monitor` (defaults to `false`). Clients get a clear error unless the exporter config explicitly opts in.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
vtz: Fixed in c311ab6c. The existing stream is now closed before retrying in `connect()`.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
vtz: Fixed in c311ab6c. `execute()` now wraps the read in `fail_after(timeout)` with a default of 30 seconds.

python/packages/jumpstarter-driver-renode/pyproject.toml
vtz: Fixed in c311ab6c. Added `jumpstarter-driver-renode = { workspace = true }` to `python/pyproject.toml`.

python/packages/jumpstarter-driver-renode/pyproject.toml
vtz: Fixed in c311ab6c. Added `jumpstarter-driver-renode` to the `jumpstarter-all` dependencies.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
vtz: Fixed in c311ab6c. `flash()` now auto-detects ELF files via magic bytes (`\x7fELF`): defaults to `sysbus LoadELF` for ELF files, `sysbus LoadBinary` otherwise.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
vtz: Fixed in c311ab6c. Prompt matching now checks against a set of registered prompt names (`monitor` by default, machine name added after `mach create`). Words like `(enabled)` or `(default)` no longer match.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
vtz: Fixed in c311ab6c. Error detection now checks each line individually, catching errors even when preceded by informational output.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/monitor.py
vtz: Fixed in c311ab6c. `execute()` now rejects commands containing `\n` or `\r` with a `ValueError`.

.github/workflows/python-tests.yaml
```diff
@@ -73,11 +73,22 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libgpiod-dev liblgpio-dev
 
+      - name: Install Renode (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          wget https://builds.renode.io/renode_1.15.3+20241210git19e40b562_amd64.deb -O /tmp/renode.deb
+          sudo apt-get install -y /tmp/renode.deb
+
```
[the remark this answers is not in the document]
mangelajo: ok, this is CI testing only, for embedding renode in our jumpstarter container we may want to be more careful, but this is okish

docs/internal/jeps/JEP-0010-renode-integration.md
```diff
@@ -0,0 +1,380 @@
...
+- **`RenodePower`** — controls the Renode process lifecycle (on/off)
+- **`RenodeFlasher`** — handles firmware loading (`sysbus LoadELF` /
+  `sysbus LoadBinary`)
+- **`PySerial` (console)** — serial access over a PTY terminal
+
+Users configure targets entirely through exporter YAML:
+
+```yaml
+export:
+  Renode:
```
mangelajo, line 93:
```suggestion
  renode:
```

docs/internal/jeps/JEP-0010-renode-integration.md
```diff
@@ -0,0 +1,380 @@
...
+2. `on()` starts the simulation, `flash()` loads firmware and resets
+
+**Decision:** Option 1 — `flash()` stores the path, `on()` loads and
+starts.
+
+**Rationale:** This matches the QEMU driver's semantic where you flash
+a disk image first, then power on. It also allows re-flashing between
+power cycles without restarting the Renode process. The `RenodeFlasher`
+additionally supports hot-loading: if the simulation is already running,
+`flash()` sends the load command and resets the machine.
```
mangelajo, line 208: not to be fixed in this PR: yep, also has the limitation of not being able to open the serial port until it's started. But I am thinking of options for qemu, and then this. We can just make the open wait on the ptty to show up.

docs/internal/jeps/JEP-0010-renode-integration.md
```diff
@@ -0,0 +1,380 @@
...
+
+**Decision:** Opt-in with `allow_raw_monitor: false` by default.
+
+**Rationale:** The Renode monitor supports commands that interact with
+the host filesystem (`logFile`, `include`, `CreateFileTerminal`). In a
+shared lab environment, unrestricted monitor access from any
+authenticated client poses a risk. The `load_command` parameter in
+`flash()` is separately validated against an allowlist of known Renode
+load commands. Newline characters are rejected in all monitor commands
+to prevent command injection.
```
mangelajo, line 225: 👍

docs/internal/jeps/JEP-0010-renode-integration.md
```diff
@@ -0,0 +1,380 @@
...
+## Implementation History
+
+- 2026-04-06: JEP proposed
+- 2026-04-09: Initial implementation merged
+  ([PR #533](https://github.com/jumpstarter-dev/jumpstarter/pull/533))
+- 2026-04-11: Address review feedback (DEVNULL, try-except cleanup,
+  async wait, RenodeMonitorError, multi-word CLI, docstrings)
+- 2026-04-15: Security hardening (load_command allowlist,
+  allow_raw_monitor, newline rejection, per-line error detection,
+  prompt matching, execute timeout, stream leak fix)
```
mangelajo:
```suggestion
- 2026-04-06 JEP proposed
- 2026-04-06: Initial implementation proposed
  ([PR #533](https://github.com/jumpstarter-dev/jumpstarter/pull/533))
- 2026-04-11: Address review feedback (DEVNULL, try-except cleanup,
  async wait, RenodeMonitorError, multi-word CLI, docstrings)
- 2026-04-15: Security hardening (load_command allowlist,
  allow_raw_monitor, newline rejection, per-line error detection,
  prompt matching, execute timeout, stream leak fix)
```

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
mangelajo: This is true, let's follow up on this, but not on this PR.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
mangelajo: IT is noted in the JEP Design decissions I believe, if not let's add it.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
mangelajo: I think it's ok, it's a self-contained driver in a single file. The on() interface should not be modified, as it's just a standard PowerInterface from jumpstarter. We could add a set_firmware_path or set_load_command to avoid looking for private methods, but IMHO it's ok since this is all the same python module boundary.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
mangelajo: I think it's checking that the power interfaces is what it is supposed to be, but it would be better to do type checking instead of looking for methods. Also adding a (non exported) monitor_execute method

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
mangelajo: Yep, I agree with this.

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
mangelajo: it's passing our complexity checks anyway :D

python/packages/jumpstarter-driver-renode/jumpstarter_driver_renode/driver.py
```diff
@@ -0,0 +1,284 @@
...
+        """Synchronous cleanup for use during driver teardown."""
+        if self._process is not None:
+            if self._monitor is not None:
+                self._monitor = None
+            self._process.terminate()
+            try:
+                self._process.wait(timeout=5)
+            except TimeoutExpired:
+                self._process.kill()
+            self._process = None
```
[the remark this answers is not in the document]
mangelajo: yes, you are agreeing with @raballew comment's up above.

mangelajo:
This is in a good mergeable state, I will do that and follow up with the collected feedback here https://github.com/jumpstarter-dev/jumpstarter/issues/556 myself

mangelajo (approved):
with collected feedback in https://github.com/jumpstarter-dev/jumpstarter/issues/556 to be handled.

.github/workflows/python-tests.yaml
```diff
@@ -73,11 +73,22 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libgpiod-dev liblgpio-dev
 
+      - name: Install Renode (Linux)
+        if: runner.os == 'Linux'
+        run: |
+          wget https://builds.renode.io/renode_1.15.3+20241210git19e40b562_amd64.deb -O /tmp/renode.deb
```
mangelajo:
```suggestion
          wget https://github.com/renode/renode/releases/download/v1.16.1/renode_1.16.1_amd64.deb -O /tmp/renode.deb
```

The other git build is not available anymore, I hope this one works.

.github/workflows/python-tests.yaml
```diff
@@ -73,11 +73,22 @@ jobs:
...
+
       - name: Install Qemu (macOS)
         if: runner.os == 'macOS'
         run: |
           brew install qemu
 
+      - name: Install Renode (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          brew install renode
```
mangelajo:
```suggestion
          brew install renode/tap/renode
```

according to renode instructions

.github/workflows/python-tests.yaml
```diff
@@ -73,11 +73,22 @@ jobs:
...
       - name: Install Qemu (macOS)
         if: runner.os == 'macOS'
         run: |
           brew install qemu
 
+      - name: Install Renode (macOS)
+        if: runner.os == 'macOS'
+        run: |
+          brew install renode/tap/renode
+
```
mangelajo, line 91:
```suggestion
# NOTE: not testing renode on macOS, as the install rebuilds renode and it is too slow.
#      - name: Install Renode (macOS)
#        if: runner.os == 'macOS'
#        run: |
#          brew install renode/tap/renode

```

PR #539, opened 2026-04-09
ci: parallelize e2e container image and wheel builds

mangelajo:
## Summary

- **Parallel build jobs**: Controller image, operator image, and Python wheels are now built in separate parallel jobs instead of being rebuilt in every test job
- **Artifact sharing**: Build outputs are saved as GitHub Actions artifacts (container tarballs, wheel files) and downloaded by test jobs that need them
- **Caching**: Each build job uses `actions/cache` keyed on source file hashes, so unchanged artifacts are reused across workflow runs

- **Pre-compiled/hashed binaries**: Instead of compiling yq/csrf/etc, we try to download the existing binaries first, keeping a hash of the version binaries to avoid supply chain issues.

## Architecture

```
changes
├── build-controller-image  (matrix: amd64, arm64)  ──┐
├── build-operator-image    (matrix: amd64, arm64)  ──┼── e2e-tests (matrix: amd64, arm64)
├── build-python-wheels     (single job)            ──┤
│                                                     ├── e2e-compat-old-controller (no images needed)
│                                                     └── e2e-compat-old-client (controller image only)
```

### Dependency optimization
- `e2e-compat-old-controller` only waits for `build-python-wheels` (uses quay.io for controller)
- `e2e-compat-old-client` only waits for `build-controller-image` + `build-python-wheels` (no operator needed)

### Supporting changes
- `controller/Makefile`: New `SKIP_BUILD` flag to skip `docker-build` and `build-operator` in the `deploy` target
- `e2e/setup-e2e.sh`: New `PREBUILT_WHEELS_DIR` env var to install from pre-built wheels instead of `make sync`
- `e2e/compat/setup.sh`: `SKIP_BUILD` support in `deploy_new_controller()` + `PREBUILT_WHEELS_DIR` in `install_jumpstarter()`

All changes are backward-compatible — without env vars set, behavior is identical to before.

## Test plan

- [x] Verify `build-controller-image` job produces valid tarballs for both architectures
- [x] Verify `build-operator-image` job produces valid tarballs for both architectures
- [x] Verify `build-python-wheels` produces installable wheels
- [x] Verify `e2e-tests` correctly loads images and passes
- [x] Verify `e2e-compat-old-controller` passes with pre-built wheels
- [x] Verify `e2e-compat-old-client` passes with pre-built controller image + wheels
- [x] Verify caching works on second run (cache hit skips builds)
- [x] Verify local dev workflow still works (`make e2e-setup && make e2e-run` without env vars)


.github/workflows/e2e.yaml
```diff
@@ -33,14 +36,147 @@ jobs:
...
+          name: python-wheels
+          path: python/dist/
+          retention-days: 1
+
+  # ===========================================================================
+  # E2E test jobs: depend on pre-built artifacts
+  # ===========================================================================
+
+  e2e-tests:
+    needs: [build-controller-image, build-operator-image, build-python-wheels]
```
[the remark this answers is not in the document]
mangelajo: I would avoid making it more complicated, it's fine

e2e/setup-e2e.sh
```diff
@@ -274,31 +271,54 @@ deploy_controller() {
...
+        cd python
+        uv venv .venv --python 3.12
+        uv pip install --python .venv/bin/python "${PREBUILT_WHEELS_DIR}"/*.whl
+        cd ..
+    else
+        cd python
+        make sync
+        cd ..
+    fi
     log_info "✓ Jumpstarter python installed"
```
raballew: [MEDIUM] The `install_jumpstarter` function here is nearly identical to the one in `e2e/compat/setup.sh` (lines 166-193). Both implement the same PREBUILT_WHEELS_DIR validation, wheel counting, venv creation, pip install, and `make sync` fallback. Only the log messages differ slightly. This kind of duplication risks drift where a bug fix in one copy gets missed in the other.

Consider extracting the shared logic into a sourced helper (e.g. `e2e/lib/install.sh`) and sourcing it from both scripts.

e2e/setup-e2e.sh
```diff
@@ -143,58 +54,143 @@ install_dependencies() {
...
+        cfssl_darwin_arm64)     echo "9a38b997ac23bc2eed89d6ad79ea5ae27c29710f66fdabdff2aa16eaaadc30d4" ;;
+        cfssljson_linux_amd64)  echo "09fbcb7a3b3d6394936ea61eabff1e8a59a8ac3b528deeb14cf66cdbbe9a534f" ;;
+        cfssljson_linux_arm64)  echo "a389793bc2376116fe2fff996b4a2f772a59a4f65048a5cfb4789b2c0ea4a7c9" ;;
+        cfssljson_darwin_amd64) echo "1529a7a163801be8cf7d7a347b0346cc56cc8f351dbc0131373b6fb76bb4ab64" ;;
+        yq_linux_amd64)         echo "75d893a0d5940d1019cb7cdc60001d9e876623852c31cfc6267047bc31149fa9" ;;
+        yq_linux_arm64)         echo "90fa510c50ee8ca75544dbfffed10c88ed59b36834df35916520cddc623d9aaa" ;;
+        yq_darwin_amd64)        echo "6e399d1eb466860c3202d231727197fdce055888c5c7bec6964156983dd1559d" ;;
+        yq_darwin_arm64)        echo "45a12e64d4bd8a31c72ee1b889e81f1b1110e801baad3d6f030c111db0068de0" ;;
+        *) echo "" ;;
+    esac
```
raballew, line 81: [MEDIUM] The `get_expected_sha256()` function provides checksums for all platform/arch combinations of `cfssl` (4 entries), `yq` (4 entries), and `cfssljson` (3 entries), but is missing the `cfssljson_darwin_arm64` entry. This is the only gap across all three tools. On macOS ARM64, the missing hash causes the native arm64 download to silently fail (empty hash), triggering a Rosetta amd64 fallback. If Rosetta is unavailable, it falls back to `go install`, which bypasses checksum verification entirely.

Add the missing entry by obtaining the SHA256 from the cfssl v1.6.5 GitHub release for `cfssljson_1.6.5_darwin_arm64`:
```
cfssljson_darwin_arm64) echo "<sha256-from-release-page>" ;;
```

e2e/setup-e2e.sh
```diff
@@ -143,58 +54,143 @@ install_dependencies() {
...
-            brew install bats-core
-        else
-            log_error "bats not found. Please install it manually:"
-            log_error "  Ubuntu/Debian: sudo apt-get install bats"
-            log_error "  Fedora/RHEL: sudo dnf install bats"
-            log_error "  macOS: brew install bats-core"
-            exit 1
-        fi
+    log_info "✓ Dependencies installed"
+}
```
raballew, line 58: [MEDIUM] The bats infrastructure was removed from the setup scripts (good!), but four `.bats` test files still exist in the repo: `e2e/tests-hooks.bats`, `e2e/tests-direct-listener.bats`, `e2e/compat/tests-old-client.bats`, and `e2e/compat/tests-old-controller.bats`. Those files require `bats-support` and `bats-assert` which are no longer installed by any setup script, and the test runners (`run_ginkgo`) never invoke bats. They are effectively dead code now.

Consider removing the four orphaned `.bats` files as part of this cleanup.

e2e/setup-e2e.sh
```diff
@@ -143,58 +54,143 @@ install_dependencies() {
...
     
-    # Create dex namespace and TLS secret
+    # Create dex namespace and TLS secret (idempotent)
     log_info "Creating dex namespace and secrets..."
-    kubectl create namespace dex
+    kubectl create namespace dex --dry-run=client -o yaml | kubectl apply -f -
+    kubectl -n dex delete secret dex-tls --ignore-not-found
     kubectl -n dex create secret tls dex-tls \
         --cert=server.pem \
         --key=server-key.pem
```
raballew: [LOW] The `dex-tls` secret is recreated with a delete-then-create pattern. If `kubectl create` fails after the delete (e.g. transient API server error), the old secret is gone with no rollback. Other resources in this same PR already use the safer dry-run-and-apply pattern (see the namespace and service account lines).

For consistency and safety, you could use:
```
kubectl -n dex create secret tls dex-tls --cert=server.pem --key=server-key.pem --dry-run=client -o yaml | kubectl apply -f -
```

e2e/setup-e2e.sh
```diff
@@ -143,58 +54,143 @@ install_dependencies() {
...
+try_download_verified() {
+    local dest="$1" url="$2" hash_key="$3"
+    local expected_hash
+    expected_hash=$(get_expected_sha256 "$hash_key")
+    if [ -z "$expected_hash" ]; then
+        return 1
     fi
-    
-    log_info "✓ Dependencies installed"
+    if curl -fsSL -o "$dest" "$url" 2>/dev/null && verify_sha256 "$dest" "$expected_hash"; then
```
raballew: [LOW] `curl -fsSL -o "$dest" "$url" 2>/dev/null` discards all curl stderr output (connection errors, DNS failures, TLS errors, HTTP error messages). When all download fallbacks fail and the final `go install` also fails, the most informative diagnostic (the curl error) has been thrown away.

Consider removing `2>/dev/null` and letting `-fsSL` handle output control, or capturing curl stderr and logging it at warn level on failure.

raballew (changes requested):

e2e/setup-e2e.sh
mangelajo: this version does not exist, that's why we don't provide it, but we can add an invalid one just in case.
"does-not-exist" for example.

e2e/setup-e2e.sh
mangelajo: ooops yes!.

e2e/setup-e2e.sh
mangelajo: Fixed (by Cursor) — extracted the shared logic into `e2e/lib/install.sh` which is now sourced by both `e2e/setup-e2e.sh` and `e2e/compat/setup.sh`.

e2e/setup-e2e.sh
mangelajo: Fixed (by Cursor) — added a `cfssljson_darwin_arm64) echo "no-prebuilt-binary-available" ;;` entry to make it explicit that no binary exists for that platform, rather than silently falling through to an empty string.

e2e/setup-e2e.sh
mangelajo: Fixed (by Cursor) — deleted all 4 orphaned `.bats` files: `e2e/tests-hooks.bats`, `e2e/tests-direct-listener.bats`, `e2e/compat/tests-old-client.bats`, and `e2e/compat/tests-old-controller.bats`.

e2e/setup-e2e.sh
mangelajo: Fixed (by Cursor) — replaced the delete-then-create pattern with the safer `--dry-run=client -o yaml | kubectl apply -f -` approach, consistent with the other resources in the script.

e2e/setup-e2e.sh
mangelajo: Fixed (by Cursor) — removed `2>/dev/null` so curl errors are preserved for debugging when all download fallbacks fail.

.github/workflows/e2e.yaml
```diff
@@ -33,14 +36,147 @@ jobs:
...
+        if: steps.cache.outputs.cache-hit != 'true'
+        working-directory: python
+        run: uv build --all --out-dir dist
+
+      - name: Upload python wheels
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-wheels
+          path: python/dist/
+          retention-days: 1
```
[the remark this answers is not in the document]
mangelajo: All the wheels in this project are python-only, no C extensions so far.

raballew (approved):

PR #555, opened 2026-04-14
qemu: add OCI flashing to qemu driver

bennyz:


python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
```diff
@@ -1,8 +1,26 @@
...
+        if isinstance(path, str) and path.startswith("oci://"):
+            oci_username = os.environ.get("OCI_USERNAME")
+            oci_password = os.environ.get("OCI_PASSWORD")
+
+            if bool(oci_username) != bool(oci_password):
+                raise ValueError("OCI authentication requires both OCI_USERNAME and OCI_PASSWORD environment variables")
+
+            return self.call("flash_oci", path, target, oci_username, oci_password)
+
+        return super().flash(path, target=target, operator=operator, compression=compression)
```
mangelajo, line 25: if we do the bifurcation on the driver side it will be easier with a plan that @kirkbrauer was proposing to allow multi-language clients :)

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -61,6 +67,71 @@ async def flash(self, source, partition: str | None = None):
...
+        if bool(oci_username) != bool(oci_password):
+            raise ValueError("OCI authentication requires both username and password")
+
+        target_path = str(self.parent.validate_partition(partition))
+
+        fls_binary = get_fls_binary(
+            fls_version=self.parent.fls_version,
+            fls_binary_url=self.parent.fls_custom_binary_url,
+            allow_custom_binaries=self.parent.fls_allow_custom_binaries,
+        )
```
mangelajo, line 123: I would go ahead and start installing the fls binary on the jumpstarter container, I was exactly looking for using fls... :)

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -61,6 +67,71 @@ async def flash(self, source, partition: str | None = None):
...
+
+        fls_cmd = [fls_binary, "from-url", oci_url, target_path]
+
+        fls_env = os.environ.copy()
+        if oci_username and oci_password:
+            fls_env["FLS_REGISTRY_USERNAME"] = oci_username
+            fls_env["FLS_REGISTRY_PASSWORD"] = oci_password
```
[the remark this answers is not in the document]
mangelajo: This is irrelevant, in almost no case those env vars will be pre-set

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -61,6 +67,71 @@ async def flash(self, source, partition: str | None = None):
...
+        self.logger.info(f"Running fls: {' '.join(fls_cmd)}")
+
+        try:
+            result = subprocess.run(
+                fls_cmd,
+                capture_output=True,
+                text=True,
+                check=True,
+                env=fls_env,
+                timeout=self.parent.flash_timeout,
```
mangelajo: I would do that later, but we need to find a standard way to stream back the output of the shell process (we can steal this from the shell driver) :)

but not now, don't want to complicate your PR

mangelajo (commented):
niiiice , bennyz == 🦸‍♂️

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
bennyz: it is there :)
https://github.com/jumpstarter-dev/jumpstarter/blob/e8e97319fee4c60b3cf59a27223e8d44d6cb3eaa/python/Dockerfile#L15

this piece of code is for overriding

mangelajo: ahh sorry I didn't see it :D

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
bennyz: added 😄

mangelajo (approved):

PR #557, opened 2026-04-15
docs/ci: Renode driver listing and conditional macOS Renode install

mangelajo:
### Documentation

- Register **Renode** (`jumpstarter-driver-renode`) in the drivers reference index and the hidden `toctree`, so `renode.md` (symlinked from the package README) is part of the Sphinx tree and the “document isn’t included in any toctree” warning goes away.
- Remove the **Design Decisions** paragraph that pointed at the old ADR path (`contributing/adr/0001-renode-integration.md`), which no longer exists and caused MyST xref failures under `make docs` (`-W`).

No change to the in-repo JEP under `docs/internal/jeps/`; we can add a link again once the team agrees on how to reference it from published docs.

### CI (Python Tests workflow)

- **macOS:** `brew install renode/tap/renode` is slow (~5+ minutes per job). It now runs only when the change set touches `python/packages/jumpstarter-driver-renode/**` (via `paths-filter`), or on **`workflow_dispatch`** so manual runs still get full coverage.
- **Linux:** unchanged; the `.deb` install remains for every matrix job (fast).

## Testing

- `make docs` (from `python/`)

### DIFF
--- a/python/packages/jumpstarter-driver-renode/README.md
+++ b/python/packages/jumpstarter-driver-renode/README.md
@@ -112,9 +112,6 @@ The `monitor` CLI subcommand is also available inside a `jmp shell` session.
 
 ## Design Decisions
 
-The architectural choices for this driver are documented in
-[ADR-0001: Renode Integration Approach](../../docs/source/contributing/adr/0001-renode-integration.md).
-
 Key decisions:
 
 - **Control interface**: Telnet monitor via `anyio.connect_tcp` (no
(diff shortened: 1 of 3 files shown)


python/packages/jumpstarter-driver-renode/README.md
```diff
@@ -112,9 +112,6 @@ The `monitor` CLI subcommand is also available inside a `jmp shell` session.
 
 ## Design Decisions
 
-The architectural choices for this driver are documented in
-[ADR-0001: Renode Integration Approach](../../docs/source/contributing/adr/0001-renode-integration.md).
-
```
mangelajo, line 117: the doc is in the internal directory and can't be grabbed by sphinx

bennyz (approved):

PR #564, opened 2026-04-15
DO NOT MERGE fix: per-invocation Listen channels to eliminate dial token loss on reconnect

### LINKED ISSUE #414: race condition in listenQueues cleanup causes intermittent 'Connection to exporter lost'
## Summary

PR #397 introduced a race condition in `controller/internal/service/controller_service.go` that intermittently causes E2E test failures with `Error: Connection to exporter lost` (tests 47 & 48).

## Root Cause

PR #397 added `defer s.cleanupListenQueue(leaseName)` to the `Listen()` gRPC handler (exporter side). This creates a race with `Dial()` (client side):

1. Exporter's `Listen()` stream fails transiently → `defer cleanupListenQueue` fires and **deletes** the queue from `listenQueues`
2. Simultaneously (or just before), `Dial()` does `LoadOrStore` → finds the old queue still in the map → **sends the router token into it**
3. `cleanupListenQueue` deletes the queue → **token is discarded**
4. Exporter reconnects, `Listen()` creates a fresh empty queue → **never receives the router address**
5. Client waits 20 seconds for a connection → `"Connection to exporter lost"`

PR #396 compounds this by resetting the exporter's retry counter after any successful data exchange, causing more aggressive reconnects and widening the race window.

## Observed Failure

```
INFO     Waiting for ready connection at /run/user/1001/jumpstarter-xxx/socket
# (20 seconds later)
INFO     Releasing Lease 019d65c4-f2c3-7b9e-a626-6f259bf280c8
Error: Connection to exporter lost
```

Example CI run: https://github.com/jumpstarter-dev/jumpstarter/actions/runs/24060390906/job/70175302716?pr=254

## Proposed Fix

In `Listen()`, instead of unconditionally deleting the queue on any exit, only delete it when the exporter is intentionally deregistering (i.e. lease released / context cancelled cleanly). On stream errors, the queue should be left intact so a reconnecting exporter can still receive the pending router token.

Alternatively, use `CompareAndDelete` to only delete the specific channel instance created by the current `Listen` invocation, preventing a reconnected `Listen` from having its newly-created queue deleted.

```go
// Instead of:
defer s.cleanupListenQueue(leaseName)

// Use CompareAndDelete to only delete the queue this invocation created:
defer func() {
    s.listenQueues.CompareAndDelete(leaseName, queue)
}()
```

## Affected Files

- `controller/internal/service/controller_service.go` — `Listen()` and `Dial()` handlers, `cleanupListenQueue`
- `python/packages/jumpstarter/jumpstarter/exporter/exporter.py` — retry reset logic (PR #396)

raballew:
DO NOT MERGE YET - I NEED CI TO CHECK FOR FLAKINESS

## Summary

Fixes https://github.com/jumpstarter-dev/jumpstarter/issues/414

### Root cause

The flaky `can lease and connect to exporters` E2E test fails because **two concurrent `Listen()` goroutines read from the same channel**, causing Dial tokens to be consumed by the stale goroutine and silently discarded.

The sequence:
1. Exporter calls `Listen(lease)` -- goroutine G1 blocks on `<-ch` in the select loop
2. The exporter's gRPC stream breaks (network hiccup)
3. G1's context cancellation hasn't propagated yet -- G1 is still in its `select`, reading from `ch`
4. Exporter reconnects, calls `Listen(lease)` again -- `LoadOrStore` returns the **same channel** -- goroutine G2 also reads from `ch`
5. Client calls `Dial(lease)` -- sends router token into `ch`
6. G1 (stale) wins the channel read. It calls `stream.Send(token)` which **fails** because the stream is broken. **Token is consumed and discarded.**
7. G2 (reconnected) never receives the token -- client times out -- `Connection to exporter lost`

This is a **logical race**, not a data race -- Go's race detector cannot catch it because channel operations are synchronized.

### Fix

Replace the shared-channel design with **per-invocation channels and done-channel supersession**:

| Before (broken) | After (fixed) |
|---|---|
| `LoadOrStore` returns the **same** shared channel to all `Listen()` goroutines | `Swap` gives each `Listen()` its **own** channel |
| Multiple goroutines compete to read from one channel | Old goroutine's `done` channel is closed, signaling it to exit |
| Stale goroutine consumes and discards Dial tokens | Only the current goroutine reads Dial tokens |
| `Dial()` uses `LoadOrStore` (creates queue if missing) | `Dial()` uses `Load` + done-check (fails fast if no active listener) |

Key details:
- Each `Listen()` call creates a fresh `listenQueue{ch, done, closeOnce}` and uses `sync.Map.Swap()` to atomically replace the previous entry
- If a previous entry existed, its `done` channel is closed, signaling the old goroutine to exit via `case <-wrapper.done: return nil`
- `sync.Once`-wrapped `closeDone()` prevents double-close panics across all exit paths
- `defer CompareAndDelete` ensures cleanup only removes the entry if this goroutine is still the current owner
- `Dial()` pre-checks the `done` channel before sending, avoiding writes to a superseded queue

## Test plan

- [x] Unit tests proving the root cause: `TestListenQueueStaleReaderConsumesDialToken` verifies stale goroutines cannot consume tokens; `TestListenQueueConcurrentReadersAreNonDeterministic` confirms 0/100 stale wins
- [x] Unit tests for supersession: `TestListenQueueSupersessionSignaling`, `TestListenQueueConcurrentSwapSupersedes` (3 goroutines), `TestListenQueueDialTokenDeliveredToNewListener`
- [x] Unit tests for cleanup safety: `TestListenQueueCompareAndDeleteOnStreamError`, `TestListenQueueCompareAndDeleteOnCleanShutdown`, `TestListenQueueReconnectPreventsStaleCleanup`, `TestListenQueueReconnectCreatesNewChannel`
- [x] All 14 controller service unit tests pass
- [ ] E2E test `can lease and connect to exporters` should no longer flake (~10% failure rate before this change)

### DIFF
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -83,8 +83,13 @@ type ControllerService struct {
 }
 
 type listenQueue struct {
-	ch   chan *pb.ListenResponse
-	done chan struct{}
+	ch        chan *pb.ListenResponse
+	done      chan struct{}
+	closeOnce sync.Once
+}
+
+func (q *listenQueue) closeDone() {
+	q.closeOnce.Do(func() { close(q.done) })
 }
 
 type wrappedStream struct {
@@ -451,8 +456,9 @@ func (s *ControllerService) Listen(req *pb.ListenRequest, stream pb.ControllerSe
 	old, loaded := s.listenQueues.Swap(leaseName, wrapper)
 	if loaded {
 		prev := old.(*listenQueue)
-		close(prev.done)
+		prev.closeDone()
 	}
+	defer wrapper.closeDone()
 	defer s.listenQueues.CompareAndDelete(leaseName, wrapper)
 	for {
 		select {
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -115,7 +115,7 @@ func (s *ControllerService) swapListenQueue(leaseName string, newQueue *listenQu
 // per-lease lock guarantees that the queue loaded here cannot be superseded
 // between the load and the send, eliminating the TOCTOU race between Dial and
 // a reconnecting Listen.
-func (s *ControllerService) sendToListener(ctx context.Context, leaseName string, response *pb.ListenResponse) error {
+func (s *ControllerService) sendToListener(_ context.Context, leaseName string, response *pb.ListenResponse) error {
 	mu := s.getLeaseLock(leaseName)
 	mu.Lock()
 	defer mu.Unlock()
@@ -130,12 +130,10 @@ func (s *ControllerService) sendToListener(ctx context.Context, leaseName string
 	default:
 	}
 	select {
-	case <-q.done:
-		return status.Errorf(codes.Unavailable, "exporter is not listening on lease %s", leaseName)
-	case <-ctx.Done():
-		return ctx.Err()
 	case q.ch <- response:
 		return nil
+	default:
+		return status.Errorf(codes.ResourceExhausted, "listener buffer full on lease %s", leaseName)
 	}
 }
 
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -83,7 +83,8 @@ type ControllerService struct {
 }
 
 type listenQueue struct {
-	ch chan *pb.ListenResponse
+	ch   chan *pb.ListenResponse
+	done chan struct{}
 }
 
 type wrappedStream struct {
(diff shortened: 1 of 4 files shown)


mangelajo:
the flakiness I suspect is not related to this https://github.com/jumpstarter-dev/jumpstarter/pull/564/changes
we need proper RCA, I suspect it's more related to the exporter remaining blocked and not announcing proper ready state

raballew:
Based on conversation with @mangelajo this might not be the root cause for the 10% flakiness we currently see but another issue we saw in roughly 0.1% of CI runs where TLS connections were terminated unexpectedly - still worth to investigate/fix.

controller/internal/service/controller_service.go
```diff
@@ -80,6 +80,61 @@ type ControllerService struct {
...
+		return status.Errorf(codes.Unavailable, "exporter is not listening on lease %s", leaseName)
+	default:
+	}
+	select {
+	case q.ch <- response:
+		return nil
+	default:
+		return status.Errorf(codes.ResourceExhausted, "listener buffer full on lease %s", leaseName)
+	}
 }
```
raballew, line 168: [MEDIUM] The old `Dial()` implementation would block on a buffered channel (with `select` on `ctx.Done()`), so a caller whose buffer was temporarily full would eventually succeed once the exporter drained it. The new `sendToListener` uses a non-blocking send and returns `ResourceExhausted` immediately when the 8-slot buffer is full.

This means a `Dial()` that previously would have succeeded after a short wait now fails outright. It also means that under context cancellation with a full buffer, clients receive `ResourceExhausted` instead of the context error they used to get.

In practice this only fires under significant backpressure (8+ concurrent Dial calls to the same lease faster than the exporter processes them), and the change is necessary to avoid mutex starvation in the new per-lease lock design. Still worth noting:

- Consider documenting this behavioral change in the PR description so reviewers and downstream consumers are aware.
- Verify that gRPC clients already retry on `ResourceExhausted`. If they do not, adding client-side retry handling would prevent user-visible regressions.

raballew:
I will start on a clean slate again. This does not fix the 10% flakes.

PR #565, opened 2026-04-16
JEP-0011: Protobuf Introspection and Interface Generation

kirkbrauer:
Propose adding protobuf introspection to allow non-Python clients to discover driver APIs. Introduces FileDescriptorProto as the schema format and enables bidirectional CLI tooling (jmp proto export, jmp proto generate, jmp proto check).

For a full reference implementation, please see: https://github.com/jumpstarter-dev/jumpstarter/tree/jep-11-protobuf-introspection

### DIFF
--- a/python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
+++ b/python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
@@ -1,24 +1,21 @@
 # JEP-0011: Protobuf Introspection and Interface Generation
 
-| Field             | Value                                                 |
-| ----------------- | ----------------------------------------------------- |
-| **JEP**           | 0011                                                  |
-| **Title**         | Protobuf Introspection and Interface Generation       |
-| **Author(s)**     | @kirkbrauer (Kirk Brauer)                             |
-| **Status**        | Discussion                                            |
-| **Type**          | Standards Track                                       |
-| **Created**       | 2026-04-06                                            |
-| **Updated**       | 2026-04-30                                            |
-| **Discussion**    | [Matrix](https://matrix.to/#/#jumpstarter:matrix.org) |
-| **Requires**      | —                                                     |
-| **Supersedes**    | —                                                     |
-| **Superseded-By** | —                                                     |
+| Field          | Value                                                              |
+| -------------- | ------------------------------------------------------------------ |
+| **JEP**        | 0011                                                               |
+| **Title**      | Protobuf Introspection and Interface Generation                    |
+| **Author(s)**  | @kirkbrauer (Kirk Brauer)                                          |
+| **Status**     | Accepted                                                           |
+| **Type**       | Standards Track                                                    |
+| **Created**    | 2026-04-06                                                         |
+| **Updated**    | 2026-05-09                                                         |
+| **Discussion** | [PR #565](https://github.com/jumpstarter-dev/jumpstarter/pull/565) |
 
 ---
 
 ## Abstract
 
-This JEP makes Jumpstarter driver interfaces discoverable to non-Python clients by introducing `.proto` files as the canonical schema artifact for each driver interface. A new CLI, `jmp interface generate`, introspects Python interface classes at development time and emits `.proto` source files that are committed to each driver package. A companion `jmp interface check` runs in CI to detect drift between Python interfaces and their committed `.proto` files. The existing gRPC Server Reflection service and the `DriverInstanceReport.file_descriptor_proto` field serve the same compiled descriptor set at runtime so that tools like `grpcurl`, Buf, and polyglot codegen can discover the driver API without reading Python source.
+This JEP makes Jumpstarter driver interfaces discoverable to non-Python clients by introducing `.proto` files as the canonical schema artifact for each driver interface. A new **codegen CLI** introspects Python interface classes at development time and emits `.proto` source files that are committed to each driver package. A companion **interface check CLI** runs in CI to detect drift between Python interfaces and their committed `.proto` files. The existing gRPC Server Reflection service and the `DriverInstanceReport.file_descriptor_proto` field serve the same compiled descriptor set at runtime so that tools like `grpcurl`, Buf, and polyglot codegen can discover the driver API without reading Python source.
 
 This JEP keeps the Jumpstarter wire protocol unchanged — `DriverCall` remains the transport. The `.proto` schemas serve as an advisory description layer that enables polyglot discovery and future native-gRPC migration. Proto-first workflows (defining interfaces as `.proto` files and generating Python scaffolding) are deferred to a follow-up JEP focused on non-Python codegen.
 
@@ -52,11 +49,13 @@ This JEP addresses three concrete gaps:
 
 This proposal adds three capabilities to Jumpstarter, all centered on committed `.proto` files as the canonical schema artifact:
 
-1. **`jmp interface generate` CLI (Python → `.proto`)** — a developer-invoked command that introspects a `DriverInterface` class and emits a `.proto` source file. The `.proto` file is committed alongside the driver package that defines the interface.
-2. **`jmp interface check` CLI (drift detection)** — runs in CI to verify the committed `.proto` file still matches the Python interface. Reports any method, parameter, return-type, or streaming-semantics mismatch as a test failure.
+1. **Codegen CLI (Python → `.proto`)** — a developer-invoked command that introspects a `DriverInterface` class and emits a `.proto` source file. The `.proto` file is committed alongside the driver package that defines the interface.
+2. **Interface check CLI (drift detection)** — runs in CI to verify the committed `.proto` file still matches the Python interface. Reports any method, parameter, return-type, or streaming-semantics mismatch as a test failure.
 3. **Runtime descriptor exposure** — the exporter loads the pre-compiled descriptor set (produced by `protoc --descriptor_set_out` from the committed `.proto` files), registers the services with gRPC Server Reflection, and embeds the raw bytes in `DriverInstanceReport.file_descriptor_proto`.
 
-The `.proto` files are the source of truth. Introspection happens once, at development time, when the author runs `jmp interface generate`; it does **not** happen at exporter startup or at Python import time. This mirrors the standard gRPC development workflow and keeps the exporter's runtime free of schema-construction work.
+The `.proto` files are the source of truth. Introspection happens once, at development time, when the author runs the codegen CLI; it does **not** happen at exporter startup or at Python import time. This mirrors the standard gRPC development workflow and keeps the exporter's runtime free of schema-construction work.
+
+**CLI naming is intentionally deferred.** This JEP does not commit to a concrete command surface for the codegen and check tools. Whether they ship as `jmp` subcommands, a separate `jmp-devel` binary, standalone executables, or some other shape is a UX decision better made during implementation, when we can weigh how much of the developer toolchain ends up under one umbrella. Throughout this document, "the codegen CLI" and "the interface check CLI" are used as descriptive names; bash code blocks use `<codegen>` and `<interface-check>` as placeholders for whatever the final invocation turns out to be.
 
 Proto-first workflows — authoring `.proto` files and generating Python interface/client/driver scaffolding from them — are **out of scope for this JEP**. They are planned as a follow-up JEP once non-Python codegen is ready to consume the committed `.proto` files.
 
@@ -68,7 +67,7 @@ The `.proto` files and `FileDescriptorProto` descriptors serve as a **descriptio
 
 In concrete terms:
 
-- **What the proto IS used for:** introspection (`GetReport`, gRPC reflection), compatibility checking (`jmp interface check`, `buf breaking`), documentation, and polyglot codegen.
+- **What the proto IS used for:** introspection (`GetReport`, gRPC reflection), compatibility checking (the interface check CLI, `buf breaking`), documentation, and polyglot codegen.
 - **What the proto is NOT used for:** actual RPC transport. The `DriverCall(uuid="...", method="on", args=[])` message continues to be the wire format.
 
 A future JEP will propose migrating to native protobuf service implementations — where `protoc`-generated stubs handle serialization directly and `DriverCall` is retired — but that is a separate, breaking change with its own migration path. A design sketch for this future work is included at the end of this JEP for context.
@@ -77,7 +76,7 @@ A future JEP will propose migrating to native protobuf service implementations �
 
 gRPC reflection will advertise services described by the committed `.proto` files — for example, `jumpstarter.interfaces.power.v1.PowerInterface.On(Empty)`. Because the wire protocol is unchanged, **those services are not backed by native gRPC handlers in this JEP**. A client that discovers the service through reflection and attempts to invoke it directly (e.g., `grpcurl -d '{}' host:port jumpstarter.interfaces.power.v1.PowerInterface/On`) will receive `UNIMPLEMENTED`.
 
-Reflection here is deliberately **advisory** — it exposes the schema so polyglot clients, codegen pipelines, and documentation tooling can discover the driver API and generate typed stubs that drive the existing `DriverCall` transport. The follow-up native-gRPC JEP will add handlers so reflected services become directly invokable without changing the proto schema produced by this JEP.
+Reflection here is deliberately **advisory** — it exposes the schema so polyglot clients, codegen pipelines, and documentation tooling can discover the driver API and generate typed stubs that drive the existing `DriverCall` transport. The follow-up native-gRPC JEP will add handlers so reflected services become directly invocable without changing the proto schema produced by this JEP.
 
 ### `FileDescriptorProto` as the Schema Format
 
@@ -89,10 +88,10 @@ Using it means there is one descriptor format throughout the entire system — g
 
 ### Build-time introspection of `@export` methods
 
-Introspection runs at `jmp interface generate` invocation time, not at import or exporter startup. The `@export` decorator itself is unchanged — it still stamps markers on the function for `DriverCall` dispatch. Type information is read directly from the live class via `inspect.signature()` when the CLI tool loads the interface module:
+Introspection runs at codegen CLI invocation time, not at import or exporter startup. The `@export` decorator itself is unchanged — it still stamps markers on the function for `DriverCall` dispatch. Type information is read directly from the live class via `inspect.signature()` when the CLI tool loads the interface module:
 
 ```python
-# inside jmp interface generate
+# inside the codegen CLI
 sig = inspect.signature(method)
 call_type = _infer_call_type(method)   # UNARY | SERVER_STREAMING | BIDI_STREAMING
 params = [
@@ -159,7 +158,7 @@ This approach means Pydantic handles ~80-85% of the type mapping automatically. 
 
 ### Build-time `.proto` generation
 
-`jmp interface generate` uses a `build_file_descriptor()` library function to construct a `google.protobuf.descriptor_pb2.FileDescriptorProto` from an interface class, then renders it as human-readable `.proto` source. The builder is a pure function — it is **not** called by the exporter at runtime or by any import-time hook.
+The codegen CLI uses a `build_file_descriptor()` library function to construct a `google.protobuf.descriptor_pb2.FileDescriptorProto` from an interface class, then renders it as human-readable `.proto` source. The builder is a pure function — it is **not** called by the exporter at runtime or by any import-time hook.
 
 ```python
 from google.protobuf.descriptor_pb2 import (
@@ -221,9 +220,10 @@ Protobuf service and message definitions carry structure — method names, param
 
 #### Interface Versioning
 
-Interface versioning follows standard protobuf package-level versioning conventions. The version is encoded in the package name (e.g., `jumpstarter.interfaces.power.v1`) and the `--version` flag on `jmp interface generate`. Breaking changes to an interface require a new package version (`v1` → `v2`), and `buf breaking` enforces backward compatibility within a version.
+Interface versioning follows standard protobuf package-level versioning conventions. The version is encoded in the package name (e.g., `jumpstarter.interfaces.power.v1`) and the `--version` flag on the codegen CLI. Breaking changes to an interface require a new package version (`v1` → `v2`), and `buf breaking` enforces backward compatibility within a version.
 
 This approach was chosen over a custom `interface_version` service option because:
+
 - It follows the standard protobuf/Buf versioning convention that all gRPC tooling already understands
 - It avoids custom annotations and the extraction logic they require
 - `buf breaking` is purpose-built for detecting incompatible proto changes
@@ -255,7 +255,7 @@ Note that `@exportstream` methods (raw byte stream constructors) do not need a c
 
 Proto comments (lines starting with `//` immediately preceding a service, method, message, or field definition) are a first-class concept in the protobuf ecosystem. They're preserved in `FileDescriptorProto` source info, rendered by `protoc-gen-doc`, displayed by `grpcurl describe`, shown in Buf Schema Registry, and emitted as language-native doc comments by standard codegen plugins (`protoc-gen-java`, `protoc-gen-ts`, etc.). There's no need to duplicate them as custom options — the standard proto comment mechanism already flows through the entire toolchain.
 
-The `build_file_descriptor()` builder and `jmp interface generate` tool extract docstrings from Python and emit them as proto comments:
+The `build_file_descriptor()` builder and the codegen CLI extract docstrings from Python and emit them as proto comments:
 
 - **Class docstrings** → comments above the `service` definition
 - **Method docstrings** → comments above each `rpc` definition
@@ -309,7 +309,7 @@ class PowerReading(BaseModel):
 
 #### Generated `.proto` output
 
-When `jmp interface generate` processes the class above, the resulting `.proto` file carries the version option and doc comments:
+When the codegen CLI processes the class above, the resulting `.proto` file carries the version option and doc comments:
 
 ```protobuf
 syntax = "proto3";
@@ -370,19 +370,19 @@ This happens for free — no custom options, no custom codegen plugins, no annot
 
 #### Doc comment round-trip consistency
 
-The `jmp interface check` tool validates doc comments bidirectionally:
+The interface check CLI validates doc comments bidirectionally:
 
 - **Python → Proto:** Verifies that docstrings in the Python source appear as proto comments in the generated `.proto` file.
 - **Proto → Python:** Verifies that proto comments in a hand-authored `.proto` file produce corresponding docstrings in the generated Python code.
 
 This ensures documentation doesn't drift regardless of which direction the developer is working from.
 
-### The `jmp interface generate` Tool (Python → Proto)
+### Codegen CLI (Python → Proto)
 
-A CLI command introspects a Python interface class and produces a canonical `.proto` source file:
+The codegen CLI introspects a Python interface class and produces a canonical `.proto` source file:
 
 ```bash
-jmp interface generate \
+<codegen> \
   --package jumpstarter-driver-power \
   --interface PowerInterface \
   --version v1 \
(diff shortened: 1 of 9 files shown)


python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1726 @@
+# JEP-0011: Protobuf Introspection and Interface Generation
+
+| Field             | Value                                                 |
+| ----------------- | ----------------------------------------------------- |
+| **JEP**           | 0011                                                  |
```
mangelajo: why 0011? :D wouldn't it be 2 ?

bennyz: or 0010 😅

mangelajo: Lol ❤️

kirkbrauer: JEP-0-9 are reserved for process improvements and JEP-10 already exists.

kirkbrauer:
@ambient-code Please update the TOCTREE and main index file for the JEPs to include this proposal.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1726 @@
...
+
+- `FlasherInterface.flash(source)` — `source` needs annotation (resource handle, type as `str`)
+- `FlasherInterface.dump(target)` — `target` needs annotation
+- `StorageMuxInterface.host()`, `.dut()`, `.off()` — need `-> None` return types
+- `StorageMuxInterface.write(src)`, `.read(dst)` — need return type annotations
+- `NetworkInterface.connect()` — needs return type annotation
+
+**Driver implementation gaps (~99 `@export` methods):**
+
+The majority (~90%) are missing `-> None` return types on void methods — a mechanical fix. ~10 methods also have missing parameter types, primarily `source` and `target` parameters on flasher implementations (resource handle UUIDs). Affected packages include: dutlink, energenie, esp32, flashers, gpiod, http, http-power, iscsi, network, noyito-relay, opendal, pi-pico, probe-rs, pyserial, qemu, ridesx, sdwire, shell, snmp, ssh, tasmota, tftp, tmt, yepkit, and composite (test file).
```
raballew: Making type annotations mandatory on every `@export` method at import time affects everything. Shouldnt it be opt-in??

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1726 @@
...
+- It requires `protoc` to be installed in the development environment, adding a native dependency.
+- The `build_file_descriptor()` approach is pure Python, runs at import time, and requires no external tooling.
+
+### Storing type info in `methods_description` strings
+
+Encoding type information into the existing `methods_description` map (e.g., as a JSON string per method) was considered. This was rejected because:
+
+- It's a hack that conflates human-readable documentation with machine-readable schema.
+- It doesn't integrate with any existing tooling.
+- The `file_descriptor_proto` field is the proper place for machine-readable schema, and `methods_description` remains for human consumption.
```
raballew, line 1267: Consider adding a rejected alternative that evaluates the build-time-only approach (generate .proto files from existing Python interfaces without adding runtime type enforcement) and explain why runtime introspection was chosen over that simpler path.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1726 @@
...
+
+The `.proto` files and `FileDescriptorProto` descriptors serve as a **description layer** on top of the existing dispatch mechanism — they describe what methods exist, what types they use, and what streaming semantics they have. They do not replace `DriverCall` with actual protobuf-native gRPC service implementations (where `PowerInterface` would be a real gRPC service with compiled request/response message stubs). That migration would be a significant breaking change to the exporter protocol, affecting every existing client and driver, and is explicitly out of scope for this JEP.
+
+In concrete terms:
+
+- **What the proto IS used for:** introspection (`GetReport`, gRPC reflection), code generation (typed clients and driver adapters), compatibility checking (`jmp interface check`, `buf breaking`), documentation, and the interface registry.
+- **What the proto is NOT used for:** actual RPC transport. The `DriverCall(uuid="...", method="on", args=[])` message continues to be the wire format.
+
```
raballew, line 72: gRPC reflection will serve FileDescriptorProtos describing driver interfaces as gRPC services, but tools like grpcurl that discover these via reflection will expect to invoke them. Calling `PowerInterface.On(Empty)` will return UNIMPLEMENTED because only `ExporterService.DriverCall` exists on the wire. The document acknowledges this but does not address the user experience when standard tools fail.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1726 @@
...
+    async def _off(self) -> None:
+        self._usb_control("off", self.port)
+
+    async def _read(self):
+        yield PowerReading(voltage=5.0, current=self._read_current())
+```
+
+The driver developer's file contains **only hardware logic**. No `@export` decorators, no `DriverClient` wiring, no `call()` / `streamingcall()` plumbing, no serialization code. The generated adapter and client handle all of that.
+
```
raballew: Food for thought: The proto-first adapter creates `@export` methods that delegate to abstract methods. For simple unary methods, the dispatch plumbing is the method. The adapter adds an extra call frame and extra method definition and a naming convention without reducing the code a driver developer writes. The Python firsst pattern feels simpler.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1726 @@
...
+        for raw in self.streamingcall("read"):
+            yield PowerReading.model_validate(raw, strict=True)
+```
+
+The generated client inherits from both `PowerInterface` (ensuring type-safety against the contract) and `DriverClient` (providing the `call()` / `streamingcall()` transport). Every method body is a one-liner that delegates to the underlying `DriverCall` RPC — serializing arguments to `google.protobuf.Value`, dispatching by method name, and deserializing the response into the typed return value. Pydantic models are deserialized using `model_validate()` rather than constructor kwargs, consistent with existing Jumpstarter client conventions.
+
+**Note:** In the current codebase, client classes inherit only from `DriverClient` (e.g., `class PowerClient(DriverClient)`). The generated pattern of inheriting from both the interface and `DriverClient` (e.g., `class PowerClient(PowerInterface, DriverClient)`) is a new convention introduced by this JEP. It provides compile-time verification that the client implements all interface methods. Existing clients are not required to change, but new auto-generated clients will follow this pattern.
```
raballew: Generated clients inherit from both the interface and DriverClient, but existing clients use single inheritance. I assume we migrate all existing clients to dual inheritance?

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
kirkbrauer: @raballew I agree, making it mandatory is probably too strict. I had originally thought it was needed, but after I created a reference implementation, we can get away with it and make it opt-in like you suggest.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
kirkbrauer: @raballew I actually think that build-time only is a cleaner approach, but that will require us to re-generate the Protobufs for all of the existing interfaces and could slow down iterative driver development. I think the dynamic proto FileDescriptor generation was an attempt to over-engineer this for backwards compat.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
kirkbrauer: @raballew Once again an attempt at over-engineering here. I think we can drop everything except the Python-first approach for now. We would only use the proto-first adapter for non-Python languages in the future which I have already created a prototype for.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
kirkbrauer: @raballew That would be correct, this will help us to make the contract more firm moving forwards and will behave similarly regardless of the language (they will always at least inherit the interface if they don't support multiple-inheritance).

kirkbrauer:
@raballew @mangelajo I have updated the JEP proposal with a more refined scope focusing on just the Protobuf generation at build-time which is much simpler, deferring the interface generation for now, and addressing additional comments.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+### User Stories
+
+- **As a** Python driver developer, **I want** an opt-in linter that flags `@export` methods missing type annotations, **so that** interfaces I choose to expose to polyglot consumers are fully typed before the `.proto` file is generated.
+
+- **As a** Java test engineer writing Android device tests, **I want to** discover all available methods on a leased device's power driver — including parameter types, return types, and streaming semantics — **so that** I can generate type-safe Kotlin stubs instead of hand-writing `DriverCall` invocations with magic string method names.
+
+- **As a** tools developer building a device management dashboard, **I want to** point standard gRPC tooling (`grpcurl`, Postman, Buf Studio) at an exporter and discover every available driver interface with full type information, **so that** I can prototype interactions without reading Python source code.
+
+- **As a** CI pipeline author, **I want to** run a compatibility check in CI that verifies the Python driver interface hasn't drifted from the committed `.proto` definition, **so that** cross-language clients don't silently break when a driver evolves.
```
mangelajo, line 44: Those are our main two points IMHO. 

Because the other two can be achieved by enforcing typechecking, and announcing parameter/types in the reports as well.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+## Proposal
+
+### Overview
+
+This proposal adds three capabilities to Jumpstarter, all centered on committed `.proto` files as the canonical schema artifact:
+
+1. **`jmp interface generate` CLI (Python → `.proto`)** — a developer-invoked command that introspects a `DriverInterface` class and emits a `.proto` source file. The `.proto` file is committed alongside the driver package that defines the interface.
```
mangelajo: probably we should standardize to a proto directory per package?

I.e. 

https://github.com/jumpstarter-dev/jumpstarter/tree/main/python/packages/jumpstarter-driver-pi-pico

could contain a proto directory at the same level as examples and jumpstarter_driver_pi_pico

this is implementation detail anyway :)

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+- **As a** tools developer building a device management dashboard, **I want to** point standard gRPC tooling (`grpcurl`, Postman, Buf Studio) at an exporter and discover every available driver interface with full type information, **so that** I can prototype interactions without reading Python source code.
+
+- **As a** CI pipeline author, **I want to** run a compatibility check in CI that verifies the Python driver interface hasn't drifted from the committed `.proto` definition, **so that** cross-language clients don't silently break when a driver evolves.
+
+## Proposal
+
+### Overview
+
+This proposal adds three capabilities to Jumpstarter, all centered on committed `.proto` files as the canonical schema artifact:
```
mangelajo, line 50: What do you think about creating a separate "jmp-devel" tool or equivalent? I am worried about overlading the jmp tool way too much making UX harder later in time.

Remember we had this back/forth with jmp-admin jmp admin etc... but devels are still a smaller subset of users.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+1. **`jmp interface generate` CLI (Python → `.proto`)** — a developer-invoked command that introspects a `DriverInterface` class and emits a `.proto` source file. The `.proto` file is committed alongside the driver package that defines the interface.
+2. **`jmp interface check` CLI (drift detection)** — runs in CI to verify the committed `.proto` file still matches the Python interface. Reports any method, parameter, return-type, or streaming-semantics mismatch as a test failure.
+3. **Runtime descriptor exposure** — the exporter loads the pre-compiled descriptor set (produced by `protoc --descriptor_set_out` from the committed `.proto` files), registers the services with gRPC Server Reflection, and embeds the raw bytes in `DriverInstanceReport.file_descriptor_proto`.
+
+The `.proto` files are the source of truth. Introspection happens once, at development time, when the author runs `jmp interface generate`; it does **not** happen at exporter startup or at Python import time. This mirrors the standard gRPC development workflow and keeps the exporter's runtime free of schema-construction work.
```
mangelajo: 👍

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+Proto-first workflows — authoring `.proto` files and generating Python interface/client/driver scaffolding from them — are **out of scope for this JEP**. They are planned as a follow-up JEP once non-Python codegen is ready to consume the committed `.proto` files.
```
mangelajo, line 60: IMHO we should try to avoid that path unless there is a very good reason. My biggest worry is making driver development harder. Now it's one of our killer "features" :) driver development is very easy and straight forward.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+### Wire Protocol: `DriverCall` Remains Unchanged
+
+An important design constraint: **this JEP does not change the wire protocol.** The existing `DriverCall` and `StreamingDriverCall` RPCs — where the client sends a method name as a string and arguments as `google.protobuf.Value` — remain the actual transport mechanism. The auto-generated client code still calls `self.call("on")` and `self.streamingcall("read")` under the hood. The auto-generated driver adapter still receives dispatch through the existing `@export` decorator and `Driver` base class machinery.
+
+The `.proto` files and `FileDescriptorProto` descriptors serve as a **description layer** on top of the existing dispatch mechanism — they describe what methods exist, what types they use, and what streaming semantics they have. They do not replace `DriverCall` with actual protobuf-native gRPC service implementations (where `PowerInterface` would be a real gRPC service with compiled request/response message stubs). That migration would be a significant breaking change to the exporter protocol, affecting every existing client and driver, and is explicitly out of scope for this JEP.
```
mangelajo, line 66: Yes, when we explore that, we will want compatibility for both methods. Either forever, or at least for a reasonable amount of time so we make sure all drivers support it.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+In concrete terms:
+
+- **What the proto IS used for:** introspection (`GetReport`, gRPC reflection), compatibility checking (`jmp interface check`, `buf breaking`), documentation, and polyglot codegen.
+- **What the proto is NOT used for:** actual RPC transport. The `DriverCall(uuid="...", method="on", args=[])` message continues to be the wire format.
+
+A future JEP will propose migrating to native protobuf service implementations — where `protoc`-generated stubs handle serialization directly and `DriverCall` is retired — but that is a separate, breaking change with its own migration path. A design sketch for this future work is included at the end of this JEP for context.
```
mangelajo: yup, migration path is important, and if it makes driver development harder, they should probably live together for a while, or for ever, i.e. available for development mode.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+#### gRPC reflection is advisory in this JEP
+
+gRPC reflection will advertise services described by the committed `.proto` files — for example, `jumpstarter.interfaces.power.v1.PowerInterface.On(Empty)`. Because the wire protocol is unchanged, **those services are not backed by native gRPC handlers in this JEP**. A client that discovers the service through reflection and attempts to invoke it directly (e.g., `grpcurl -d '{}' host:port jumpstarter.interfaces.power.v1.PowerInterface/On`) will receive `UNIMPLEMENTED`.
```
mangelajo: How will we tell which UUID / driver instance are we calling? passing the UUID in the call metadata?

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
mangelajo: Or should we keep them organized under the base https://github.com/jumpstarter-dev/jumpstarter/tree/main/protocol/proto/jumpstarter ? having a drivers/<name> directory for each driver?

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+**Migration:** The standard in-tree clients (PowerClient, NetworkClient, StorageMuxClient, FlasherClient, CompositeClient, and the virtual-power client) are migrated to dual inheritance alongside the `DriverInterface` migration (Phase 1b). Drivers with clients that provide client-side orchestration (e.g., `FlasherClient` with `OpendalAdapter`, `StorageMuxFlasherClient.flash()`) keep their hand-written orchestration — dual inheritance does not change the methods, only the declared bases.
+
+### Proto-first workflow (deferred)
+
+An earlier revision of this JEP described a `jmp interface implement` command that took a `.proto` file and generated a Python interface class, client class, and driver adapter. That capability is **not part of this JEP** and is deferred to a follow-up JEP focused on non-Python codegen.
```
mangelajo: thanks 👍  :)

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+The `.proto` schema format defined by this JEP is stable enough that the follow-up JEP can build on it without revisiting the schema.
+
+### The `jmp interface check` Tool (CI drift detection)
+
+Because the `.proto` files are committed and reviewed, CI needs a way to detect when a Python interface change makes the committed `.proto` stale. `jmp interface check` is that gate:
+
+```bash
+jmp interface check \
+  --proto python/packages/jumpstarter-driver-power/proto/power/v1/power.proto \
+  --interface jumpstarter_driver_power.interface.PowerInterface
```
mangelajo: should we have a list somewhere in the project (yaml or similar)? so we can just run jmp interface check? or have it auto discover the usage of interfaces across the codebase?

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+**Rationale:** Committed schemas give reviewers a visible diff, CI a concrete artifact for `buf breaking`, and polyglot consumers a stable reference. Dynamic generation has no diff, couples dispatch to schema at import time, and shifts the drift-detection problem onto the exporter. A `jmp interface check` CI gate against a committed `.proto` is both simpler and more informative than runtime reconstruction. This decision reflects PR #565 review feedback (@raballew, 2026-04-30): dynamic generation was over-engineered for backwards compatibility that the committed artifact already provides.
+
+### DD-2: Opt-in annotation validation, not mandatory
+
+**Alternatives considered:**
+
+1. **Mandatory at decoration time** — `@export` raises `TypeError` for any method without complete annotations. Forces the entire codebase (~111 methods across 25 packages) to be fully typed before anything builds.
+2. **Opt-in via `@export(strict=True)` / `JMP_EXPORT_STRICT=1`** — `@export` in default mode emits `DeprecationWarning`. Teams enable strict mode per package. `jmp interface generate` always requires full annotations — enforcement moves to the tool.
```
mangelajo: 👍

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+### DD-3: Python-first only; proto-first deferred
+
+**Alternatives considered:**
+
+1. **Bidirectional tooling in Phase 1** — ship both `jmp interface generate` (Python → `.proto`) and `jmp interface implement` (`.proto` → Python interface + client + driver adapter).
+2. **Python-first only** — ship only `jmp interface generate` and `jmp interface check`. Proto-first is deferred to a follow-up JEP focused on non-Python codegen.
+
+**Decision:** Option 2 — Python-first only.
+
+**Rationale:** For Python drivers, the proto-first adapter pattern adds an inheritance layer and an underscore-prefixed abstract-method indirection without materially reducing the code the author writes. Its main value is producing clients and servicers for **non-Python** languages — that design is orthogonal to the Python introspection work and benefits from a dedicated JEP. Shrinking scope unblocks Phase 1 and avoids committing to a Python adapter pattern before non-Python codegen design is complete. A reference prototype for non-Python codegen already exists (author's work) and will be the basis for the follow-up JEP. Matches PR #565 review feedback (@raballew, 2026-04-30).
```
mangelajo: We also must consider how all this works for out of tree drivers, how do they ship the .protos, etc.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+| `PowerInterface` | `jumpstarter-driver-power` | ABCMeta, fully typed | Straightforward migration |
+| `VirtualPowerInterface` | `jumpstarter-driver-power` | ABCMeta, fully typed | Separate from PowerInterface; `off(destroy: bool = False)` differs |
+| `NetworkInterface` | `jumpstarter-driver-network` | ABCMeta | `connect()` missing return type annotation |
+| `FlasherInterface` | `jumpstarter-driver-opendal` | ABCMeta | `flash(source)` and `dump(target)` missing param types |
+| `StorageMuxInterface` | `jumpstarter-driver-opendal` | ABCMeta | 5 methods missing return types |
+| `StorageMuxFlasherInterface` | `jumpstarter-driver-opendal` | Inherits StorageMuxInterface | No own methods; just overrides `client()` |
+| `CompositeInterface` | `jumpstarter-driver-composite` | **No metaclass (plain class)** | Empty interface, no abstract methods |
+
+**Explicitly out of scope:** `FlasherClientInterface` (`jumpstarter-driver-opendal/client.py`) is a client-side ABC with complex types (`PathBuf`, `Operator`, `Compression`). It is not a driver interface contract and does not need migration to `DriverInterface`. The introspection system targets driver-side interfaces only.
```
mangelajo, line 823: Yes, this one will need a exporter-side rewrite to enable gRPC usage.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+        self.call("off")
+    def cycle(self, wait=2):     # NOT in PowerInterface — pure client-side logic
+        self.off()
+        time.sleep(wait)
+        self.on()
+```
+
+`cycle()` is a convenience method that composes `off()` + `sleep()` + `on()` — it doesn't correspond to an `@export` method on the driver. Similarly, `StorageMuxFlasherClient` has `flash()` and `dump()` methods that orchestrate multiple interface calls with `OpendalAdapter` logic.
+
+Convenience methods are **not represented in the proto** because they don't cross the wire — they're client-side compositions of methods that do. The proto describes only the interface contract. Clients are free to add orchestration methods alongside the interface-mandated ones; keeping them in the same class is fine because the proto / `jmp interface check` layer only cares about the methods declared on the `DriverInterface`.
```
mangelajo: I would say that some of those convenience methods (like this one which is simple) must me moved to the Power driver itself, I guess we cannot add real methods to the interface, unless we stop making them purely ABC interfaces.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+### Concurrency and Thread-Safety
+
+`build_file_descriptor()` is a pure function (no side effects, no mutation of inputs) and safe to call from any thread — but it is only called at `jmp interface generate` time, so concurrency is not relevant at runtime. The exporter's descriptor-set load is a single file read during startup before the gRPC server begins accepting connections. The gRPC reflection service is thread-safe by design (`grpcio-reflection` handles concurrent requests internally).
+
+### Security Implications
+
+gRPC Server Reflection exposes the full interface schema to any client that can reach the exporter's gRPC port. In Jumpstarter's architecture, the exporter is already behind the operator's authentication and lease system — only clients with a valid lease can dial the exporter. Reflection does not bypass this; it's registered on the same `grpc.Server` that serves `ExporterService` and inherits its transport security (mTLS via cert-manager).
```
mangelajo:
```suggestion
gRPC Server Reflection exposes the full interface schema to any client that can reach the exporter's gRPC port. In Jumpstarter's architecture, the exporter is already behind the controller's authentication and lease system — only clients with a valid lease can dial the exporter. Reflection does not bypass this; it's registered on the same `grpc.Server` that serves `ExporterService` and inherits its transport security (mTLS via cert-manager).
```

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
mangelajo: it is documented in line 1097

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
mangelajo: ahh yes, I see it later down the JEP :)

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+        # Map: (service_name, method_name) -> {uuid: servicer}
+        self._servicers: dict[str, dict[UUID, grpc.GenericRpcHandler]] = {}
+
+    def register(self, uuid: UUID, servicer, service_name: str):
+        self._servicers.setdefault(service_name, {})[uuid] = servicer
+
+    async def intercept_service(self, continuation, handler_call_details):
+        # Extract UUID from metadata
+        metadata = dict(handler_call_details.invocation_metadata)
+        uuid_str = metadata.get("x-jumpstarter-driver-uuid")
```
mangelajo, line 1459: 👍

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+stub.on(Empty.getDefaultInstance())
+stub.read(Empty.getDefaultInstance()).collect { reading ->
+    println("Voltage: ${reading.voltage}V, Current: ${reading.current}A")
+}
+```
+
+#### Backward compatibility: dual-path dispatch
+
+During the transition, the exporter serves both protocols simultaneously:
```
mangelajo, line 1576: I would really keep it "forever", unless we get comfortable with it, we see it's easy to bring drivers into native gRPC, and that testability (i.e. the serve(xxx) pattern) is not impacted.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+Native client ────> │  On(Empty) + UUID metadata  │ ──┘
+                    └─────────────────────────────┘
+```
+
+#### Migration phases
+
+1. **This JEP:** Generate `FileDescriptorProto` and `.proto` files. Wire protocol unchanged. Polyglot clients can use `DynamicMessage` with `DriverCall` and the descriptor.
+2. **Future JEP — dual path:** Exporter registers native gRPC services alongside `DriverCall`. Compile `.proto` files to stubs. New clients choose native path. Old clients unchanged.
+3. **Deprecation:** Mark `DriverCall` as deprecated. Migration guide published.
+4. **Removal:** Remove `DriverCall` in a major version bump. All clients use native gRPC.
```
mangelajo:
```suggestion
To be decided based on the discoveries during implementation of this JEP and the dual-path JEP
3. **Deprecation:** Mark `DriverCall` as deprecated. Migration guide published.
4. **Removal:** Remove `DriverCall` in a major version bump. All clients use native gRPC.
```
I keep repeating me.. but ... I want to experience how it impacts testability, driver development, out of tree drivers etc. :-)

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
```diff
@@ -0,0 +1,1526 @@
...
+
+## Implementation Phases
```
mangelajo, line 1614: nice I was about to ask for this. I think it's important that we do it in small PR chunks
to avoid reviewers getting stuck, or rebases becoming a problem.

But this will require diligence in terms of PR reviewing, to avoid stalling the process.

mangelajo (commented):
Thank you !, very substantial work, and positive (of course with some risks of making things a little bit more complicated) but my opinion after fully reading it is that it's worth it.

My main question at this point is if we should really commit to deprecate the DriverCall or just keep it around forever to make development and out of tree drivers easier. I'd prefer to defer that decision until we are more familiar with this new path :)

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
raballew: I second this thought.

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
mangelajo: I comment later on this.. IMO I prefer that we evaluate later and decide, not now

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
kirkbrauer: We can make this a follow-up JEP I think. But, I think this is a good feature to have. Maybe auto-discovery is the best way to handle this?

python/docs/source/internal/jeps/JEP-0011-protobuf-introspection-interface-generation.md
kirkbrauer: @mangelajo I'm not sure which solution is best, but I'm inclined towards a `proto/` directory in each driver source.

kirkbrauer:
@mangelajo I've added the final refinements based on the comments, LGTM

raballew (approved):
lgtm

PR #616, opened 2026-04-17
jumpstarter-driver-gpio: Add read method required by PowerInterface

anobli:
This addresses the following issue:
TypeError: Can't instantiate abstract class PowerSwitch without an implementation for abstract method 'read'

This adds the read method which return 0 power because we can't get this kind of data from a gpio.

### DIFF
--- a/python/packages/jumpstarter-driver-gpiod/jumpstarter_driver_gpiod/driver.py
+++ b/python/packages/jumpstarter-driver-gpiod/jumpstarter_driver_gpiod/driver.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import time
+from collections.abc import Generator
 from dataclasses import dataclass, field
 
 try:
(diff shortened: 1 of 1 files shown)


python/packages/jumpstarter-driver-gpiod/jumpstarter_driver_gpiod/driver.py
```diff
@@ -236,3 +238,8 @@ def on(self) -> None:
     def off(self) -> None:
         """Switch off the power"""
         DigitalOutput.off(self)
+
+    @export
```
mangelajo, line 242: oh, good catch, too bad we didn't have a test for this.

Can you do it like in this one?

https://github.com/jumpstarter-dev/jumpstarter/blob/main/python/packages/jumpstarter-driver-noyito-relay/jumpstarter_driver_noyito_relay/driver.py#L187

Instead of returning 0/0, if somebody tries to read power, will get an error that makes it clear that the driver can't implement this.

mangelajo: I will set this patch for backport to release-0.8 / v0.8.2

mangelajo (changes requested):
Thank you very much for the bugfix.

mangelajo (approved):
Thank you so much

PR #598, opened 2026-04-17
fix: release lease on unused timeout when hooks are configured

raballew:
When a lease times out without any client connection and hooks are configured, _cleanup_after_lease skipped the afterLease hook (correct) but also skipped calling _request_lease_release (bug). This left the exporter permanently stuck in LeaseReady status because the controller was never notified that the lease should be freed.

Add _request_lease_release() call in the else branch of _cleanup_after_lease so the controller always frees the lease, regardless of whether the afterLease hook ran.

Fixes: #237

PR #617, opened 2026-04-17
JEP-0012: Explicit lease lifecycle FSM for exporter

mangelajo:
## Summary
- add `JEP-0012` under `docs/internal/jeps/` proposing an explicit per-lease lifecycle FSM for exporter orchestration
- follow the PR-based JEP process from #423 and keep this proposal as a JEP only, with no separate ADR documents
- capture the motivation from the recent exporter lifecycle fixes while proposing a cleaner long-term internal design

## Test plan
- [x] Docs-only change reviewed locally
- [x] Checked the new JEP file for diagnostics
- [ ] No automated tests run (docs-only proposal)

### DIFF
--- a/python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
+++ b/python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
@@ -9,7 +9,7 @@
 | **Type**           | Standards Track |
 | **Created**        | 2026-04-17 |
 | **Updated**        | 2026-04-17 |
-| **Discussion**     | Pending PR creation |
+| **Discussion**     | [PR #617](https://github.com/jumpstarter-dev/jumpstarter/pull/617) |
 
 ---
 
--- a/python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
+++ b/python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
@@ -1,0 +1,462 @@
+# JEP-0012: Explicit Lease Lifecycle State Machine for Exporter
+
+| Field              | Value |
+|--------------------|-------|
+| **JEP**            | 0012 |
+| **Title**          | Explicit Lease Lifecycle State Machine for Exporter |
+| **Author(s)**      | @mangelajo |
+| **Status**         | Discussion |
+| **Type**           | Standards Track |
+| **Created**        | 2026-04-17 |
+| **Updated**        | 2026-04-17 |
+| **Discussion**     | Pending PR creation |
+
+---
+
+## Abstract
+
+This JEP proposes replacing the exporter's implicit lease lifecycle
+coordination with an explicit per-lease state machine.
+
+Today the exporter coordinates lease setup and teardown through a mix of
+`LeaseContext` events, exporter status values, hook callbacks, and task
+cancellation timing. That design works for the common path, but recent
+fixes around premature lease end and unused lease timeout have shown that
+the current model is difficult to reason about and easy to break when new
+paths are added.
+
+The proposal introduces a small explicit lease lifecycle state machine,
+owned by the exporter, that becomes the single source of truth for:
+
+- when connections may start
+- when cleanup may begin
+- when after-lease work is allowed to run
+- when a lease is fully complete and the next lease may start
+
+This is an internal refactor only. It does not change the gRPC protocol
+or the external `ExporterStatus` API.
+
+## Motivation
+
+The current exporter lease lifecycle is already an implicit state
+machine, but its state is spread across multiple mechanisms:
+
+- `LeaseContext.before_lease_hook`
+- `LeaseContext.after_lease_hook_started`
+- `LeaseContext.after_lease_hook_done`
+- `LeaseContext.lease_ended`
+- `LeaseContext.skip_after_lease_hook`
+- `ExporterStatus` transitions such as `BEFORE_LEASE_HOOK`,
+  `LEASE_READY`, `AFTER_LEASE_HOOK`, and `AVAILABLE`
+- cancellation timing across `serve()`, `handle_lease()`,
+  `_handle_end_session()`, and the hook runner
+
+This makes the lifecycle hard to audit because the answer to "what state
+is this lease in?" depends on several fields at once.
+
+Recent fixes exposed the problem:
+
+- the no-hook deadlock fixed in PR #569
+- unused-lease cleanup behavior adjusted in PR #598
+- the follow-up semantic bug around `before_lease_hook` meaning both
+  "beforeLease is done" and "cleanup must not deadlock", addressed in
+  PR #614
+
+These changes were individually correct, but they all touched the same
+underlying issue: the exporter does not have one explicit lifecycle model
+for leases.
+
+### User Stories
+
+- **As an exporter maintainer**, I want one authoritative lease state so
+  that I can change setup and cleanup logic without reasoning about
+  several loosely-coupled events.
+- **As a contributor adding hook behavior**, I want the rules for
+  readiness and cleanup to be explicit so that adding a new edge case
+  does not create another race.
+- **As an exporter user**, I want leases to transition reliably through
+  setup and teardown so that short-lived leases, unused leases, and hook
+  failures do not leave the exporter stuck.
+
+### Constraints
+
+- No gRPC protocol changes
+- No change to the externally visible meaning of `ExporterStatus`
+- Keep `HookExecutor` focused on executing hooks, not owning exporter
+  lifecycle policy
+- Preserve current behavior for no-hook, before-hook, after-hook, and
+  standalone exporters
+- The refactor must remain compatible with anyio task cancellation
+  semantics
+
+## Proposal
+
+Introduce a dedicated per-lease lifecycle controller, tentatively named
+`LeaseLifecycle`, owned by the exporter and attached to each
+`LeaseContext`.
+
+The lifecycle controller defines an explicit state enum and a small set
+of validated transition methods. Instead of using the same `Event` to
+mean multiple things, the exporter will move the lease through named
+states and expose wait points with dedicated semantics.
+
+### Proposed Internal States
+
+The initial state set should stay intentionally small:
+
+- **`CREATED`** - Lease context exists, but session setup has not
+  completed.
+- **`STARTING`** - Session sockets and Listen stream are being prepared.
+- **`BEFORE_LEASE`** - A configured `before_lease` hook is running.
+- **`READY`** - Connections may be routed and driver calls are allowed.
+- **`ENDING`** - Lease end has been requested; no new work should begin.
+- **`AFTER_LEASE`** - The after-lease phase is running, or being
+  intentionally skipped according to policy.
+- **`RELEASING`** - The exporter is releasing the lease and finalizing
+  cleanup.
+- **`DONE`** - Lease cleanup is complete; the exporter may accept the
+  next lease.
+- **`FAILED`** - Terminal local failure in lifecycle orchestration.
+
+The exporter may still track orthogonal flags such as `has_client`,
+`stop_requested`, or `skip_after_lease_hook`, but these flags should not
+replace the lifecycle state itself.
+
+### Proposed Wait Points
+
+The FSM replaces overloaded wait semantics with dedicated signals:
+
+- **connections ready** - set exactly when the lifecycle enters `READY`
+- **cleanup complete** - set exactly when the lifecycle enters `DONE` or
+  `FAILED`
+
+These signals replace the current overloading of
+`before_lease_hook`/`after_lease_hook_done` as generic coordination
+primitives.
+
+### Proposed Ownership Boundaries
+
+- `Exporter` owns lifecycle state and transition policy
+- `LeaseLifecycle` validates and records transitions
+- `LeaseContext` holds lease resources and lifecycle references
+- `HookExecutor` only executes hook bodies and reports results back to
+  the exporter
+
+In particular, `HookExecutor` should not set lifecycle completion flags
+directly. Instead, the exporter should observe hook completion and apply
+the appropriate transition.
+
+## API / Protocol Changes
+
+No gRPC protocol changes are proposed.
+
+No user-visible YAML configuration changes are proposed.
+
+This JEP changes only internal exporter orchestration.
+
+## Design Decisions
+
+### DD-1: Use an explicit internal FSM instead of adding more event checks
+
+**Alternatives considered:**
+
+1. **Keep the current event-based model** and continue tightening edge
+   cases with narrower condition checks and more tests.
+2. **Introduce an explicit internal FSM** with validated transitions and
+   named wait points.
+
+**Decision:** Introduce an explicit internal FSM.
+
+**Rationale:** The exporter is already acting as a state machine, just
+without a single source of truth. The recent bugs were not caused by
+lack of tests alone; they were caused by the same event representing
+different phases depending on context. An explicit FSM addresses that
+root cause directly.
+
+### DD-2: Keep `ExporterStatus` separate from internal lifecycle state
+
+**Alternatives considered:**
+
+1. **Reuse `ExporterStatus` as the lifecycle state.**
+2. **Keep a separate internal lifecycle enum and derive outward status
+   from it.**
+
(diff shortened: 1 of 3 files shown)


bennyz (approved):

mangelajo:
@kirkbrauer lol, this could be our JEP-0012 then we have to rename the other two? are you ok with merging ? :D

mangelajo:
> @kirkbrauer lol, this could be our JEP-0012 then we have to rename the other two? are you ok with merging ? :D

ahh no this one was stepped up to JEP-0012, ok I will take 0013 for the metrics then :) avoid conflicts.

kirkbrauer (approved):

kirkbrauer:
@ambient-code Please link this to the main README.md file for the JEPs which is missing.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
+# JEP-0012: Explicit Lease Lifecycle State Machine for Exporter
+
+| Field              | Value |
+|--------------------|-------|
+| **JEP**            | 0012 |
+| **Title**          | Explicit Lease Lifecycle State Machine for Exporter |
+| **Author(s)**      | @mangelajo |
```
raballew: [MEDIUM] The Author(s) field contains only `@mangelajo`, but the JEP template specifies the format `@github-handle (Full Name <email>)`. Other JEPs follow this convention too, e.g. JEP-0010 uses `@vtz (Vinicius Zein)`. The README.md index entry has the same gap.

Consider updating to something like `@mangelajo (Miguel Angel Ajo)` or `@mangelajo (Miguel Angel Ajo <email>)` in both the JEP document and the README index, so it stays consistent with the template and existing JEPs.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+- **`BEFORE_LEASE`** - A configured `before_lease` hook is running.
+- **`READY`** - Connections may be routed and driver calls are allowed.
+- **`ENDING`** - Lease end has been requested; no new work should begin.
+- **`AFTER_LEASE`** - The after-lease phase is running, or being
+  intentionally skipped according to policy.
+- **`RELEASING`** - The exporter is releasing the lease and finalizing
+  cleanup.
+- **`DONE`** - Lease cleanup is complete; the exporter may accept the
+  next lease.
+- **`FAILED`** - Terminal local failure in lifecycle orchestration.
```
raballew, line 119: While I agree a FSM makes sense the JEP proposes an FSM but does not include the FSM's most essential artifact: a complete state transition diagram or table. The doc defines 9 states and walks through a few scenarios narratively (happy path, early-end, after-lease-only), but there's no formal enumeration of all valid (source_state, trigger, target_state) triples.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+
+`LeaseContext` would then hold:
+
+- lease identity and session/socket references
+- client metadata
+- status message cache
+- lifecycle controller reference
+
+The goal is to reduce `LeaseContext` from a bag of coordination events to
+a clearer resource holder plus lifecycle object.
```
raballew, line 260: The JEP identifies that the exporter runs concurrent tasks: serve(), handle_lease(), _handle_end_session() and the hook runner. But the JEP never specifies how the FSM handles concurrent transition attempts. Replacing events with an FSM doesn't automatically eliminate races - the existing bugs (#569, #598, #614) were all race conditions iiuc.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+The initial state set should stay intentionally small:
+
+- **`CREATED`** - Lease context exists, but session setup has not
+  completed.
+- **`STARTING`** - Session sockets and Listen stream are being prepared.
+- **`BEFORE_LEASE`** - A configured `before_lease` hook is running.
+- **`READY`** - Connections may be routed and driver calls are allowed.
+- **`ENDING`** - Lease end has been requested; no new work should begin.
+- **`AFTER_LEASE`** - The after-lease phase is running, or being
+  intentionally skipped according to policy.
```
raballew, line 114: BEFORE_LEASE and AFTER_LEASE states involve running hooks, but the JEP doesn't specify what happens when a hook raises an exception or times out.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+1. session and Listen startup begin
+2. a real `before_lease` hook starts
+3. the lease ends before the hook finishes
+4. cleanup must not start `after_lease` early
+5. the exporter must still eventually complete cleanup
+
+Under the FSM, this is expressed as:
+
+- lifecycle stays in `BEFORE_LEASE`
+- an `end_requested` intent is recorded
```
raballew, line 293: The "Early-End During Before-Hook" section describes recording an "end_requested intent" while lifecycle stays in BEFORE_LEASE. This is effectively a sub-state. Consider adding BEFORE_LEASE_ENDING as a distinct state.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+- **`CREATED`** - Lease context exists, but session setup has not
+  completed.
+- **`STARTING`** - Session sockets and Listen stream are being prepared.
+- **`BEFORE_LEASE`** - A configured `before_lease` hook is running.
+- **`READY`** - Connections may be routed and driver calls are allowed.
+- **`ENDING`** - Lease end has been requested; no new work should begin.
+- **`AFTER_LEASE`** - The after-lease phase is running, or being
+  intentionally skipped according to policy.
+- **`RELEASING`** - The exporter is releasing the lease and finalizing
+  cleanup.
```
raballew, line 116: RELEASING is described as "releasing the lease and finalizing cleanup" but the JEP doesn't specify what happens if the release operation fails (e.g., controller unreachable), whether RELEASING has a timeout, or how it interacts with anyio task cancellation.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+- client `EndSession` and controller lease-end both converge on the same
+  terminal lifecycle state
+
+### Regression Tests
+
+Preserve focused regressions for the scenarios addressed by:
+
+- PR #569
+- PR #598
+- PR #614
```
raballew: [MEDIUM] The test plan covers unit, integration, and regression testing but has no E2E testing section. JEP-0010 includes an E2E test section. For an internal refactor of this scope, at least one E2E test exercising the full lease lifecycle through the gRPC interface would provide the strongest behavioral preservation signal.

raballew, line 369: For an FSM, property-based testing or exhaustive state exploration would be a natural fit: changes to the transition table would automatically produce test cases rather than relying on manually-enumerated tests. Worth considering as an addition to the test strategy, especially since it would align well with the goal of making the FSM fully verifiable.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
+# JEP-0012: Explicit Lease Lifecycle State Machine for Exporter
+
+| Field              | Value |
+|--------------------|-------|
+| **JEP**            | 0012 |
```
raballew: [LOW] JEP-0000 states numbers are assigned sequentially. No JEP-0011 exists and the gap isn't documented. Consider either documenting why 0011 was skipped in the README, or using 0011 if it is unassigned.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
```diff
@@ -0,0 +1,462 @@
...
+coverage.
+
+That approach can solve individual failures, but it does not address the
+root design problem that multiple events and status fields currently
+encode overlapping notions of readiness, cleanup, and completion.
+
+Another rejected alternative is to introduce a generic third-party FSM
+framework. The exporter does not need a full workflow engine; it needs a
+small explicit lifecycle model that is easy to read and debug in this
+codebase.
```
raballew: [LOW] The Rejected Alternatives section describes two approaches (continue current design, third-party FSM). For an FSM proposal, coroutine-based coordination and context-manager-stack approaches are natural alternatives worth brief mention to show they were considered.

raballew (changes requested):

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
mangelajo: not necessary, we are just explaining that we want to move to an FSM not to define all states and transitions

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
mangelajo: same.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
mangelajo: I think that is part of the work that needs to be done to implement this, the result of moving this JEP to implemented should include python code and the transition table.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
mangelajo: This is implementation detail.

Normally on an FSM software implementation would be responsible for looking at any signals that trigger a transition to a next state. I think this is again implementation detail.

We could go deep in doing the exercise here on a document, but what is the value? , I'd rather do that analysis at implementation time.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
mangelajo: Yeah, that would make sense.

python/docs/source/internal/jeps/JEP-0012-exporter-lease-lifecycle-fsm.md
mangelajo: I don't like FSMs with transition tables, that was the first attempt of claude at imlpementing it, and it's very hard to follow, read and maintain.

mangelajo (commented):
Agreeing with some of the comments, but in general I'd prefer to agree on the intent "we want to move this to FSM", and not try to design the state machine before hand. 

Just settle the agreement now, and during implementation we can document the state machine, perhaps even try to get auto documenting python code.

bennyz:
## Suggested improvements from protocol-level race condition work

We recently completed FizzBee formal verification of the client-exporter-controller protocol ([protocol spec](protocol/spec/jumpstarter_protocol_fixed.fizz)), proving two race conditions and implementing fixes across Go/Python. This work adds protocol-level constraints that JEP-0012's internal FSM should be aware of. The JEP's core design is sound — these are additive improvements.

### 1. New section: Protocol Constraints

The controller now actively validates ExporterStatus transitions. This is a hard constraint on the internal FSM's "ExporterStatus projection" (DD-2):

**Controller transition validation** — `ValidExporterTransitions` in `controller/api/v1alpha1/exporter_types.go` rejects invalid transitions with gRPC `FAILED_PRECONDITION`:

```
Offline               → Available
Available             → BeforeLeaseHook, LeaseReady, Offline
BeforeLeaseHook       → LeaseReady, BeforeLeaseHookFailed
LeaseReady            → AfterLeaseHook
AfterLeaseHook        → Available, AfterLeaseHookFailed
BeforeLeaseHookFailed → Offline
AfterLeaseHookFailed  → Offline
```

Note: `Available → LeaseReady` is allowed for no-hook exporters that skip `BEFORE_LEASE_HOOK`.

**Fallible status reporting** — `_report_status()` now returns `bool`. `True` = controller accepted, `False` = rejected (FAILED_PRECONDITION). The internal FSM must check this before advancing.

**TOCTOU mitigation** — Controller re-checks exporter status + lease before `sendToListener`. Reported ExporterStatus is load-bearing for connection routing correctness.

### 2. Suggested design decisions

**DD-6: Internal FSM must produce valid ExporterStatus projections**

The internal FSM is a refinement of the controller's transition table. Every internal state transition that changes the projected ExporterStatus must map to a valid transition in `ValidExporterTransitions`. The FizzBee `StatusAgreement` invariant (verified across 13640 states) proves this pattern prevents status divergence between exporter and controller.

**DD-7: Handle `_report_status()` rejection as first-class FSM concern**

When `_report_status()` returns `False`, the FSM must NOT advance to the target state. It should stay in the current state and either retry (if lease is still active) or transition to FAILED. This is the Python-side equivalent of the FizzBee model's `last_report_ok` check pattern.

**DD-8: Drain pending connections during ENDING state**

FizzBee proved that tokens queued via `SendDial` during `LEASE_READY` can persist in the controller's queue after lease expiry. The `ENDING` state should stop accepting new dial tokens and drain in-flight RPCs before proceeding to `AFTER_LEASE`, preventing driver connections from being active when after-lease hooks run.

### 3. Explicit internal transition table

The JEP lists states and a suggested flow, but an exhaustive transition table would prevent the same class of bugs the JEP cites. Key rows to note:

| From | To | Condition |
|------|-----|-----------|
| BEFORE_LEASE | READY | Only if `_report_status(LEASE_READY)` returns True |
| BEFORE_LEASE | ENDING | `end_requested` recorded; waits for hook completion before firing |
| STARTING | READY | No before_lease hook configured (skips BEFORE_LEASE) |

### 4. ExporterStatus projection mapping

DD-2 says "derive outward status from internal state" — a concrete mapping table would make the constraint from §1 verifiable:

| Internal State | Projected ExporterStatus |
|---------------|-------------------------|
| BEFORE_LEASE | BEFORE_LEASE_HOOK |
| READY | LEASE_READY |
| ENDING | (none — internal only) |
| AFTER_LEASE | AFTER_LEASE_HOOK |
| RELEASING | AVAILABLE (with release_lease) |
| FAILED | *_FAILED or OFFLINE (phase-dependent) |

**No-hook path note:** Exporters without a `before_lease` hook go directly from STARTING → READY, projecting `Available → LeaseReady`. The controller allows this. The JEP should decide whether to keep this shortcut or always project `BEFORE_LEASE_HOOK` for consistency.

### 5. Additional acceptance criteria

- ExporterStatus projection produces only transitions valid per `ValidExporterTransitions`
- `_report_status()` returning False prevents FSM state advance
- ENDING drains pending connections before AFTER_LEASE
- Internal transition table is exhaustive (every from/to pair explicitly allowed or disallowed)

### 6. Prior art additions

- **FizzBee formal verification** — models the protocol across 13640 states with 6 invariants. Key insight: `StatusAgreement` tracks `last_accepted` (what controller told exporter) rather than `controller.status` (which changes unilaterally via lease expiry). This same principle should guide the internal FSM's relationship with `_report_status()`.
- **Controller-side transition validation** — `ValidExporterTransitions` makes the controller an active validator, not just a passive recorder.
- **Fallible `_report_status()`** — state update moved AFTER RPC, rejection handling added.
- **Dial TOCTOU mitigation** — re-checks before `sendToListener`.

### 7. Test plan additions

- ExporterStatus projection validity for all lifecycle paths (with hooks, without hooks, after-only)
- `_report_status()` rejection prevents FSM advance
- Connection drain timeout in ENDING state
- Consider extending the FizzBee model to include internal lifecycle states, or use Hypothesis property tests for projection validity

PR #648, opened 2026-05-05
feat(cli): CLI Authentication in Dev Spaces (Device Flow)

mickume:
# CLI Authentication in Dev Spaces (Device Flow)

## Problem

The `jumpstarter-cli` and `caib-cli` use the OIDC **Authorization Code flow with a local
callback server**. When the user runs the CLI from a local machine this works fine: Keycloak
redirects the browser to `http://localhost:PORT/callback`, and the CLI's local HTTP server
picks up the authorization code.

When the CLI runs inside **OpenShift Dev Spaces** (or any cloud/container-based IDE), the
flow breaks:

1. The browser is redirected to `http://localhost:PORT/callback`.
2. `localhost` in the browser refers to the **user's laptop**, not the Dev Spaces pod.
3. Nothing is listening on that port on the laptop → browser shows a connection error.

This PR adds the Device Flow, see https://github.com/mickume/jumpstarter/issues/13


python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+            try:
+                token = await run_sync(
+                    lambda c=client: c.fetch_token(
+                        token_endpoint,
+                        grant_type=DEVICE_FLOW_GRANT_TYPE,
+                        device_code=device_code,
+                    )
+                )
+                return token
+            except Exception as e:
```
mangelajo, line 243: can we be more specific about the exceptions we capture?. Capturing wide exceptions and then matching text is a bad pattern and could end up hiding the wrong exceptions.

mangelajo (commented):
very nice! thank you!!! :)

Just a couple of comments

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+
+        raise click.ClickException("Device code expired (timeout). Please try again.")
+
+
+def should_use_device_flow(device_flow_flag: bool | None = None) -> bool:
+    if device_flow_flag is not None:
+        return device_flow_flag
+    if os.environ.get(JMP_OIDC_DEVICE_FLOW, "").strip() == "1":
+        return True
+    if os.environ.get("VSCODE_INJECTION", "").strip() == "1":
```
raballew, line 264: I am wondering: The `should_use_device_flow()` function checks `VSCODE_INJECTION` for auto-detection, but `VSCODE_INJECTION` is set in any VS Code remote container environment (Remote-SSH, Remote-Containers, Codespaces), not just OpenShift Dev Spaces.  @mangelajo @mickume Should we consider checking `DEVWORKSPACE_ID` instead (or in addition), since that is specific to Dev Spaces. Also, `VSCODE_INJECTION` is used as a raw string literal while the other env var (`JMP_OIDC_DEVICE_FLOW`) is properly declared as a constant in `env.py`.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+            f"\nOpen the following URL in your browser and enter the code:"
+            f"\n  {verification_uri}"
+            f"\n  Code: {user_code}\n"
+        )
+
+        token_endpoint = config["token_endpoint"]
+        deadline = time.monotonic() + expires_in
+
+        while time.monotonic() < deadline:
+            await anyio.sleep(interval)
```
raballew, line 231: [HIGH] `interval = device_data.get("interval", DEVICE_FLOW_POLL_INTERVAL)` accepts any value from the server without a minimum bound. When `interval=0`, `anyio.sleep(0)` yields but does not actually wait, creating a tight polling loop that could last up to `expires_in` seconds (default 600). RFC 8628 Section 3.2 specifies a minimum default of 5 seconds. The test `test_successful_device_flow` explicitly sets `"interval": 0`, which masks this in tests but would cause a tight loop in production. A malicious or buggy OIDC server could weaponize this.

Suggested fix: `interval = max(device_data.get("interval", DEVICE_FLOW_POLL_INTERVAL), DEVICE_FLOW_POLL_INTERVAL)`. Update tests that set `"interval": 0` to use the constant or a realistic value.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+                if "authorization_pending" in error_msg:
+                    continue
+                if "slow_down" in error_msg:
+                    interval += 5
+                    continue
+                if "expired_token" in error_msg or "expired" in error_msg:
+                    raise click.ClickException("Device code expired. Please try again.") from e
+                if "access_denied" in error_msg:
+                    raise click.ClickException("Authorization was denied by the user.") from e
+                raise click.ClickException(f"Device flow token request failed: {e}") from e
```
raballew, line 254: [HIGH] Error handling uses `except Exception as e` with substring checks like `if "authorization_pending" in error_msg` against the full `str(e)`. Authlib's `OAuthError.__str__()` formats as `"error_code: description_text"`, so substring matching checks both the error code and the free-text description. Concrete false positive: `OAuthError('invalid_client', 'Client certificate has expired')` produces `str(e) = 'invalid_client: Client certificate has expired'` where `"expired" in str(e)` is True, causing a TLS certificate error to be misreported as "Device code expired".

Authlib's `OAuthError` exposes a structured `.error` attribute containing only the RFC error code. Consider using `error_code = getattr(e, "error", None) or str(e)` and matching against that instead.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+        device_endpoint = config.get("device_authorization_endpoint")
+        if not device_endpoint:
+            raise click.ClickException(
+                "OIDC provider does not support Device Authorization Grant. "
+                "Ensure 'oauth2DeviceAuthorizationGrantEnabled' is enabled on the OIDC client."
+            )
+
+        ssl_context: ssl.SSLContext | bool = False if self.insecure_tls else _get_ssl_context()
+        connector = aiohttp.TCPConnector(ssl=ssl_context)
+        async with aiohttp.ClientSession(connector=connector) as session:
```
raballew, line 202: The `aiohttp.ClientSession` created here has no `timeout` parameter. If the endpoint is unresponsive, the CLI hangs indefinitely.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+            f"\n  Code: {user_code}\n"
+        )
+
+        token_endpoint = config["token_endpoint"]
+        deadline = time.monotonic() + expires_in
+
+        while time.monotonic() < deadline:
+            await anyio.sleep(interval)
+
+            client = self.client()
```
raballew, line 233: `client = self.client()` inside the `while` loop creates a new `OAuth2Session` (wrapping a `requests.Session` with connection pool) on every poll iteration without closing the previous one. All other grant methods in this class create the client once before the operation.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc_test.py
```diff
@@ -1,6 +1,21 @@
 import ssl
```
raballew, line 1: [MEDIUM] Missing test for `slow_down` response. The `slow_down` error handling at oidc.py:247-249 increases the polling interval by 5 seconds per RFC 8628 Section 3.5, but no test in `TestDeviceCodeGrant` covers this behavior. Tests exist for `authorization_pending` and `access_denied` but not `slow_down`. Consider adding `test_increases_interval_on_slow_down` that raises `Exception("slow_down")` on the first call, then succeeds on the second, and asserts `anyio.sleep` was called with the increased interval (original + 5).

raballew, line 1: [MEDIUM] Missing test for `expired_token` error response. The `expired_token` handling at oidc.py:250-251 raises `click.ClickException` with "Device code expired" message, but no corresponding test exists. Consider adding `test_raises_on_expired_token` that raises `Exception("expired_token")` from `fetch_token` and asserts `click.ClickException` with "Device code expired" message.

raballew, line 1: [MEDIUM] Missing test for device code timeout (deadline expiry). The timeout path at oidc.py:256 (`raise click.ClickException("Device code expired (timeout)")`) is reached when the polling loop exceeds the deadline, but no test covers this path. Consider adding `test_raises_on_timeout` that patches `time.monotonic` to simulate deadline expiry and asserts the timeout error is raised.

python/packages/jumpstarter-cli/jumpstarter_cli/login.py
```diff
@@ -382,7 +385,10 @@ async def relogin_client(config: ClientConfigV1Alpha1):
             except Exception:
                 pass
 
-        tokens = await oidc.authorization_code_grant()
+        if should_use_device_flow():
+            tokens = await oidc.device_code_grant()
+        else:
+            tokens = await oidc.authorization_code_grant()
```
raballew, line 391: The `login.py` changes add the `device_flow` parameter and `should_use_device_flow` branching in both `login()` and `relogin_client()`, but no test file was modified or created for these integration points. Consider adding tests verifying that the login command routes to device flow when `should_use_device_flow` returns True, and that `relogin_client` routes to the correct grant method.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+                if resp.status != 200:
+                    body = await resp.text()
+                    raise click.ClickException(f"Device authorization request failed (HTTP {resp.status}): {body}")
+                device_data = await resp.json()
+
+        device_code = device_data["device_code"]
+        user_code = device_data["user_code"]
+        verification_uri = device_data.get("verification_uri_complete") or device_data["verification_uri"]
+        expires_in = device_data.get("expires_in", 600)
+        interval = device_data.get("interval", DEVICE_FLOW_POLL_INTERVAL)
```
raballew, line 219: After a successful HTTP 200, the JSON response is accessed via direct key lookups (`device_data["device_code"]`, `device_data["user_code"]`) without schema validation. If the OIDC provider returns 200 with unexpected JSON (e.g., missing `device_code`), the caller gets a raw `KeyError` with no context. Compare with `_validate_auth_config_payload` in `login.py` which explicitly validates required fields.

Suggested fix:
```python
for field in ("device_code", "user_code"):
    if field not in device_data:
        raise click.ClickException(
            f"Device authorization response missing required field '{field}'"
        )
```

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+                if "slow_down" in error_msg:
+                    interval += 5
+                    continue
+                if "expired_token" in error_msg or "expired" in error_msg:
+                    raise click.ClickException("Device code expired. Please try again.") from e
+                if "access_denied" in error_msg:
+                    raise click.ClickException("Authorization was denied by the user.") from e
+                raise click.ClickException(f"Device flow token request failed: {e}") from e
+
+        raise click.ClickException("Device code expired (timeout). Please try again.")
```
raballew, line 256: The `device_code_grant` method is ~67 lines handling endpoint discovery/validation, device code request, user output, and token polling with error handling. Consider extracting the device authorization request into `_request_device_code()` and the polling loop into `_poll_for_token()` to keep things clean.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +187,84 @@ async def callback(request):
...
+            async with session.post(
+                device_endpoint,
+                data={
+                    "client_id": self.client_id,
+                    "scope": " ".join(self._scopes()),
+                },
+            ) as resp:
+                if resp.status != 200:
+                    body = await resp.text()
+                    raise click.ClickException(f"Device authorization request failed (HTTP {resp.status}): {body}")
```
raballew, line 212: The full HTTP response body from the OIDC server is included in the error message: `raise click.ClickException(f"Device authorization request failed (HTTP {resp.status}): {body}")`. A misconfigured OIDC server could return verbose error details.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc_test.py
```diff
@@ -23,3 +38,253 @@ class TestGetSslContext:
...
+    def test_explicit_false_flag(self) -> None:
+        assert should_use_device_flow(False) is False
+
+    def test_env_var_jmp_oidc_device_flow(self, monkeypatch) -> None:
+        monkeypatch.setenv("JMP_OIDC_DEVICE_FLOW", "1")
+        assert should_use_device_flow(None) is True
+
+    def test_env_var_vscode_injection(self, monkeypatch) -> None:
+        monkeypatch.setenv("VSCODE_INJECTION", "1")
+        assert should_use_device_flow(None) is True
```
raballew, line 56: `test_env_var_vscode_injection` sets `VSCODE_INJECTION=1` but does not clear `JMP_OIDC_DEVICE_FLOW`. If the test runner's environment has `JMP_OIDC_DEVICE_FLOW=1`, the test passes for the wrong reason. Consider adding `monkeypatch.delenv("JMP_OIDC_DEVICE_FLOW", raising=False)` to `test_env_var_vscode_injection` and vice versa to improve test isolation.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc_test.py
raballew, line 1: [LOW] Missing test for the unknown/generic error path in the polling loop. The catch-all `raise click.ClickException(f"Device flow token request failed: {e}")` at oidc.py:254 is not tested. Consider adding `test_raises_on_unknown_token_error` that triggers `Exception("server_error")` and verifies the generic `ClickException`.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -46,6 +50,12 @@ def opt_oidc(f):
         default=True,
         help="Request offline_access scope (refresh token)",
     )
+    @click.option(
+        "--device-flow/--no-device-flow",
+        "device_flow",
+        default=None,
+        help="Use OAuth 2.0 Device Authorization Grant (auto-detected in Dev Spaces)",
+    )
```
raballew, line 58: The `opt_oidc` decorator gained a `--device-flow/--no-device-flow` option, but no test verifies it is correctly wired (default `None`, parameter name `device_flow`). Consider adding a test that applies `opt_oidc` to a test function, invokes it via Click's test runner, and verifies the parameter propagation.

raballew (changes requested):

PR #649, opened 2026-05-06
oci: support existing OCI credentials

bennyz:
Fallback to using existing OCI credentials, if OCI_USERNAME/OCI_PASSWORD are not set

### DIFF
--- a/python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
+++ b/python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
@@ -1,4 +1,3 @@
-import os
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Dict, Optional
@@ -51,8 +50,7 @@ class RideSXClient(FlasherClient, CompositeClient):
                 self.storage.write_from_path(filename, path_buf, operator=operator)
             except Exception as e:
                 raise RideSXFlashError(
-                    f"Failed to transfer '{file_path}' to exporter storage as '{filename}' "
-                    f"(scheme={operator_scheme})"
+                    f"Failed to transfer '{file_path}' to exporter storage as '{filename}' (scheme={operator_scheme})"
                 ) from e
         else:
             self.logger.info(f"File {filename} already exists in storage with matching hash, skipping upload")
@@ -78,9 +76,7 @@ class RideSXClient(FlasherClient, CompositeClient):
             try:
                 remote_files[partition] = self._upload_file_if_needed(file_path, operator)
             except Exception as e:
-                raise RideSXFlashError(
-                    f"Failed preparing image for partition '{partition}': {file_path}"
-                ) from e
+                raise RideSXFlashError(f"Failed preparing image for partition '{partition}': {file_path}") from e
 
         self.logger.info("Checking for fastboot devices on Exporter...")
         detection_result = self.call("detect_fastboot_device", 5, 2.0)
@@ -201,8 +197,7 @@ class RideSXClient(FlasherClient, CompositeClient):
                     if "/" in before_colon:
                         # registry/path:tag — likely an OCI ref missing the oci:// prefix
                         raise click.ClickException(
-                            f"OCI URLs must start with oci://, got: {path}\n"
-                            f"Usage: j storage flash oci://{path}"
+                            f"OCI URLs must start with oci://, got: {path}\nUsage: j storage flash oci://{path}"
                         )
                     # partition:something — likely a partition:path mapping
                     raise click.ClickException(
@@ -280,22 +275,14 @@ class RideSXClient(FlasherClient, CompositeClient):
 
         return self._execute_flash_operation(_flash_operation, power_off=power_off)
 
-    def _read_oci_credentials(self):
-        """Read OCI registry credentials from environment variables.
+    def _read_oci_credentials(self, oci_url: str):
+        """Read OCI registry credentials from environment variables or auth files."""
+        from jumpstarter.common.oci import resolve_oci_credentials
 
-        Returns:
-            Tuple of (username, password), both None if not set.
-
-        Raises:
-            click.ClickException: If only one of username/password is set.
-        """
-        username = os.environ.get("OCI_USERNAME")
-        password = os.environ.get("OCI_PASSWORD")
+        username, password = resolve_oci_credentials(oci_url)
 
         if bool(username) != bool(password):
-            raise click.ClickException(
-                "OCI authentication requires both OCI_USERNAME and OCI_PASSWORD environment variables"
-            )
+            raise click.ClickException("OCI authentication requires both username and password")
 
         return username, password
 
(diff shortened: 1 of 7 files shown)


python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
```diff
@@ -280,22 +275,14 @@ def _flash_operation():
...
 
-        Returns:
-            Tuple of (username, password), both None if not set.
-
-        Raises:
-            click.ClickException: If only one of username/password is set.
-        """
-        username = os.environ.get("OCI_USERNAME")
-        password = os.environ.get("OCI_PASSWORD")
+        username, password = resolve_oci_credentials(oci_url)
```
raballew, line 282: Consider wrapping the test body with `with patch("jumpstarter.common.oci.resolve_oci_credentials", return_value=(None, None)):` to keep it isolated, similar to what qemu's `test_flash_oci_no_credentials` does. I dont want to leak my credentials testing this.

python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
```diff
@@ -280,22 +275,14 @@ def _flash_operation():
...
 
         if bool(username) != bool(password):
-            raise click.ClickException(
-                "OCI authentication requires both OCI_USERNAME and OCI_PASSWORD environment variables"
-            )
+            raise click.ClickException("OCI authentication requires both username and password")
 
         return username, password
```
raballew, line 287: Worth adding tests in the driver that verify: (1) delegation to `resolve_oci_credentials` with the OCI URL, and (2) the partial-credential error when `bool(username) != bool(password)`.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -0,0 +1,221 @@
...
+        Registry hostname (with port if present), e.g. ``quay.io`` or
+        ``registry.example.com:5000``.
+    """
+    url = oci_url
+    if url.startswith("oci://"):
+        url = url[len("oci://") :]
+
+    # Strip any tag or digest suffix for parsing purposes
+    # e.g. "quay.io/org/repo:tag" -> we just need "quay.io"
+    # The registry is the first path component
```
raballew, line 44: Several comments here just restate what the code does. The function and variable names already communicate the intent pretty clearly. The priority-ordering comments are fine since they clarify non-obvious ordering, but the rest could be removed to reduce noise.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -0,0 +1,221 @@
...
+    # Remove tag/digest if someone passed just "registry:tag" with no path
+    if "/" not in url and ":" in registry:
+        # Could be registry:port or image:tag — if the part after : is numeric
+        # it's a port, otherwise it's a tag on a Docker Hub image
+        host_port = registry.split(":", 1)
+        if host_port[1].isdigit():
+            return registry  # registry:port
+        return "docker.io"  # bare image like "ubuntu:latest"
+
+    return registry
```
raballew, line 57: `parse_oci_registry` doesn't handle digest references. For something like `oci://ubuntu@sha256:abc123` (bare image with digest, no slash), the `@sha256:abc123` part gets included in the registry string and the `:` triggers port/tag disambiguation, corrupting the result.

Stripping the digest early with `url = url.split("@")[0]` before parsing should fix it. Images with a path (e.g., `oci://myregistry/repo@sha256:abc123`) already work fine since the `@` ends up in the path segment.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -0,0 +1,221 @@
...
+    # Try to find a matching entry — normalize all keys for comparison
+    for key, value in auths.items():
+        if _normalize_registry(key) == registry:
+            # The "auth" field is base64(username:password)
+            auth_b64 = value.get("auth")
+            if auth_b64:
+                try:
+                    decoded = base64.b64decode(auth_b64).decode("utf-8")
+                    username, password = decoded.split(":", 1)
+                    return username, password
```
raballew, line 145: When base64 auth decodes to `:password`, `split(":", 1)` returns `("", "password")`. The function returns that as-is, but its contract says it returns `(None, None)` when credentials aren't found. The caller's `if username and password:` check prevents an actual runtime bug, but the function's return violates its own contract.

Adding `if not username or not password: continue` after the split would keep the return value consistent with the documented behavior.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -0,0 +1,221 @@
...
+    # 3. Podman fallback: ~/.config/containers/auth.json
+    paths.append(Path.home() / ".config" / "containers" / "auth.json")
+
+    # 4. Docker: $DOCKER_CONFIG/config.json or ~/.docker/config.json
+    docker_config = os.environ.get("DOCKER_CONFIG")
+    if docker_config:
+        paths.append(Path(docker_config) / "config.json")
+    paths.append(Path.home() / ".docker" / "config.json")
+
+    return paths
```
raballew, line 87: `_get_auth_file_paths` is always mocked in existing tests. There's no test that actually verifies it reads `REGISTRY_AUTH_FILE`, `XDG_RUNTIME_DIR`, `DOCKER_CONFIG` correctly or produces the right ordered path list. A parameterized test class setting various env var combinations and checking the returned paths would close that gap.

python/packages/jumpstarter/jumpstarter/common/oci_test.py
```diff
@@ -0,0 +1,253 @@
...
+                assert username == "fileuser"
+                assert password == "filepass"
+
+    def test_returns_none_when_no_source(self):
+        env_clean = {k: v for k, v in os.environ.items() if k not in ("OCI_USERNAME", "OCI_PASSWORD")}
+        with patch.dict(os.environ, env_clean, clear=True):
+            with patch("jumpstarter.common.oci._get_auth_file_paths", return_value=[]):
+                username, password = resolve_oci_credentials("oci://quay.io/org/image:latest")
+                assert username is None
+                assert password is None
```
raballew, line 253: There's no test for the `except (ValueError, UnicodeDecodeError)` branch at oci.py:146-148 that handles invalid base64 in auth entries. A test with an auth file containing garbage base64 in an auth value would verify that malformed entries are properly skipped rather than crashing.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -0,0 +1,221 @@
...
+    }
+    # Strip port if default
+    host = registry.split(":")[0]
+    if host in docker_hub_aliases or host == "docker.io":
+        return "docker.io"
+
+    return registry
+
+
+def _lookup_credentials_in_auth_data(auth_data: dict, registry: str) -> tuple[str | None, str | None]:
```
raballew, line 122: The type annotation for `auth_data` is bare `dict` rather than `dict[str, Any]`.

mangelajo (approved):

raballew (commented):
@bennyz please follow up on this.

bennyz:
posted #654

PR #654, opened 2026-05-07
oci: add OciCredentials model and harden credential resolution

bennyz:
Introduce a Pydantic OciCredentials model that normalizes empty/whitespace strings and enforces both-or-neither validation. Update all callers to use the new model. Add tests for digest parsing, Docker Hub bare images, and whitespace env var handling.

follow up on #649

### DIFF
--- a/python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
+++ b/python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
@@ -1,3 +1,5 @@
+from __future__ import annotations
+
 import base64
 import hashlib
 import json
@@ -12,7 +14,11 @@ from contextlib import contextmanager
 from dataclasses import dataclass
 from pathlib import Path, PosixPath
 from queue import Queue
+from typing import TYPE_CHECKING
 from urllib.parse import urlparse
+
+if TYPE_CHECKING:
+    from jumpstarter.common.oci import OciCredentials
 
 import click
 import pexpect
@@ -137,7 +143,9 @@ class BaseFlasherClient(FlasherClient, CompositeClient):
 
         if headers:
             headers = self._validate_header_dict(headers)
-        oci_username, oci_password = self._resolve_oci_credentials(path, oci_username, oci_password)
+        oci_creds = self._resolve_oci_credentials(path, oci_username, oci_password)
+        oci_username = oci_creds.username
+        oci_password = oci_creds.plain_password
         should_download_to_httpd = True
         image_url = ""
         original_http_url = None
@@ -1291,34 +1299,17 @@ class BaseFlasherClient(FlasherClient, CompositeClient):
 
         return token
 
-    def _validate_oci_credentials(self, username: str | None, password: str | None) -> tuple[str | None, str | None]:
-        if username is not None:
-            username = username.strip()
-        if password is not None:
-            password = password.strip()
-
-        if username == "":
-            username = None
-        if password == "":
-            password = None
-
-        if bool(username) != bool(password):
-            raise click.ClickException(
-                "OCI authentication requires both OCI_USERNAME and OCI_PASSWORD "
-                "environment variables (or both oci_username and oci_password arguments)"
-            )
-
-        return username, password
-
     def _resolve_oci_credentials(
         self, path: PathBuf, username: str | None, password: str | None
-    ) -> tuple[str | None, str | None]:
-        if username is None and password is None and path.startswith("oci://"):
-            from jumpstarter.common.oci import resolve_oci_credentials
+    ) -> "OciCredentials":
+        from jumpstarter.common.oci import OciCredentials, resolve_oci_credentials
 
-            username, password = resolve_oci_credentials(str(path))
-
-        return self._validate_oci_credentials(username, password)
+        if username is not None or password is not None or str(path).startswith("oci://"):
+            try:
+                return resolve_oci_credentials(str(path), username=username, password=password)
+            except ValueError as err:
+                raise click.ClickException(str(err)) from err
+        return OciCredentials()
 
     def _fls_oci_auth_env(self, path: PathBuf, creds_file: str | None) -> str:
         if not str(path).startswith("oci://") or not creds_file:
--- a/python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
+++ b/python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
@@ -105,19 +105,9 @@ class QemuFlasher(FlasherInterface, Driver):
         if not oci_url.startswith("oci://"):
             raise ValueError(f"OCI URL must start with oci://, got: {oci_url}")
 
-        # If explicit credentials were provided, validate immediately
-        if oci_username or oci_password:
-            if bool(oci_username) != bool(oci_password):
-                raise ValueError("OCI authentication requires both username and password")
-        else:
-            # Fall back to env vars, then container auth files
-            from jumpstarter.common.oci import resolve_oci_credentials
+        from jumpstarter.common.oci import resolve_oci_credentials
 
-            oci_username, oci_password = resolve_oci_credentials(oci_url)
-            if oci_username and oci_password:
-                self.logger.info("Using OCI registry credentials from environment or auth file")
-            elif oci_username or oci_password:
-                raise ValueError("OCI authentication requires both username and password")
+        creds = resolve_oci_credentials(oci_url, username=oci_username, password=oci_password)
 
         target_path = str(self.parent.validate_partition(partition))
 
@@ -130,10 +120,10 @@ class QemuFlasher(FlasherInterface, Driver):
         fls_cmd = [fls_binary, "from-url", oci_url, target_path]
 
         fls_env = None
-        if oci_username and oci_password:
+        if creds.is_authenticated:
             fls_env = os.environ.copy()
-            fls_env["FLS_REGISTRY_USERNAME"] = oci_username
-            fls_env["FLS_REGISTRY_PASSWORD"] = oci_password
+            fls_env["FLS_REGISTRY_USERNAME"] = creds.username
+            fls_env["FLS_REGISTRY_PASSWORD"] = creds.plain_password
 
         self.logger.info(f"Running fls: {' '.join(fls_cmd)}")
 
--- a/python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
+++ b/python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
@@ -1,6 +1,11 @@
+from __future__ import annotations
+
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Dict, Optional
+from typing import TYPE_CHECKING, Dict, Optional
+
+if TYPE_CHECKING:
+    from jumpstarter.common.oci import OciCredentials
 
 import click
 from jumpstarter_driver_composite.client import CompositeClient
@@ -275,16 +280,11 @@ class RideSXClient(FlasherClient, CompositeClient):
 
         return self._execute_flash_operation(_flash_operation, power_off=power_off)
 
-    def _read_oci_credentials(self, oci_url: str):
+    def _read_oci_credentials(self, oci_url: str) -> OciCredentials:
         """Read OCI registry credentials from environment variables or auth files."""
         from jumpstarter.common.oci import resolve_oci_credentials
 
-        username, password = resolve_oci_credentials(oci_url)
-
-        if bool(username) != bool(password):
-            raise click.ClickException("OCI authentication requires both username and password")
-
-        return username, password
+        return resolve_oci_credentials(oci_url)
 
     def _flash_oci_auto_impl(
         self,
@@ -292,7 +292,7 @@ class RideSXClient(FlasherClient, CompositeClient):
         partitions: Dict[str, str] | None = None,
     ):
         """Core implementation of OCI flash without wrapper logic."""
-        oci_username, oci_password = self._read_oci_credentials(oci_url)
+        creds = self._read_oci_credentials(oci_url)
 
         self.logger.info("Checking for fastboot devices on Exporter...")
         detection_result = self.call("detect_fastboot_device", 5, 2.0)
@@ -307,8 +307,8 @@ class RideSXClient(FlasherClient, CompositeClient):
             "flash_oci_image",
             oci_url,
             partitions,
-            oci_username,
-            oci_password,
+            creds.username,
+            creds.plain_password,
         )
 
         # Display FLS output to user
--- a/python/packages/jumpstarter/jumpstarter/common/oci.py
+++ b/python/packages/jumpstarter/jumpstarter/common/oci.py
@@ -13,20 +13,143 @@ Supported auth file locations (checked in order):
 """
 
 import base64
+import binascii
 import json
 import logging
 import os
+import re
+import tomllib
+from functools import lru_cache
 from pathlib import Path
+from typing import Any
 from urllib.parse import urlparse
 
+from pydantic import BaseModel, ConfigDict, Field, SecretStr, ValidationError, field_validator, model_validator
+
 logger = logging.getLogger(__name__)
+
+__all__ = [
+    "OciCredentials",
+    "parse_oci_registry",
+    "read_auth_file_credentials",
+    "resolve_oci_credentials",
+]
+
+
+class OciCredentials(BaseModel):
+    """Resolved OCI registry credentials.
+
+    Construction enforces that username and password are both set or both None.
+    Passing only one raises ``ValidationError``.
+    """
+
+    model_config = ConfigDict(frozen=True)
+
+    username: str | None = None
+    password: SecretStr | None = Field(default=None)
+
+    @field_validator("username", "password", mode="before")
+    @classmethod
+    def _normalize_empty(cls, v: str | SecretStr | None) -> str | None:
+        if isinstance(v, SecretStr):
+            v = v.get_secret_value()
+        if isinstance(v, str):
+            v = v.strip()
+            return v if v else None
+        return v
+
+    @property
+    def plain_password(self) -> str | None:
+        """Return the raw password string, or None if unset."""
+        return self.password.get_secret_value() if self.password else None
+
+    @property
+    def is_authenticated(self) -> bool:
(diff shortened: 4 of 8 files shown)


python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -39,20 +80,22 @@ def parse_oci_registry(oci_url: str) -> str:
...
-        host_port = registry.split(":", 1)
-        if host_port[1].isdigit():
-            return registry  # registry:port
-        return "docker.io"  # bare image like "ubuntu:latest"
+    if "/" not in url:
+        if ":" in registry:
+            host_port = registry.split(":", 1)
+            if host_port[1].isdigit():
+                return registry  # registry:port
+            return "docker.io"  # bare image like "ubuntu:latest"
```
raballew: there are also local images that are just in a registry per se but only in container storage - not relevant for credentials but i this this default to docker.io is not correct - it depends on the hosts setup iirc

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -105,19 +105,14 @@ async def flash_oci(
         if not oci_url.startswith("oci://"):
             raise ValueError(f"OCI URL must start with oci://, got: {oci_url}")
 
-        # If explicit credentials were provided, validate immediately
         if oci_username or oci_password:
             if bool(oci_username) != bool(oci_password):
                 raise ValueError("OCI authentication requires both username and password")
```
raballew: This code path still uses the old pattern

python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
```diff
@@ -279,12 +279,8 @@ def _read_oci_credentials(self, oci_url: str):
...
         from jumpstarter.common.oci import resolve_oci_credentials
 
-        username, password = resolve_oci_credentials(oci_url)
-
-        if bool(username) != bool(password):
-            raise click.ClickException("OCI authentication requires both username and password")
-
-        return username, password
+        creds = resolve_oci_credentials(oci_url)
+        return creds.username, creds.password
```
raballew: `_read_oci_credentials` calls `resolve_oci_credentials(oci_url)` which returns an `OciCredentials` model, then immediately destructures it into `return creds.username, creds.password`. The caller `_flash_oci_auto_impl` then works with bare `(username, password)` tuples, discarding the typed model benefits like `is_authenticated` and `repr=False` password protection.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -13,48 +13,143 @@
...
+    Passing only one raises ``ValidationError``.
+    """
 
-    Returns:
-        Registry hostname (with port if present), e.g. ``quay.io`` or
-        ``registry.example.com:5000``.
+    model_config = ConfigDict(frozen=True)
+
+    username: str | None = None
+    password: str | None = Field(default=None, repr=False)
```
raballew: The password field uses `repr=False` which hides it from `repr()` output, but the value is still a plain `str` that could show up in tracebacks or debug sessions. Pydantic's `SecretStr` would provide defense-in-depth by masking the value in all string contexts.

raballew (changes requested):

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -105,19 +105,18 @@ async def flash_oci(
...
-        # If explicit credentials were provided, validate immediately
+        from pydantic import SecretStr
+
+        from jumpstarter.common.oci import OciCredentials, resolve_oci_credentials
+
         if oci_username or oci_password:
-            if bool(oci_username) != bool(oci_password):
-                raise ValueError("OCI authentication requires both username and password")
+            password = SecretStr(oci_password) if oci_password is not None else None
+            creds = OciCredentials(username=oci_username, password=password)
```
raballew: The QEMU driver doesn't catch `ValidationError` from `OciCredentials`, so if someone passes partial credentials (e.g. username but no password), they get a raw multi-line Pydantic validation dump instead of a clean error message. The flashers driver already handles this correctly by catching `ValidationError` and wrapping it in a `ClickException`. Consider wrapping the `OciCredentials` construction here with `try/except ValidationError` and re-raising as `ValueError("OCI authentication requires both username and password")`. This is kind of important as soon as we have telemetry enabled as I am sure we do not want to leak those credentials into logs.

python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
```diff
@@ -1291,34 +1299,25 @@ def _validate_bearer_token(self, token: str | None) -> str | None:
...
-        if username is None and password is None and path.startswith("oci://"):
-            from jumpstarter.common.oci import resolve_oci_credentials
+            ) from err
 
-            username, password = resolve_oci_credentials(str(path))
+        if not creds.is_authenticated and path.startswith("oci://"):
+            return resolve_oci_credentials(str(path))
 
-        return self._validate_oci_credentials(username, password)
+        return creds
```
raballew: Each driver implements credential resolution differently: flashers constructs `OciCredentials` with `ValidationError` handling then falls back, QEMU constructs without error handling, and ridesx delegates entirely with no explicit credential support. Users end up getting different error messages for the same partial-credential mistake depending on which driver they use. Consider extending `resolve_oci_credentials` with optional `username`/`password` parameters so the full precedence chain (explicit -> env -> auth files) lives in one place.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -203,15 +302,19 @@ def resolve_oci_credentials(oci_url: str) -> tuple[str | None, str | None]:
...
+    if password is not None:
+        password = password.strip() or None
+
     if username and password:
         logger.info("Using OCI registry credentials from environment variables")
-        return username, password
+        return OciCredentials(username=username, password=password)
 
     if username or password:
         logger.warning(
```
raballew: [MEDIUM] The credential resolution precedence is split across layers. `resolve_oci_credentials` only implements levels 2-3 (env vars, auth files), while level 1 (explicit args) is independently reimplemented by each driver with different logic. Adding optional `username`/`password` parameters to `resolve_oci_credentials` would consolidate the full three-level precedence in one place.

python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
```diff
@@ -137,7 +143,9 @@ def flash(  # noqa: C901
 
         if headers:
             headers = self._validate_header_dict(headers)
-        oci_username, oci_password = self._resolve_oci_credentials(path, oci_username, oci_password)
+        oci_creds = self._resolve_oci_credentials(path, oci_username, oci_password)
+        oci_username = oci_creds.username
+        oci_password = oci_creds.password.get_secret_value() if oci_creds.password else None
```
raballew: [MEDIUM] The expression `creds.password.get_secret_value() if creds.password else None` is duplicated across all three drivers (flashers, QEMU, ridesx). Callers shouldn't need to know `SecretStr` internals. A `plain_password` property on `OciCredentials` would eliminate this repetition.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -203,15 +302,19 @@ def resolve_oci_credentials(oci_url: str) -> tuple[str | None, str | None]:
...
-        are found from any source.
+        OciCredentials with both fields set, or both None.
     """
     username = os.environ.get("OCI_USERNAME")
     password = os.environ.get("OCI_PASSWORD")
 
+    if username is not None:
+        username = username.strip() or None
+    if password is not None:
+        password = password.strip() or None
```
raballew: [MEDIUM] `resolve_oci_credentials` manually strips whitespace from env var values, but `OciCredentials._normalize_empty` already does this via its field validator. The manual stripping is redundant. Consider removing it and restructuring the warning logic to check raw env var values before constructing the model.

python/packages/jumpstarter/jumpstarter/common/oci_test.py
```diff
@@ -240,14 +257,371 @@ def test_partial_env_falls_back_to_auth_file(self, tmp_path):
...
+        conf = tmp_path / "registries.conf"
+        conf.write_text('[registries.search]\nregistries = ["old-format"]\n')
+
+        with patch(
+            "jumpstarter.common.oci._get_registries_conf_paths",
+            return_value=[conf],
+        ):
+            _get_unqualified_search_registries.cache_clear()
+            result = _get_unqualified_search_registries()
+            assert result == ["docker.io"]
```
raballew: TestUnqualifiedSearchRegistries` calls `cache_clear()` before each test but not after. The last test's cached value persists and could affect subsequent test classes. A class-level autouse fixture that clears the cache after each test would prevent this leakage.

python/packages/jumpstarter/jumpstarter/common/oci_test.py
```diff
@@ -240,14 +257,371 @@ def test_partial_env_falls_back_to_auth_file(self, tmp_path):
...
+    def test_resolve_returns_oci_credentials_type(self, tmp_path):
+        auth_path = tmp_path / "auth.json"
+        auth_path.write_text(_make_auth_json({"quay.io": {"auth": _encode_auth("user", "pass")}}))
+
+        env_clean = {k: v for k, v in os.environ.items() if k not in ("OCI_USERNAME", "OCI_PASSWORD")}
+        with patch.dict(os.environ, env_clean, clear=True):
+            with patch("jumpstarter.common.oci._get_auth_file_paths", return_value=[auth_path]):
+                result = resolve_oci_credentials("oci://quay.io/org/image:latest")
+                assert isinstance(result, OciCredentials)
+                assert result.is_authenticated
```
raballew, line 507: There is no test for `OciCredentials(username="user", password="")`

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -119,7 +216,7 @@ def _normalize_registry(registry: str) -> str:
     return registry
 
 
-def _lookup_credentials_in_auth_data(auth_data: dict, registry: str) -> tuple[str | None, str | None]:
+def _lookup_credentials_in_auth_data(auth_data: dict[str, Any], registry: str) -> tuple[str | None, str | None]:
```
raballew: [MEDIUM] `_lookup_credentials_in_auth_data` returns a `tuple[str | None, str | None]` while the caller immediately wraps it into `OciCredentials`. The positional tuple is an implicit contract at a model boundary. Returning `OciCredentials` directly would make the contract explicit.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -13,48 +13,145 @@
...
     if url.startswith("oci://"):
         url = url[len("oci://") :]
 
-    # Strip any tag or digest suffix for parsing purposes
-    # e.g. "quay.io/org/repo:tag" -> we just need "quay.io"
-    # The registry is the first path component
+    # Strip digest before parsing — "ubuntu@sha256:abc" would otherwise
+    # have the colon corrupt port/tag disambiguation.
+    if "@" in url.split("/", 1)[0]:
+        url = url.split("@")[0]
```
raballew: [LOW] The digest-stripping logic uses `if "@" in url.split("/", 1)[0]` which strips on any at-sign, not only digest references. A crafted URL like `oci://private.registry@attacker.com/image:tag` could theoretically misdirect credential lookup. As a defense-in-depth measure, consider tightening this to only strip recognized digest formats (`@sha256:`, `@sha384:`, `@sha512:`).

raballew: Not sure how relevant this is, keeping it here for your reference @bennyz

python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
raballew: Thats a good point but more of a nit.

raballew (changes requested):

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -13,48 +13,150 @@
...
+            with open(conf_path, "rb") as f:
+                data = tomllib.load(f)
+            registries = [r for r in data.get("unqualified-search-registries", []) if isinstance(r, str) and r.strip()]
+            if registries:
+                logger.debug("Read unqualified-search-registries from %s: %s", conf_path, registries)
+                return registries
+        except (OSError, tomllib.TOMLDecodeError) as e:
+            logger.debug("Skipping registries.conf %s: %s", conf_path, e)
+            continue
+    return ["docker.io"]
```
raballew: [MEDIUM] `@lru_cache(maxsize=1)` on `_get_unqualified_search_registries()` returns a `list[str]`. Since the same list reference is returned to every caller, any caller that mutates it (e.g. `.append("evil.io")`) would poison the cached value for everyone else. `_parse_registries_for_url` returns this list directly (lines 131, 133, 139).

No current caller mutates it, but the API contract doesn't prevent it. Consider returning `tuple[str, ...]` instead of `list[str]` to make the cached value immutable. Change the return type annotation and wrap returns in `tuple(...)`.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -13,48 +13,150 @@
...
+
+    @property
+    def is_authenticated(self) -> bool:
+        return self.username is not None and self.password is not None
+
+    @model_validator(mode="after")
+    def _check_both_or_neither(self) -> "OciCredentials":
+        if bool(self.username) != bool(self.password):
+            raise ValueError("OCI authentication requires both username and password")
+        return self
```
raballew, line 74: The `plain_password` encapsulates `self.password.get_secret_value() if self.password else None`, but callers in other driver packages still use `creds.password.get_secret_value() if creds.password else None` directly (e.g. `flashers/client.py`, `qemu/driver.py`). I think you should migrate to match behaviour.

python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
```diff
@@ -279,20 +279,15 @@ def _read_oci_credentials(self, oci_url: str):
         """Read OCI registry credentials from environment variables or auth files."""
         from jumpstarter.common.oci import resolve_oci_credentials
 
-        username, password = resolve_oci_credentials(oci_url)
-
-        if bool(username) != bool(password):
-            raise click.ClickException("OCI authentication requires both username and password")
-
-        return username, password
+        return resolve_oci_credentials(oci_url)
```
raballew, line 287: Missing return type annotation. The flasher client's equivalent method correctly uses `-> "OciCredentials"`. Also add the corresponding `TYPE_CHECKING` import guard.

python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
```diff
@@ -307,8 +302,8 @@ def _flash_oci_auto_impl(
             "flash_oci_image",
             oci_url,
             partitions,
-            oci_username,
-            oci_password,
+            creds.username,
+            creds.plain_password,
```
raballew, line 311: [MEDIUM] `creds.plain_password` exercises the `OciCredentials.plain_password` property, but both `test_flash_oci_auto_success` and `test_flash_oci_auto_error_cases` mock `resolve_oci_credentials` to return `OciCredentials()` (empty/unauthenticated), so the password is always `None` in tests. The branch where `plain_password` actually calls `get_secret_value()` is never exercised.

Consider adding a test that mocks `resolve_oci_credentials` to return `OciCredentials(username="user", password=SecretStr("pass"))` and verifies that `flash_oci_image` is called with the plain-text password value.

python/packages/jumpstarter/jumpstarter/common/oci.py
```diff
@@ -140,82 +242,109 @@ def _lookup_credentials_in_auth_data(auth_data: dict, registry: str) -> tuple[st
...
 
             # Some auth files use separate username/password fields
             username = value.get("username")
             password = value.get("password")
             if username and password:
-                return username, password
+                return OciCredentials(username=username, password=password)
 
-    return None, None
+    return OciCredentials()
```
raballew, line 265: [LOW] `_lookup_credentials_in_auth_data` calls `value.get("auth")` assuming each `value` in `auths.items()` is a dict. If an auth file contains a non-dict value (e.g. `{"auths": {"quay.io": "string"}}`), an `AttributeError` is raised and not caught, halting credential resolution.

Consider adding `if not isinstance(value, dict): continue` at the start of the loop body.

python/packages/jumpstarter/jumpstarter/common/oci_test.py
```diff
@@ -240,14 +257,417 @@ def test_partial_env_falls_back_to_auth_file(self, tmp_path):
         env_partial["OCI_USERNAME"] = "partialuser"
         with patch.dict(os.environ, env_partial, clear=True):
             with patch("jumpstarter.common.oci._get_auth_file_paths", return_value=[auth_path]):
-                username, password = resolve_oci_credentials("oci://quay.io/org/image:latest")
-                assert username == "fileuser"
-                assert password == "filepass"
+                result = resolve_oci_credentials("oci://quay.io/org/image:latest")
+                assert result.username == "fileuser"
+                assert result.password.get_secret_value() == "filepass"
```
raballew, line 275: [LOW] `test_partial_env_falls_back_to_auth_file` only tests `OCI_USERNAME` set without `OCI_PASSWORD`. The symmetric case (`OCI_PASSWORD` set without `OCI_USERNAME`) is not tested.

Consider parametrizing or adding a companion test covering the `OCI_PASSWORD`-only scenario.

python/packages/jumpstarter/jumpstarter/common/oci.py
raballew: defense in depth, not strictly necessary but worth reading ;)

raballew (changes requested):

raballew (approved):

raballew:
@ambient-code rebase onto main to resolve merge conflicts

PR #656, opened 2026-05-07
feat(cli): OAuth 2.0 Device Authorization Grant, silent token refresh, and TLS fixes

mickume:
## Summary

- **OAuth 2.0 Device Authorization Grant**: Add `--device-flow` / `--no-device-flow` CLI option to `jmp login`, enabling headless authentication via the RFC 8628 device code flow. Auto-detected when running inside VS Code Dev Spaces (`VSCODE_INJECTION=1`) or when `JMP_OIDC_DEVICE_FLOW=1` is set.
- **Silent token refresh**: Background token-expiry monitor no longer prints noisy status messages; warnings are downgraded to `logger.debug`/`logger.info`. Token refresh also works correctly with `--insecure-tls` now by passing the TLS flag to the OIDC config.
- **Automatic retry on expired tokens**: `handle_exceptions_with_reauthentication` now silently re-authenticates and retries the wrapped function once, instead of asking the user to manually retry.
- **Warning suppression**: Suppress `urllib3.InsecureRequestWarning` (when user already opted into `--insecure-tls`) and transient `authlib.jose` deprecation warnings.
- **Lint fixes**: Resolved ruff E402 (import ordering), C901 (function complexity), and W293 (trailing whitespace) violations.

## Test plan

- [ ] Verify `jmp login --device-flow` prompts with a verification URI and polls until the user completes browser auth
- [ ] Verify auto-detection in a Dev Spaces environment (`VSCODE_INJECTION=1`)
- [ ] Verify `jmp login` without `--device-flow` still uses the authorization code grant
- [ ] Verify expired-token re-auth + retry works transparently (no "please try again" message)
- [ ] Verify `--insecure-tls` no longer emits `InsecureRequestWarning` on every request
- [ ] Run existing OIDC and exception unit tests (`oidc_test.py`, `exceptions_test.py`, `shell_test.py`)


python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -16,7 +18,27 @@
...
+)
+
+# When the user opts into --insecure-tls, we set verify=False on the requests session.
+# urllib3 emits InsecureRequestWarning for every such request; suppress it since the
+# user has already acknowledged the risk.
+warnings.filterwarnings(
+    "ignore",
+    message=r"Unverified HTTPS request is being made.*",
+    module=r"urllib3\..*",
+)
```
raballew, line 38: The `InsecureRequestWarning` suppression runs unconditionally at module import time, which means any code that imports `oidc` silently hides legitimate TLS errors even when the user never passed `--insecure-tls`.

Consider moving the `warnings.filterwarnings` call into `Config.client()` or `Config.__init__`, gated on `self.insecure_tls`, so the warning is only suppressed when the user explicitly opted in.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc_test.py
```diff
@@ -1,6 +1,21 @@
...
+    should_use_device_flow,
+)
+
+
+@pytest.fixture
+def anyio_backend():
+    return "asyncio"
 
 
 class TestConfigInsecureTls:
```
raballew, line 21: No tests covering the `expired_token` error path or the deadline-exceeded timeout path in `device_code_grant`. Both are critical failure modes.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc_test.py
```diff
@@ -23,3 +38,253 @@ class TestGetSslContext:
     def test_returns_ssl_context(self) -> None:
```
raballew, line 38: There is no test verifying the interval actually increases. A regression here would silently violate the RFC.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +205,84 @@ async def callback(request):
...
+                if "authorization_pending" in error_msg:
+                    continue
+                if "slow_down" in error_msg:
+                    interval += 5
+                    continue
+                if "expired_token" in error_msg or "expired" in error_msg:
+                    raise click.ClickException("Device code expired. Please try again.") from e
+                if "access_denied" in error_msg:
+                    raise click.ClickException("Authorization was denied by the user.") from e
+                raise click.ClickException(f"Device flow token request failed: {e}") from e
```
raballew, line 272: The authlib library typically provides structured `OAuthError` with an `.error` attribute. Consider checking for specific exception types and using structured attributes, falling back to string matching only for unknown types.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +205,84 @@ async def callback(request):
...
+            f"\n  Code: {user_code}\n"
+        )
+
+        token_endpoint = config["token_endpoint"]
+        deadline = time.monotonic() + expires_in
+
+        while time.monotonic() < deadline:
+            await anyio.sleep(interval)
+
+            client = self.client()
```
raballew, line 251: The device flow polling loop creates a new `self.client()` (OAuth2Session wrapping a requests.Session with a connection pool) on every iteration.

python/packages/jumpstarter-cli/jumpstarter_cli/login.py
```diff
@@ -382,7 +385,10 @@ async def relogin_client(config: ClientConfigV1Alpha1):
             except Exception:
                 pass
```
raballew, line 386: When `refresh_token_grant` fails in `relogin_client`, a bare `except Exception: pass` silently swallows the error with no logging. This makes debugging refresh failures during re-authentication very difficult.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -177,6 +205,84 @@ async def callback(request):
             lambda: client.fetch_token(config["token_endpoint"], authorization_response=authorization_response)
         )
 
+    async def device_code_grant(self):
```
raballew, line 208: This is a lot of responsibility in one class.

Consider extracting grant-specific logic into standalone functions or strategy classes. For example, `device_code_grant` could be a standalone function that receives a Config and returns tokens.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc_test.py
```diff
@@ -23,3 +38,253 @@ class TestGetSslContext:
...
+        oidc_config = {
+            "token_endpoint": "https://auth.example.com/token",
+            "device_authorization_endpoint": "https://auth.example.com/device",
+        }
+        device_response = {
+            "device_code": "test-device-code",
+            "user_code": "ABCD-1234",
+            "verification_uri": "https://auth.example.com/device",
+            "expires_in": 600,
+            "interval": 0,
```
raballew, line 120: This is essentially dead code. Using a non-zero interval (e.g., 5) and verifying mock sleep was called with the correct value would make these tests more meaningful.

raballew (changes requested):
Some inconsistencies but a good first second try.

python/packages/jumpstarter-cli-common/jumpstarter_cli_common/oidc.py
```diff
@@ -46,6 +68,12 @@ def opt_oidc(f):
         default=True,
         help="Request offline_access scope (refresh token)",
     )
+    @click.option(
+        "--device-flow/--no-device-flow",
+        "device_flow",
+        default=None,
+        help="Use OAuth 2.0 Device Authorization Grant (auto-detected in Dev Spaces)",
```
kirkbrauer, line 75: Oh nice! I didn't know this would be auto-detected in DevSpaces.

PR #662, opened 2026-05-11
fix: preserve percent-encoding in presigned URLs to prevent signature invalidation

bennyz:
yarl.URL() normalizes %XX sequences (e.g. %40 → @), which breaks S3/CloudFront presigned URL signatures that are computed over the encoded form. This adds `_make_url(encoded=True)` to preserve encoding, manual redirect following to avoid aiohttp re-encoding Location headers, and an original_url passthrough so already-signed HTTP URLs bypass OpenDAL's presign (which strips query params).

### DIFF
--- a/python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/adapter.py
+++ b/python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/adapter.py
@@ -64,9 +64,19 @@ async def OpendalAdapter(
     path: str,  # file path in storage backend relative to the storage root
     mode: Literal["rb", "wb"] = "rb",  # binary read or binary write mode
     compression: Compression | None = None,  # compression algorithm
+    original_url: str | None = None,  # original HTTP URL, bypasses OpenDAL presign to avoid path re-encoding
 ):
+    if original_url is not None and compression is not None:
+        raise ValueError("compression is not supported with direct HTTP URLs (original_url)")
+    if mode == "wb" and original_url is not None:
+        raise ValueError("original_url is not supported in write mode")
+    if mode == "rb" and compression is None and original_url is not None:
+        yield PresignedRequestResource(
+            headers={}, url=original_url, method="GET"
+        ).model_dump(mode="json")
+        return
     # if the access mode is binary read, and the storage backend supports presigned read requests
-    if mode == "rb" and operator.capability().presign_read and compression is None:
+    elif mode == "rb" and operator.capability().presign_read and compression is None:
         # create presigned url for the specified file with a 60 second expiration
         presigned = await operator.to_async_operator().presign_read(path, expire_second=60)
         yield PresignedRequestResource(
--- a/python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/client.py
+++ b/python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/client.py
@@ -51,7 +51,7 @@ def operator_for_path(path: PathBuf) -> tuple[PathBuf, Operator, str]:
         - the operator for the given path
         - the scheme of the operator.
     """
-    if type(path) is str and path.startswith(("http://", "https://")):
+    if isinstance(path, str) and path.startswith(("http://", "https://")):
         parsed_url = urlparse(path)
         operator = Operator("http", root="/", endpoint=f"{parsed_url.scheme}://{parsed_url.netloc}")
         return Path(parsed_url.path), operator, "http"
@@ -79,10 +79,13 @@ class OpendalFile:
         """
         Write into remote file with content from local file
         """
+        original_url = None
         if operator is None:
+            if isinstance(path, str) and path.startswith(("http://", "https://")):
+                original_url = path
             path, operator, _ = operator_for_path(path)
 
-        with OpendalAdapter(client=self.client, operator=operator, path=path) as handle:
+        with OpendalAdapter(client=self.client, operator=operator, path=path, original_url=original_url) as handle:
             return self.__write(handle)
 
     @validate_call(validate_return=True, config=ConfigDict(arbitrary_types_allowed=True))
@@ -630,10 +633,15 @@ class FlasherClient(FlasherClientInterface, DriverClient):
         compression: Compression | None,
     ):
         """Flash image to DUT"""
+        original_url = None
         if operator is None:
+            if isinstance(image, str) and image.startswith(("http://", "https://")):
+                original_url = image
             image, operator, _ = operator_for_path(image)
 
-        with OpendalAdapter(client=self, operator=operator, path=image, mode="rb", compression=compression) as handle:
+        with OpendalAdapter(
+            client=self, operator=operator, path=image, mode="rb", compression=compression, original_url=original_url
+        ) as handle:
             return self.call("flash", handle, target)
 
     def flash(
@@ -763,10 +771,15 @@ class StorageMuxFlasherClient(FlasherClient, StorageMuxClient):
 
         self.host()
 
+        original_url = None
         if operator is None:
+            if isinstance(path, str) and path.startswith(("http://", "https://")):
+                original_url = path
             path, operator, _ = operator_for_path(path)
 
-        with OpendalAdapter(client=self, operator=operator, path=path, mode="rb", compression=compression) as handle:
+        with OpendalAdapter(
+            client=self, operator=operator, path=path, mode="rb", compression=compression, original_url=original_url
+        ) as handle:
             try:
                 return self.write(handle)
             finally:
--- a/python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/driver_test.py
+++ b/python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/driver_test.py
@@ -12,6 +12,7 @@ from opendal import Operator
 
 from .common import PresignedRequest
 from .driver import MockFlasher, MockStorageMux, MockStorageMuxFlasher, Opendal
+from jumpstarter.client.core import DriverError
 from jumpstarter.common.utils import serve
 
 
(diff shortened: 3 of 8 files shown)


mangelajo (approved):

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,51 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+                    async with AiohttpStreamReaderStream(reader=resp.content) as stream:
+                        yield ProgressStream(stream=stream, logging=True)
+                        return
+                location = resp.headers.get("Location", "")
+                if not location:
+                    raise RuntimeError(
+                        f"Presigned HTTP GET redirect missing Location header for {current_url}"
+                    )
+                current_url = location
+        raise RuntimeError(f"Too many redirects ({max_redirects}) for {url}")
```
raballew: [HIGH] The `_presigned_get` method follows redirects manually but forwards the same `headers` dict to every hop, regardless of whether the destination origin changes. Standard HTTP clients strip sensitive headers like `Authorization` and `x-amz-security-token` on cross-origin redirects to prevent credential leakage. Since this manual redirect loop bypasses that protection, if OpenDAL-generated presigned request headers contain auth tokens and a redirect points to a different domain, those tokens leak.

Suggestion: compare the origin of the redirect target against the original URL before following, and strip `Authorization` and `x-amz-*` headers on cross-origin redirects.

python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/client.py
```diff
@@ -47,7 +47,7 @@ def _upload_file_if_needed(self, file_path: str, operator: Operator | None = Non
                 self.logger.info(f"Uploading {file_path} to storage as {filename}")
 
             try:
-                self.storage.write_from_path(filename, path_buf, operator=operator)
+                self.storage.write_from_path(filename, file_path)
```
raballew: [HIGH] The PR changes the upload call from `self.storage.write_from_path(filename, path_buf, operator=operator)` to `self.storage.write_from_path(filename, file_path)`, which silently drops the caller-provided `operator`. When `operator` is not None (provided via `flash_local -> flash_images -> _upload_file_if_needed`), it is still used for hash-checking at line 43 but dropped for the actual upload. The downstream `write_from_path` will then auto-detect the operator via `operator_for_path`, creating a generic one without the caller's custom credentials or settings.

python/packages/jumpstarter/jumpstarter/driver/base.py
raballew: [HIGH] The two defensive error paths in `_presigned_get` have no test coverage: (1) the `max_redirects=10` safety limit that raises RuntimeError on redirect loops, and (2) the RuntimeError for redirect responses missing the Location header. These are new branches in a new public method.

Suggestion: add two tests: one that simulates a redirect loop (server responds with 302 pointing back to itself) and asserts RuntimeError after exceeding the limit, and one with a 302 response missing the Location header asserting RuntimeError.

python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/client.py
```diff
@@ -79,10 +79,13 @@ def write_from_path(self, path: PathBuf, operator: Operator | None = None):
         """
         Write into remote file with content from local file
         """
+        original_url = None
         if operator is None:
+            if isinstance(path, str) and path.startswith(("http://", "https://")):
+                original_url = path
             path, operator, _ = operator_for_path(path)
```
raballew, line 86: Maybe extract a helper function that encapsulates the `original_url` detection and `operator_for_path` call, returning a tuple like `(path, operator, scheme, original_url)`.

python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/adapter.py
```diff
@@ -64,9 +64,15 @@ async def OpendalAdapter(
     path: str,  # file path in storage backend relative to the storage root
     mode: Literal["rb", "wb"] = "rb",  # binary read or binary write mode
     compression: Compression | None = None,  # compression algorithm
+    original_url: str | None = None,  # original HTTP URL, bypasses OpenDAL presign to avoid path re-encoding
 ):
+    if mode == "rb" and compression is None and original_url is not None:
+        yield PresignedRequestResource(
+            headers={}, url=original_url, method="GET"
+        ).model_dump(mode="json")
+        return
```
raballew, line 77: Restructure so the `original_url` bypass happens at the call site before entering `OpendalAdapter`, or make `operator`/`path` explicitly optional when `original_url` is provided. Raise an explicit error when `original_url` and `compression` are both set.

python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/driver_test.py
```diff
@@ -322,3 +322,164 @@ def test_copy_and_rename_tracking(tmp_path):
...
+            )
+            assert received_paths[0] == "/hop1"
+            assert "%40" in received_paths[1], (
+                f"Hop 2 received decoded path {received_paths[1]!r}"
+            )
+            assert "%40" in received_paths[2], (
+                f"Hop 3 received decoded path {received_paths[2]!r}"
+            )
+        finally:
+            server.shutdown()
```
raballew, line 490: All redirect tests use HTTP 302 only. Redirect status codes 307 and 308 have different semantics (they must preserve the HTTP method). The PR claims to handle all five redirect status codes (301, 302, 303, 307, 308), but test coverage exists only for 302.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,51 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+            ) as resp:
+                if resp.status not in (301, 302, 303, 307, 308):
+                    async with AiohttpStreamReaderStream(reader=resp.content) as stream:
+                        yield ProgressStream(stream=stream, logging=True)
+                        return
+                location = resp.headers.get("Location", "")
+                if not location:
+                    raise RuntimeError(
+                        f"Presigned HTTP GET redirect missing Location header for {current_url}"
+                    )
```
raballew, line 294: The defensive RuntimeError for a redirect response missing the Location header is untested. This is a distinct error path from the max-redirects limit. I think you could add a test where the server returns 302 without a Location header and assert that RuntimeError is raised.

python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/client.py
```diff
@@ -79,10 +79,13 @@ def write_from_path(self, path: PathBuf, operator: Operator | None = None):
         """
         Write into remote file with content from local file
         """
+        original_url = None
         if operator is None:
+            if isinstance(path, str) and path.startswith(("http://", "https://")):
+                original_url = path
```
raballew, line 85: All call sites that set `original_url` check only for the `http://`/`https://` scheme prefix, without verifying that the URL contains query parameters. This means plain HTTP URLs (without presigned query strings) also bypass the normal OpenDAL presign path, which could break non-presigned HTTP URL handling. No test distinguishes query-parameter URLs from plain HTTP URLs.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,51 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+                current_url = location
```
raballew, line 296: [LOW] When following redirects, the code uses the Location header value directly as the next URL. If the Location header contains a relative URI, `yarl.URL` with `encoded=True` but no scheme/host will cause an `InvalidUrlClientError`. While presigned URL redirects use absolute URLs in practice, RFC 7231 permits relative Location headers.

Suggestion: resolve relative Location headers against the current request URL using yarl's URL join.

raballew, line 296: [LOW] The redirect loop calls `aiohttp.request()` per hop, which internally creates and destroys a complete `ClientSession` on each iteration. For the primary use case (e.g. CloudFront redirecting to S3), this means 2 full TCP+TLS handshakes instead of 1. No resource leak, but unnecessary connection churn.

Suggestion: use a single `aiohttp.ClientSession` for the entire redirect chain to reuse the connection pool.

python/packages/jumpstarter/jumpstarter/driver/base.py
raballew: [LOW] Error messages in `_presigned_get` include the full presigned URL with cryptographic signature tokens and potentially access key IDs. These could end up in logs. This follows an existing pattern in `_resource_from_presigned` on the base branch, but it would be good to address it here.

Suggestion: truncate or redact query parameters from URLs in error messages.

raballew (changes requested):

python/packages/jumpstarter/jumpstarter/driver/base.py
bennyz: not sure there is an actual attack vector here, but added for good measure

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,67 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+                        yield ProgressStream(stream=stream, logging=True)
+                        return
+                location = resp.headers.get("Location", "")
+                if not location:
+                    raise RuntimeError(
+                        f"Presigned HTTP GET redirect missing Location header for {current_url}"
+                    )
+                current_headers = self._strip_sensitive_headers(current_headers, current_url, location)
+                current_url = location
+        raise RuntimeError(f"Too many redirects ({max_redirects}) for {url}")
```
raballew: Consider sanitizing URLs before embedding them in error messages, e.g.:

```python
from urllib.parse import urlparse, urlunparse
safe_url = urlunparse(urlparse(current_url)._replace(query="[REDACTED]"))
```

he same pattern appears in `_resource_from_presigned` further down. If these errors propagate to logs, temporary credentials get leaked.

python/packages/jumpstarter-driver-opendal/jumpstarter_driver_opendal/adapter.py
```diff
@@ -64,9 +64,17 @@ async def OpendalAdapter(
...
     compression: Compression | None = None,  # compression algorithm
+    original_url: str | None = None,  # original HTTP URL, bypasses OpenDAL presign to avoid path re-encoding
 ):
+    if original_url is not None and compression is not None:
+        raise ValueError("compression is not supported with direct HTTP URLs (original_url)")
+    if mode == "rb" and compression is None and original_url is not None:
+        yield PresignedRequestResource(
+            headers={}, url=original_url, method="GET"
+        ).model_dump(mode="json")
+        return
```
raballew, line 77: Consider adding a guard to protect against fall through

```python
if mode == "wb" and original_url is not None:
    raise ValueError("original_url is not supported in write mode")
```

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,67 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+        dest = yarl.URL(redirect_url)
+        if (orig.scheme, orig.host, orig.port) == (dest.scheme, dest.host, dest.port):
+            return headers
+        return {
+            k: v for k, v in headers.items()
+            if not k.lower().startswith(cls._SENSITIVE_HEADER_PREFIXES)
+        }
+
+    @asynccontextmanager
+    async def _presigned_get(self, url: str, headers, timeout: aiohttp.ClientTimeout, max_redirects: int = 10):
```
raballew: Consider annotating it as `headers: dict[str, str]`.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,67 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+        yarl.URL() normalizes %XX sequences (e.g. %40 → @), which breaks
+        signed redirect URLs (CloudFront, S3) whose signatures cover the
+        encoded form.  Using encoded=True keeps the raw string intact.
+        """
+        return yarl.URL(url, encoded=True)
+
+    _SENSITIVE_HEADER_PREFIXES = ("authorization", "x-amz-", "x-ms-", "x-goog-")
+
+    @classmethod
+    def _strip_sensitive_headers(cls, headers: dict, original_url: str, redirect_url: str) -> dict:
```
raballew: Consider changing to `headers: dict[str, str]` and `-> dict[str, str]` to match actual usage.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -235,22 +236,67 @@ async def _resource_from_client_stream(self, resource_uuid: UUID, content_encodi
...
+    def _make_url(url: str) -> yarl.URL:
+        """Construct a yarl.URL preserving percent-encoding in the path.
+
+        yarl.URL() normalizes %XX sequences (e.g. %40 → @), which breaks
+        signed redirect URLs (CloudFront, S3) whose signatures cover the
+        encoded form.  Using encoded=True keeps the raw string intact.
+        """
+        return yarl.URL(url, encoded=True)
+
+    _SENSITIVE_HEADER_PREFIXES = ("authorization", "x-amz-", "x-ms-", "x-goog-")
```
raballew: Consider adding `"cookie"` and `"proxy-authorization"` to the tuple.

raballew (changes requested):
fix the credential leaks, otherwise only a few nits left

raballew (approved):
looking much better now. i still would improve test coverage for guards and error paths though.

mangelajo (approved):

PR #695, opened 2026-05-19
feat: add internal token rotation for clients

bennyz:
Wire the existing but unused tokenLifetime config into Signer.Token(), add a RotateToken RPC to ClientService so clients can rotate their own internal JWT without being deleted and recreated, and add an admin path (jmp admin rotate client) that works via K8s API for expired tokens.

- `jmp auth rotate` client self-rotates via gRPC (needs valid token)
- `jmp admin rotate client <name> --save` admin rotates via K8s API


mangelajo:
Thanks benny!, this is awesome, can we add a bunch of E2E tests to verify admin and client rotation?, leave docs for later as @raballew is doing a small doc refactor :)

mangelajo (commented):
can we add some E2E tests for the admin and client sides? looks awesome

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/rotate.py
```diff
@@ -0,0 +1,84 @@
...
+                click.echo(f"Token rotated for client '{name}'")
+
+            if save or out is not None:
+                if ClientConfigV1Alpha1.exists(name):
+                    config = ClientConfigV1Alpha1.load(name)
+                    config.token = new_token
+                    ClientConfigV1Alpha1.save(config, out)
+                else:
+                    client_config = await api.get_client_config(name, allow=[], unsafe=False)
+                    ClientConfigV1Alpha1.save(client_config, out)
```
raballew, line 77: Else branch ignores the `new_token` that `rotate_client_token` just returned. If the controller reconciles between the rotate call and the `get_client_config` call, the saved token could be stale. `if` branch (uses `new_token` correctly imho

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/clients.py
```diff
@@ -161,6 +162,30 @@ async def get_client_config(self, name: str, allow: list[str], unsafe=False) ->
...
+        while count < CREATE_CLIENT_COUNT:
+            try:
+                secret = await self.core_api.read_namespaced_secret(secret_name, self.namespace)
+                if secret.data and "token" in secret.data:
+                    return base64.b64decode(secret.data["token"]).decode("utf8")
+            except ApiException as e:
+                if e.status != 404:
+                    raise
+            except Exception:
+                raise
```
raballew: Isnt this functionally identical to not having the handler at all?

controller/internal/oidc/op.go
```diff
@@ -99,11 +106,15 @@ func (k *Signer) Validate(token string) error {
...
+	if lifetime == 0 {
+		lifetime = defaultTokenLifetime
+	}
 	return jwt.NewWithClaims(jwt.SigningMethodES256, jwt.RegisteredClaims{
 		Issuer:    k.issuer,
 		Subject:   subject,
 		Audience:  []string{k.audience},
 		IssuedAt:  jwt.NewNumericDate(time.Now()),
-		ExpiresAt: jwt.NewNumericDate(time.Now().Add(365 * 24 * time.Hour)), // FIXME: rotate keys on expiration
+		ExpiresAt: jwt.NewNumericDate(time.Now().Add(lifetime)),
```
raballew: `IssuedAt` and `ExpiresAt` each call `time.Now()` separately, so the actual expiry window will be slightly longer than `lifetime` by the time elapsed between the two calls.

controller/internal/service/client/v1/client_service.go
```diff
@@ -386,3 +391,56 @@ func (s *ClientService) DeleteLease(ctx context.Context, req *cpb.DeleteLeaseReq
...
+	var expiry *timestamppb.Timestamp
+	if claims.ExpiresAt != nil {
+		expiry = timestamppb.New(claims.ExpiresAt.Time)
+	}
+
+	return &cpb.RotateTokenResponse{
+		Token:  token,
+		Expiry: expiry,
+	}, nil
+}
```
raballew, line 449: add something like `log.Info("token rotated", "client", jclient.Name, "namespace", namespace)` after successful rotation

raballew (changes requested):

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/rotate.py
mangelajo: What do you mean? I am not sure if @bennyz handled it already.

mangelajo (approved):
look at the remaining @raballew comment, I think it could be stale

raballew (approved):

PR #712, opened 2026-05-29
feat: default exporter configs to ~/.config/jumpstarter

### LINKED ISSUE #76: consolidate --save output directories
`jmp admin create exporter --save` stores exporter configs in `/etc/jumpstarter/exporters/`
`jmp admin create client --save` stores clients in  `~/.config/jumpstarter/clients/`

It would be great if both would be stored in `~/.config/jumpstarter/`.

mmahut:
Fixes #76.
  
Exporter configs created via the CLI now default to `~/.config/jumpstarter/exporters/` (consistent with clients), instead of the hardcoded `/etc/jumpstarter/exporters/` which needs root. `/etc` is kept as a read fallback, so existing systemd/container deployments keep working.

### DIFF
--- a/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/delete_test.py
+++ b/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/delete_test.py
@@ -176,17 +176,20 @@ EXPORTER_CONFIG = ExporterConfigV1Alpha1(
 
 @patch.object(ExporterConfigV1Alpha1, "delete")
 @patch.object(ExporterConfigV1Alpha1, "exists")
+@patch.object(ExporterConfigV1Alpha1, "user_config_exists")
 @patch.object(ExportersV1Alpha1Api, "delete_exporter")
 @patch.object(ExportersV1Alpha1Api, "_load_kube_config")
 def test_delete_exporter(
     _mock_load_kube_config,
     mock_delete_exporter: AsyncMock,
+    mock_user_config_exists: Mock,
     mock_config_exists: Mock,
     mock_config_delete: Mock,
 ):
     runner = CliRunner()
 
     # Delete exporter object and config does not exist
+    mock_user_config_exists.return_value = False
     mock_config_exists.return_value = False
     result = runner.invoke(delete, ["exporter", EXPORTER_NAME])
     assert result.exit_code == 0
@@ -195,11 +198,13 @@ def test_delete_exporter(
     mock_delete_exporter.assert_called_once_with(EXPORTER_NAME)
     mock_config_delete.assert_not_called()
 
+    mock_user_config_exists.reset_mock()
     mock_config_exists.reset_mock()
     mock_delete_exporter.reset_mock()
     mock_config_delete.reset_mock()
 
     # Delete exporter object and config exists, delete = n
+    mock_user_config_exists.return_value = True
     mock_config_exists.return_value = True
     result = runner.invoke(delete, ["exporter", EXPORTER_NAME], input="n\n")
     assert result.exit_code == 0
@@ -208,11 +213,13 @@ def test_delete_exporter(
     mock_delete_exporter.assert_called_once_with(EXPORTER_NAME)
     mock_config_delete.assert_not_called()
 
+    mock_user_config_exists.reset_mock()
     mock_config_exists.reset_mock()
     mock_delete_exporter.reset_mock()
     mock_config_delete.reset_mock()
 
     # Delete exporter object and config exists, delete = Y
+    mock_user_config_exists.return_value = True
     mock_config_exists.return_value = True
     result = runner.invoke(delete, ["exporter", EXPORTER_NAME], input="Y\n")
     assert result.exit_code == 0
@@ -221,11 +228,13 @@ def test_delete_exporter(
     mock_delete_exporter.assert_called_once_with(EXPORTER_NAME)
     mock_config_delete.assert_called_with(EXPORTER_NAME)
 
+    mock_user_config_exists.reset_mock()
     mock_config_exists.reset_mock()
     mock_delete_exporter.reset_mock()
     mock_config_delete.reset_mock()
 
     # Delete exporter object nointeractive
+    mock_user_config_exists.return_value = True
     mock_config_exists.return_value = True
     result = runner.invoke(delete, ["exporter", EXPORTER_NAME, "--nointeractive"])
     assert result.exit_code == 0
@@ -234,11 +243,13 @@ def test_delete_exporter(
     mock_delete_exporter.assert_called_once_with(EXPORTER_NAME)
     mock_config_delete.assert_not_called()
 
+    mock_user_config_exists.reset_mock()
     mock_config_exists.reset_mock()
     mock_delete_exporter.reset_mock()
     mock_config_delete.reset_mock()
 
     # Delete exporter object output name
+    mock_user_config_exists.return_value = True
     mock_config_exists.return_value = True
     result = runner.invoke(delete, ["exporter", EXPORTER_NAME, "--nointeractive", "--output", "name"])
     assert result.exit_code == 0
@@ -246,6 +257,7 @@ def test_delete_exporter(
     mock_delete_exporter.assert_called_once_with(EXPORTER_NAME)
     mock_config_delete.assert_not_called()
 
+    mock_user_config_exists.reset_mock()
     mock_config_exists.reset_mock()
     mock_delete_exporter.reset_mock()
     mock_config_delete.reset_mock()
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
@@ -8,9 +8,19 @@ from jumpstarter_cli_common.opt import (
 )
 from jumpstarter_cli_common.print import model_print
 
+from jumpstarter.common.exceptions import ConfigurationError
 from jumpstarter.config.exporter import ExporterConfigV1Alpha1, ObjectMeta
 
-arg_alias = click.argument("alias", default="default")
+
+def _validate_alias_param(ctx, param, value):
+    try:
+        ExporterConfigV1Alpha1.validate_alias(value)
+    except ConfigurationError as e:
+        raise click.BadParameter(str(e)) from e
+    return value
+
+
+arg_alias = click.argument("alias", default="default", callback=_validate_alias_param)
 
 
 @click.group("exporter")
@@ -29,11 +39,9 @@ def config_exporter():
 @arg_alias
 def create_exporter_config(alias, namespace, name, endpoint, token, output: PathOutputType):
     """Create an exporter config."""
-    try:
-        ExporterConfigV1Alpha1.load(alias)
-    except FileNotFoundError:
-        pass
-    else:
+    # Guard against overwriting an existing user-level config (the write target).
+    # A same-named config in the system location (/etc) is allowed to be shadowed.
+    if ExporterConfigV1Alpha1.user_config_exists(alias):
         raise click.ClickException(f'exporter "{alias}" exists')
 
     config = ExporterConfigV1Alpha1(
@@ -57,9 +65,18 @@ def delete_exporter_config(alias, output: PathOutputType):
         ExporterConfigV1Alpha1.load(alias)
     except FileNotFoundError as err:
         raise click.ClickException(f'exporter "{alias}" does not exist') from err
-    path = ExporterConfigV1Alpha1.delete(alias)
+    try:
+        path = ExporterConfigV1Alpha1.delete(alias)
+    except ConfigurationError as err:
+        raise click.ClickException(str(err)) from err
     if output == OutputMode.PATH:
         click.echo(path)
+    if ExporterConfigV1Alpha1.exists(alias):
+        click.echo(
+            f"Warning: {path} deleted, but a system config at "
+            f"{ExporterConfigV1Alpha1.resolve_path(alias)} still exists and will now be used.",
+            err=True,
+        )
 
 
 @config_exporter.command("edit")
--- a/python/packages/jumpstarter/jumpstarter/config/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/config/exporter.py
@@ -1,5 +1,8 @@
 from __future__ import annotations
 
+import errno
+import os
+import tempfile
 from contextlib import asynccontextmanager, contextmanager, suppress
 from pathlib import Path
 from typing import TYPE_CHECKING, Any, ClassVar, Literal, Optional, Self
@@ -9,7 +12,7 @@ import yaml
 from anyio.from_thread import start_blocking_portal
 from pydantic import BaseModel, ConfigDict, Field, RootModel
 
-from .common import ObjectMeta
+from .common import CONFIG_PATH, ObjectMeta
 from .grpc import call_credentials
 from .tls import TLSConfigV1Alpha1
 from jumpstarter.common.exceptions import ConfigurationError, MissingDriverError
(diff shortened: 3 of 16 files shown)


mmahut:
@coderabbitai review

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -148,7 +148,12 @@ def from_str(cls, config: str) -> ExporterConfigV1Alpha1DriverInstance:
 
 
 class ExporterConfigV1Alpha1(BaseModel):
-    BASE_PATH: ClassVar[Path] = Path("/etc/jumpstarter/exporters")
+    # Default location for exporter configs created via the CLI. Lives under the
+    # user config dir (e.g. ~/.config/jumpstarter/exporters), consistent with clients.
+    BASE_PATH: ClassVar[Path] = CONFIG_PATH / "exporters"
+    # Legacy system-wide location, kept as a read fallback so existing deployments
```
mangelajo:
```suggestion
    # System-wide location, kept as a read fallback so production deployments
```

mangelajo: The write path you are fixing is good, it doesn't make any sense to write to /etc/jumpstarter/exporters by default, but.

In production, this is how it's expected, when you have a linux host serving an exporter in production, it should not be a user workload, so this is not a legacy location, just the production location.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,11 +178,29 @@ class ExporterConfigV1Alpha1(BaseModel):
 
     @classmethod
     def _get_path(cls, alias: str):
+        """Return the default write path (user config dir) for an alias."""
         return (cls.BASE_PATH / alias).with_suffix(".yaml")
 
+    @classmethod
+    def _resolve_path(cls, alias: str) -> Path:
+        """Return an alias's config path, preferring the user dir over the legacy system location.
```
mangelajo:
```suggestion
        """Return an alias's config path, preferring the user dir over the production system location.
```

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -188,16 +211,25 @@ def load_path(cls, path: Path):
...
-        config = cls.load_path(cls._get_path(alias))
+        """Load the config for an alias, searching the user dir then the system fallback."""
+        config = cls.load_path(cls._resolve_path(alias))
         config.alias = alias
         return config
 
     @classmethod
     def list(cls) -> ExporterConfigListV1Alpha1:
+        """List exporter configs from the user and system locations (user takes precedence)."""
+        # Aliases in the user config dir take precedence over the legacy system location.
```
mangelajo:
```suggestion
        # Aliases in the user config dir take precedence over the production system location.
```

mangelajo (changes requested):
@mmahut thanks for the commit, it looks good, just some comments. 

Also you need to update the e2e tests as they are failing to locate the files in the system path https://github.com/jumpstarter-dev/jumpstarter/actions/runs/26632549825/job/78506862972?pr=712#step:12:62

Can you update it so we write in /etc/jumpstarter/exporters/ during the tests?, and perhaps make sure that we also test the user based path as well?.

Thanks a lot.

mmahut:
@coderabbitai review

mmahut:
@coderabbitai review

mmahut:
@coderabbitai resume

mmahut:
@mangelajo, thank you so much for your review! I've rebased the branch with the requested changes. I believe the rabbit will pass this time, but sadly he's had enough of me for today :)

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
...
         # Set the config path before saving
         if path is None:
             config.path = cls._get_path(config.alias)
-            config.path.parent.mkdir(parents=True, exist_ok=True)
         else:
             config.path = Path(path)
+        config.path.parent.mkdir(parents=True, exist_ok=True)
         with config.path.open(mode="w") as f:
             yaml.safe_dump(config.model_dump(mode="json", by_alias=True, exclude={"alias", "path"}), f, sort_keys=False)
         return config.path
```
raballew, line 292: `save` writes config files (which contain the `token` field) using `config.path.open(mode="w")` with no explicit permission restriction. The sibling `ClientConfigV1Alpha1.save` already handles this correctly. Suggest mirroring that pattern here.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
...
 
     @classmethod
     def delete(cls, alias: str) -> Path:
-        path = cls._get_path(alias)
+        """Delete the resolved config file for an alias and return its path."""
+        path = cls._resolve_path(alias)
         path.unlink(missing_ok=True)
         return path
```
raballew, line 307: `delete()` should only remove from `_get_path` (user dir). If the alias only exists in the system location, it should warn the user and refuse.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
...
+        config.path.parent.mkdir(parents=True, exist_ok=True)
         with config.path.open(mode="w") as f:
             yaml.safe_dump(config.model_dump(mode="json", by_alias=True, exclude={"alias", "path"}), f, sort_keys=False)
         return config.path
 
     @classmethod
     def delete(cls, alias: str) -> Path:
-        path = cls._get_path(alias)
+        """Delete the resolved config file for an alias and return its path."""
+        path = cls._resolve_path(alias)
```
raballew: Suggested TCs: deleting a config that exists only in the system location, deleting a config that exists in both locations removes only the user-dir copy,  deleting a non-existent alias is a no-op

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
...
+        exists, the user-dir path is returned so callers raise a ``FileNotFoundError`` pointing at
+        the current default location.
+        """
+        user_path = cls._get_path(alias)
+        if user_path.exists():
+            return user_path
+        system_path = (cls.SYSTEM_CONFIG_PATH / alias).with_suffix(".yaml")
+        if system_path.exists():
+            return system_path
+        return user_path
```
raballew, line 216: An alias like `../../tmp/evil` would resolve outside the config directory. Impact is limited and the attacker must be the local CLI user, but one should validate that `alias` contains no path separators (`/`, `\`) and is not `.` or `..`.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
...
+                    if entry.suffix == ".yaml":
+                        aliases.setdefault(entry.stem, None)
         exporters = []
-        with suppress(FileNotFoundError):
-            for entry in cls.BASE_PATH.iterdir():
-                exporters.append(cls.load(entry.stem))
+        for alias in aliases:
+            with suppress(FileNotFoundError):
+                exporters.append(cls.load(alias))
         return ExporterConfigListV1Alpha1(items=exporters)
```
raballew, line 259: Existing tests cover system fallback (single config) and deduplication (same alias in both), but there is no test where user dir has alias "A" and system dir has alias "B" to verify `list()` returns both.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
...
+        system_path = (cls.SYSTEM_CONFIG_PATH / alias).with_suffix(".yaml")
+        if system_path.exists():
+            return system_path
+        return user_path
+
     @classmethod
     def exists(cls, alias: str):
-        return cls._get_path(alias).exists()
+        """Return True if a config for the alias exists in either the user or system location."""
+        return cls._resolve_path(alias).exists()
```
raballew, line 229: Add `-> Path`, `-> bool`, and `-> Self` return type annotations respectively to match `ClientConfigV1Alpha1`.

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
```diff
@@ -29,11 +29,9 @@ def config_exporter():
...
 def create_exporter_config(alias, namespace, name, endpoint, token, output: PathOutputType):
     """Create an exporter config."""
-    try:
-        ExporterConfigV1Alpha1.load(alias)
-    except FileNotFoundError:
-        pass
-    else:
+    # Guard against overwriting an existing user-level config (the write target).
+    # A same-named config in the system location (/etc) is allowed to be shadowed.
+    if ExporterConfigV1Alpha1._get_path(alias).exists():
```
raballew: Calling underscore-prefixed methods creates implicit coupling between `jumpstarter-cli` and internal implementation details of `jumpstarter`. I think expose public methods like `user_config_exists(alias) -> bool` and `resolve_path(alias) -> Path` might work better.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -173,57 +196,89 @@ class ExporterConfigV1Alpha1(BaseModel):
 
     @classmethod
     def _get_path(cls, alias: str):
+        """Return the default write path (user config dir) for an alias."""
         return (cls.BASE_PATH / alias).with_suffix(".yaml")
```
raballew: The PR adds many docstrings whose names already convey their purpose. As a nit, I would suggest to remove the tautological docstrings and keep only those communicating non-obvious behavior.

raballew (changes requested):
Thank you for tackling this issue. I ran into this issue myself multiple times but now I remember why I never found the time to work on this. `save()` and `delete()` are critical findings, the rest is mostly nits.

python/packages/jumpstarter/jumpstarter/config/exporter.py
mmahut: Indeed, I was trying to make coderabbit happy but it is little bit awkward :)

raballew: Yeah, @mangelajo can we tweak it a bit more? I mean for user facing interfaces it makes a lot of sense but this kind of feedback is misleading. I trust human judgement to reject those kinds of ideas but as a first time contributor it might be difficult to guess if "coderabbit happiness" is required for PR acceptance (which, in case it's unclear, is not the case).

mmahut: That’s true. As a first-time contributor, my main goal was just to make the repo checks and tests pass, since that’s what’s usually expected in other projects.

I’m glad to see there are still projects that haven’t lost trust in human judgement :)

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -172,54 +188,122 @@ class ExporterConfigV1Alpha1(BaseModel):
...
+        if not path.exists():
+            system_path = cls.SYSTEM_CONFIG_PATH / f"{alias}.yaml"
+            if system_path.exists():
+                raise ConfigurationError(
+                    f"Exporter config '{alias}' exists only in the system location"
+                    f" '{system_path}' and cannot be deleted."
+                )
+            return path
+        path.unlink()
         return path
```
raballew, line 307: @mangelajo do we want to implement a system-location deletion with appropriate permission checks or keep it as is (the current implementation is already better compared to what we had prior)

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
```diff
@@ -57,9 +65,18 @@ def delete_exporter_config(alias, output: PathOutputType):
...
         click.echo(path)
+    if ExporterConfigV1Alpha1.exists(alias):
+        click.echo(
+            f"Warning: {path} deleted, but a system config at "
+            f"{ExporterConfigV1Alpha1.resolve_path(alias)} still exists and will now be used.",
+            err=True,
+        )
 
 
 @config_exporter.command("edit")
```
raballew, line 82: @mangelajo If we want to implement permission checks etc, i think the tests here must change as well.

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/delete_test.py
```diff
@@ -234,18 +243,21 @@ def test_delete_exporter(
...
     result = runner.invoke(delete, ["exporter", EXPORTER_NAME, "--nointeractive", "--output", "name"])
     assert result.exit_code == 0
     assert result.output == f"exporter.jumpstarter.dev/{EXPORTER_NAME}\n"
     mock_delete_exporter.assert_called_once_with(EXPORTER_NAME)
     mock_config_delete.assert_not_called()
 
+    mock_user_config_exists.reset_mock()
     mock_config_exists.reset_mock()
     mock_delete_exporter.reset_mock()
     mock_config_delete.reset_mock()
```
raballew, line 263: The admin `delete_exporter` test does not cover the branch where `user_config_exists=False` AND `exists=True` with `output=None`, which would trigger the system config warning on stderr.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -172,54 +188,122 @@ class ExporterConfigV1Alpha1(BaseModel):
...
+                os.fsync(f.fileno())
+            os.replace(temp_path, config.path)
+            os.chmod(config.path, 0o600)
+        finally:
+            try:
+                os.unlink(temp_path)
+            except OSError as e:
+                if e.errno != errno.ENOENT:
+                    raise
         return config.path
```
raballew, line 292: A quick assertion in a test like `assert config.path.stat().st_mode & 0o777 == 0o600` after `save()` would confirm the permission behavior. I do not want this to degrade silently over time as those files imo as somewhat sensitive.

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
```diff
@@ -8,9 +8,19 @@
...
+from jumpstarter.common.exceptions import ConfigurationError
 from jumpstarter.config.exporter import ExporterConfigV1Alpha1, ObjectMeta
 
-arg_alias = click.argument("alias", default="default")
+
+def _validate_alias_param(ctx, param, value):
+    try:
+        ExporterConfigV1Alpha1.validate_alias(value)
+    except ConfigurationError as e:
+        raise click.BadParameter(str(e)) from e
```
raballew, line 19:
```python
def _validate_alias_param(ctx: click.Context, param: click.Parameter, value: str) -> str:
```

raballew (changes requested):
@mmahut looks better! just some test coverage missing.

@mangelajo please provide feedback on two of my comments in the latest review or just resolve them if you think its good to go.

python/packages/jumpstarter/jumpstarter/config/exporter.py
```diff
@@ -172,54 +188,122 @@ class ExporterConfigV1Alpha1(BaseModel):
...
         return ExporterConfigListV1Alpha1(items=exporters)
 
     @classmethod
     def dump_yaml(self, config: Self) -> str:
+        """Serialize a config to a YAML string, omitting internal fields (alias, path)."""
         return yaml.safe_dump(config.model_dump(mode="json", by_alias=True, exclude={"alias", "path"}), sort_keys=False)
 
     @classmethod
     def save(cls, config: Self, path: Optional[str] = None) -> Path:
+        """Save the config to disk, defaulting to the user config dir when no path is given."""
```
mangelajo, line 268: 👍

python/packages/jumpstarter/jumpstarter/config/exporter.py
mangelajo: I don't think it's a wide/important use case, we can do it later if somebody misses that in any way, and I suspect nobody will :D

mangelajo (approved):
Look great, the test paul comments would do no harm, but the code looks good.

mangelajo (dismissed):
oh code looks good but I see some tests still failing.

raballew:
this could be related to https://status.redhat.com/incidents/jn6r256zc62c

raballew (approved):
Only nits left.

PR #721, opened 2026-06-01
fix: raise ValueError for unknown label selector operators

### LINKED ISSUE #718: _label_satisfies_expression silently returns False for unknown operators
## Description

`_label_satisfies_expression` silently returns `False` when given an unrecognized operator string (including empty string `''`). This masks bugs in callers -- an invalid operator produces the same result as a legitimate non-match, with no indication that something is wrong.

## Steps to reproduce

```python
from jumpstarter.client.selectors import _label_satisfies_expression

result = _label_satisfies_expression({"key": "value"}, "key", "", ["value"])
# Returns False silently -- no error raised
```

## Expected behavior

Should raise `ValueError` for unrecognized operator strings, so callers are alerted to the invalid input rather than silently getting a `False` match result.

## Actual behavior

Returns `False` without any error. The function only recognizes `"in"`, `"notin"`, `"exists"`, `"!exists"`, and `"!="`, and falls through to `return False` for anything else.

## Location

`python/packages/jumpstarter/jumpstarter/client/selectors.py`, `_label_satisfies_expression` -- the final `return False` should be `raise ValueError(f"unknown label selector operator: {operator!r}")`.

raballew:
## Summary

- `_label_satisfies_expression` now raises `ValueError` for unrecognized operators instead of silently returning `False`
- `selector_contains` falls back to `_label_satisfies_expression` when a required matchExpression has no matching expression in the selector but matchLabels may satisfy it
- Regression tests for empty string, invalid operators, and all valid operator paths

Closes #718

## Test plan

- [x] `_label_satisfies_expression("", ...)` raises `ValueError`
- [x] `_label_satisfies_expression("bogus", ...)` raises `ValueError` with operator in message
- [x] All valid operators (`in`, `notin`, `exists`, `!exists`, `!=`) still work correctly
- [x] `selector_contains` label-to-expression fallback works for `in`/`notin`
- [x] Existing selector tests pass unchanged

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/client/selectors.py
+++ b/python/packages/jumpstarter/jumpstarter/client/selectors.py
@@ -85,10 +85,11 @@ def _label_satisfies_expression(sel_labels: dict[str, str], key: str, operator: 
 
 
 def selector_contains(selector: str, requirements: str) -> bool:
-    """Check if selector contains all criteria from requirements.
+    """Check if selector satisfies all criteria from requirements.
 
-    Returns True if all matchLabels and matchExpressions in `requirements`
-    are present in `selector`.
+    Returns True if all matchLabels in `requirements` are present in `selector`
+    and all matchExpressions in `requirements` are satisfied by `selector`
+    (either by exact match in matchExpressions or by evaluation against matchLabels).
     """
     if not requirements or not requirements.strip():
         return True
--- a/python/packages/jumpstarter/jumpstarter/client/selectors.py
+++ b/python/packages/jumpstarter/jumpstarter/client/selectors.py
@@ -67,9 +67,7 @@ def extract_match_labels_filter(selector: str | None) -> str | None:
     return ",".join(f"{k}={v}" for k, v in match_labels.items())
 
 
-def _label_satisfies_expression(
-    sel_labels: dict[str, str], key: str, operator: str, values: list[str]
-) -> bool:
+def _label_satisfies_expression(sel_labels: dict[str, str], key: str, operator: str, values: list[str]) -> bool:
     if key not in sel_labels:
         return False
     label_value = sel_labels[key]
@@ -106,10 +104,7 @@ def selector_contains(selector: str, requirements: str) -> bool:
     # All required matchExpressions must be satisfied by selector's
     # matchExpressions or matchLabels
     for r_key, r_op, r_vals in req_exprs:
-        found = any(
-            s_key == r_key and s_op == r_op and set(s_vals) == set(r_vals)
-            for s_key, s_op, s_vals in sel_exprs
-        )
+        found = any(s_key == r_key and s_op == r_op and set(s_vals) == set(r_vals) for s_key, s_op, s_vals in sel_exprs)
         if not found:
             found = _label_satisfies_expression(sel_labels, r_key, r_op, r_vals)
         if not found:
--- a/python/packages/jumpstarter/jumpstarter/client/selectors.py
+++ b/python/packages/jumpstarter/jumpstarter/client/selectors.py
@@ -68,6 +68,8 @@ def extract_match_labels_filter(selector: str | None) -> str | None:
 
 
 def _label_satisfies_expression(sel_labels: dict[str, str], key: str, operator: str, values: list[str]) -> bool:
+    if operator == "!exists":
+        return key not in sel_labels
     if key not in sel_labels:
         return False
     label_value = sel_labels[key]
@@ -79,8 +81,6 @@ def _label_satisfies_expression(sel_labels: dict[str, str], key: str, operator: 
         return label_value not in values
     if operator == "notin":
         return label_value not in values
-    if operator == "!exists":
-        return False
     raise ValueError(f"unknown label selector operator: {operator!r}")
 
 
--- a/python/packages/jumpstarter/jumpstarter/client/selectors.py
+++ b/python/packages/jumpstarter/jumpstarter/client/selectors.py
@@ -67,6 +67,23 @@ def extract_match_labels_filter(selector: str | None) -> str | None:
     return ",".join(f"{k}={v}" for k, v in match_labels.items())
 
 
+def _label_satisfies_expression(
+    sel_labels: dict[str, str], key: str, operator: str, values: list[str]
+) -> bool:
+    if key not in sel_labels:
+        return False
+    label_value = sel_labels[key]
+    if operator == "in":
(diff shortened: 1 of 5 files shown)


python/packages/jumpstarter/jumpstarter/client/selectors.py
```diff
@@ -84,13 +101,12 @@ def selector_contains(selector: str, requirements: str) -> bool:
...
+    # matchExpressions or matchLabels
     for r_key, r_op, r_vals in req_exprs:
-        found = False
-        for s_key, s_op, s_vals in sel_exprs:
-            if s_key == r_key and s_op == r_op and set(s_vals) == set(r_vals):
-                found = True
-                break
+        found = any(s_key == r_key and s_op == r_op and set(s_vals) == set(r_vals) for s_key, s_op, s_vals in sel_exprs)
+        if not found:
+            found = _label_satisfies_expression(sel_labels, r_key, r_op, r_vals)
```
maboras-rh, line 108: `ValueError` may be raised here, but it is neither documented nor handled by the callers (`selector_contains` / `filter_by_selector`)

python/packages/jumpstarter/jumpstarter/client/selectors_test.py
maboras-rh: design question - should `selector_contains("board=rpi", "!experimental")` be`True`? The existing test says `False`, but kubernetes semantics say `True`.

python/packages/jumpstarter/jumpstarter/client/selectors.py
```diff
@@ -67,6 +67,23 @@ def extract_match_labels_filter(selector: str | None) -> str | None:
...
+        return label_value in values
+    if operator == "exists":
+        return True
+    if operator == "!=":
+        return label_value not in values
+    if operator == "notin":
+        return label_value not in values
+    if operator == "!exists":
+        return False
+    raise ValueError(f"unknown label selector operator: {operator!r}")
```
maboras-rh, line 79: `_label_satisfies_expression` has an early-return at the top
```
if key not in sel_labels:
    return False
```
this fires before the `!exists` reached, so `_label_satisfies_expression({}, "key", "!exists", [])` returns `False`. The semantics of `!exists` are "the key must not be present", so when the key is absent the correct result is True.
There's no test for the absent-key case. The fix is to handle `!exists` before the guard:



```suggestion
def _label_satisfies_expression(sel_labels, key, operator, values):
    if operator == "!exists":
        return key not in sel_labels
    if key not in sel_labels:
        return False
    label_value = sel_labels[key]
    if operator == "in":
        return label_value in values
    if operator == "exists":
        return True
    if operator in ("!=", "notin"):
        return label_value not in values
    raise ValueError(f"unknown label selector operator: {operator!r}")
```

mangelajo (changes requested):
see the comments

python/packages/jumpstarter/jumpstarter/client/selectors_test.py
raballew: Test updated to assert true and !exist handled before absence guard.

python/packages/jumpstarter/jumpstarter/client/selectors.py
raballew: I am not sure how this _today_ could cause a bug but if will try to catch at filter_by_selector so that a bad lease can not break the entire list operation.

raballew:
You create two exporters:

- Exporter A: labels board=sa, selector requires !nonexistent
- Exporter B: labels board=sa, selector requires !production

You run `jmp get leases --selector board=sa,!production`

@maboras-rh @mangelajo what do you expect to see?

mangelajo:
> You create two exporters:
> 
> * Exporter A: labels board=sa, selector requires !nonexistent
> * Exporter B: labels board=sa, selector requires !production
> 
> You run `jmp get leases --selector board=sa,!production`
> 
> @maboras-rh @mangelajo what do you expect to see?

I think we should just do whatever kubernetes does in this case, which... I don't know what it is :D

maboras-rh:
> You create two exporters:
> 
> * Exporter A: labels board=sa, selector requires !nonexistent
> * Exporter B: labels board=sa, selector requires !production
> 
> You run `jmp get leases --selector board=sa,!production`
> 
> @maboras-rh @mangelajo what do you expect to see?

  this is a valid design question. The fallback treats the selector's `matchLabels` as the complete label set. But a lease selector is a constraint, not a full description of an exporter's labels. Exporter A's selector
  (board=sa,!nonexistent) only tells us:
  - The exporter has board=sa
  - The exporter doesn't have a nonexistent label

It says nothing about whether the exporter has a production label. the actual exporter behind that lease could have `production=true` - the selector just didn't check for it. So the fallback's conclusion that "production doesn't exist" is an assumption that may not hold.
my expectation is that only Exporter B should be shown. We can only guarantee that B's exporter doesn't have production (its selector explicitly requires `!production`). For A, we simply don't know(and showing it could be a false positive)

raballew:
and even this might not be true because the exporter object could have changed in the meantime - so the lease holds a snapshot of an old exporter? In this case the lease  would need to carry its own context and we have two options:

1. snapshot the exporters labels at lease creation time
2. remove the fallback

raballew:
k8s has no concept of "one selector contains another". It only evaluates selectors against concrete label sets. selector_contains is a custom abstraction - it treats one selector's matchLabels as a proxy for resource labels and evaluates the other selector's expressions against them which has no k8s counterpart.

raballew:
The latest commit fixes Kubernetes compatibility for `notin` and `!=` when the key is absent from the label set.

In Kubernetes, `NotIn` and `NotEquals` return `true` when the key is missing — the resource trivially satisfies "not in" or "not equal" because it has no value for that key. The previous implementation returned `false` for all operators (except `!exists`) when the key was absent.

Concrete examples that were wrong before:
- `selector_contains("board=rpi", "env notin (prod)")` returned `False`, now returns `True`
- `selector_contains("board=rpi", "env!=prod")` returned `False`, now returns `True`

The fix moves `notin` and `!=` before the key-absence guard in `_label_satisfies_expression`, matching the behavior of Go's `k8s.io/apimachinery/pkg/labels` package.

Tests now cover all 7 Kubernetes label selector operators (`=`, `==`, `!=`, `in`, `notin`, `exists`, `!exists`) with both key-present and key-absent cases.

mangelajo (approved):
Thanks for the patience, looks great now, and great coverage.

PR #722, opened 2026-06-01
fix: guard DurationParamType against OverflowError on large values

### LINKED ISSUE #717: DurationParamType.convert raises OverflowError on large numeric strings
## Description

`DurationParamType.convert` raises an unhandled `OverflowError` when given a string that parses as a very large integer (e.g. `'86400000000000'`). The method attempts `timedelta(seconds=int_value)` without guarding against values that exceed `timedelta`'s maximum (`days=999999999`).

## Steps to reproduce

```python
from jumpstarter_cli.common import DURATION

DURATION.convert('86400000000000', None, None)
# OverflowError: days=1000000000; must have magnitude <= 999999999
```

## Expected behavior

Should raise `click.exceptions.BadParameter` with a user-friendly message, consistent with how other invalid inputs are handled.

## Actual behavior

```
OverflowError: days=1000000000; must have magnitude <= 999999999
```

The `OverflowError` propagates uncaught because the `except ValueError` block at the call site does not cover it.

## Location

`python/packages/jumpstarter-cli/jumpstarter_cli/common.py`, `DurationParamType.convert` method -- the `timedelta(seconds=int_value)` call needs an `OverflowError` guard.

raballew:
## Summary
- Fixes `OverflowError` crash in `DurationParamType.convert()` when extremely large integer values (e.g. `86400000000000`) are passed, which cause `timedelta(seconds=...)` to overflow
- Catches `OverflowError` in both the integer and string-integer parsing paths and converts it to a user-friendly `click.BadParameter` error
- Extracts string parsing logic into `_parse_string()` for clarity and adds `OverflowError` to the `pytimeparse2` exception handling
- Adds regression tests for large positive integers, large string integers, and large negative string integers

Closes #717

## Test plan
- [x] `test_large_string_integer_raises_bad_parameter_not_overflow` -- verifies string `"86400000000000"` raises `BadParameter` not `OverflowError`
- [x] `test_large_integer_raises_bad_parameter_not_overflow` -- verifies int `86400000000000` raises `BadParameter` not `OverflowError`
- [x] `test_large_negative_string_raises_bad_parameter_not_overflow` -- verifies string `"-86400000000000"` raises `BadParameter` not `OverflowError`
- [x] All existing `DurationParamType` tests continue to pass

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/common.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/common.py
@@ -36,40 +36,43 @@ class DurationParamType(click.ParamType):
         super().__init__()
         self.minimum = minimum
 
+    def _parse_string(self, value: str, param, ctx) -> timedelta:
+        try:
+            int_value = int(value)
+            try:
+                return timedelta(seconds=int_value)
+            except OverflowError:
+                self.fail(f"{value!r} exceeds the maximum allowed duration", param, ctx)
+        except ValueError:
+            pass
+
+        try:
+            seconds = parse_duration(value)
+            if seconds is not None and isinstance(seconds, (int, float)):
+                return timedelta(seconds=seconds)
+        except (ValueError, TypeError, OverflowError):
+            pass
+
+        try:
+            return TypeAdapter(timedelta).validate_python(value)
+        except (ValueError, ValidationError):
+            self.fail(
+                f"{value!r} is not a valid duration "
+                "(e.g., '30m', '3h30m', '1d', '1d3h40m', 'PT1H30M', '01:30:00')",
+                param,
+                ctx,
+            )
+
     def convert(self, value, param, ctx):
         if isinstance(value, timedelta):
             td = value
         elif isinstance(value, int):
-            # Integer as seconds (backward compatibility)
-            td = timedelta(seconds=value)
-        elif isinstance(value, str):
-            # Try parsing as plain integer first (backward compatibility)
             try:
-                int_value = int(value)
-                td = timedelta(seconds=int_value)
-            except ValueError:
-                # Parse with pytimeparse2 first (supports human-readable formats)
-                td = None
-                try:
-                    seconds = parse_duration(value)
-                    if seconds is not None and isinstance(seconds, (int, float)):
-                        td = timedelta(seconds=seconds)
-                except (ValueError, TypeError):
-                    pass
-
-                # Fall back to pydantic/speedate for ISO 8601 and other formats
-                if td is None:
-                    try:
-                        td = TypeAdapter(timedelta).validate_python(value)
-                    except (ValueError, ValidationError):
-                        self.fail(
-                            (
-                                f"{value!r} is not a valid duration "
-                                "(e.g., '30m', '3h30m', '1d', '1d3h40m', 'PT1H30M', '01:30:00')"
-                            ),
-                            param,
-                            ctx,
-                        )
+                td = timedelta(seconds=value)
+            except OverflowError:
+                self.fail(f"{value!r} exceeds the maximum allowed duration", param, ctx)
+        elif isinstance(value, str):
+            td = self._parse_string(value, param, ctx)
         else:
             self.fail(
                 f"{value!r} is not a valid duration (e.g., '30m', '3h30m', '1d', '1d3h40m')",
@@ -77,12 +80,9 @@ class DurationParamType(click.ParamType):
                 ctx,
             )
 
-        # Validate minimum if specified
         if self.minimum is not None and td < self.minimum:
             min_seconds = int(self.minimum.total_seconds())
-            self.fail(
-                f"{value!r} must be at least {min_seconds} seconds", param, ctx
-            )
+            self.fail(f"{value!r} must be at least {min_seconds} seconds", param, ctx)
 
         return td
 
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/common.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/common.py
@@ -56,12 +56,13 @@ class DurationParamType(click.ParamType):
 
         try:
             return TypeAdapter(timedelta).validate_python(value)
-        except (ValueError, ValidationError):
+        except (ValueError, ValidationError) as exc:
             self.fail(
                 f"{value!r} is not a valid duration (e.g., '30m', '3h30m', '1d', '1d3h40m', 'PT1H30M', '01:30:00')",
                 param,
                 ctx,
             )
+            raise exc
 
     def convert(self, value, param, ctx):
         if isinstance(value, timedelta):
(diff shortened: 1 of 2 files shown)


python/packages/jumpstarter-cli/jumpstarter_cli/common.py
```diff
@@ -36,53 +37,53 @@ def __init__(self, minimum: timedelta | None = None):
...
+                f"{value!r} is not a valid duration (e.g., '30m', '3h30m', '1d', '1d3h40m', 'PT1H30M', '01:30:00')",
+                param,
+                ctx,
+            )
+            raise exc
+
     def convert(self, value, param, ctx):
         if isinstance(value, timedelta):
             td = value
         elif isinstance(value, int):
```
evakhoni: seems like we don't actually need separate handling for value type `int`, as simply converting `int(value)` like we do in `_parse_string` should handle both a string and an integer regardless. IMO we can simplify it. also if we do, we won't need a separate helper method `_parse_string` anymore, as the code would be simple enough to be handled inline

python/packages/jumpstarter-cli/jumpstarter_cli/common.py
```diff
@@ -36,53 +37,53 @@ def __init__(self, minimum: timedelta | None = None):
...
+
+        try:
+            return TypeAdapter(timedelta).validate_python(value)
+        except (ValueError, ValidationError) as exc:
+            self.fail(
+                f"{value!r} is not a valid duration (e.g., '30m', '3h30m', '1d', '1d3h40m', 'PT1H30M', '01:30:00')",
+                param,
+                ctx,
+            )
+            raise exc
```
evakhoni: seems to be dead code as we're already raising via self.fail above

python/packages/jumpstarter-cli/jumpstarter_cli/common.py
```diff
@@ -36,53 +37,53 @@ def __init__(self, minimum: timedelta | None = None):
...
+                self.fail(f"{value!r} exceeds the maximum allowed duration", param, ctx)
+        except ValueError:
+            pass
+
+        try:
+            seconds = parse_duration(value)
+            if seconds is not None and isinstance(seconds, (int, float)):
+                return timedelta(seconds=seconds)
+        except (ValueError, TypeError, OverflowError):
+            pass
```
evakhoni, line 59: in case of a float overflow we'll pass silently and then get `is not a valid duration` message from the next try block, instead of an `exceeds the maximum allowed duration`. just a nitpick though.

evakhoni (commented):
all in all looking good! we can simplify a bit as I commented below:

python/packages/jumpstarter-cli/jumpstarter_cli/common.py
```diff
@@ -36,53 +37,52 @@ def __init__(self, minimum: timedelta | None = None):
...
+                return timedelta(seconds=seconds)
+        except OverflowError:
+            self.fail(f"{value!r} exceeds the maximum allowed duration", param, ctx)
+            raise  # satisfy ty: self.fail is NoReturn but ty cannot verify it
+        except (ValueError, TypeError):
+            pass
+
+        try:
+            return TypeAdapter(timedelta).validate_python(value)
+        except (ValueError, ValidationError):
```
maboras-rh: when `TypeAdapter.validate_python` raise `OverflowError` is not caught. the three blocks protect against `OverflowError` in the `int→timedelta` path and the `parse_duration→timedelta` path. The `TypeAdapter(timedelta).validate_python(value)` path (the pydantic/speedate fallback for ISO 8601) does not have an `OverflowError` catch:


```suggestion
        except (ValueError, ValidationError, OverflowError):
```

python/packages/jumpstarter-cli/jumpstarter_cli/common.py
```diff
@@ -36,53 +37,52 @@ def __init__(self, minimum: timedelta | None = None):
...
+            try:
+                return timedelta(seconds=int_value)
+            except OverflowError:
+                self.fail(f"{value!r} exceeds the maximum allowed duration", param, ctx)
+                raise  # satisfy ty: self.fail is NoReturn but ty cannot verify it
+        except ValueError:
+            pass
+
+        try:
+            seconds = parse_duration(value)
```
maboras-rh, line 52: `parse_duration` returning `None` with `_parse_string` falling through silently. after the `parse_duration` block, if seconds is `None` (string not recognized), the code falls through to TypeAdapter. this was the same behavior before, so it's not a regression, but it's worth noting the path: a value like "not-a-duration" would not raise from `parse_duration`, would fall through, and would then be rejected by `TypeAdapter` with the "is not a valid duration" error

evakhoni: seems like it's not a real issue as getting None here is unlikely, and probably a sign of invalid duration anyway, so even with the unusual flow path, the end result is ok IMO

python/packages/jumpstarter-cli/jumpstarter_cli/common.py
evakhoni: low severity as afaik pydantic currently raises ValidationError for out-of-range ISO 8601 values rather than OverflowError, however it costs nothing to address, and would nice having it here in case that pydantic devs decide to change the behavior. i'm in favor of this one too

python/packages/jumpstarter-cli/jumpstarter_cli/common.py
```diff
@@ -36,53 +37,52 @@ def __init__(self, minimum: timedelta | None = None):
...
-                        self.fail(
-                            (
-                                f"{value!r} is not a valid duration "
-                                "(e.g., '30m', '3h30m', '1d', '1d3h40m', 'PT1H30M', '01:30:00')"
-                            ),
-                            param,
-                            ctx,
-                        )
+        elif isinstance(value, (str, int)):
+            td = self._parse_string(str(value), param, ctx)
```
evakhoni, line 75: all in all PR looks good. the only suggestion from my side, is we may want to inline the code that we moved to `_parse_string` back into here. this line is the only consumer of the method, and keeping it in two separate methods adds no value logically as well since `convert` is already pretty much dedicated. I would suggest removing `_parse_string` and moving the code into here. otherwise LGTM

raballew: it allows for proper unit testing even if its just used once. if you dont mind, if would keep it as is.

evakhoni: it's a styling choice. if you prefer it that way I won't try to convince you otherwise 👍 

another thing in the line above, we're checking (str, int) but we may want to include float as well?

raballew: already included

evakhoni (approved):
ok looks good

PR #723, opened 2026-06-01
fix: raise TypeError in V1Alpha1Lease.from_dict for non-dict spec

### LINKED ISSUE #719: V1Alpha1Lease.from_dict raises AttributeError when spec is not a dict
## Description

`V1Alpha1Lease.from_dict` assumes `dict["spec"]` is always a dictionary and calls `.get("selector", {})` on it. If `spec` is present but has a non-dict value (e.g. a list or string), this raises an `AttributeError` instead of a clean validation error.

## Steps to reproduce

```python
from jumpstarter_kubernetes.leases import V1Alpha1Lease

V1Alpha1Lease.from_dict({"spec": []})
# AttributeError: 'list' object has no attribute 'get'
```

## Expected behavior

Should raise `TypeError`, `ValueError`, or `KeyError` (consistent with other malformed-input handling in the method), or defensively handle the case where `spec` is not a dict.

## Actual behavior

```
AttributeError: 'list' object has no attribute 'get'
```

## Location

`python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py`, `V1Alpha1Lease.from_dict` -- line `selector_data = dict["spec"].get("selector", {})` needs a type check or defensive `.get()` chain.

raballew:
## Summary
- Add type check in `V1Alpha1Lease.from_dict` to raise `TypeError` when `spec` is not a dict, instead of leaking an `AttributeError` from `.get()`
- Extract `spec` into a local variable to avoid repeated `dict["spec"]` lookups
- Add regression tests covering list, string, integer, None, boolean, and empty dict inputs

Fixes https://github.com/jumpstarter-dev/jumpstarter/issues/719

## Test plan
- [x] `test_from_dict_raises_type_error_when_spec_is_a_list`
- [x] `test_from_dict_raises_type_error_when_spec_is_a_string`
- [x] `test_from_dict_raises_type_error_when_spec_is_an_integer`
- [x] `test_from_dict_raises_type_error_when_spec_is_none`
- [x] `test_from_dict_raises_type_error_when_spec_is_a_boolean`
- [x] `test_from_dict_does_not_raise_type_error_when_spec_is_an_empty_dict`
- [x] Existing `test_leases.py` tests still pass

### DIFF
--- a/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
+++ b/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
@@ -1,3 +1,4 @@
+from collections.abc import Mapping
 from typing import Literal, Optional
 
 from kubernetes_asyncio.client.models import V1Condition, V1ObjectMeta, V1ObjectReference
@@ -37,7 +38,10 @@ class V1Alpha1Lease(JsonBaseModel):
 
     @staticmethod
     def from_dict(dict: dict):
-        selector_data = dict["spec"].get("selector", {})
+        spec = dict["spec"]
+        if not isinstance(spec, Mapping):
+            raise TypeError(f"spec must be a dict, got {type(spec).__name__}")
+        selector_data = spec.get("selector", {})
         return V1Alpha1Lease(
             api_version=dict["apiVersion"],
             kind=dict["kind"],
@@ -70,10 +74,10 @@ class V1Alpha1Lease(JsonBaseModel):
                 ],
             ),
             spec=V1Alpha1LeaseSpec(
-                client=V1ObjectReference(name=dict["spec"]["clientRef"]["name"])
-                if "clientRef" in dict["spec"]
+                client=V1ObjectReference(name=spec["clientRef"]["name"])
+                if "clientRef" in spec
                 else None,
-                duration=dict["spec"]["duration"] if "duration" in dict["spec"] else None,
+                duration=spec["duration"] if "duration" in spec else None,
                 selector=V1Alpha1LeaseSelector(match_labels=selector_data.get("matchLabels", {})),
             ),
         )
--- a/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
+++ b/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
@@ -74,9 +74,7 @@ class V1Alpha1Lease(JsonBaseModel):
                 ],
             ),
             spec=V1Alpha1LeaseSpec(
-                client=V1ObjectReference(name=spec["clientRef"]["name"])
-                if "clientRef" in spec
-                else None,
+                client=V1ObjectReference(name=spec["clientRef"]["name"]) if "clientRef" in spec else None,
                 duration=spec["duration"] if "duration" in spec else None,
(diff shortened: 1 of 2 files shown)


raballew:
Probably a bit overengineered as k8s should always returns the right type.

maboras-rh (changes requested):

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
```diff
@@ -37,7 +38,10 @@ class V1Alpha1Lease(JsonBaseModel):
 
     @staticmethod
     def from_dict(dict: dict):
```
maboras-rh: Shadowing python builtin `dict` - the parameter `def from_dict(dict: dict)` shadows the built-in `dict`. This is pre-existing (not introduced by this PR), so not a blocker, but worth noting as a future cleanup opportunity - renaming to `data` or `raw` would be more conventional.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
```diff
@@ -37,7 +38,10 @@ class V1Alpha1Lease(JsonBaseModel):
...
-        selector_data = dict["spec"].get("selector", {})
+        spec = dict["spec"]
+        if not isinstance(spec, Mapping):
+            raise TypeError(f"spec must be a dict, got {type(spec).__name__}")
```
maboras-rh: minor point - error message could include the value - `f"spec must be a dict, got {type(spec).__name__}"` tells you the type but not the value. For debugging malformed API responses, including a `repr` of the value could help. 
```suggestion
            raise TypeError(f"spec must be a dict, got {type(spec).__name__}: {repr(spec)}")
```

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/test_leases.py
```diff
@@ -152,3 +153,33 @@ def test_lease_from_dict_without_match_labels_name_targeted():
...
+    with pytest.raises(TypeError):
+        V1Alpha1Lease.from_dict({"spec": None})
+
+
+def test_from_dict_raises_type_error_when_spec_is_a_boolean():
+    with pytest.raises(TypeError):
+        V1Alpha1Lease.from_dict({"spec": True})
+
+
+def test_from_dict_does_not_raise_type_error_when_spec_is_an_empty_dict():
```
maboras-rh: The empty-dict test name is misleading - `test_from_dict_does_not_raise_type_error_when_spec_is_an_empty_dict` asserts that a KeyError is raised (because apiVersion, kind, metadata, status are all missing).
The name says "does not raise type error" which is technically true, but a reader might expect the call to succeed.
consider renaming it to something like `test_from_dict_with_empty_spec_raises_key_error_not_type_error` to make the intent clearer.

```suggestion
def test_from_dict_with_empty_spec_raises_key_error_not_type_error():
```

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/leases.py
raballew: fixed in the last commit

maboras-rh (approved):
lgtm, clean, well-scoped fix. all three comments have been addressed
approved from my side

mangelajo (approved):

PR #727, opened 2026-06-01
feat: use generateName for leases to allow alias reuse after deletion

bennyz:
Leases created with --lease-id now use K8s generateName instead of a fixed Name. The user-provided alias is stored as a label and resolved transparently in Get/Delete/Update via a name-or-alias lookup. This prevents the "already exists" error when re-creating a lease with the same --lease-id after deletion (since deletion is a soft-delete that leaves the K8s object in etcd).


mangelajo:
I am not sure about the extra layer of complexity, what if we just trigger a generate name on the lease name if there is a conflict?

The current implementation has some level of race condition (2 leases could end up with the same label) ...

raballew:
> I am not sure about the extra layer of complexity, what if we just trigger a generate name on the lease name if there is a conflict?
> 
> The current implementation has some level of race condition (2 leases could end up with the same label) ...

I was about to comment the same thing. I think this just moves the problem one layer of complexity up. maybe we run generated name with the prefix of the user provided input if it conflicts, or we could also simply raise an error and clean up leases properly - the latter is what i prefer @mangelajo

bennyz:
> I am not sure about the extra layer of complexity, what if we just trigger a generate name on the lease name if there is a conflict?
> 
> The current implementation has some level of race condition (2 leases could end up with the same label) ...

the problem with this is that it's essentially the same behavior we already have and the user won't get the name they request

python/packages/jumpstarter-protocol/jumpstarter_protocol/jumpstarter/client/v1/client_pb2_grpc.pyi
```diff
@@ -234,20 +236,28 @@ class ClientServiceStub(typing.Generic[_ClientServiceGetExporterType, _ClientSer
...
+    """Create a new lease for an exporter."""
 
     UpdateLease: _ClientServiceUpdateLeaseType
+    """Update an existing lease."""
 
     DeleteLease: _ClientServiceDeleteLeaseType
+    """Delete a lease by resource name."""
 
     RotateToken: _ClientServiceRotateTokenType
+    """Rotate the authentication token for the client."""
```
mangelajo, line 260: something happened, the PR is including a lot of unrelated changes.

mangelajo:
let's close for now. Users got around this limitation

PR #730, opened 2026-06-02
ci: collect controller, router and exporter logs from e2e runs

raballew:
## Summary
- Adds log collection steps to all three e2e jobs (`e2e-tests`, `e2e-compat-old-controller`, `e2e-compat-old-client`)
- Collects controller logs, router logs, pod listing, and pod descriptions as downloadable artifacts
- Artifacts retained for 7 days
- Currently runs on every e2e execution (`if: always()`) to verify the collection works — will be narrowed to `if: failure()` in a follow-up once confirmed

## Test plan
- [x] Trigger the e2e workflow and verify artifacts appear in the Actions run summary
- [ ] Download the artifacts and confirm they contain meaningful controller/router logs
- [ ] Once verified, follow up to change `if: always()` to `if: failure()`

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -246,12 +246,21 @@ jobs:
         if: always()
         run: |
           mkdir -p /tmp/e2e-logs
-          kubectl -n jumpstarter-lab logs -l component=controller --all-containers --prefix --tail=-1 > /tmp/e2e-logs/controller.log 2>&1 || true
-          kubectl -n jumpstarter-lab logs -l component=controller --all-containers --prefix --previous > /tmp/e2e-logs/controller-previous.log 2>&1 || true
-          kubectl -n jumpstarter-lab logs -l component=router --all-containers --prefix --tail=-1 > /tmp/e2e-logs/router.log 2>&1 || true
-          kubectl -n jumpstarter-lab logs -l component=router --all-containers --prefix --previous > /tmp/e2e-logs/router-previous.log 2>&1 || true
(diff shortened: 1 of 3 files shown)


raballew:
@mangelajo I have set the CI to only trigger on failure, https://github.com/jumpstarter-dev/jumpstarter/actions/runs/26820799379/job/79074701210 has collected all the logs for a successful run.

.github/workflows/e2e.yaml
```diff
@@ -242,6 +242,12 @@ jobs:
         env:
           CI: true
 
+      - name: Collect and upload e2e logs
+        if: failure()
+        uses: ./.github/actions/collect-e2e-logs
+        with:
+          artifact-name: e2e-logs-${{ matrix.arch }}
+
```
mangelajo, line 250: In the past I found it useful to compare "good logs" with "failure logs" to pinpoint issues in other projects, because of that reason I would always collect the logs which aren't too big :)

it's also helpful to scan the past (as long as we have them stored) to look for things that started showing off in logs (even if it didn't fail)

Can we remove the `if: failure()` ?

mangelajo (commented):
thank you , it looks great, just one nit.

mangelajo:
@evakhoni this is what Paul was working on to start collecting the exporter logs and so on :)

.github/workflows/e2e.yaml
raballew: sure, removing it :)

mangelajo:
@raballew this looks awesome!

Thanks a lot

image

mangelajo (approved):

PR #731, opened 2026-06-02
ci: extract e2e artifact loading into a reusable composite action

raballew:
## Summary
- Extracts the duplicated download + docker-load sequence (controller image, operator image, Python wheels) into `.github/actions/load-e2e-artifacts/`
- Replaces 4 steps with a single `uses:` reference in both `e2e-tests` and `e2e-compat-old-client` jobs
- The composite action takes an `arch` input (`amd64` or `arm64`) to select the correct image artifacts

## Test plan
- [ ] Verify e2e-tests job passes on both amd64 and arm64
- [ ] Verify e2e-compat-old-client job passes (uses hardcoded `amd64`)

### DIFF
--- a/.github/actions/load-e2e-artifacts/action.yaml
+++ b/.github/actions/load-e2e-artifacts/action.yaml
@@ -1,0 +1,36 @@
+name: Load e2e artifacts
+description: Download and load prebuilt controller/operator images and Python wheels for e2e tests
+
+inputs:
+  arch:
+    description: Architecture suffix for container image artifacts (amd64 or arm64)
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Download controller image
+      uses: actions/download-artifact@v4
+      with:
+        name: controller-image-${{ inputs.arch }}
+        path: /tmp/artifacts
+
+    - name: Download operator image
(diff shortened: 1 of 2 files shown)


.github/actions/load-e2e-artifacts/action.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+inputs:
+  arch:
+    description: Architecture suffix for container image artifacts (amd64 or arm64)
+    required: true
+
+runs:
+  using: composite
+  steps:
+    - name: Download controller image
```
mangelajo, line 12: We already have caching for this. If the go-files didn.t change we don't regenerate the containers and use caching.

mangelajo:
search for "cache" in the e2e.yaml

raballew:
am i just misunderstanding something here but isnt this just deduping code thats twice in the yaml? i am not touching the caching behavior at all.

.github/actions/load-e2e-artifacts/action.yaml
mangelajo: ignore, thanks for the explanation. 

Sorry

mangelajo (approved):

PR #741, opened 2026-06-03
fix: deduplicate NotIn values in ParseLabelSelector

### LINKED ISSUE #728: ParseLabelSelector round-trip instability with duplicate NotEquals values
## Summary

`ParseLabelSelector` produces a `NotIn` expression with duplicate values when the same key appears in multiple `!=` requirements. This breaks round-trip stability because the Kubernetes label parser uses a set internally, which silently drops duplicates.

## Steps to reproduce

1. Parse `"key!=,key!="` with `ParseLabelSelector`
2. Format the result with `metav1.FormatLabelSelector` -- produces `"key notin (,)"` (two empty strings joined by comma)
3. Parse `"key notin (,)"` again -- the Kubernetes parser stores values in a set, so the two empty strings collapse to one
4. Format again -- produces `"key notin ()"` instead of `"key notin (,)"`

## Root cause

`lease_helpers.go:129` blindly appends to the `notEqualsByKey` slice without deduplicating:

```go
notEqualsByKey[key] = append(notEqualsByKey[key], values[0])
```

When the same key-value pair appears twice (e.g. `key!=,key!=`), the resulting `Values` slice contains `["", ""]`. The Kubernetes `labels.Parse` function uses `sets.String` (a map) for values, which can only hold one copy of each value. So the first format is faithful to the slice, but the second parse deduplicates, and the second format reflects that.

## Suggested fix

Deduplicate values in `notEqualsByKey` before building the `NotIn` expression, so the slice never contains duplicates that the set-based re-parser would collapse.

## Found by

The `FuzzParseLabelSelector` fuzz test in PR #720.

raballew:
## Summary
- Adds a `slices.Contains` guard in `ParseLabelSelector` to prevent duplicate values in `NotIn` expressions when the same key appears in multiple `!=` requirements
- Fixes round-trip instability: `ParseLabelSelector` -> `FormatLabelSelector` -> `ParseLabelSelector` now produces stable output
- Adds 4 tests covering duplicate dedup and round-trip stability

Closes #728

## Test plan
- [x] `TestParseLabelSelector_DeduplicatesNotEquals` -- duplicate `!=` values collapse to single entry
- [x] `TestParseLabelSelector_DeduplicatesNotEqualsMultipleValues` -- mixed duplicate/unique values handled correctly
- [x] `TestParseLabelSelector_RoundTripNotEquals` -- `key!=,key!=` round-trips stably
- [x] `TestParseLabelSelector_RoundTripNotIn` -- `key notin (a,b)` round-trips stably
- [x] All 44 existing controller tests pass

### DIFF
--- a/controller/api/v1alpha1/lease_helpers_test.go
+++ b/controller/api/v1alpha1/lease_helpers_test.go
@@ -235,6 +235,26 @@ var _ = Describe("ParseLabelSelector", func() {
 			Expect(selector.MatchExpressions[0].Operator).To(Equal(metav1.LabelSelectorOpNotIn))
 			Expect(selector.MatchExpressions[0].Values).To(ConsistOf("value1", "value2"))
 		})
+
+		It("should deduplicate NotIn values when the same key-value pair appears multiple times", func() {
+			selector, err := ParseLabelSelector("key!=value1,key!=value1")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(selector).NotTo(BeNil())
+			Expect(selector.MatchExpressions).To(HaveLen(1))
+			Expect(selector.MatchExpressions[0].Key).To(Equal("key"))
+			Expect(selector.MatchExpressions[0].Operator).To(Equal(metav1.LabelSelectorOpNotIn))
+			Expect(selector.MatchExpressions[0].Values).To(Equal([]string{"value1"}))
+		})
+
+		It("should deduplicate NotIn values with empty strings", func() {
+			selector, err := ParseLabelSelector("key!=,key!=")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(selector).NotTo(BeNil())
+			Expect(selector.MatchExpressions).To(HaveLen(1))
+			Expect(selector.MatchExpressions[0].Key).To(Equal("key"))
+			Expect(selector.MatchExpressions[0].Operator).To(Equal(metav1.LabelSelectorOpNotIn))
+			Expect(selector.MatchExpressions[0].Values).To(Equal([]string{""}))
+		})
 	})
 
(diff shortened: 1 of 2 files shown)


controller/api/v1alpha1/lease_helpers_test.go
```diff
@@ -235,6 +235,26 @@ var _ = Describe("ParseLabelSelector", func() {
...
+		It("should deduplicate NotIn values when the same key-value pair appears multiple times", func() {
+			selector, err := ParseLabelSelector("key!=value1,key!=value1")
+			Expect(err).NotTo(HaveOccurred())
+			Expect(selector).NotTo(BeNil())
+			Expect(selector.MatchExpressions).To(HaveLen(1))
+			Expect(selector.MatchExpressions[0].Key).To(Equal("key"))
+			Expect(selector.MatchExpressions[0].Operator).To(Equal(metav1.LabelSelectorOpNotIn))
+			Expect(selector.MatchExpressions[0].Values).To(Equal([]string{"value1"}))
+		})
+
```
maboras-rh, line 248: Consider adding one more test case as following :-
```suggestion

		It("should deduplicate NotIn values while preserving unique values", func() {
			selector, err := ParseLabelSelector("key!=value1,key!=value2,key!=value1")
			Expect(err).NotTo(HaveOccurred())
			Expect(selector).NotTo(BeNil())
			Expect(selector.MatchExpressions).To(HaveLen(1))
			Expect(selector.MatchExpressions[0].Values).To(ConsistOf("value1", "value2"))
		})
```

maboras-rh (approved):
lgtm, one minor nit for adding one more test.
approved from my end

mangelajo (approved):

PR #742, opened 2026-06-03
flashers: add default CA to be inject into flashing command

bennyz:
if `ca` field is set on the exporter, inject it by default to the falshing command

### DIFF
--- a/python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
+++ b/python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
@@ -37,6 +37,7 @@ from opendal import Metadata, Operator
 
 from jumpstarter_driver_flashers.bundle import FlasherBundleManifestV1Alpha1
 
+from jumpstarter.client.core import DriverMethodNotImplemented
 from jumpstarter.client.decorators import driver_click_group
 from jumpstarter.common.exceptions import ArgumentError, JumpstarterException
 from jumpstarter.common.fls import get_fls_github_url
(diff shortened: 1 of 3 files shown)


python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
```diff
@@ -529,13 +530,32 @@ def _perform_flash_operation(
             else:
                 self.logger.info("Leaving target powered on")
 
+    def _store_cacert_on_dut(self, console, manifest, cacert_content) -> str:
```
mangelajo, line 535: We had code for this already, let me look it up.

mangelajo: https://github.com/jumpstarter-dev/jumpstarter/blob/main/python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py#L491

ahh you are actually modifying the called method :)

Note this, will be very slow on nxp board that have cps limit, I wonder if we can remove that.

mangelajo (commented):
It looks good. I wonder if we can remove the cps limit on nxp targets now because this will be *very* slow :D

mangelajo (approved):

mangelajo (changes requested):
After talking to eballetbo, and looking at our nxp boards that limit serial cps input, if we enable this by default we should probably pass the CA cert via the http server.

Can we change that?

Thanks

bennyz:
tested in the lab

python/packages/jumpstarter-driver-flashers/jumpstarter_driver_flashers/client.py
```diff
@@ -529,13 +530,37 @@ def _perform_flash_operation(
...
+            cacert_content: CA certificate content (str or bytes)
+
+        Returns:
+            Path to stored CA certificate on the DUT
+        """
+        stored_cacert = "/tmp/cacert.crt"
+        cacert_bytes = cacert_content.encode() if isinstance(cacert_content, str) else cacert_content
+        self.http.storage.write_bytes("cacert.crt", cacert_bytes)
+        cacert_url = self.http.get_url() + "/cacert.crt"
+        self.logger.info(f"Downloading CA certificate to DUT from {cacert_url}")
```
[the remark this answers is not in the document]
bennyz: this is not a possible use case, it runs once in initramfs to prepare for flashing

kirkbrauer:
@bennyz @mangelajo I think this will be useful for us as well. We have a lot of complex certificate situations in our environment that could be mitigated by allowing custom CAs at all levels.

mangelajo (approved):

PR #744, opened 2026-06-03
docs(jep-0014): simplified virtual exporter design

mangelajo:
## Summary

JEP-0014 Virtual Scalable Exporters — revised per **2026-06 team review**.

- **2-CRD model:** namespaced `VirtualTargetClass` + `ExporterSet` only — no typed `*VirtualTarget` claim CRDs
- **Dictionary config:** nested `parameters` on class + set with deep-merge (not flat `map[string]string`)
- **Flash-at-lease (DD-7):** warm pool provides baseline runtime; users flash/boot like physical targets — no `ExporterSet.spec.ttl` in v1
- **End-to-end flow:** phased QEMU warm-pool walkthrough (admin / controller / user), flash-at-lease session
- **External provisioning:** off-cluster bare-metal example (`qemu-baremetal.jumpstarter.dev`)
- **Implementation plan:** Phase 2 QEMU reference → Phase 3 off-cluster → deferred Corellium/Android/lifecycle controller

## Design decisions

- **DD-3:** pluggable provisioner + dictionary parameters; reject typed claims
- **DD-6:** namespaced `VirtualTargetClass`, same-namespace references
- **DD-7:** defer TTL/image refresh; future cross-cutting lifecycle controller

## Test plan

- [ ] Review 2-CRD model and nested `parameters` merge semantics
- [ ] Verify E2E flow and flash-at-lease alignment with physical targets (DD-7)
- [ ] Confirm implementation phases (QEMU-first) match v1 scope
- [ ] Sphinx docs build clean

### DIFF
--- a/python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
+++ b/python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
@@ -8,7 +8,7 @@
 | **Status**        | Draft                                                          |
 | **Type**          | Standards Track                                                |
 | **Created**       | 2026-06-03                                                     |
-| **Updated**       | 2026-06-03                                                     |
+| **Updated**       | 2026-06-18                                                     |
 | **Discussion**    | https://github.com/jumpstarter-dev/jumpstarter/issues/41       |
 | **Requires**      |                                                                |
 | **Supersedes**    |                                                                |
@@ -19,15 +19,16 @@
 ## Abstract
 
 This JEP proposes a Virtual Scalable Exporter subsystem for Jumpstarter that
-manages pools of virtual targets with configurable autoscaling. Each virtual
-target definition declares a minimum and maximum number of instances; the system
-maintains a warm pool of pre-spawned exporters ready for immediate lease
-fulfillment — avoiding the 10-60s cold-start latency of VM boot and exporter
-registration — and scales up or down based on demand. This enables low-latency
-lease acquisition, massive scalability, resource efficiency, and simplified
-orchestration of mixed physical/virtual test topologies — while allowing
-administrators to tune the trade-off between responsiveness and resource
-consumption on a per-target basis.
+manages pools of virtual targets with configurable autoscaling. Conceptually,
+the system scales **virtual targets**; the **Exporter** is the scheduling and
+leasing unit (the Pod analog). Each `ExporterSet` declares scaling bounds using
+familiar Kubernetes vocabulary (`minReplicas`, `maxReplicas`,
+`minAvailableReplicas`); the controller maintains a warm pool of ready exporters
+to absorb the 10-60s cold-start latency of VM boot and exporter registration.
+This enables low-latency lease acquisition, massive scalability, resource
+efficiency, and simplified orchestration of mixed physical/virtual test
+topologies — while allowing administrators to tune the trade-off between
+responsiveness and resource consumption on a per-target basis.
 
 ## Motivation
 
@@ -53,6 +54,24 @@ This model has several limitations:
 The core problem is that virtual targets lack a pool manager that can maintain a
 configurable warm pool while autoscaling to meet demand.
 
+### Fidelity / Cost Ladder
+
+One logical target can be served by multiple backends at different fidelity and
+cost tiers. Users select via labels through `jmp lease`; the same workflow
+applies regardless of backend:
+
+| class (provisioner) | fidelity | scale/cost | role |
+| --- | --- | --- | --- |
+| container sim (`qemu.jumpstarter.dev`) | low | cheap / CI-scale | functional checks |
+| cloud virtual device (`corellium.jumpstarter.dev`) | high | metered | higher-fidelity behavior |
+| real hardware (Exporter) | full | scarce | ground truth |
+
+For example, a target that needs GPU or specialized I/O can run functional
+checks cheaply on a QEMU class in CI, validate higher-fidelity behavior on a
+cloud-backed virtual device, and use real hardware as ground truth. The
+`VirtualTargetClass` / `*VirtualTarget` abstraction makes this ladder explicit
+without changing the lease experience.
+
 ### User Stories
 
 - **As a** CI pipeline author, **I want to** lease N virtual targets instantly
@@ -63,162 +82,264 @@ configurable warm pool while autoscaling to meet demand.
   physical board's properties with near-zero wait time, **so that** I can
   iterate quickly without waiting for scarce hardware.
 
-- **As a** platform engineer, **I want to** declare a virtual target pool with
-  `minWarmInstances: 2, maxTotalInstances: 20`, **so that** there are always warm
+- **As a** platform engineer, **I want to** declare an `ExporterSet` with
+  `minAvailableReplicas: 2, maxReplicas: 20`, **so that** there are always warm
   instances ready while the system scales up on demand and scales down when idle.
 
-- **As a** cost-conscious operator, **I want to** set `minWarmInstances: 0` for
-  rarely-used target types, **so that** they consume no resources until actually
-  requested, accepting a cold-start delay.
+- **As a** cost-conscious operator, **I want to** set `minAvailableReplicas: 0`
+  for rarely-used target types, **so that** they consume no resources until
+  actually requested, accepting a cold-start delay.
 
 ## Proposal
 
 The proposal introduces **Virtual Scalable Exporters** — a controller-managed
 pool of virtual target instances with configurable autoscaling. Rather than
-treating virtual targets as purely on-demand or purely static, each virtual
-target definition declares scaling parameters that let administrators tune the
+treating virtual targets as purely on-demand or purely static, each
+`ExporterSet` declares scaling parameters that let administrators tune the
 trade-off between instant availability and resource consumption.
 
-### Core Concept: Managed Pools with Scaling
+### Resource Hierarchy
 
-Each provider type has its own CRD (e.g., `QEMUExporterPool`,
-`AndroidExporterPool`, `CorelliumExporterPool`) with provider-specific
-configuration fields alongside shared scaling parameters. This gives each
-provider a strongly-typed schema rather than a generic bag of config.
+Virtual scalable exporters are modeled on familiar Kubernetes workload
+primitives:
 
-**Example: QEMU pool**
+```text
+VirtualTargetClass  ←── referenced by ──  ExporterSet
+                                              │
+                                              ▼
+                                         Exporter ──► Pod
+                              (exporter sidecar + target runtime)
+
+# API-backed / static / multi-device cases may also use:
+VirtualTargetClass  ←── referenced by ──  *VirtualTarget (typed claim)
+                                              ↑
+ExporterSet ──► Exporter ────────────────────┘
+```
+
+- **`VirtualTargetClass`** — cluster-scoped configuration for a backend
+  (`provisioner`, credentials, scheduling, binding mode, device parameters).
+  Admins own classes; claim authors never touch credentials.
+- **`*VirtualTarget`** — optional strongly-typed claim for backends where each
+  instance has distinct identity (API-backed devices, static benches). Not
+  required for homogeneous container-backed pools.
+- **`ExporterSet`** — generic scaling resource with `selector` + inline
+  `template`. References a `VirtualTargetClass` (or optionally a `*VirtualTarget`
+  claim). One mental model for all backends.
+- **`Exporter`** — the minimum leased unit. Exposes drivers that connect to the
+  virtual target provisioned from the class (or claim).
+
+### Core Concept: ExporterSet with Kubernetes-Native Scaling
+
+`ExporterSet` is a generic CRD (ReplicaSet + HPA analog) with familiar scaling
+vocabulary. Provider typing lives in `VirtualTargetClass` and `*VirtualTarget`,
+not in the pool CRD itself.
+
+**Example: VirtualTargetClass (cluster-scoped, StorageClass analog)**
 
 ```yaml
 apiVersion: jumpstarter.dev/v1alpha1
-kind: QEMUExporterPool
+kind: VirtualTargetClass
+metadata:
+  name: qemu-rpi4
+spec:
+  provisioner: qemu.jumpstarter.dev
+  bindingMode: Immediate              # warm pool; WaitForFirstConsumer = on-demand
+  reclaimPolicy: Delete
+  scheduling:                         # inherited by rendered exporter Pods
+    nodeSelector:
+      kubernetes.io/arch: arm64
+    tolerations:
+      - key: jumpstarter.dev/kvm
+        operator: Exists
+        effect: NoSchedule
+    resources:
+      limits:
+        devices.kubevirt.io/kvm: "1"
+  parameters:
+    machineType: virt
+    firmware: registry.example.com/firmware/rpi4:latest
+    cpu: 4
+    memory: 4Gi
+    storage: 16Gi
+```
+
+**Example: QEMUVirtualTarget (optional typed claim)**
+
+For homogeneous QEMU pools, admins configure `VirtualTargetClass` + `ExporterSet`
+only (see *End-to-End Flow*). A per-instance claim is optional — useful for
+static benches or when per-instance sizing differs from the class defaults:
+
+```yaml
+apiVersion: jumpstarter.dev/v1alpha1
+kind: QEMUVirtualTarget
+metadata:
+  name: rpi4-target-01
+  namespace: jumpstarter
+spec:
+  virtualTargetClassName: qemu-rpi4
+  resources:
+    cpu: 8                           # override class default
+    memory: 8Gi
+    storage: 32Gi
+```
+
+**Example: ExporterSet (generic scaling resource)**
+
+```yaml
+apiVersion: jumpstarter.dev/v1alpha1
+kind: ExporterSet
 metadata:
   name: rpi4-virtual
   namespace: jumpstarter
 spec:
-  # Scaling configuration (shared across all pool CRDs)
-  minWarmInstances: 2        # Always keep 2 warm instances ready
-  maxTotalInstances: 20       # Scale up to 20 under load
-  
-  # Node scheduling (shared across all pool CRDs, optional)
-  nodeSelector:
-    node.kubernetes.io/instance-type: bare-metal
-    jumpstarter.dev/nested-virt: "true"
-  
-  # Labels exposed on each instance (for lease matching)
-  labels:
-    board: rpi4
-    arch: aarch64
-    virtual: "true"
-  
-  # Pod overrides (shared across all pool CRDs, optional)
-  podTemplate:
-    resources:
-      requests:
-        cpu: "4"
-        memory: 5Gi
-      limits:
-        cpu: "4"
-        memory: 5Gi
-  
-  # QEMU-specific configuration
-  machineType: virt
-  firmware: registry.example.com/firmware/rpi4:latest
-  resources:
-    cpu: 4
-    memory: 4Gi
-    storage: 16Gi
-    
-  # Exporter template (drivers exposed by each instance)
-  exporterTemplate:
-    drivers:
-      - type: jumpstarter_driver_power.driver.QemuPower
-      - type: jumpstarter_driver_network.driver.TcpNetwork
-        config:
-          port: 22
-      - type: jumpstarter_driver_serial.driver.QemuSerial
+  minReplicas: 0
+  maxReplicas: 20
+  minAvailableReplicas: 2            # PDB-style warm buffer (ready & unleased)
+  scaleDownCooldown: 5m
+  recycleStrategy: ExitAndReplace    # or InPlaceReuse
+  virtualTargetClassName: qemu-rpi4  # references VirtualTargetClass above
+  selector:
+    matchLabels:
+      board: rpi4
+  template:                          # embedded template (Deployment idiom)
+    metadata:
+      labels:
+        board: rpi4
+        arch: aarch64
+        virtual: "true"
+    spec:
+      drivers:
+        - type: jumpstarter_driver_power.driver.QemuPower
+        - type: jumpstarter_driver_network.driver.TcpNetwork
+          config:
+            port: 22
+        - type: jumpstarter_driver_serial.driver.QemuSerial
+status:
+  replicas: 5
+  readyReplicas: 3
+  availableReplicas: 1             # warm (ready & unleased)
+  leasedReplicas: 2
+# scale subresource: specReplicasPath=.spec.maxReplicas
 ```
 
-**Example: Android Emulator pool**
+**Example: Corellium VirtualTargetClass + claim**
 
 ```yaml
 apiVersion: jumpstarter.dev/v1alpha1
-kind: AndroidExporterPool
+kind: VirtualTargetClass
 metadata:
-  name: pixel7-emulator
-  namespace: jumpstarter
+  name: corellium-kronos
 spec:
-  minWarmInstances: 0        # Fully on-demand (cold-start OK for this target)
-  maxTotalInstances: 10
-  
-  labels:
-    device: pixel7
-    os: android
-    api-level: "34"
-    virtual: "true"
-  
-  # Android-specific configuration
-  systemImage: system-images;android-34;google_apis;arm64-v8a
-  avdProfile: pixel_7
-  gpu: swiftshader
-  
-  exporterTemplate:
-    drivers:
-      - type: jumpstarter_driver_android.driver.AdbDriver
-      - type: jumpstarter_driver_power.driver.EmulatorPower
-```
-
-**Example: Corellium pool**
-
-The Corellium driver (`jumpstarter_driver_corellium.driver.Corellium`) manages
-the full virtual instance lifecycle through the Corellium REST API — it creates
-instances on power-on and destroys them on power-off. It exposes a power
-interface and a websocket-based serial console. The pool controller manages the
-exporter Pod and injects API credentials via environment variables
-(`CORELLIUM_API_HOST`, `CORELLIUM_API_TOKEN`) from the referenced Secret.
-
-```yaml
+  provisioner: corellium.jumpstarter.dev
+  credentialsSecretRef:
+    name: corellium-creds
+    namespace: jumpstarter
+  bindingMode: WaitForFirstConsumer  # provision on lease
+  reclaimPolicy: Delete
+  parameters:
+    apiHost: app.corellium.com
+    projectId: "778f00af-5e9b-40e6-8e7f-c4f14b632e9c"
+---
 apiVersion: jumpstarter.dev/v1alpha1
-kind: CorelliumExporterPool
+kind: CorelliumVirtualTarget
 metadata:
-  name: rd1ae-corellium
+  name: rd1ae-kronos-01
   namespace: jumpstarter
 spec:
-  minWarmInstances: 1
-  maxTotalInstances: 5
-  
-  labels:
-    board: rd1ae
-    flavor: kronos
-    virtual: "true"
-  
-  # Corellium-specific configuration
-  apiHost: app.corellium.com
-  apiCredentialsSecret: corellium-api-credentials  # Secret with keys: token
-  projectId: "778f00af-5e9b-40e6-8e7f-c4f14b632e9c"
-  
-  # Device/instance parameters
+  virtualTargetClassName: corellium-kronos
   deviceFlavor: kronos
   deviceOs: "1.1.1"
   deviceBuild: "Critical Application Monitor (Baremetal)"
   consoleName: "Primary Compute Non-Secure"
-  
-  exporterTemplate:
-    drivers:
-      - type: jumpstarter_driver_corellium.driver.Corellium
-        config:
-          device_name: "{{ .InstanceName }}"
 ```
 
-The pool controller automatically injects the Corellium-specific CRD spec fields
-(`projectId`, `deviceFlavor`, `deviceOs`, `deviceBuild`, `consoleName`) into the
-driver config at instance creation time. Only fields that vary per instance (like
-`device_name` using the `{{ .InstanceName }}` template variable) need to be
-specified explicitly in `exporterTemplate`.
+The Corellium driver (`jumpstarter_driver_corellium.driver.Corellium`) manages
+the full virtual instance lifecycle through the Corellium REST API — it creates
+instances on power-on and destroys them on power-off. The provisioner injects
+API credentials from `VirtualTargetClass.credentialsSecretRef` into the
+exporter Pod; claim authors never see credentials.
 
-A pool with `minWarmInstances: 0` consumes no resources until a lease is
-requested, accepting cold-start latency. A pool with `minWarmInstances: 3`
-always has 3 ready-to-lease instances — leases are fulfilled instantly from
-the warm pool, and the controller scales up if more are needed.
+**Example: Android ExporterSet**
+
+```yaml
+apiVersion: jumpstarter.dev/v1alpha1
+kind: ExporterSet
+metadata:
+  name: pixel7-emulator
+  namespace: jumpstarter
+spec:
+  minReplicas: 0
+  maxReplicas: 10
+  minAvailableReplicas: 0            # fully on-demand
+  selector:
+    matchLabels:
+      device: pixel7
+  template:
+    metadata:
+      labels:
+        device: pixel7
+        os: android
+        api-level: "34"
+        virtual: "true"
+    spec:
+      virtualTargetRef:
+        apiVersion: jumpstarter.dev/v1alpha1
+        kind: AndroidVirtualTarget
+        name: pixel7-template
+      drivers:
+        - type: jumpstarter_driver_android.driver.AdbDriver
+        - type: jumpstarter_driver_power.driver.EmulatorPower
+```
+
+An `ExporterSet` with `minAvailableReplicas: 0` consumes no resources until a
+lease is requested, accepting cold-start latency. An `ExporterSet` with
+`minAvailableReplicas: 3` always has 3 ready-to-lease exporters — leases are
+fulfilled instantly from the warm pool, and the controller scales up if more are
+needed.
+
+### Container-Backed Targets: Sidecar Pattern
+
+For container-backed provisioners (`qemu.jumpstarter.dev`, Android emulator, etc.),
+the provisioner renders each instance Pod from independently shipped artifacts:
+
+```yaml
+# rendered by qemu.jumpstarter.dev provisioner
+spec:
+  initContainers:
+    - name: exporter                 # native sidecar (starts first, drains last)
+      restartPolicy: Always
+      image: quay.io/jumpstarter-dev/exporter:latest
+  containers:
+    - name: target-runtime           # QEMU/Cuttlefish — independent image
+      image: quay.io/jumpstarter-dev/qemu-runtime:latest
+      volumeMounts:
+        - name: os
+          mountPath: /os
+        - name: shared
+          mountPath: /shared
+  volumes:
+    - name: os
+      image:
+        reference: registry.example.com/os/rpi4:latest   # OS as OCI artifact
+    - name: shared
+      emptyDir: {}
+```
+
+Benefits:
+
+- **Independent release cadence** — exporter, runtime, and OS image version
+  independently.
+- **Fault isolation** — exporter survives target-runtime crashes and can drain
+  or report failure.
+- **Standard interfaces** — drivers attach over virtio (serial/SPI/CAN/GPIO) or
+  Unix sockets on shared volumes; same driver code works physical + virtual.
+- **Unprivileged Pods** — virtio-backed guests avoid privileged containers when
+  the host supports it.
+
+The exporter sidecar communicates with the target-runtime container via Unix
+sockets on a shared `emptyDir` volume (QMP for QEMU control, serial console,
+launcher socket for dynamic argv). API-backed provisioners (`corellium`, `ec2`)
+skip the runtime container and connect out to external APIs.
 
 ### User Experience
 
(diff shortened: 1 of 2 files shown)


python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+| **Superseded-By** |                                                                |
+
+---
+
+## Abstract
+
+This JEP proposes a Virtual Scalable Exporter subsystem for Jumpstarter that
+manages pools of virtual targets with configurable autoscaling. Each virtual
+target definition declares a minimum and maximum number of instances; the system
+maintains a warm pool of pre-spawned exporters ready for immediate lease
```
raballew: What is the rationale for having a warm pool? Couldnt we just spawn them on demand or is the warm up time too high?

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+- **As a** CI pipeline author, **I want to** lease N virtual targets instantly
+  from a warm pool, **so that** my pipeline doesn't block on provisioning
+  latency during burst periods.
+
+- **As a** developer, **I want to** lease a virtual target matching a known
+  physical board's properties with near-zero wait time, **so that** I can
+  iterate quickly without waiting for scarce hardware.
+
+- **As a** platform engineer, **I want to** declare a virtual target pool with
+  `minInstances: 2, maxInstances: 20`, **so that** there are always warm
```
raballew: shouldnt it be minWarmInstances and maxTotalInstances then?

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+
+# Lease explicitly virtual targets
+jmp lease -l board=rpi4,virtual=true
+
+```
+
+The guiding principle is: **"Get me a target that matches my requirements."** The
+distinction between physical and virtual is an implementation detail, not a
+primary concern for the user. Virtual exporters simply appear in the same pool
+as physical ones, differentiated only by labels.
```
raballew, line 341: yeah, thats a reasonable design decision - this would allow you to easily switch between virtual and physical by merely adding a label

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+                  ┌────────────────────────────────────┐
+                  │          Kubernetes API            │
+                  │  (Lease CRs, Exporter CRs,         │
+                  │   *ExporterPool CRs)               │
+                  └─┬──────────────┬──────────────┬────┘
+                    │              │              │
+         watches    │   watches    │   watches    │
+      Leases +      │  Leases +    │  Leases +    │
+      Exporters     │  Exporters   │  Exporters   │
+                    │              │              │
```
raballew, line 693: Have we looked into things like crossplane.io

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+                                   ▼
+                          Kubernetes API (Exporters)
+```
+
+**Scaling Inputs — Watches on Leases and Exporters:**
+
+Each pool controller watches two key resources to make scaling decisions:
+
+1. **Leases** — The controller watches for pending Leases whose label selectors
+   match the pool's labels. Pending leases with no available exporter signal
```
raballew: what about scheduled leases?

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+excess idle instances beyond `minInstances` for a sustained period, scale down.
+
+**Per-Provider Deployments (single image by default):** All provider
+controllers are compiled into a single binary. Each Deployment in the cluster
+passes a `--provider=<type>` flag to activate the corresponding reconciler.
+This gives each provider isolated logs and independent restarts while
+maintaining a single image to build and release. The per-provider `image`
+override in the operator CR allows administrators to substitute a custom image
+for a specific provider (e.g., a third-party provider distributed as its own
+image) without affecting other providers.
```
raballew: +1

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+   `status.leaseRef` to remain empty).
+3. Pool controller deletes the Pod and Exporter CR.
+
+### Hardware Considerations
+
+This proposal is specifically designed to reduce reliance on physical hardware
+for scalable testing. However:
+
+- Virtual targets must faithfully emulate the interfaces exposed by physical
+  hardware (serial, network, storage, power) through the existing driver model.
```
raballew, line 956: +inf

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+
+1. **Pool-based with configurable min/max** — Maintain a warm pool of
+   pre-spawned instances; scale between `minInstances` and `maxInstances`.
+2. **Purely on-demand** — Spawn a new instance only when a lease request arrives;
+   destroy it when the lease is released.
+
+**Decision:** Pool-based with configurable min/max.
+
+**Rationale:** Purely on-demand provisioning introduces unacceptable latency for
+CI pipelines (VM boot + exporter registration can take 30-120s). A warm pool
```
raballew: why does it take 2min?

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+**Rationale:** A single image is cheaper to build, test, and productize — there
+is one CI pipeline, one vulnerability scan, one release artifact. Deploying it
+as separate Deployments (one per provider) gives operational benefits: each
+provider has isolated logs, independent scaling, and can be restarted without
+affecting other providers. The `--provider` flag makes it explicit which CRD
+a given Deployment reconciles. Adding a new provider type means adding a new
+Deployment manifest pointing to the same image with a different flag — no new
+image build required.
+
+### DD-3: CRD per provider vs. generic CRD
```
raballew: I want to throw in another idea here. lets call it composite exporters which are logically close to this per provider / generic discussion because at some point you might want to link multiple exporters (and thus exporterpools?!) to each other.

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+- Providers like QEMU/Renode require `/dev/kvm` access for acceptable
+  performance on the host nodes.
+- Timing-sensitive tests (USB/IP latency, boot ROM timeouts) may behave
+  differently on virtual targets — the system should expose labels indicating
+  whether a target is physical or virtual so users can filter when fidelity
+  matters.
+
+## Design Decisions
+
+### DD-1: Pool-based scaling vs. purely on-demand provisioning
```
raballew, line 1079: HPA, KEDA, ... was not discussed

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+    (never below minInstances)
+```
+
+### Instance States
+
+Each virtual exporter instance transitions through:
+
+```
+Provisioning → Ready (warm pool) → Leased → Ready
+                                              └→ Terminating → (deleted if available instances>min)
```
raballew: does that me we are reusing virtual instances (even just the exporter?)

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: How would crossplane apply here?

I think leases and exporters are already CRDs and we have the operator framework to reconcile all this. This is really close to tiny operators :)

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: +1 we need to ignore them until it's time to make them effective.

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: hehe the LLM was a little bit creative in this, I think it needs some work

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: heh, I discused it with the LLM and ditched it out :D what do you want from that? :)

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: 1. create exporter
2. create the pod, node is assigned...
3. the image is downloaded
4. the exporter boots
5. connects back and becomes ready.

May be it's more around 10-15 seconds.

But could be driver dependent, i.e. renode takes some time to initialize.

But it really doesn't make much difference, with the current design you can chose to set minInstances to 0 and ... get 0 warmth instances, or more if you want any :) it becomes an admin decission and it doesn't really change a lot the underlying code design.

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: I think this is something we will be able to acomplish with composite leases and connectivity between DUTs, starting multiple DUTs in parallel (physical or virtual) and inter-connecting their networks.

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: ok, so I am proposing "0" for no limits... (at your own risk) :D

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
+# JEP-0014: Virtual Scalable Exporters
```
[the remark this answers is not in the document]
mangelajo: Reviewed — the existing `index.md` on `main` intentionally dropped the toctree after the directory rename. The JEP files are reachable via inline table links, which matches the pattern used by the live docs. No toctree re-added.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+  ┌──────────────────┐ ┌───────────────────────┐ ┌────────────────────────┐
+  │ Warm Pool        │ │ Warm Pool             │ │ Warm Pool              │
+  │ [inst1][inst2].. │ │ [inst1][inst2]..      │ │ [inst1]..              │
+  └────────┬─────────┘ └───────────┬───────────┘ └────────────┬───────────┘
+           │                       │                          │
+           └───────────────────────┼──────────────────────────┘
+                                   │ register as standard Exporter CRs
+                                   ▼
+                          Kubernetes API (Exporters)
+```
```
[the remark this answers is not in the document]
mangelajo: Done — added `text` language identifiers to the two untyped fenced code blocks (reconciliation pseudocode and instance state diagram).

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,878 @@
...
+- **Scaling storm:** Rate limiting on scale-up prevents creating too many
+  instances simultaneously.
+
+## Test Plan
+
+<!-- TODO: Detail specific test cases -->
+
+### Unit Tests
+Unit tests should meet the project test coverage requirements.
+
```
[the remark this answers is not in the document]
mangelajo: The TODO placeholder has been removed. The test plan section now stands as-is; concrete test cases will be added as implementation progresses.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Addressed — the Abstract and DD-1 now explicitly explain the warm pool rationale: virtual targets (especially QEMU VMs) have cold-start latencies of 10–60 seconds (image pull, VM boot, exporter registration). Warm pools absorb this cost so leases are fulfilled near-instantly.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Done — renamed throughout the document: `minInstances` → `minWarmInstances` and `maxInstances` → `maxTotalInstances`. CRD spec comments updated to reflect the semantics.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Addressed — added Crossplane to the "Prior Art" section. The rationale: Jumpstarter already has its own CRDs (Exporter, Lease) and operator framework, so the pool controllers are essentially tiny purpose-built operators. Crossplane's generic infrastructure-provisioning model doesn't map well to Jumpstarter's domain-specific lifecycle (warm pools, lease-aware scaling, driver injection).

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Addressed — moved "Scheduled leases" from Unresolved to Resolved Questions. The controller already supports `Spec.BeginTime` on Lease CRs; the pool controller simply ignores leases whose `BeginTime` is in the future when counting demand, and only pre-provisions instances as the scheduled time approaches.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Done — `maxTotalInstances` is now optional; omitting it or setting it to 0 means no upper limit.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Updated DD-1 — the cold-start estimate is now 10–60 seconds (not 2 minutes). The breakdown: image pull + VM boot + exporter registration. The previous number was too generous.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Great idea — added "Composite exporters" and "Cross-pool scheduling" to the Future Possibilities section. This captures the scenario where multiple exporters (or pools) are logically linked together.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Addressed — added DD-5 (Built-in Scaling vs. HPA/KEDA) to the Design Decisions section. The rationale: the pool controller needs Jumpstarter-specific knowledge (lease states, warm pool targets, exporter readiness) that generic metric-based autoscalers like HPA/KEDA cannot express without a complex custom-metrics adapter that would duplicate most of the controller logic anyway.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Addressed — clarified instance reuse in the Component Interaction section. Added a `recycleStrategy` field to the common pool spec with two options:

- **ExitAndReplace** (default): the exporter process exits after lease release; the Deployment/ReplicaSet respawns a fresh Pod.
- **InPlaceReuse**: the exporter performs internal cleanup and re-registers as ready for the next lease without restarting.

_This comment was generated from a Cursor session._

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+2. **Purely on-demand** — Spawn a new instance only when a lease request arrives;
+   destroy it when the lease is released.
+
+**Decision:** Pool-based with configurable min/max.
+
+**Rationale:** Purely on-demand provisioning introduces noticeable latency for
+CI pipelines (Pod scheduling + image pull + VM boot + exporter registration
+typically takes 10-15s, and up to 60s with cold image pulls or heavy
+providers). A warm pool
+provides instant lease fulfillment for the common case. Setting `minWarmInstances: 0`
```
raballew: are you referring to the exporter image or the "disk image" that the exporter will run?

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
raballew: i was referring to crossplane for integration into external services (corellium) that do not run on the cluster

mangelajo: I am not sure how that would work, you need something that connects to corellium and exposes their APIs as an exporter in jumpstarter. What role would crossplane play?

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
raballew: I am not sure what you are referring to here. the comment seems to be not related to my support for proper interfaces

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
raballew: virtual instances should be throw away imo

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Pull of the container image that runs the exporter in a pod. I guess that in most cases will already be part of the node.

raballew: then this is a no-op wiht time only spend on scheduling, boot and registration

kirkbrauer:
@mangelajo I would also propose a sidecar pattern for the exporters by default. This would prevent losing the main pod from bringing down the exporter too and add more flexibility for multi-device virtual benches in the future.

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+### Phase 3: Additional providers
+
+Add support for additional provider types using the same binary with different
+`--provider` flags.
+
+**Deliverables:**
+
+- [ ] `AndroidExporterPool` CRD and reconciler
+- [ ] Provider authoring guide documenting how to add a new `*ExporterPool`
+
```
mangelajo, line 1558: Future ideas:

* Priority selectors : I want "this", if not available "this other thing" ... otherwise ....

mangelajo: Should this be admin configured or user configured?

This is probably related to Device Classes

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+| **Discussion**    | https://github.com/jumpstarter-dev/jumpstarter/issues/41       |
+| **Requires**      |                                                                |
+| **Supersedes**    |                                                                |
+| **Superseded-By** |                                                                |
+
+---
+
+## Abstract
+
+This JEP proposes a Virtual Scalable Exporter subsystem for Jumpstarter that
```
kirkbrauer, line 21: We might want to specify that the thing that is really scaling here is not the exporter per say, but actually the targets themselves. For example, one exporter may have multiple targets, but I do understand this from the perspective of the "exporter" being the basic unit of scheduling such as a Pod in k8s.

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+
+## Proposal
+
+The proposal introduces **Virtual Scalable Exporters** — a controller-managed
+pool of virtual target instances with configurable autoscaling. Rather than
+treating virtual targets as purely on-demand or purely static, each virtual
+target definition declares scaling parameters that let administrators tune the
+trade-off between instant availability and resource consumption.
+
+### Core Concept: Managed Pools with Scaling
```
kirkbrauer: Suggestion: consider re-modeling the pool on native Kubernetes workload primitives so it reads like standard k8s to cluster admins. Replace the provider-typed `*ExporterPool` with a generic **`ExporterSet`** (a ReplicaSet+HPA analog): `spec.selector` + an inline `spec.template`, HPA/PDB scaling vocab (`minReplicas`/`maxReplicas`/`minAvailableReplicas`), Deployment-style status (`replicas`/`readyReplicas`/`availableReplicas`/`leasedReplicas`), and the `scale` subresource. This reuses the existing lease-selector→Exporter-label matching unchanged and lets `kubectl scale`/HPA/KEDA interoperate.

```yaml
kind: ExporterSet                  # ≈ ReplicaSet + HPA
spec:
  minReplicas: 0
  maxReplicas: 20
  minAvailableReplicas: 2          # PDB-style warm buffer (ready & unleased)
  selector:
    matchLabels:
      board: rpi4
  template:                        # embedded template (Deployment idiom)
    metadata:
      labels:
        board: rpi4
        virtual: "true"
    spec:
      drivers:
        - type: jumpstarter_driver_power.driver.QemuPower
        # ...
status:                            # Deployment-style counters
  replicas: 5
  readyReplicas: 3
  availableReplicas: 3             # warm
  leasedReplicas: 2
# scale subresource: specReplicasPath=.spec.maxReplicas
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+  # QEMU-specific configuration
+  machineType: virt
+  firmware: registry.example.com/firmware/rpi4:latest
+  resources:
+    cpu: 4
+    memory: 4Gi
+    storage: 16Gi
+    
+  # Exporter template (drivers exposed by each instance)
+  exporterTemplate:
```
kirkbrauer: Suggestion: rather than the pool *being* the provider-typed device, separate the **device** into a first-class `VirtualTarget`, keeping **`Exporter`** as the minimum leased unit (Pod-equivalent). The drivers/`exporterTemplate` here would move under a typed `*VirtualTarget` (e.g. `QEMUVirtualTarget`) the `Exporter` owns. Keeps the lease flow unchanged + unified for physical/virtual, localizes provider typing, and leaves room for one Exporter to expose multiple VirtualTargets later (multi-device benches).

```text
ExporterSet (generic)            ~ ReplicaSet + HPA
  └ Exporter (leasable; a Pod)   ~ Pod   ← minimum leased unit
      └ QEMUVirtualTarget        ~ the device (provider-typed)
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+  maxTotalInstances: 5
+  
+  labels:
+    board: rd1ae
+    flavor: kronos
+    virtual: "true"
+  
+  # Corellium-specific configuration
+  apiHost: app.corellium.com
+  apiCredentialsSecret: corellium-api-credentials  # Secret with keys: token
```
kirkbrauer: Suggestion: instead of inlining credentials/provisioning per pool, adopt the CSI `StorageClass`/PVC pattern. A cluster-scoped **`VirtualTargetClass`** holds the `provisioner`, an inline `credentialsSecretRef`, opaque `parameters` (apiHost/projectId/region), `bindingMode` (Immediate=warm vs WaitForFirstConsumer=provision-on-lease), and `reclaimPolicy`; the typed `*VirtualTarget` is the *claim* naming the class. Admins own classes + secrets; claim authors never touch credentials — like a PVC naming a StorageClass.

```yaml
# cluster-scoped (StorageClass analog) — admins own it
kind: VirtualTargetClass
metadata:
  name: corellium-kronos
spec:
  provisioner: corellium.jumpstarter.dev
  credentialsSecretRef:
    name: corellium-creds
    namespace: jumpstarter
  parameters:
    apiHost: app.corellium.com
    projectId: "778f..."
  bindingMode: WaitForFirstConsumer    # Immediate = pre-warmed pool
  reclaimPolicy: Delete
---
# the typed claim just names the class (PVC analog) — no creds here
kind: CorelliumVirtualTarget
spec:
  virtualTargetClassName: corellium-kronos
  deviceFlavor: kronos
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+**Rationale:** A single image is cheaper to build, test, and productize — there
+is one CI pipeline, one vulnerability scan, one release artifact. Deploying it
+as separate Deployments (one per provider) gives operational benefits: each
+provider has isolated logs, independent scaling, and can be restarted without
+affecting other providers. The `--provider` flag makes it explicit which CRD
+a given Deployment reconciles. Adding a new provider type means adding a new
+Deployment manifest pointing to the same image with a different flag — no new
+image build required.
+
+### DD-3: CRD per provider vs. generic CRD
```
kirkbrauer: Suggestion: a cleaner framing than "CRD per provider vs generic" (borrowed from CSI/CRI/Cluster-API): keep orchestration generic and make the device backend **pluggable via a `provisioner`** named on the class. The typed `*VirtualTarget` stays strongly-typed (your DD-3 win), while one `provisioner` string selects the backend — k8s container, EC2 instance, Corellium/REST, a vendor cloud-device API — all behind one interface, so the Exporter/lease experience is identical regardless of where the device runs. New backends add a claim kind + a provisioner, no pool-tier changes.

```text
VirtualTargetClass.provisioner →
  qemu.jumpstarter.dev        →  k8s container (+ OS OCI image volume)
  ec2.jumpstarter.dev         →  AWS API
  corellium.jumpstarter.dev   →  Corellium REST API
# one typed *VirtualTarget claim interface; backend is pluggable
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+must parse and validate provider-specific keys, driver templates must support
+runtime overrides, and the interaction between parameters and pool defaults
+(override vs. merge) must be defined and tested. All of this for a use case
+that is already satisfied by creating multiple pools with different resource
+profiles and letting users select via labels. The pool-flavors approach keeps
+the Lease API unchanged, requires no controller modifications, and is
+immediately understandable. Per-lease parameters can be revisited in a future
+JEP if the pool-flavors model proves insufficient.
+
+### DD-5: Built-in scaling vs. HPA / KEDA
```
kirkbrauer, line 1163: Suggestion (mostly endorsing the direction): keeping scaling **built-in and backend-aware** is the right call — each provisioner should implement autoscaling appropriate to its backend (local container churn vs. EC2 quotas/boot latency vs. an external device API all behave very differently), so a single generic autoscaler wouldn't fit. The only suggestion is to keep a **consistent scaling API/vocabulary across provisioners** — the same `minReplicas`/`maxReplicas`/`minAvailableReplicas` surface (and optionally the `scale` subresource, for `kubectl scale` familiarity) — even though each provisioner's scaling *logic* differs. One mental model for users; each backend still free to scale however suits it.

```yaml
# identical scaling surface across provisioners…
spec:
  minReplicas: 0
  maxReplicas: 20
  minAvailableReplicas: 2
# …but the logic differs per backend:
#   qemu      → fast local container churn
#   ec2       → quota / boot-latency aware
#   corellium → external API-rate aware
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+
+**Scaling Logic:** Each pool controller monitors its instances and scales based
+on available (unleased) instances:
+
+- If available instances drop below a threshold (e.g., `minWarmInstances`), scale up.
+- If available instances exceed demand for a cooldown period, scale down (never
+  below `minWarmInstances`).
+- Never exceed `maxTotalInstances` (if set; 0 or omitted means no upper bound).
+
+**Instance Lifecycle:**
```
kirkbrauer, line 769: Suggestion: for container-backed targets, compose the instance Pod from independently shipped artifacts — **exporter as a native sidecar** (`restartPolicy: Always`), an independent **target-runtime container** (QEMU/Cuttlefish), and the **OS image as an OCI artifact** via a Kubernetes image volume — instead of one container spawning QEMU via `Popen`. Benefits: independent release cadence, fault isolation (exporter survives a workload crash and can drain), and drivers attach over **standard interfaces (serial/SPI/CAN/GPIO via virtio)** so the same driver code works physical+virtual (socket/vsock backends also avoid privileged containers). API-backed targets skip the runtime container and connect out.

```yaml
# container-backed instance Pod (rendered by the provisioner)
spec:
  initContainers:
    - name: exporter            # native sidecar (starts first, drains last)
      restartPolicy: Always
  containers:
    - name: target-runtime      # QEMU/Cuttlefish — independent image
      volumeMounts:
        - name: os
          mountPath: /os
  volumes:
    - name: os
      image:
        reference: registry.example.com/os/rpi4:latest   # OS as OCI artifact
# drivers attach over virtio (serial/SPI/CAN/GPIO) → same code as physical
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+metadata:
+  name: rpi4-virtual
+  namespace: jumpstarter
+spec:
+  # Scaling configuration (shared across all pool CRDs)
+  minWarmInstances: 2        # Always keep 2 warm instances ready
+  maxTotalInstances: 20       # Scale up to 20 under load
+  
+  # Node scheduling (shared across all pool CRDs, optional)
+  nodeSelector:
```
kirkbrauer: Suggestion: node scheduling (this `nodeSelector`) is really a property of the *backend*, so consider folding it into the `VirtualTargetClass` as a `scheduling` block — `nodeSelector`/`nodeAffinity` plus **`tolerations`** (tainted KVM/GPU/baremetal/ARM nodes) and **device resource requests** (`kubernetes.io/arch`, `devices.kubevirt.io/kvm`, `nvidia.com/gpu`). CSI precedent: `StorageClass.allowedTopologies`. The rendered Pod inherits it, with optional per-`ExporterSet` override.

```yaml
kind: VirtualTargetClass
spec:
  scheduling:                    # inherited by the rendered exporter Pod
    nodeSelector:
      kubernetes.io/arch: arm64
    tolerations:
      - key: jumpstarter.dev/kvm
        operator: Exists
        effect: NoSchedule
    resources:
      limits:
        devices.kubevirt.io/kvm: "1"   # or nvidia.com/gpu
```

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+```
+
+The graceful scale-down sequence becomes:
+
+1. Pool controller sets `enabled: false` on the target exporter.
+2. Pool controller waits to confirm no lease was assigned (watches for
+   `status.leaseRef` to remain empty).
+3. Pool controller deletes the Pod and Exporter CR.
+
+### Hardware Considerations
```
kirkbrauer, line 950: Suggestion: add a concrete fidelity/cost ladder showing one logical target served by multiple classes — a container-backed software emulator/simulator (cheap, CI-scale), an API/cloud-backed virtual device (higher fidelity, metered), and real hardware (full fidelity, scarce) — all selected via labels through one `jmp lease`. For example, a target that needs a GPU or specialized I/O device: a software-emulated class runs functional checks cheaply in CI, a cloud-backed virtual device adds higher fidelity, and real hardware is the ground truth. This illustrates why keeping the `VirtualTarget`/class abstraction generic pays off across fidelity tiers.

One logical target, selected via labels through `jmp lease`:

| class (provisioner) | fidelity | scale/cost | role |
| --- | --- | --- | --- |
| container sim (qemu) | low | cheap / CI | functional checks |
| cloud virtual device (api) | high | metered | higher-fidelity behavior |
| real hardware (exporter) | full | scarce | ground truth |

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -0,0 +1,920 @@
...
+- **Scheduled (future-dated) leases:** The existing `jumpstarter-controller`
+  already supports `Spec.BeginTime` for scheduled leases. The controller does
+  not attempt to acquire an exporter until `BeginTime` arrives; it requeues
+  with a delay. Once `BeginTime` passes and no exporter is available, the
+  controller sets a `Pending` condition (e.g., reason `NotAvailable`). Pool
+  controllers watch for pending leases with matching labels as their scaling
+  input, so they naturally do not scale up for future-dated leases until the
+  controller makes them effective.
+
+## Future Possibilities
```
kirkbrauer, line 1444: Suggestion: list a few forward-compat items so the model stays open to them: disaggregated/cross-node accelerators (ARM64 runtime bridged to a remote GPU via virtio-gpu/RDMA), a separate reusable `*ProviderConfig` CRD (multi-account credential reuse/rotation), a realized-instance CRD (PV analog) for static/pre-provisioned devices, an `ExporterDeployment` rollout tier (Deployment analog), multiple/spawned-on-lease VirtualTargets per Exporter, and a universal physical+virtual `Target` abstraction.

---
*🤖 Drafted with AI assistance (Claude) and reviewed by the PR reviewer before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
kirkbrauer: @raballew I think if we can use virtio here, we might be able to get very high fidelity for guests that run natively on virtio targets and we can avoid privileged Pods in this case :)

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
kirkbrauer: @mangelajo Humm, yeah we'll need to think about this if this is related to a DeviceClass or ExporterClass since the Exporter is the primary scheduling unit in this proposal.

kirkbrauer (commented):
Thanks for putting this together, @mangelajo — the warm-pool + autoscaling direction is great and the design decisions are well argued. I've left a set of inline suggestions exploring whether we can lean harder on native Kubernetes primitives (so the proposal reads like standard k8s to cluster admins) and make the provider/device model more extensible.

The throughline of the inline comments:

- **Orchestration → `ExporterSet`** (a ReplicaSet+HPA analog): `selector` + an embedded `template`, HPA/PDB scaling vocab (`minReplicas`/`maxReplicas`/`minAvailableReplicas`), the `scale` subresource, and Deployment-style status. Reuses the existing lease-selector→Exporter-label matching unchanged.
- **Separate the device from the exporter**: keep `Exporter` as the minimum leased unit (Pod analog); move the provider-typed device into a first-class `VirtualTarget`.
- **CSI-style class + claim**: a `VirtualTargetClass` (StorageClass analog) with a **pluggable `provisioner`** (k8s container / EC2 / Corellium / a vendor cloud-device API), an inline `credentialsSecretRef`, `bindingMode` (warm vs provision-on-lease), `reclaimPolicy`, and node `scheduling`; the typed `*VirtualTarget` is the claim.
- **Endorse per-provisioner, backend-aware autoscaling** — the only ask is a consistent scaling API across provisioners.
- **Packaging**: exporter as a native sidecar + an independent runtime container + the OS image as an OCI artifact (image volume); drivers attach over standard interfaces (serial/SPI/CAN/GPIO via virtio), reused across physical/virtual.
- **Node scheduling on the class** (arch/KVM/GPU via tolerations + device resources).
- **A fidelity/cost ladder** of multiple classes (software-emulated → cloud virtual device → real hardware), all selectable through one `jmp lease`.
- A few **Future Possibilities** to keep the model open (cross-node accelerators, a `*ProviderConfig` for multi-account creds, a realized-instance CRD, an `ExporterDeployment` rollout tier, multi-target-per-exporter, a universal `Target`).

To make this concrete rather than abstract, I pushed a worked rewrite of the JEP on a separate branch so we can diff and discuss specifics:

- **Diff vs this PR:** https://github.com/jumpstarter-dev/jumpstarter/compare/jep-0014-virtual-scalable-exporters...jep-0014-k8s-remodel
- **Rendered file:** https://github.com/jumpstarter-dev/jumpstarter/blob/jep-0014-k8s-remodel/python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md

These are suggestions for discussion, not blockers — happy to scope any of them down to Future Possibilities.

---
*🤖 This review summary and the inline suggestions were drafted with AI assistance (Claude) and reviewed by me (the PR reviewer) before posting.*

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: done

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: More than the benefit of fault isolation I see the benefit of making the distribution of images simpler. We may need to figure out how can we properly call from one container to the other, as there is no way to exec from one side to the other, you can mount unix sockets, which may help.

I am taking this, but we must weight at implementation time, if this gets way too complicated we may want to convert this into a future improvement (I am feeling inclined to that as I write) :D

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Done

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Applied

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
mangelajo: Done

python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
raballew: @mangelajo I think we will need to add a conformance test because both virtual and physical targets would require the same base capabilities and thus drivers that cause the same outcomes. for example if you have a script that flashes, boots and accesses serial you expect it to work the exact same way on both targets our you would need a ton of if elses in your testing logic.

raballew (approved):
lgtm

kirkbrauer (approved):

PR #748, opened 2026-06-05
feat: driver/sdwire: support unprogrammed FT200X EEPROM + macOS storage/mux fixes

mmahut:
Recently, I got an unprogrammed unit from [3mdeb](https://shop.3mdeb.com/product/sdwire/), which did not work out of the box. It had the factory-default FTDI FT200X EEPROM (`0x0403`/`0x6015`) rather than the Samsung VID/PID set by `sd-mux-ctrl --init`. Also, previously the driver was Linux-only (pyudev).

To elaborate:

**1. Unprogrammed FT200X support**
Also matches `0x0403`/`0x6015`, but **only when an explicit `serial` is configured** -- a bare FT200X has no reliable runtime signature, so requiring a serial avoids false-positive matches on unrelated FTDI devices. Also raises a clear error whe multiple programmed SD Wires are present with no serial (instead of silently binding the first).

**2. macOS storage discovery**
`effective_storage_device()` now uses `system_profiler SPUSBDataType` on Darwin (pyudev is Linux-only). The SD reader is correlated to *its* FT200X as a sibling under the same internal hub, and the disk node is read from the reader's nested
`Media[].bsd_name`, so the right disk is targeted when several SD Wires are attached. Reads/writes retry discovery (USB re-enumerates after a switch) and fail with a clear error instead of a `TypeError`.

**3. macOS `dut()` ejects before switching**
The mux only switches when the SD bus is idle; `diskutil eject` (SCSI STOP UNIT) idles it. If the disk can't be determined, `dut()` aborts rather than risk corrupting a mounted volume.

**4. macOS `host()` power-cycles the reader's hub port**
`eject` leaves the SMSC reader stopped, so `host()` routes the card back first and then power-cycles port 1 of this SD Wire's internal hub (scoped by USB topology, so other attached devices aren't disturbed) to force a clean re-enumeration without a physical replug.

### DIFF
--- a/python/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
+++ b/python/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
@@ -1,8 +1,12 @@
 from __future__ import annotations
 
+import platform
+import subprocess
+import time
 from dataclasses import dataclass, field
 
-import pyudev
+import anyio
+import anyio.to_thread
 import usb.core
 import usb.util
 from jumpstarter_driver_opendal.driver import StorageMuxFlasherInterface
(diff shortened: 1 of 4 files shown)


raballew:
@mangelajo assigned you for reviews as it seems to be macos focused which i can not really test

python/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -1,77 +1,251 @@
...
+        context = pyudev.Context()
+        for udevice in (
+            context.list_devices(subsystem="usb")
+            .match_attribute("busnum", dev.bus)
+            .match_attribute("devnum", dev.address)
+        ):
+            for block in filter(lambda d: d.subsystem == "block", udevice.parent.children):
+                for storage_device in filter(lambda link: link.startswith("/dev/disk/by-diskseq/"), block.device_links):
+                    return storage_device
+    except Exception:
```
mangelajo: Can we be more explicit about Exception capture?  this will also probably fail our linter.

python/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -1,77 +1,251 @@
...
+        pass
+    return None
+
+
+def _node_is_controller(node: dict, serial: str | None) -> bool:
+    if serial is not None:
+        return node.get("serial_num") == serial
+    vid = node.get("vendor_id", "").lower()
+    pid = node.get("product_id", "").lower()
+    return ("0x04e8" in vid and "0x6001" in pid) or ("0x0403" in vid and "0x6015" in pid)
```
mangelajo: can we create/use constants for the VID/PIDs ?

python/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -1,77 +1,251 @@
...
+
+
+def _reader_bsd(node: dict) -> str | None:
+    # The disk node lives under the reader's nested "Media" list, not at top level.
+    vid = node.get("vendor_id", "").lower()
+    pid = node.get("product_id", "").lower()
+    if not ("0x0424" in vid and "0x4050" in pid):
```
mangelajo: same

python/packages/jumpstarter-driver-sdwire/jumpstarter_driver_sdwire/driver.py
```diff
@@ -1,77 +1,251 @@
...
+    return None
+
+
+def _find_storage_device_macos(serial: str | None) -> str | None:
+    # pyudev is Linux-only; on macOS walk system_profiler and pick the SD reader
+    # that is a sibling of this SD Wire's FT200X.
+    try:
+        import json
+
+        out = subprocess.check_output(
```
mangelajo, line 131: nice! :)

mangelajo (changes requested):
Thank you @mmahut this looks great, only a few comments, and we need to make the linter + ty + typos CI work.

I didn't even dare to fight the OSX battle for this, but I see there was more to it than I even expected. Thanks for making this better!

mangelajo:
@mmahut where can I ship you a bunch of Jumpstarter stickers? :D totally deserved! (ping me at majopela@redhat.com ) if you're interested.

Cheers and thanks again!

mmahut:
@mangelajo thank you for the great review! I have addressed these comments and force pushed.

Also, I will never say no to free stickers! :D

mmahut:
@mangelajo with regards to the failing CI right now:

* The typo us a false positive on a variable `pn` (port numbers), I have renamed it to `port_numbers` to avoid false positives
* e2e looks like timeout/network failures
* check-warnings looks unrelated to PR's changes

raballew:
We have been bumping a whole bunch of dependencies last/this week. It also increased the nitpickyness of the typos checker, check warnings is related to a sphinx version bump and addressed in https://github.com/jumpstarter-dev/jumpstarter/pull/778, e2e flakiness are likely failures in either hooks which @evakhoni is looking at or some transient infrastructure related issues (AWS was having trouble last week, GitHub CI in the beginning of this week) - after all it looks good to me

kirkbrauer (commented):
I think this looks really interesting, could be helpful for these macOS edge cases that can be tough to get out of without a Linux device lying around 😄

mmahut:
> I think this looks really interesting, could be helpful for these macOS edge cases that can be tough to get out of without a Linux device lying around 😄

As I'm waiting for a replacement of my framework laptop, this was indeed my case. 🥲

mangelajo (approved):

mangelajo:
I will run a quick check here, Linux + macOS, then merge

mangelajo:
I will test on Linux today, I was having some issues in macOS I will try to fix in a separate PR, since this works for you it’s a step forward :-)

raballew:
@mangelajo is this ready to merge?

raballew:
I dont see a reason why this should not be merged. Code looks good. Its basically working and @mangelajo and I approved. Lets merge and fix any left over bugs if we find them.

mangelajo:
We need to verify there is no regression in linux. I will open an issue to make sure I test it before 0.9

PR #776, opened 2026-06-08
fix: restrict Fedora to stable and group Go toolchain updates

raballew:
## Summary
- Restrict Fedora Docker image to `<=44` to prevent Renovate from proposing rawhide (Fedora 45). `followTag` does not work with Docker datasources since Docker tags lack the dist-tag-to-version mapping that npm provides.
- Create a `go-toolchain` group that bundles Go version directives, `.go-version`, and the `go-toolset` Docker image so they update in a single PR.
- Add a versioning regex for `go-toolset` so only Go-version-style tags (`1.x.y`) are tracked instead of UBI build tags.

## Test plan
- [ ] Verify Renovate no longer reports "Can't find version with tag latest for docker package fedora"
- [ ] Verify Renovate groups `.go-version`, Go version directives, and `go-toolset` image updates into a single `go-toolchain` PR


raballew:
Recreating with synced fork

PR #782, opened 2026-06-11
docs: add driver client/exporter compatibility rule

mangelajo:
## Summary

Adds a new project rule, `.claude/rules/driver-compatibility.md` (stored as `.cursor/rules/driver-compatibility.mdc` and symlinked, matching the existing pattern), and indexes it in `CLAUDE.md`.

It documents the backwards-compatibility invariants between independently-released clients and exporters, so future driver changes — human or AI-assisted — don't break mixed-version deployments:

- **`client()` class path stability**: the exporter reports this string and the client imports it from its own environment. Changing it for a released driver breaks old clients with `ImportError` — a failure mode invisible to unit tests. New drivers must always override `client()` explicitly.
- **Stable wire surfaces**: RPC method names/argument shapes and the `ClientStreamResource` / `PresignedRequestResource` handle types handled by `Driver.resource()`.
- **Client API deprecation policy**: deprecate public method parameters before removing them; keep `j` CLI commands and flags stable.
- **Dependency pinning**: released sdists exact-pin `jumpstarter*` deps via `hatch-pin-jumpstarter`, so no manual bounds in `pyproject.toml`.
- **Opendal migration checklist**: the pattern established in #535 for removing opendal from the remaining drivers (#441), including the `operator=` kwarg migration note for release notes.
- **Testing expectations**: end-to-end tests via `jumpstarter.common.utils.serve` when touching file-transfer paths.

## Why

While reviewing #535 (QEMU opendal removal), it turned out the migration was cross-version safe only because `QemuFlasher` explicitly overrides `client()` — drivers relying on a base-class default would have silently started reporting a class path that old clients can't import. With more drivers queued for the opendal removal series (#441), this invariant should be written down before the next migration rather than rediscovered.

## Notes for reviewers

- Doc-only change; no code is touched.
- Two judgement calls worth a look: the deprecation policy (one release of warnings before removing a public client API parameter) is proposed here as project policy — adjust if you want a different commitment; and the opendal checklist lives in this rule rather than in #441 because it's only actionable next to the `client()` invariant that makes migrations safe.

Related #441

### DIFF
--- a/.cursor/rules/driver-compatibility.mdc
+++ b/.cursor/rules/driver-compatibility.mdc
@@ -1,0 +1,81 @@
+---
+description: when modifying driver client classes, exported RPC methods, client() class paths, or migrating drivers away from opendal
+alwaysApply: false
+---
+# Driver Client/Exporter Compatibility
+
+Exporters and clients are released and upgraded independently. A lab may run an
+older exporter for months while clients update weekly, or vice versa. Any change
+to a driver must keep both directions working: old client ↔ new exporter and
+new client ↔ old exporter.
+
+## How client classes are resolved (the critical invariant)
+
+The exporter reports the string returned by the driver's `client()` classmethod
+(e.g. `"jumpstarter_driver_qemu.client.QemuFlasherClient"`). The client imports
+that class dynamically **from its own environment**.
+
+- **NEVER change the string returned by `client()` for a released driver.**
+  If an old client receives a class path that does not exist in its installed
+  packages, it fails with `ImportError` when leasing the exporter. This breakage
+  is invisible to unit tests — it only appears with mixed versions in the field.
+- When moving an interface ABC between packages (e.g. opendal → core), check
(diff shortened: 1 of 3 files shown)


.cursor/rules/driver-compatibility.mdc
```diff
@@ -0,0 +1,81 @@
+---
+description: when modifying driver client classes, exported RPC methods, client() class paths, or migrating drivers away from opendal
+alwaysApply: false
+---
+# Driver Client/Exporter Compatibility
+
+Exporters and clients are released and upgraded independently. A lab may run an
+older exporter for months while clients update weekly, or vice versa. Any change
```
kirkbrauer, line 8: I thought we defined a specific upgrade path here, not that both could be outdated?

kirkbrauer (commented):
Added some notes

mangelajo:
Close for now.

PR #785, opened 2026-06-12
fix: skip uninstalled child drivers in composite CLI

### LINKED ISSUE #784: Composite driver CLI crashes with ImportError when a child driver is not installed
## Summary

When the client connects to an exporter that exposes a driver not installed on the client side, the composite driver's `cli()` method crashes with an unhandled `ImportError` instead of gracefully skipping the unavailable driver.

## Error

```
ImportError: Driver 'jumpstarter_driver_dut_network.client.DutNetworkClient' is not installed.

This usually indicates a version mismatch between your client and the exporter.
Please try to update your client to the latest version and ensure the exporter has the
correct version installed.
```

## Root Cause

In `jumpstarter_driver_composite/client.py:55`, the `cli()` method iterates over child drivers and calls `hasattr(v, "cli")`. For stub clients (drivers not installed on the client), this triggers `__getattr__` (`jumpstarter/client/base.py:157`), which calls `_raise_missing_error()` (`base.py:126`), raising `ImportError`.

```python
# jumpstarter_driver_composite/client.py
for k, v in self.children.items():
    if hasattr(v, "cli"):       # <-- triggers __getattr__ on stub → ImportError
        base.add_command(v.cli(), k)
```

The `hasattr()` call is expected to return `False` for missing attributes, but the stub client's `__getattr__` raises `ImportError` instead of `AttributeError`, so `hasattr()` does not catch it (Python only catches `AttributeError` in `hasattr`).

## Expected Behavior

When a child driver is not installed on the client:
- The `cli()` method should skip that driver and only register CLI commands for available drivers
- A warning should be logged (the warning is already emitted during connection setup)
- The CLI should remain functional for all other installed drivers

## Suggested Fix

Either:

1. **Catch `ImportError` in the composite `cli()` loop:**
```python
for k, v in self.children.items():
    try:
        if hasattr(v, "cli"):
            base.add_command(v.cli(), k)
    except ImportError:
        pass  # skip unavailable drivers
```

2. **Or have stub clients raise `AttributeError` instead of `ImportError` in `__getattr__`**, which is the Python convention for missing attributes and would make `hasattr()` work correctly.

## Environment

- Client version: `0.8.1rc4.dev12`
- Python: 3.14
- Exporter exposing: `jumpstarter_driver_dut_network.client.DutNetworkClient`

## Traceback

<details>
<summary>Full traceback (click to expand)</summary>

```
jumpstarter_driver_composite/client.py:55 in cli
  54 │   │   for k, v in self.children.items():
❱ 55 │   │   │   if hasattr(v, "cli"):
  56 │   │   │   │   base.add_command(v.cli(), k)

jumpstarter/client/base.py:157 in __getattr__
  154 │   │   This ensures that calls like .on(), .off(), .write() etc. on stub clients
  155 │   │   raise a helpful ImportError instead of AttributeError.
  156 │   │   """
❱ 157 │   │   self._raise_missing_error()

jumpstarter/client/base.py:126 in _raise_missing_error
  123 │   │   """Raise ImportError with installation instructions."""
  124 │   │   class_path = self._get_missing_class_path()
  125 │   │   message = _format_missing_driver_message(class_path)
❱ 126 │   │   raise ImportError(message)
```

</details>

mangelajo:
## Summary
- Fix `CompositeClient.cli()` crashing with `ImportError` when a child driver is not installed on the client (fixes #784)
- Stub clients raise `ImportError` from `__getattr__` instead of `AttributeError`, so `getattr(v, "cli")` needs to catch both exception types to skip missing drivers
- The `ImportError` catch is scoped only to the attribute lookup — errors from `v.cli()` or `add_command()` still propagate normally
- Fix unrelated typos checker false positives on base64-encoded test tokens

## Test plan
- [x] `make pkg-test-jumpstarter-driver-composite` (6 tests pass)
- [x] Test: mixed installed + stub children — only installed commands registered
- [x] Test: all-stubs edge case — produces valid empty click group
- [x] Test: all-installed — every child command registered
- [x] Connect to an exporter with a composite driver where one child package is not installed on the client; verify CLI loads and available subcommands work

### DIFF
--- a/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
+++ b/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
@@ -5,6 +5,7 @@ import click
 from rich import traceback
 
 from jumpstarter.client import DriverClient
+from jumpstarter.client.base import StubDriverClient
 from jumpstarter.client.decorators import driver_click_group
 
 
@@ -52,12 +53,9 @@ class CompositeClient(DriverClient):
             pass
 
         for k, v in self.children.items():
-            try:
-                if hasattr(v, "cli"):
-                    base.add_command(v.cli(), k)
-            except ImportError:
-                # Stub clients raise ImportError from __getattr__ instead of
-                # AttributeError, so hasattr() does not catch missing drivers.
+            if isinstance(v, StubDriverClient):
                 continue
+            if hasattr(v, "cli"):
+                base.add_command(v.cli(), k)
 
         return base
--- a/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
+++ b/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
@@ -52,7 +52,12 @@ class CompositeClient(DriverClient):
             pass
 
         for k, v in self.children.items():
-            if hasattr(v, "cli"):
-                base.add_command(v.cli(), k)
+            try:
+                if hasattr(v, "cli"):
+                    base.add_command(v.cli(), k)
+            except ImportError:
+                # Stub clients raise ImportError from __getattr__ instead of
+                # AttributeError, so hasattr() does not catch missing drivers.
+                continue
 
         return base
--- a/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
+++ b/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
@@ -110,6 +110,36 @@ def test_cli_skips_uninstalled_child_drivers():
         assert "missing" not in cli_group.commands
 
 
+def test_cli_all_stubs_produces_empty_group():
+    """Composite CLI with only stub children should produce a valid empty group."""
+    with serve(
+        Composite(
+            children={
+                "missing1": MissingClientDriver(),
+                "missing2": MissingClientDriver(),
+            },
+        )
+    ) as client:
+        cli_group = client.cli()
+        assert cli_group.commands == {}
+
+
+def test_cli_all_installed_registers_all_commands():
+    """Composite CLI with all installed children should register every command."""
+    with serve(
+        Composite(
+            children={
+                "power0": MockPower(),
+                "power1": MockPower(),
+            },
+        )
+    ) as client:
+        cli_group = client.cli()
+        assert "power0" in cli_group.commands
+        assert "power1" in cli_group.commands
+        assert len(cli_group.commands) == len(client.children)
+
+
 def test_proxy_in_parent_child():
     """Test that parent driver can call methods on Proxy child (RideSX scenario)"""
     # Server-side test: verify parent accessing self.children["serial"].method()
--- a/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
+++ b/python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
@@ -2,8 +2,17 @@ from jumpstarter_driver_power.driver import MockPower
 from pydantic.dataclasses import dataclass
 
 from .driver import Composite, Proxy
+from jumpstarter.client.base import StubDriverClient
 from jumpstarter.common.utils import serve
 from jumpstarter.driver import Driver, export
+
+
+class MissingClientDriver(Driver):
+    """Test driver that returns a non-existent client class path."""
+
+    @classmethod
+    def client(cls) -> str:
+        return "nonexistent_driver_package.client.NonExistentClient"
 
 
 # Mock serial driver with a connect() method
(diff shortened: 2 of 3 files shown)


python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
```diff
@@ -52,7 +52,12 @@ def base():
...
         for k, v in self.children.items():
-            if hasattr(v, "cli"):
-                base.add_command(v.cli(), k)
+            try:
+                if hasattr(v, "cli"):
+                    base.add_command(v.cli(), k)
+            except ImportError:
+                # Stub clients raise ImportError from __getattr__ instead of
+                # AttributeError, so hasattr() does not catch missing drivers.
+                continue
```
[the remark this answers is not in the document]
mangelajo: hasattr will trigger the issue as well

python/pyproject.toml
```diff
@@ -97,6 +97,10 @@ known-local-folder = ["packages", "examples", "jumpstarter"]
 
 [tool.typos.default]
 locale = "en-us"
+extend-ignore-re = [
```
mangelajo: we had a problem with the latest update of the typos checker. This fixes it although it's not directly related to this patch, cannot be merged otherwise. It's a separate commit though

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
raballew, line 57: Consider narrowing the catch to only cover the `hasattr` call:

```python
for k, v in self.children.items():
    try:
        has_cli = hasattr(v, "cli")
    except ImportError:
        continue
    if has_cli:
        base.add_command(v.cli(), k)
```

Alternatively, you could skip stub clients explicitly via `isinstance(v, StubDriverClient)` to avoid the try/except entirely.

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
```diff
@@ -2,10 +2,19 @@
...
 from jumpstarter.common.utils import serve
 from jumpstarter.driver import Driver, export
 
 
+class MissingClientDriver(Driver):
+    """Test driver that returns a non-existent client class path."""
+
+    @classmethod
+    def client(cls) -> str:
+        return "nonexistent_driver_package.client.NonExistentClient"
```
raballew, line 15: There is no test for the all-stubs edge case. When every child is a `StubDriverClient`, the loop runs ImportError handling for each one, and the result should be a valid but empty click group.

raballew, line 15: There is no regression test for the all-installed case.

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
```diff
@@ -2,10 +2,19 @@
 from pydantic.dataclasses import dataclass
 
 from .driver import Composite, Proxy
+from jumpstarter.client.base import StubDriverClient
```
raballew, line 5: Consider either exporting `StubDriverClient` from `jumpstarter.client.__init__` (it is a meaningful public concept for handling missing drivers) or using duck-typing assertions in the test instead.

raballew (changes requested):

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
mangelajo: What do you mean?

mangelajo: good catch!

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/client.py
mangelajo: Done.

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
raballew: Test exists: some children are stubs (missing drivers) -> they get skipped
Test exists: all children are stubs -> empty CLI group
Test missing: all children are real, installed drivers -> every child registers a CLI command

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
mangelajo: Done

python/packages/jumpstarter-driver-composite/jumpstarter_driver_composite/driver_test.py
mangelajo: nice, covered by test_cli_all_installed_registers_all_commands now

python/pyproject.toml
raballew: why does it need to be defined twice? once here and again in typos.toml?

mangelajo: The GitHub Actions typos check uses ./typos.toml (the root config), not python/pyproject.toml. So we actually do need the base64 regex in the root typos.toml. The python linter (that can be ran locally as well) uses the pyproject one...

kirkbrauer (approved):

raballew (approved):

PR #789, opened 2026-06-13
feat: add jumpstarter-driver-obd for OBD-II vehicle diagnostics

mmahut:
This adds an OBD-II driver for live diagnostics off a real vehicle through one of those cheap ELM327 USB adapters. It wraps [python-obd](https://python-obd.readthedocs.io/) and was tested in my garage! :) Ignition off through engine running, a full sweep of all ~100 PIDs it advertised, and DTC reads tested.

obd-car-rpi
  
 What it can do:

  | Method | What it does |
  |--------|--------------|
  | `query(name)` | Read a PID by name (`RPM`, `SPEED`, …); `None` if the ECU doesn't answer |
  | `clear_dtc()` | Clear stored fault codes / freeze frame - destructive, see below |
  | `status()` | Connection state (`Car Connected`, `ELM Connected`, …) |
  | `supported_commands()` | Sorted list of PIDs the ECU advertises |
  | `is_connected()` | True only when a car is actually on the bus |

  A few decisions worth flagging for review:

  - **`query()` is read-only.** Clearing codes is OBD mode 04 — it wipes stored DTCs *and* resets the emissions readiness monitors, so it's behind its own `clear_dtc()` rather than reachable by passing `"CLEAR_DTC"` to `query()`.
  - A couple of PIDs don't stringify nicely — `VIN` comes back as a bytearray and `STATUS` as an object with no `__str__`; so those are handled explicitly.
 - `fast` mode is off by default: it's quicker but flaky on the clone adapters most people actually have (like the cheap one I have). If you have a professional one, you can still use the fast mode.

### DIFF
--- a/python/packages/jumpstarter-driver-obd/README.md
+++ b/python/packages/jumpstarter-driver-obd/README.md
@@ -1,0 +1,119 @@
+# OBD-II Driver
+
+`jumpstarter-driver-obd` wraps [python-obd](https://python-obd.readthedocs.io/en/latest/)
+to query On-Board Diagnostics (OBD-II) PIDs from a vehicle ECU via an ELM327 adapter.
+
+## Installation
+
+```{code-block} console
+:substitutions:
+$ pip3 install --extra-index-url {{index_url}} jumpstarter-driver-obd
+```
+
+## Hardware Requirements
+
+- An ELM327 USB adapter (e.g. PremiumCord ELM327 USB)
+- A vehicle with an OBD-II port (petrol cars from 2001+, diesel from 2004+, all US cars from 1996+)
+- **macOS only**: ELM327 USB cables use a CH340 or CP2102 USB-serial chip that may need a kernel
+  extension — install the appropriate driver if the port does not appear after plugging in
+
+## Configuration
+
+### Auto-detect (recommended for a single adapter)
+
+Leave `port` unset or `null` to let python-obd scan available serial ports and connect to the first
+ELM327 it finds:
+
+```yaml
+export:
+  obd:
+    type: jumpstarter_driver_obd.driver.OBD
+    config:
+      port: null
+```
+
+```{note}
+Auto-detect is reliable on Linux (e.g. `/dev/ttyUSB0`). On **macOS** it is not: python-obd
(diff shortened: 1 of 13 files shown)


python/packages/jumpstarter-driver-obd/jumpstarter_driver_obd/driver.py
```diff
@@ -0,0 +1,107 @@
...
+
+        Returns None if the ECU doesn't answer. Destructive commands such as
+        CLEAR_DTC are rejected here; use clear_dtc() instead.
+        """
+        if command_name in DESTRUCTIVE_COMMANDS:
+            raise ValueError(
+                f"{command_name} is destructive (clears stored DTCs and resets "
+                f"readiness monitors) and cannot be sent via query(); use clear_dtc()"
+            )
+        cmd = getattr(obd.commands, command_name, None)
```
bennyz: can we use has_name here? https://python-obd.readthedocs.io/en/latest/Command%20Lookup/

bennyz:
cool!

python/packages/jumpstarter-driver-obd/README.md
```diff
@@ -0,0 +1,113 @@
...
+|-----------|-------------|------|----------|---------|
+| port | Serial port path; `null` to auto-detect an ELM327 adapter | str \| null | no | null |
+| baudrate | ELM327 baud rate | int | no | 38400 |
+| fast | Enable ELM327 fast mode (~100–400 ms faster per query); unreliable on cheap clone adapters, so it is opt-in | bool | no | false |
+
+## Usage
+
+```python
+# Check connection state
+print(obd.status())           # e.g. "Car Connected"
```
bennyz: would it make sense to have something like:
```python
class OBDConnectionStatus(StrEnum):
      NOT_CONNECTED = "Not Connected"
      ELM_CONNECTED = "ELM Connected"
      OBD_CONNECTED = "OBD Connected"
      CAR_CONNECTED = "Car Connected"
```

then clients could match against the enum variant

mmahut: That's a good idea!

mangelajo (approved):

mangelajo:
Thank you so much for this contribution!

PR #790, opened 2026-06-13
feat: parse power measurements in http-power and add a power read CLI command

mmahut:
Previously the http-power driver's `read()` always returned dummy `(0.0, 0.0)` values.

`read()` now parses the JSON the read endpoint returns and pulls voltage/current out of it. By default it looks for top-level `voltage`/`current` keys, but you can point `voltage_path`/`current_path` at a dotted path (`emeter.voltage`, `StatusSNS.ENERGY.Voltage`, `meters.0.voltage`) for devices that nest them.

Tested against a real Shelly Plug S Gen3. 

j-power


python/packages/jumpstarter-driver-http-power/jumpstarter_driver_http_power/driver.py
```diff
@@ -92,20 +105,40 @@ def off(self):
...
+        try:
+            value = _json_path(data, key)
+        except (KeyError, IndexError, TypeError, ValueError):
+            if path is not None:
+                raise ValueError(f"configured path {key!r} not found in read response") from None
+            return 0.0
+        try:
+            return float(value)
+        except (TypeError, ValueError):
+            raise ValueError(f"value at {key!r} is not numeric: {value!r}") from None
```
raballew: The error message includes `{value!r}`, which dumps the full Python repr of whatever sits at the targeted path. Can a device response contain sensitive fields that should not leak into logs or client-visible error messages? If so consider including only the type name instead:

```python
raise ValueError(f"value at {key!r} is not numeric (got {type(value).__name__})") from None
```

python/packages/jumpstarter-driver-http-power/jumpstarter_driver_http_power/driver.py
```diff
@@ -8,11 +9,23 @@
 from jumpstarter.driver import Driver, export
 
 
+def _json_path(data, path: str):
```
raballew: Annotations missing:

```python
def _json_path(data: Any, path: str) -> Any:
def _extract_reading(data: Any, path: Optional[str], default_key: str) -> float:
```

python/packages/jumpstarter-driver-http-power/README.md
```diff
@@ -106,8 +127,18 @@ http_power_client.off()
...
+voltage and current from `voltage_path` / `current_path` (defaulting to top-level
+`voltage` and `current` keys). A field the device doesn't report reads as `0.0`; a
+configured path that isn't found raises an error.
+
+```yaml
+      power_read:
+        url: "http://192.168.1.65/cm?cmnd=Status%2010"   # Tasmota
+        voltage_path: "StatusSNS.ENERGY.Voltage"
+        current_path: "StatusSNS.ENERGY.Current"
+```
```
mangelajo, line 153: Can you provide the examples for the shelly? I have one and would love to test it :)

python/packages/jumpstarter-driver-power/jumpstarter_driver_power/client.py
```diff
@@ -58,6 +58,15 @@ def cycle(wait):
             click.echo(f"Power cycling with {wait} seconds wait time...")
             self.cycle(wait)
 
+        @base.command()
+        def read():
+            """Read power measurements"""
+            for reading in self.read():
```
mangelajo: for the CLI we should probably perform a single read. Otherwise this will remain in a loop reading powers without any throttling.

Another option is to provide a number of readings, and the interval between them, may be default to 1s ?

mangelajo (changes requested):
just a couple of comments but looks great :) thank you so much!

raballew (dismissed):

python/packages/jumpstarter-driver-http-power/jumpstarter_driver_http_power/driver.py
mmahut: That is a good comment, I did not think of sensitive data, but we can never be sure.

mangelajo (approved):
Looks great, thanks a lot @mmahut

PR #791, opened 2026-06-14
feat: add Ctrl-] x3 hotkey to power cycle board from serial console

evakhoni:
## Summary

- Add Ctrl-] x3 hotkey to the serial console that power cycles the board without exiting the session
- Auto-discover the power driver from the DUT tree using label-based matching against known power clients (PowerClient, VirtualPowerClient)
- Configurable via `power_control_ref` (explicit device name) and `power_control_method` (method sequence with sleep support)
- Disables safely on ambiguity (multiple power drivers), missing ref, or invalid methods — set `power_control_method` to `[]` or `null` to disable explicitly
- Reconnect the console automatically if the serial stream drops after the power cycle
- Documented new config parameters in README and field docstrings

## How it works

The root client is back-referenced onto every driver client after tree construction. `PySerialClient` walks the tree matching `jumpstarter.dev/client` labels against a known-power-clients allowlist. If exactly one match is found, it's used; multiple matches disable the hotkey with a warning. The method sequence is pre-validated at console start — invalid methods or malformed `sleep:` values disable the hotkey immediately rather than failing at press time. Configuration is passed from exporter to client via `extra_labels()`.

## Test plan

- [ ] With a power driver in the DUT tree: power cycle hint appears, Ctrl-] x3 triggers a power cycle, console survives
- [ ] Without a power driver: no hint, Ctrl-] x3 has no effect
- [ ] Multiple power drivers: warning logged, hotkey disabled
- [ ] `power_control_ref` set to explicit device: uses that device directly
- [ ] `power_control_method: null` or `[]`: hotkey disabled
- [ ] Ctrl-B x3 exits cleanly

### DIFF
--- a/python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
+++ b/python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
@@ -1,4 +1,5 @@
 import sys
+import time
 from contextlib import contextmanager
 from typing import Optional
 
@@ -8,7 +9,7 @@ from anyio.streams.file import FileReadStream
 from jumpstarter_driver_network.adapters import PexpectAdapter
 from pexpect.fdpexpect import fdspawn
 
-from .console import Console
+from .console import Console, ConsoleStreamDrop
 from jumpstarter.client import DriverClient
 from jumpstarter.client.decorators import driver_click_group
 
@@ -126,9 +127,10 @@ class PySerialClient(DriverClient):
         return bytes_read, bytes_sent
 
     def _find_power_client(self):
-        if self.root is None:
+        root = getattr(self, 'root', None)
+        if root is None:
             return None
-        return self._search_power(self.root)
+        return self._search_power(root)
 
     def _search_power(self, client):
         for child in client.children.values():
@@ -163,8 +165,18 @@ class PySerialClient(DriverClient):
             click.echo("\nStarting serial port console ... exit with CTRL+B x 3 times\n")
             if on_power_cycle is not None:
                 click.echo("Power cycle: CTRL+] x 3 times\n")
-            console = Console(serial_client=self, on_power_cycle=on_power_cycle)
-            console.run()
+            retries = 0
+            while retries < 30:
+                console = Console(serial_client=self, on_power_cycle=on_power_cycle)
+                try:
+                    console.run()
+                    break
+                except ConsoleStreamDrop:
+                    click.echo("\r\nSerial connection lost, reconnecting...\n", err=True)
+                    retries += 1
+                    time.sleep(1)
+            else:
+                click.echo("\nSerial connection lost (reconnect attempts exhausted).\n", err=True)
 
         @base.command()
         @click.option(
--- a/python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
+++ b/python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
@@ -3,7 +3,7 @@ from contextlib import contextmanager
 from typing import Optional
 
 import click
-from anyio import BrokenResourceError, EndOfStream, create_task_group, open_file
+from anyio import BrokenResourceError, EndOfStream, create_task_group, open_file, sleep, to_thread
 from anyio.streams.file import FileReadStream
 from jumpstarter_driver_network.adapters import PexpectAdapter
 from pexpect.fdpexpect import fdspawn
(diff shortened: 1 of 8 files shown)


LuuOW (commented):
Technical audit: Implementation verified for architectural consistency and engineering integrity.

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
```diff
@@ -125,6 +126,31 @@ async def _stdin_to_serial(self, stream) -> tuple[int, int]:
...
+        return self._search_power(root)
+
+    def _search_power(self, client):
+        for child in client.children.values():
+            if hasattr(child, "cycle") or (hasattr(child, "on") and hasattr(child, "off")):
+                return child
+            result = self._search_power(child)
+            if result is not None:
+                return result
+        return None
```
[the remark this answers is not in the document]
evakhoni: I doubt that we have any power capability in the root node directly.

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/console_test.py
```diff
@@ -0,0 +1,89 @@
...
+            console = Console(serial_client=client, on_power_cycle=on_power_cycle)
+            try:
+                console.run()
+            except Exception as e:
+                result["exc"] = e
+
+    t = threading.Thread(target=_run, daemon=True)
+    t.start()
+    return master_fd, t, result
+
```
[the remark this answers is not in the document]
evakhoni: sounds good i'll fix that

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
```diff
@@ -125,6 +126,31 @@ async def _stdin_to_serial(self, stream) -> tuple[int, int]:
 
         return bytes_read, bytes_sent
 
+    def _find_power_client(self):
```
mangelajo, line 140: I think we should provide a config entry to let admins specify the power device via a "ref". While I think that finding it automatically is cool, and will work out of the box in most cases, imagine environments where you have multiple power controls , and the power controls could not be what you are expecting, or the reset mechanism is different.

Even in some cases the method to be called could be "reset()", i.e. in the esp32 controller.


So I would take a config with "ref" + method  in the serial config

mangelajo: i.e.
```
export:
   serial:
      type:....
      config:
             ....
      reset_control:
         device:
            ref: "esp32"
         commands:
              - "reset()"
```

```
export:
   serial:
      type:....
      config:
             ....
      reset_control:
         device:
            ref: "power"
         commands:
              - "on()"
              - "sleep 2"
              - "off()"
```

or something like this.

mangelajo (changes requested):

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
evakhoni: yeah config sounds good. but if unset let it fallback to the auto discovery WDYT? otherwise I guess the user base would be rather small..
+1 on the reset(), i thought I seen it somewhere but forgot it was esp32.

mangelajo: IMHO auto detect continues to be risky, it could pick up the wrong power device if you had multiple doing different things in your setup. We could provide a flag for auto-detection, but explain very clearly what it does, and should be disabled by default.

mangelajo: or even enabled by default.. but a flag :D

mangelajo (changes requested):
This one is close @evakhoni  :D

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
evakhoni: Yep, that's a valid concern. I think I found a middle ground: conservative auto-discovery that disables on first trouble + explicit config when needed.

Added `power_control_ref` and `power_control_method` fields — you can now specify which device and what sequence to call:
```yaml
power_control_ref: "esp32"
power_control_method: ["hard_reset"]
```

Auto-discovery now uses a strict allowlist (only `PowerClient`/`VirtualPowerClient` labels) and disables itself if it finds multiple candidates. No flag needed since it's conservative by design.

Ready for re-review when you have time!

evakhoni: Btw if needed - there's an option to disable it via the config, by setting 
```
  - power_control_method: [] ✅
  - power_control_method: "" ✅
  - power_control_method: null ✅
```

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/client.py
```diff
@@ -125,6 +134,77 @@ async def _stdin_to_serial(self, stream) -> tuple[int, int]:
...
+                    )
+                    return None
+
+        async def _cycle():
+            for step in methods:
+                if step.startswith("sleep:"):
+                    delay = float(step.split(":", 1)[1])
+                    await sleep(delay)
+                else:
+                    await to_thread.run_sync(getattr(power_client, step))
```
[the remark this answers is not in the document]
evakhoni: Fixed. Now validating with callable(getattr(...)) and parsing sleep: values at construction time. Validation errors surface immediately instead of at hotkey press.

raballew:
@mangelajo @evakhoni whats the plan with this one? It seems quite old. Do we want to merge or close this one?

PR #797, opened 2026-06-16
fix: construct image URL after HTTP server starts

bennyz:
get_url() now returns _bound_port (since #740) which is 0 until start() is called. The flash client was calling get_url() before _services_up() started the server, producing http://host:0/image URLs.

Move image_url construction inside the _services_up() context so get_url() runs after the server binds and _bound_port is set.


eballetbo (approved):
LGTM - I reported the problem and after applying this PR I can say that fixes the issue. Logs report the correct port now and the flashing workflow is able to find the file in the http server and flash it.

mangelajo (approved):

PR #803, opened 2026-06-17
feat: add description field to access policy rules

bennyz:
Surface policy descriptions in NoAccess lease errors so clients can see which policies blocked their access and why.

```shell
jmp shell --client repro-client -l pool=demos
[06/17/26 10:14:49] INFO     [jumpstarter.client.lease] Acquiring lease 019ed46e-c223-7b5f-b9b9-5c11bf6f6f2a for selector pool=demos for duration 0:30:00
                    INFO     Releasing Lease 019ed46e-c223-7b5f-b9b9-5c11bf6f6f2a
Error: the lease cannot be satisfied: While there are 1 exporters matching the selector, none of them are approved by any policy for your client. Matching policies: Requires demo-team=true label on your client. Contact admins if you need the label
```

### DIFF
--- a/controller/internal/controller/lease_controller.go
+++ b/controller/internal/controller/lease_controller.go
@@ -255,17 +255,27 @@ func (r *LeaseReconciler) reconcileStatusExporterRef(
 			matchingExporters = listed.Items
 		}
 
-		approvedExporters, err := r.attachMatchingPolicies(ctx, lease, matchingExporters)
+		approvedExporters, unmatchedDescriptions, err := r.attachMatchingPolicies(ctx, lease, matchingExporters)
 		if err != nil {
 			return fmt.Errorf("reconcileStatusExporterRef: failed to handle policy approval: %w", err)
 		}
 
 		if len(approvedExporters) == 0 {
-			lease.SetStatusUnsatisfiable(
-				"NoAccess",
-				"While there are %d exporters matching the selector, none of them are approved by any policy for your client",
-				len(matchingExporters),
-			)
+			if len(unmatchedDescriptions) > 0 {
+				desc := strings.Join(unmatchedDescriptions, "; ")
+				if len(desc) > 4096 {
+					desc = desc[:4096] + "..."
+				}
+				lease.SetStatusUnsatisfiable("NoAccess",
+					"While there are %d exporters matching the selector, none of them are approved by any policy for your client. Matching policies: %s",
+					len(matchingExporters), desc,
+				)
+			} else {
+				lease.SetStatusUnsatisfiable("NoAccess",
+					"While there are %d exporters matching the selector, none of them are approved by any policy for your client",
+					len(matchingExporters),
+				)
+			}
 			return nil
 		}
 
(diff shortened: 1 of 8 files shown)


controller/internal/controller/lease_controller.go
```diff
@@ -128,40 +122,45 @@ func (r *LeaseReconciler) reconcileStatusEnded(
...
-				expiration = lease.Status.BeginTime.Add(lease.Spec.Duration.Duration)
-			}
 
-			if expiration.Before(now) {
-				lease.Expire(ctx)
-				return nil
-			}
-			result.RequeueAfter = expiration.Sub(now)
-			return nil
+	// if lease has status condition unsatisfiable or invalid, we mark it as ended to avoid reprocessing
```
mangelajo: nice cleanup, just checked that it's the same, but separated ifs to limit indent.

controller/internal/controller/lease_controller.go
```diff
@@ -255,17 +254,41 @@ func (r *LeaseReconciler) reconcileStatusExporterRef(
 			matchingExporters = listed.Items
 		}
 
-		approvedExporters, err := r.attachMatchingPolicies(ctx, lease, matchingExporters)
+		approvedExporters, policyDescriptions, err := r.attachMatchingPolicies(ctx, lease, matchingExporters)
```
mangelajo:
```suggestion
		approvedExporters, unmatchedDescriptions, err := r.attachMatchingPolicies(ctx, lease, matchingExporters)
```

mangelajo: otherwise it's a bit confusing, on first read I was understanding that it was matched policies and it didn't make sense.

controller/internal/controller/lease_controller.go
```diff
@@ -255,17 +254,41 @@ func (r *LeaseReconciler) reconcileStatusExporterRef(
...
+					if length == 0 {
+						length = len(d)
+					} else {
+						length += len("; ") + len(d)
+					}
+				}
+				desc := strings.Join(included, "; ")
+				if remaining := len(policyDescriptions) - len(included); remaining > 0 {
+					desc = fmt.Sprintf("%s (%d more)", desc, remaining)
+				}
```
mangelajo: can we extract this to a separate function?

controller/internal/controller/lease_controller.go
```diff
@@ -563,7 +593,6 @@ func filterOutOfflineExporters(approvedExporters []ApprovedExporter) []ApprovedE
 			)
 		},
 	)
-	return onlineExporters
```
mangelajo, line 566: nit, no need to change it now, but... please ask the LLM to not make unrelated changes. While I understand this can be simplified, the variable served as a "comment" :) to make it easier to understand what was being down.

python/packages/jumpstarter/jumpstarter/client/lease_test.py
```diff
@@ -353,6 +353,92 @@ def test_throttling_not_applied_when_console_available(self):
             assert mock_spinner.update.call_count == 4
 
 
+class TestAcquireUnsatisfiable:
```
mangelajo: those tests aren't really interacting with the server. just Unit tests in python, pure hallucination I believe :-P :) can you check? :)

mangelajo (changes requested):
nice improvement, can you check the comments?

python/packages/jumpstarter/jumpstarter/client/lease_test.py
bennyz: yeah it wasn't supposed to be committed 🤦🏻

mangelajo (approved):

PR #804, opened 2026-06-17
fix(renovate): store exact Python version, migrate devfile images to uv

raballew:
## Summary
- Switch `.py-version` from `MAJOR.MINOR` (`3.12`) to `MAJOR.MINOR.PATCH` (`3.12.11`) and update the Renovate regex to match three version segments -- fixes the `renovate/python-3.x` branch creation failure where the two-segment regex couldn't verify a three-segment replacement (`expectedValue: "3.14.6", foundValue: "14.6"`)
- Migrate both devfile Containerfiles from `dnf install python3.X` / `pip install` to `uv python install` / `uv pip install`, reading the Python version from `.py-version` instead of hardcoding it
- Add `python-base` intermediary stage in `Containerfile.client` for layer caching (system deps cached separately from wheel installation)
- Replace bash alias (`alias python='uv run python'`) with `uv venv` + `ENV PATH` so `python` is available in all contexts (scripts, podman exec, Makefiles), not just interactive bash shells
- Pin the builder stage to the same `.py-version` Python via `uv python install` + `UV_PYTHON_PREFERENCE=only-managed` so wheels are built with the same interpreter as the runtime
- Set group ownership (GID 0) on `.local`, `.cache`, and `.venv` in both Containerfiles for OpenShift arbitrary UID support
- Update `name` labels from `udi9` to `udi10` to match the `ubi10-latest` base image

## Test plan
- [x] Validated `renovate.jsonc` parses correctly (`npx json5`)
- [x] Ran Renovate v43.218.0 dry-run against fork -- confirmed `.py-version` matched, `currentValue: "3.12.11"` extracted, and `DRY-RUN: Would commit files to branch renovate/python-3.x` (no "Value is not updated" error)
- [x] Build `python/.devfile/Containerfile.client` and verify `uv python install` succeeds
- [x] Build `python/.devfile/Containerfile.client` and verify wheel + pytest installation into venv
- [x] `podman run --rm jumpstarter-client:test python --version` returns Python 3.12.11
- [x] `podman run --rm jumpstarter-client:test uv pip list --python /home/user/.venv/bin/python | grep jumpstarter` lists all packages
- [x] Verified builder stage uses Python 3.12.11 (via `UV_VERBOSE=1`) with and without `UV_PYTHON_PREFERENCE=only-managed`


raballew:
I had a bit of trouble with the build in Containerfile.client - I have seen it using the system python version once so I made sure it always prefers the UV managed one.

python/.devfile/Containerfile.client
```diff
@@ -3,41 +3,38 @@ COPY --from=ghcr.io/astral-sh/uv:0.11.21@sha256:ff07b86af50d4d9391d9daf4ff89ce42
...
     rm -rf /var/cache/dnf
+COPY .py-version /tmp/.py-version
+RUN uv python install "$(cat /tmp/.py-version)"
+ENV UV_PYTHON_PREFERENCE=only-managed
 ADD . /src
 RUN make -C /src/python build
 
-
-FROM quay.io/devfile/base-developer-image:ubi10-latest@sha256:6b9d2a3b90798e6d92c1c6b05cd2ce733b989cc6f0b3ff4e0820b45ad19fe474
+FROM quay.io/devfile/base-developer-image:ubi10-latest@sha256:6b9d2a3b90798e6d92c1c6b05cd2ce733b989cc6f0b3ff4e0820b45ad19fe474 AS python-base
```
mangelajo, line 12: we should not tag a latest tracking sha256

.py-version
```diff
@@ -1 +1 @@
-3.12
+3.12.11
```
mangelajo, line 1: why going so specific with the python version?

Isn't this too much? :D

python/.devfile/Containerfile
```diff
@@ -1,22 +1,25 @@
 FROM quay.io/devfile/base-developer-image:ubi10-latest@sha256:6b9d2a3b90798e6d92c1c6b05cd2ce733b989cc6f0b3ff4e0820b45ad19fe474
 LABEL maintainer="Jumpstarter.dev"
 
-LABEL name="devfile/udi9/jumpstarter"
+LABEL name="devfile/udi10/jumpstarter"
```
mangelajo, line 4: niiice

mangelajo (changes requested):

.py-version
raballew: you need to pin the version for CPython and a few other things that could be flaky. I dont think its too much tbh.

python/.devfile/Containerfile.client
raballew: good point! i have to admit that those red hat provided container images have the worst versioning scheme of all (same for ubi's)  - they are tracking the os version (9->10) and the dev base image version in it which can not be properly handled. how do you know when you can move the os from 9->10 without breaking anything in the dev base image? really really annoying if you ask me.

raballew: i will address this in a follow up PR

.py-version
mangelajo: ok, let's try

mangelajo (approved):

PR #809, opened 2026-06-18
test(qemu): replace Fedora Cloud image with Alpine Linux tiny image

mangelajo:
## Motivation

Fedora Cloud qcow2 images used in QEMU driver tests weigh ~556 MB (x86_64) + ~519 MB (aarch64) = **~1.1 GB total**, taking ~1 minute to download on every CI run. Alpine Linux 3.22.4 UEFI tiny images reduce that to **~278 MB total** — a 74% reduction — and download in ~6 s on GitHub Actions runners, making a cache unnecessary.

| | Fedora 43 | Alpine 3.22.4 tiny | Reduction |
|---|---|---|---|
| x86\_64 | 556 MB | 127 MB | 77% |
| aarch64 | 519 MB | 151 MB | 71% |
| **Total** | **~1,076 MB** | **~278 MB** | **74%** |

Alpine's `nocloud` tiny images support UEFI boot on both architectures and use `tiny-cloud` (a minimal cloud-init alternative) that reads the standard NoCloud CIDATA vfat volume the QEMU driver already generates.

## Driver changes (`driver.py`)

`tiny-cloud` differs from full cloud-init in two ways:

1. **Hostname**: tiny-cloud reads `@hostname` from the YAML key `hostname` in `meta-data`, not `local-hostname`. Add `hostname` alongside `local-hostname` so both implementations set it correctly.

2. **Password**: tiny-cloud ignores `plain_text_passwd` in the `users` stanza. Add a `runcmd` entry to set the password via `chpasswd`. Credentials are escaped with `shlex.quote()` so special characters are safe. Full cloud-init images are unaffected.

**`hostport=0` support**: `Hostfwd` now allows `hostport=0`, which tells QEMU's user-mode networking to pick a free port automatically. After QEMU starts, the driver queries the actual assigned port via QMP `human-monitor-command info usernet`, logs it at INFO level:

```
hostfwd 'ssh': resolved port 0 -> 127.0.0.1:<port> (guest port 22)
```

and stores it in `_resolved_hostports` so `get_hostfwd_port()` can return it to the client side.

## Client changes (`client.py`)

`shell()` now checks for an `ssh` hostfwd entry by calling `get_hostfwd_port('ssh')`. If present, it opens a direct `fabric.Connection` to `127.0.0.1:<resolved_port>`, bypassing the jumpstarter streaming layer entirely. If no ssh hostfwd is configured (`KeyError`), it falls back to the original `FabricAdapter` path over vsock.

This is necessary because **Alpine uses OpenRC, not systemd**, so `systemd-ssh-generator` never runs and sshd only listens on TCP. Fedora cloud images ship `systemd-ssh-generator` (systemd ≥ 256) which automatically binds sshd to `AF_VSOCK` port 22 inside VMs — that is why the original vsock path worked with Fedora but not Alpine.

## Test changes (`driver_test.py`)

- Pass `hostfwd={'ssh': {'hostport': 0, 'guestport': 22}}` so `qemu.shell()` uses the TCP path with a QEMU-assigned port; no probe socket needed.
- Wait for `bootstrap_complete: done` on the serial console before attempting login, guaranteeing tiny-cloud has set the password and sshd is ready (avoids the race that existed with Fedora too, where the extra `setenforce` console round-trip happened to provide enough delay).
- Press Enter if the GRUB countdown appears to skip the 10 s wait.
- Update post-login prompt from bash `[user@host ~]$` → Alpine ash `host:~$`.
- Remove `sudo setenforce 0` (no SELinux on Alpine).
- Relax `uname -r` assertion to a non-empty check.

## CI changes (`python-tests.yaml`)

Remove the `actions/cache` steps entirely. Alpine images download in ~6 s on GitHub runners (measured from CI logs), which is faster than cache restore overhead for files this size.

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -120,19 +120,11 @@ jobs:
         run: |
           brew install renode/tap/renode
 
-      - name: Cache Fedora Cloud images
-        id: cache-fedora-cloud-images
-        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5
-        with:
-          path: python/packages/jumpstarter-driver-qemu/images
-          key: fedora-cloud-43-1.6
-
-      - name: Download Fedora Cloud images
-        if: steps.cache-fedora-cloud-images.outputs.cache-hit != 'true'
+      - name: Download Alpine cloud images
         run: |
           for arch in aarch64 x86_64; do
-            curl -L --fail --output "python/packages/jumpstarter-driver-qemu/images/Fedora-Cloud-Base-Generic-43-1.6.${arch}.qcow2" \
-              "https://iad.mirror.rackspace.com/fedora/releases/43/Cloud/${arch}/images/Fedora-Cloud-Base-Generic-43-1.6.${arch}.qcow2"
+            curl -L --fail --output "python/packages/jumpstarter-driver-qemu/images/nocloud_alpine-3.22.4-${arch}-uefi-tiny-r0.qcow2" \
+              "https://dl-cdn.alpinelinux.org/alpine/v3.22/releases/cloud/nocloud_alpine-3.22.4-${arch}-uefi-tiny-r0.qcow2"
           done
 
       - name: Run pytest
--- a/python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
+++ b/python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
@@ -4,6 +4,7 @@ import sys
 from contextlib import contextmanager
 
 import click
+from fabric import Connection
 from jumpstarter_driver_composite.client import CompositeClient
 from jumpstarter_driver_network.adapters import FabricAdapter, NovncAdapter
 
@@ -75,12 +76,25 @@ class QemuClient(CompositeClient):
 
     @contextmanager
     def shell(self):
-        with FabricAdapter(
-            client=self.ssh,
-            user=self.username,
-            connect_kwargs={"password": self.password},
-        ) as conn:
-            yield conn
+        # If the driver has an 'ssh' hostfwd entry, fetch the actual host port
+        # (resolving any port=0 assignment) and connect directly over TCP.
+        # Otherwise fall back to tunnelling through the jumpstarter stream (vsock).
+        try:
+            port = int(self.call("get_hostfwd_port", "ssh"))
+            with Connection(
+                host="127.0.0.1",
+                port=port,
+                user=self.username,
+                connect_kwargs={"password": self.password},
+            ) as conn:
+                yield conn
+        except KeyError:
+            with FabricAdapter(
+                client=self.ssh,
+                user=self.username,
+                connect_kwargs={"password": self.password},
+            ) as conn:
+                yield conn
 
     def cli(self):
         # Get the base group from CompositeClient which includes all child commands
--- a/python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
+++ b/python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
@@ -5,6 +5,7 @@ import json
 import logging
 import os
 import platform
+import shlex
 import shutil
 from collections.abc import AsyncGenerator
 from dataclasses import dataclass, field
@@ -381,6 +382,26 @@ class QemuPower(PowerInterface, Driver):
         Path(self.parent._pty).unlink(missing_ok=True)
         Path(self.parent._pty).symlink_to(pty)
 
+        # Resolve any hostport=0 hostfwd entries to the actual port QEMU chose.
+        # Parse 'info usernet': lines look like "TCP[HOST_FORWARD] fd addr port addr port ..."
+        # Store resolved ports on the parent so get_hostfwd_port() can return them to clients.
+        zero_fwds = {k: v for k, v in self.parent.hostfwd.items() if v.hostport == 0}
+        if zero_fwds:
+            usernet = await qmp.execute("human-monitor-command", {"command-line": "info usernet"})
+            self.logger.debug("info usernet output:\n%s", usernet)
+            for line in usernet.splitlines():
+                parts = line.split()
+                if len(parts) >= 6 and "HOST_FORWARD" in parts[0]:
+                    # parts: Protocol[State] fd hostaddr hostport guestaddr guestport ...
+                    actual_hostaddr, actual_hostport, actual_guestport = parts[2], int(parts[3]), int(parts[5])
+                    for k, v in zero_fwds.items():
+                        if v.hostaddr == actual_hostaddr and v.guestport == actual_guestport:
+                            self.logger.info(
+                                "hostfwd '%s': resolved port 0 -> %s:%d (guest port %d)",
+                                k, actual_hostaddr, actual_hostport, actual_guestport,
+                            )
+                            self.parent._resolved_hostports[k] = actual_hostport
+
         await qmp.execute("system_reset")
         await qmp.disconnect()
 
@@ -410,7 +431,7 @@ class QemuPower(PowerInterface, Driver):
 class Hostfwd(BaseModel):
     protocol: Literal["tcp"] = "tcp"
     hostaddr: str = "127.0.0.1"
-    hostport: int = Field(ge=1, le=65535)
+    hostport: int = Field(ge=0, le=65535)  # 0 = let QEMU pick a free port
     guestport: int = Field(ge=1, le=65535)
 
 
@@ -440,6 +461,8 @@ class Qemu(Driver):
     flash_timeout: int = field(default=30 * 60)  # 30 minutes
 
     _tmp_dir: TemporaryDirectory = field(init=False, default_factory=TemporaryDirectory)
+    # Maps hostfwd key -> actual host port after QEMU resolves port 0 assignments
+    _resolved_hostports: dict[str, int] = field(init=False, default_factory=dict)
 
     @classmethod
     def client(cls) -> str:
@@ -512,6 +535,7 @@ class Qemu(Driver):
                 {
                     "instance-id": str(self.uuid),
                     "local-hostname": self.hostname,
+                    "hostname": self.hostname,
                 }
             )
         )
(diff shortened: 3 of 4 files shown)


python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -528,6 +529,13 @@ def cidata(self) -> TemporaryDirectory:
                             "sudo": "ALL=(ALL) NOPASSWD:ALL",
                         }
                     ],
+                    # runcmd sets the password explicitly for cloud-init implementations
+                    # that do not support plain_text_passwd (e.g. Alpine's tiny-cloud).
+                    # cloud-init ignores runcmd entries it doesn't understand, so this
+                    # is safe to include unconditionally.
+                    "runcmd": [
+                        f"echo '{self.username}:{self.password}' | chpasswd",
+                    ],
```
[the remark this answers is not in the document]
mangelajo: done

raballew (changes requested):
On a high level, I don't like hardcoding this in so many different places plus we are completely missing out the renovate process here. I think we will need to ensure that this image gets constantly updated as well.

mangelajo:
> On a high level, I don't like hardcoding this in so many different places plus we are completely missing out the renovate process here. I think we will need to ensure that this image gets constantly updated as well.

no real need for this, this does not go into production or anything, is just an image that we boot to run the tests. But that's an existing problem from the past, nothing new.

.github/workflows/python-tests.yaml
```diff
@@ -120,19 +120,11 @@ jobs:
         run: |
           brew install renode/tap/renode
 
-      - name: Cache Fedora Cloud images
```
mangelajo, line 123: the cache was actually slower than the download.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
```diff
@@ -75,12 +76,25 @@ def novnc(self):
...
-        with FabricAdapter(
-            client=self.ssh,
-            user=self.username,
-            connect_kwargs={"password": self.password},
-        ) as conn:
-            yield conn
+        # If the driver has an 'ssh' hostfwd entry, fetch the actual host port
+        # (resolving any port=0 assignment) and connect directly over TCP.
+        # Otherwise fall back to tunnelling through the jumpstarter stream (vsock).
+        try:
```
mangelajo, line 82: we had to implement this because alpine does not setup ssh over vhost, and we didn't want to pick a port that could potentially conflict with whatever is on the testing host.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
```diff
@@ -75,12 +76,25 @@ def novnc(self):
...
+            port = int(self.call("get_hostfwd_port", "ssh"))
+            with Connection(
+                host="127.0.0.1",
+                port=port,
+                user=self.username,
+                connect_kwargs={"password": self.password},
+            ) as conn:
+                yield conn
+        except KeyError:
```
raballew, line 91: Okay this one is a little backwards but hear me out: I think `shell()` method catches `KeyError` to fall back from tcp to vsock, but this never actually works. On the driver side, `get_hostfwd_port()` raises `KeyError`, but the gRPC transport catches all exceptions and reraises them as `DriverError` via `StatusCode.UNKNOWN` but `DriverError` doesn't inherit from `KeyError` so VMs without SSH hostfwd will crash with an unhandled `DriverError` instead of falling back to vsock.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -381,6 +382,26 @@ async def on(self) -> None:  # noqa: C901
...
+                    for k, v in zero_fwds.items():
+                        if v.hostaddr == actual_hostaddr and v.guestport == actual_guestport:
+                            self.logger.info(
+                                "hostfwd '%s': resolved port 0 -> %s:%d (guest port %d)",
+                                k, actual_hostaddr, actual_hostport, actual_guestport,
+                            )
+                            self.parent._resolved_hostports[k] = actual_hostport
+
         await qmp.execute("system_reset")
         await qmp.disconnect()
```
raballew, line 407: A `ValueError` from malformed output or a QMP error leaves the connection open and the QEMU process orphaned. You could wrap the QMP operation block in try/finally to guarantee `await qmp.disconnect()` runs.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -410,7 +431,7 @@ def close(self):
 class Hostfwd(BaseModel):
     protocol: Literal["tcp"] = "tcp"
     hostaddr: str = "127.0.0.1"
-    hostport: int = Field(ge=1, le=65535)
+    hostport: int = Field(ge=0, le=65535)  # 0 = let QEMU pick a free port
     guestport: int = Field(ge=1, le=65535)
```
raballew, line 437: If `on()` fails after process creation but before usernet resolution completes, `get_hostfwd_port()` returns a stale port. I think adding `self.parent._resolved_hostports.clear()` at the beginning of `on()` before starting the new QEMU process, and also in `off()` alongside the existing cleanup should resolve this.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -410,7 +431,7 @@ def close(self):
 class Hostfwd(BaseModel):
     protocol: Literal["tcp"] = "tcp"
     hostaddr: str = "127.0.0.1"
-    hostport: int = Field(ge=1, le=65535)
+    hostport: int = Field(ge=0, le=65535)  # 0 = let QEMU pick a free port
```
raballew, line 434: [HIGH] `hostaddr: str = "127.0.0.1"` has no format validation. The value is interpolated unsanitized into a comma-separated QEMU `-netdev` argument. Since QEMU uses commas to delimit netdev options, a crafted `hostaddr` like `127.0.0.1,hostfwd=tcp:0.0.0.0:4444-:4444` would inject an additional hostfwd rule. The `hostport` and `guestport` fields have `Field(ge=..., le=...)` constraints, but `hostaddr` accepts arbitrary strings.

Suggested fix: add a Pydantic `field_validator` using `ipaddress.ip_address()` to ensure `hostaddr` is a valid IP address.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -528,12 +552,30 @@ def cidata(self) -> TemporaryDirectory:
...
 
+    @export
+    @validate_call(validate_return=True)
+    def get_hostfwd_port(self, key: str) -> int:
+        """Return the actual host port for a hostfwd entry (resolves port 0 assignments)."""
+        if key in self._resolved_hostports:
+            return self._resolved_hostports[key]
+        if key in self.hostfwd:
+            return self.hostfwd[key].hostport
+        raise KeyError(f"hostfwd key {key!r} not found")
```
raballew, line 577: [HIGH] `get_hostfwd_port()` is a new exported method with three code paths (resolved port, configured port, KeyError), and the `info usernet` parsing logic uses positional string splitting with `int(parts[3])` and `int(parts[5])`. Neither has any unit test. The only coverage is the full integration test that boots a real VM. Extracting the parsing logic into a standalone function (e.g., `_parse_usernet_hostfwd(usernet_output, zero_fwds)`) would make it straightforward to add parameterized unit tests covering all three `get_hostfwd_port()` code paths as well as edge cases in the usernet output format.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -440,6 +461,8 @@ class Qemu(Driver):
     flash_timeout: int = field(default=30 * 60)  # 30 minutes
 
     _tmp_dir: TemporaryDirectory = field(init=False, default_factory=TemporaryDirectory)
+    # Maps hostfwd key -> actual host port after QEMU resolves port 0 assignments
+    _resolved_hostports: dict[str, int] = field(init=False, default_factory=dict)
 
     @classmethod
     def client(cls) -> str:
```
raballew, line 468: Defer TcpNetwork child creation for hostport=0 entries until after port resolution in `on()`, or update the child's port after resolution. Use a different key namespace for hostfwd children to avoid overwriting vsock children.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -381,6 +382,26 @@ async def on(self) -> None:  # noqa: C901
...
+                if len(parts) >= 6 and "HOST_FORWARD" in parts[0]:
+                    # parts: Protocol[State] fd hostaddr hostport guestaddr guestport ...
+                    actual_hostaddr, actual_hostport, actual_guestport = parts[2], int(parts[3]), int(parts[5])
+                    for k, v in zero_fwds.items():
+                        if v.hostaddr == actual_hostaddr and v.guestport == actual_guestport:
+                            self.logger.info(
+                                "hostfwd '%s': resolved port 0 -> %s:%d (guest port %d)",
+                                k, actual_hostaddr, actual_hostport, actual_guestport,
+                            )
+                            self.parent._resolved_hostports[k] = actual_hostport
```
raballew, line 403: The `info usernet` parser uses `int(parts[3])` and `int(parts[5])` without try/except. A malformed line would raise `ValueError

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
```diff
@@ -4,6 +4,7 @@
 from contextlib import contextmanager
 
 import click
+from fabric import Connection
```
raballew, line 7: `from fabric import Connection` is a new import added by this PR, but `fabric` is not listed in dependencies.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/client.py
```diff
@@ -75,12 +76,25 @@ def novnc(self):
...
+                connect_kwargs={"password": self.password},
+            ) as conn:
+                yield conn
+        except KeyError:
+            with FabricAdapter(
+                client=self.ssh,
+                user=self.username,
+                connect_kwargs={"password": self.password},
+            ) as conn:
+                yield conn
```
raballew, line 97: Narrow the try/except to only wrap the `self.call("get_hostfwd_port", "ssh")` call, then branch on the result.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver_test.py
```diff
@@ -88,16 +93,22 @@ def test_driver_qemu(tmp_path, ovmf):
...
             p.expect_exact("Password:")
             p.sendline(password)
-            p.expect_exact(f"[{username}@{hostname} ~]$")
-            p.sendline("sudo setenforce 0")
-            p.expect_exact(f"[{username}@{hostname} ~]$")
+            p.expect_exact(f"{hostname}:~$")
 
         with qemu.shell() as s:
-            assert s.run("uname -r").stdout.strip() == f"6.17.1-300.fc43.{arch}"
+            assert s.run("uname -r").stdout.strip() != ""
```
raballew, line 111: Also assert the return code, e.g., `result = s.run("uname -r"); assert result.ok; assert result.stdout.strip() != ""`.

mangelajo:
I need to review the qemu/ssh changes more carefully

raballew (changes requested):

mangelajo:
will close for now

PR #814, opened 2026-06-19
ci(lint): add ASCII character check using ripgrep

raballew:
## Summary
- Adds an `ascii-check` job to the lint workflow that uses `rg '[^\p{ASCII}]'` to flag non-ASCII characters (em-dashes, en-dashes, curly quotes, emoji, etc.) that render poorly or break in terminals
- Excludes `.cast` and `.html` files which legitimately contain non-ASCII content
- Currently there are ~1,104 lines with non-ASCII characters across tracked files that need to be fixed before this check can be enforced

## Test plan
- [ ] Verify the `ascii-check` job runs in CI (it will fail on existing violations -- expected)
- [ ] Incrementally fix non-ASCII findings across the codebase
- [ ] Once clean, the job gates future PRs from introducing new non-ASCII characters


mangelajo:
Why do we hate UTF8? :D I loved when we got free of the ASCII limitations

raballew:
until your yaml depends on it

kirkbrauer:
@raballew Humm, I get the idea behind this, but I think this is a lot of changes to merge just for a character transition. One of the nice things about languages such as Python, Rust, and Go is the ability to incorporate UTF-8 characters into comments and logs, for example, adding emojis to rich log outputs. So therefore, I think we can apply this, but limit the file types if applies to or the domains we enforce this in.

PR #815, opened 2026-06-19
CI: per-package test logs, parallel runs, and uboot OOM fix

mangelajo:
## Summary

- **Per-package test log capture**: When `LOGS_DIR` is set, each `pkg-test-%` target captures stdout/stderr to a separate `.log` file via `tee` (visible during run). Failures are tracked with `.failed` markers based on exit code. A `test-report` target prints full logs for failed packages. Local dev behavior is unchanged.
- **`fail-fast: false`**: All matrix jobs (runner × Python version) run to completion even when one fails.
- **Parallel test runs**: `make test -j4` runs package tests concurrently.
- **Log noise suppression**: `--log-level=CRITICAL --log-cli-level=CRITICAL` in CI-only `PYTEST_ADDOPTS`.
- **Artifact upload**: Per-package logs uploaded as GitHub Actions artifacts (7-day retention).
- **uboot test OOM fix**: `rpmfile` decompresses the entire 31MB zstd RPM into 572MB of memory to extract a 1.6MB `u-boot.bin`. Under parallel runs on resource-constrained VMs this triggers the OOM killer. Now uses streaming `rpm2cpio | cpio` on Linux, falling back to `rpmfile` when unavailable.

## Test plan

- [ ] All matrix jobs complete (none cancelled by fail-fast)
- [ ] Per-package `.log` files appear in uploaded artifacts
- [ ] Failed packages show full output in the `test-report` section
- [ ] uboot test passes without hanging on extraction
- [ ] Local `make test` still works without `LOGS_DIR`

### DIFF
--- a/.github/workflows/python-tests.yaml
+++ b/.github/workflows/python-tests.yaml
@@ -77,7 +77,7 @@ jobs:
           sudo chmod 0666 /dev/kvm /dev/vhost-vsock /dev/vhost-net
 
           sudo apt-get update
-          sudo apt-get install -y qemu-system-arm qemu-system-x86
+          sudo apt-get install -y qemu-system-arm qemu-system-x86 rpm2cpio cpio
 
       - name: Install libgpiod-dev (Linux)
         if: runner.os == 'Linux'
@@ -138,9 +138,18 @@ jobs:
       - name: Run pytest
         working-directory: python
         env:
-          PYTEST_ADDOPTS: "--cov-report=xml"
+          PYTEST_ADDOPTS: "--cov-report=xml --log-level=CRITICAL --log-cli-level=CRITICAL"
         run: |
-          make test
+          make test -j4 LOGS_DIR=${{ runner.temp }}/test-logs
+
+      - name: Upload test logs
+        if: always()
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: test-logs-${{ matrix.runs-on }}-py${{ matrix.python-version }}
+          path: ${{ runner.temp }}/test-logs/
+          if-no-files-found: ignore
+          retention-days: 7
 
       # Diff-coverage is only checked on Linux. Several packages (e.g.
       # jumpstarter-driver-dut-network) are Linux-only, so their tests are
@@ -156,7 +165,7 @@ jobs:
             echo "::error::No coverage.xml files found"
             exit 1
           fi
-          uv run diff-cover $coverage_files --compare-branch=origin/${{ github.base_ref }} --fail-under=80 --exclude '*_pb2.py' '*_pb2_grpc.py' '**/conftest.py'
+          uv run diff-cover $coverage_files --compare-branch=origin/${{ github.base_ref }} --fail-under=80 --exclude '*_pb2.py' '*_pb2_grpc.py' '**/conftest.py' '**/test_*.py' '**/*_test.py'
 
   # https://github.com/orgs/community/discussions/26822
   pytest:
--- a/python/packages/jumpstarter-driver-uboot/jumpstarter_driver_uboot/driver_test.py
+++ b/python/packages/jumpstarter-driver-uboot/jumpstarter_driver_uboot/driver_test.py
@@ -1,5 +1,7 @@
 import os
 import platform
+import shutil
+import subprocess
 
 import pytest
 import requests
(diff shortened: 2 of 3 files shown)


.github/workflows/python-tests.yaml
```diff
@@ -138,9 +139,18 @@ jobs:
       - name: Run pytest
         working-directory: python
         env:
-          PYTEST_ADDOPTS: "--cov-report=xml"
+          PYTEST_ADDOPTS: "--cov-report=xml --log-level=CRITICAL --log-cli-level=CRITICAL"
```
mangelajo, line 141: This reduces a lot of unnecessary noise on the tests without completely disabling logs.

python/packages/jumpstarter-driver-uboot/jumpstarter_driver_uboot/driver_test.py
```diff
@@ -11,24 +13,45 @@
...
-
-    url = "https://kojipkgs.fedoraproject.org/packages/uboot-tools/2025.10/1.fc43/noarch/uboot-images-armv8-2025.10-1.fc43.noarch.rpm"
-
-    with requests.get(url, stream=True) as r:
-        r.raise_for_status()
-        with (tmp_path / "uboot-images-armv8.rpm").open("wb") as f:
-            for chunk in r.iter_content(chunk_size=8192):
-                f.write(chunk)
-
-    with rpmfile.open(tmp_path / "uboot-images-armv8.rpm") as rpm:
```
mangelajo, line 26: this was extracting the RPM to RAM, then extracting the file, not very optimal, and in parallelization used a lot of ram.

.github/workflows/python-tests.yaml
```diff
@@ -42,6 +42,7 @@ jobs:
     if: needs.changes.outputs.should_run == 'true' || github.event_name == 'workflow_dispatch'
     runs-on: ${{ matrix.runs-on }}
     strategy:
+      fail-fast: false
```
raballew: isnt it likely that an error in one matrix job will also surface on others hence failing fast kinda makes sense.

python/packages/jumpstarter-driver-uboot/jumpstarter_driver_uboot/driver_test.py
```diff
@@ -11,24 +13,45 @@
 from jumpstarter.common.utils import serve
 
 
+UBOOT_RPM_URL = "https://kojipkgs.fedoraproject.org/packages/uboot-tools/2025.10/1.fc43/noarch/uboot-images-armv8-2025.10-1.fc43.noarch.rpm"
```
mangelajo, line 15: Again, this is not an image we include in the containers, or the product, It's just used to test a real uboot on QEMU and properly validate the uboot driver.

.github/workflows/python-tests.yaml
mangelajo: yes and no, I wanted to try without it, but may be it's better to leave it on. For example I was having a lot of mac failures and that stopped the linux ones where I could not see if it was going to fail or not as well... or the other way around.

mangelajo: Putting it back, we should probably discuss that separately, or disable on a need basis.

python/pyproject.toml
```diff
@@ -106,7 +106,7 @@ mosquitto = "mosquitto"
 
 [tool.coverage.run]
 branch = true
-omit = ["conftest.py", "test_*.py", "*_test.py", "*_pb2.py", "*_pb2_grpc.py"]
+omit = ["**/conftest.py", "**/test_*.py", "**/*_test.py", "**/*_pb2.py", "**/*_pb2_grpc.py"]
```
mangelajo: Apparently coverage was complaining for my _test.py file increased python surface. Which unveiled that we probably have better coverage than we thought, and that this may need patters to pickup the subpackages test files. I am not sure if each subpackage may need replication of this.

.github/workflows/python-tests.yaml
```diff
@@ -156,7 +165,7 @@ jobs:
             echo "::error::No coverage.xml files found"
             exit 1
           fi
-          uv run diff-cover $coverage_files --compare-branch=origin/${{ github.base_ref }} --fail-under=80 --exclude '*_pb2.py' '*_pb2_grpc.py' '**/conftest.py'
+          uv run diff-cover $coverage_files --compare-branch=origin/${{ github.base_ref }} --fail-under=80 --exclude '*_pb2.py' '*_pb2_grpc.py' '**/conftest.py' '**/test_*.py' '**/*_test.py'
```
mangelajo, line 168: We were considering _test and test_ files as part of our coverage, who tests the test problem?

raballew (approved):

Issue #733, opened 2026-06-03
Flaky test_exec_bash in hooks_test.py on macOS due to PTY output race condition


mangelajo:
## Summary

`TestHookExecutor::test_exec_bash` in `jumpstarter/exporter/hooks_test.py` is flaky on macOS. It fails intermittently on the `pytest-matrix (macos-15, 3.11)` CI job because the PTY output from a fast-completing bash subprocess is lost before the reader can capture it.

**Failing CI run:** https://github.com/jumpstarter-dev/jumpstarter/actions/runs/26871055542/job/79246255810

**Commit:** `e654084d39501b057489923687636b264ec2cbe2` (main branch)

**Result:** 1 failed, 515 passed — only this test fails.

## Failure Details

```
FAILED jumpstarter/exporter/hooks_test.py::TestHookExecutor::test_exec_bash - assert False
```

The test runs a bash script `V="hello_world"; echo "BASH_OK: ${V:6:5}"` via the hook executor with `exec_="/bin/bash"`, then asserts that `"BASH_OK: world"` appears in the mocked logger's `info` calls. The assertion fails because the output was never captured.

## Root Cause Analysis

The hook executor (`hooks.py`) uses PTY-based subprocess execution with async non-blocking I/O. The execution flow is:

1. Subprocess spawns, writes `"BASH_OK: world\n"` to stdout via PTY, and exits
2. `read_pty_output` async task reads from PTY parent_fd using non-blocking I/O in a loop (0.1s `wait_readable` timeout per iteration)
3. `wait_for_process()` returns (process exited with code 0)
4. `await anyio.sleep(0.2)` — brief pause to let reader catch up
5. `pty_state.reader_stop = True` — signals reader to stop
6. Reader exits loop, enters `finally` drain block (non-blocking `os.read`)

**The race:** On macOS, when bash executes a very fast script, the process can write output and exit before the PTY kernel buffer delivers all data to the parent_fd. The 0.2s sleep after `wait_for_process()` is not always sufficient. When the reader enters its drain phase, it does a non-blocking `os.read` — if data hasn't propagated through the macOS PTY layer yet, `BlockingIOError` is raised and the drain breaks immediately, losing the output.

This is a known characteristic of macOS PTY behavior where output buffering is less synchronous with process exit compared to Linux.

## Environment

- **Platform:** macOS 15 (Apple Silicon, GitHub Actions runner)
- **Python:** 3.11.9
[...]

Issue #560, opened 2026-04-15
test_exec_bash hook test fails on macOS due to PTY output not captured by mocked logger


raballew:
## Description

The `test_exec_bash` test in `jumpstarter-common` fails consistently on macOS CI runners (macos-15, Python 3.11) while passing on Linux. The root cause is that PTY output written by hook subprocesses is not fully drained before the test assertions run, causing the mocked logger to miss expected log lines.

## Root Cause

In `python/packages/jumpstarter/jumpstarter/exporter/hooks.py`, the `HookExecutor._run_hook` method uses a PTY for subprocess output capture. On macOS, the PTY I/O timing differs from Linux -- the async reader task may not have consumed all output by the time it is cancelled after the subprocess exits. There is no drain step to read remaining buffered data from the PTY file descriptor after the reader task stops.

## Affected CI

- **Workflow**: Python CI (`python.yml`)
- **Runner**: `macos-15`
- **Python version**: 3.11
- **Test**: `test_exec_bash` in `python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py`
- **Failure pattern**: Persistent (fails on every macOS run)

## Reproduction

Run the test suite for `jumpstarter-common` on macOS:

```bash
cd python && make pkg-test-jumpstarter_common
```

The test passes on Linux but fails on macOS because the mocked logger does not receive the expected output lines.

## Expected Behavior

`test_exec_bash` should pass on both Linux and macOS, with all hook subprocess output captured by the logger regardless of platform PTY timing differences.

## Proposed Fix

1. After the async reader task is cancelled, drain any remaining data from the PTY file descriptor synchronously before closing it
2. Bound the drain loop to prevent indefinite blocking (max iterations + timeout)
3. Properly handle `OSError` during drain (PTY may already be closed)

mangelajo:
FYI — this issue is still reproducing despite the fixes in #560, #733, and #826. See #821 (re-opened) for the latest details and the list of tests now marked as `xfail` on macOS. Targeted for investigation in 0.10.0.

PR #826, opened 2026-06-23
fix: retry PTY drain on empty select() to prevent data loss on macOS

### LINKED ISSUE #821: Flaky test_exec_bash in hooks_test.py on macOS still reproducing after previous fixes
## Summary

`TestHookExecutor::test_exec_bash` in `jumpstarter/exporter/hooks_test.py` is still flaky on macOS despite previous fixes in #560, #733, and #826. The PTY output race condition on macOS has not been fully resolved.

**Latest failing CI run:** https://github.com/jumpstarter-dev/jumpstarter/actions/runs/28084374100/job/83146665529

**Workflow:** Python Tests — `pytest-matrix (macos-15)`

This is the 4th occurrence of this issue (#560, #733, #821, #826). Each fix has reduced the frequency but not eliminated the race condition.

## Root Cause

The bug is in the PTY drain logic in `hooks.py` (the `finally` block of `read_pty_output`). On macOS, the PTY kernel buffer may not deliver data synchronously with process exit. Despite increasing `DRAIN_MAX_EMPTY_POLLS` to 10 in #826, the race still occurs under heavy CI load.

The CI runs `make test -j4` (4 parallel test suites) with concurrent `uv` package installations, creating enough I/O and CPU contention for the macOS PTY buffer to miss the drain window.

## Previous Fix Attempts

- #560 — Added synchronous drain after async reader cancellation (closed 2026-04-16)
- #733 — Added `select()`-based drain with timeout in the finally block (closed 2026-06-05)
- #826 — Increased DRAIN_MAX_EMPTY_POLLS from 1 to 10 (~1s window) (merged 2026-06-23)

All fixes reduced the frequency but did not eliminate the race.

## Tests Marked as `xfail` on macOS

The following tests are marked with `@macos_pty_xfail` (`strict=False`) to unblock CI while the root cause is investigated. All of these spawn real subprocesses via PTY and assert on captured logger output:

### `TestHookExecutor` (output capture tests)
- `test_hook_environment_variables`
- `test_real_time_output_logging`
- `test_post_lease_hook_execution_on_completion`
- `test_exec_bash`
- `test_exec_python3`
- `test_script_file_sh`
- `test_script_file_py_autodetects_python`
- `test_script_file_py_exec_override`
- `test_noninteractive_environment`
- `test_drain_captures_output_without_trailing_newline`

### `TestHookExecutor` (drain behavior tests with patched PTY)
- `test_drain_reads_data_remaining_in_pty_buffer`
- `test_drain_select_oserror_exits_gracefully`
- `test_drain_select_valueerror_exits_gracefully`
- `test_drain_exits_when_deadline_exceeded_before_select`
- `test_drain_exception_is_suppressed`
- `test_drain_retries_empty_select_then_captures_data`
- `test_drain_terminates_after_max_empty_polls`
- `test_drain_empty_counter_resets_on_data`

### `TestHookExecutorPRRegressions`
- `test_infrastructure_messages_at_debug_not_info`

## Environment

- **Platform:** macOS 15 (Apple Silicon, GitHub Actions runner)
- **Python:** 3.11, 3.13
- **Passes on:** Linux runners (same Python versions)

## Goal for 0.10.0

Investigate the root cause properly and either:
1. Find a reliable fix for macOS PTY timing (possibly using `waitpid` + blocking drain, or pipe-based output capture instead of PTY)
2. Or redesign hook output capture to not depend on PTY timing

Once fixed, remove the `macos_pty_xfail` markers from all affected tests.

mangelajo:
## Problem

`test_exec_bash` is flaky on macOS CI runners (macos-15, Python 3.11). The PTY drain logic gives up after a single empty `select()` poll (100ms), but on loaded CI runners, macOS PTY kernel buffer delivery can be delayed beyond 100ms after subprocess exit.

**Failing CI run:** https://github.com/jumpstarter-dev/jumpstarter/actions/runs/27971589029/job/82778910902

This is the third occurrence of this issue (#560, #733, #821). The `select()`-based drain from #733 was the right approach but insufficient — it breaks after a single empty poll.

## Root Cause

In the `read_pty_output` finally block, the drain loop does:

```python
if not readable:
    # Timed out with no data — drain is complete
    break  # <-- gives up after ONE empty 100ms poll
```

CI runs `make test -j4` (4 parallel test suites) with concurrent `uv` package installations, creating enough I/O and CPU contention for the macOS PTY buffer to miss this single 100ms window.

## Fix

Add a `DRAIN_MAX_EMPTY_POLLS = 10` constant and track consecutive empty `select()` polls, only giving up after 10 in a row (~1s worst-case). This is well within the existing `DRAIN_TIMEOUT_SECONDS` (2s) deadline and gives ample headroom for heavily-loaded CI runners.

## Reproduction

Isolated the drain logic and tested with delayed PTY writes:

```
Delayed PTY write: original drain vs fixed drain
   Delay      Original         Fixed
--------------------------------------
     0ms          PASS          PASS
    50ms          PASS          PASS
    80ms          PASS          PASS
   105ms          0/20          PASS
   120ms          0/20          PASS
   150ms          0/20          PASS
   200ms          0/20          PASS
   250ms          0/20          PASS
```

When data arrives after 105ms, the original drain has a **100% failure rate**. The fix passes all cases.

## Testing

- All 52 hooks tests pass (Python 3.11, macOS)
- 100/100 pass rate for `test_exec_bash` in isolation (Python 3.11)
- 41/41 pass rate under CPU + I/O stress (Python 3.11)

Fixes #821

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
@@ -902,6 +902,128 @@ class TestHookExecutor:
             result = await executor.execute_before_lease_hook(lease_scope)
             assert result is None
 
+    async def test_drain_retries_empty_select_then_captures_data(self, lease_scope) -> None:
+        """Verify that the drain retries after empty select() calls and still
+        captures data that arrives later.
+
+        Patches select.select to return empty for the first N calls (where
+        N < DRAIN_MAX_EMPTY_POLLS), then reports the fd as readable.  The
+        hook output should still be captured despite the initial empty polls.
+        """
+        import select as select_mod
+
+        original_select = select_mod.select
+        state = _PtyTracker()
+        empty_count = 0
+        empties_before_data = DRAIN_MAX_EMPTY_POLLS - 2  # e.g. 8 empties then data
+
+        def select_with_delayed_ready(rlist, wlist, xlist, timeout=None):
+            nonlocal empty_count
+            if state.eof_seen and rlist and rlist[0] == state.parent_fd:
+                empty_count += 1
+                if empty_count <= empties_before_data:
+                    return ([], [], [])  # simulate delayed data
+            return original_select(rlist, wlist, xlist, timeout)
+
+        hook_config = HookConfigV1Alpha1(
+            before_lease=HookInstanceConfigV1Alpha1(
+                script="echo DELAYED_DRAIN_OK", timeout=10,
+            ),
+        )
+        executor = HookExecutor(config=hook_config)
+
+        with (
+            patch("pty.openpty", side_effect=state.tracking_openpty),
+            patch("os.read", side_effect=state.os_read_with_drain_data),
+            patch("jumpstarter.exporter.hooks.select.select", side_effect=select_with_delayed_ready),
+            patch("jumpstarter.exporter.hooks.logger") as mock_logger,
+        ):
+            result = await executor.execute_before_lease_hook(lease_scope)
+            assert result is None
+            info_calls = [str(call) for call in mock_logger.info.call_args_list]
+            assert any("DELAYED_DRAIN_OK" in call for call in info_calls)
+
+    async def test_drain_terminates_after_max_empty_polls(self, lease_scope) -> None:
+        """Verify the drain loop terminates after DRAIN_MAX_EMPTY_POLLS
+        consecutive empty select() results.
+
+        Patches select.select to always return empty during the drain phase.
+        The hook should still complete (no hang) and the drain data should
+        not appear since it's never read.
+        """
+        import select as select_mod
+
+        original_select = select_mod.select
+        state = _PtyTracker(return_drain_data=False)
+
+        def select_always_empty(rlist, wlist, xlist, timeout=None):
+            if state.eof_seen and rlist and rlist[0] == state.parent_fd:
+                return ([], [], [])  # always empty
+            return original_select(rlist, wlist, xlist, timeout)
+
+        hook_config = HookConfigV1Alpha1(
+            before_lease=HookInstanceConfigV1Alpha1(
+                script="echo MAX_EMPTY_TEST", timeout=10,
+            ),
+        )
+        executor = HookExecutor(config=hook_config)
+
+        with (
+            patch("pty.openpty", side_effect=state.tracking_openpty),
+            patch("os.read", side_effect=state.os_read_with_drain_data),
+            patch("jumpstarter.exporter.hooks.select.select", side_effect=select_always_empty),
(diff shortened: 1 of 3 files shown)


python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
```diff
@@ -904,6 +905,7 @@ def flush_lines_with_drain_error(buffer, output_lines):
     async def test_drain_constants_are_reasonable(self) -> None:
         assert MAX_DRAIN_BYTES == 256 * 1024
         assert DRAIN_TIMEOUT_SECONDS == 2.0
```
raballew, line 1029: It might be worth adding tests that patch select.select to simulate: a) N empty returns followed by data (verify data is captured), b) DRAIN_MAX_EMPTY_POLL consecutive empties (verify loop terminates) and c) empty-data-empty pattern to verify the counter resets. 

Apart from this is looks good.

mangelajo: We don't need to test the test code :D where would the test chain end? :D

As long as the test is executing it should be good IMHO. or do you mean something else?

mangelajo: ahh no, !! sorry ignore me .. :D it's not all on the _test.py, I see what you mean.

mangelajo: Added all three tests in 29dc538a:

- `test_drain_retries_empty_select_then_captures_data` — patches select to return empty for N calls (N < DRAIN_MAX_EMPTY_POLLS), then reports readable. Verifies data is still captured.
- `test_drain_terminates_after_max_empty_polls` — patches select to always return empty during drain. Verifies the loop terminates (no hang).
- `test_drain_empty_counter_resets_on_data` — simulates empty-data-empty pattern (5 empties, data, 5 empties, data). Total empties (10) >= DRAIN_MAX_EMPTY_POLLS but never consecutive, so drain keeps going. Verifies counter reset works.

kirkbrauer (approved):

kirkbrauer:
@mangelajo It seems like the beforeLease hook seems to be failing in the Ubuntu 24 tests here now...

mangelajo:
> @mangelajo It seems like the beforeLease hook seems to be failing in the Ubuntu 24 tests here now...

it's the BLE test that fails :D (we are fixing it on the other PR...)

bennyz (approved):

kirkbrauer:
@mangelajo Humm, now macOS pytests are failing due to failed assertions of output.

Ahh, the BLE makes sense :)

kirkbrauer (approved):

PR #823, opened 2026-06-23
fix(hooks): release lease when beforeLease hook fails with onFailure=endLease

mangelajo:
## Problem

When a `beforeLease` hook fails with `on_failure=endLease`, the exporter reports `BEFORE_LEASE_HOOK_FAILED` status but **never actually releases the lease**. This leaves the exporter stuck in `BeforeLeaseHookFailed` state and the lease active indefinitely. Clients that attempt to connect get `Connection to exporter lost`.

Similarly, when a `beforeLease` hook fails with `on_failure=exit`, the exporter called `shutdown(wait_for_lease_exit=True)` which deferred exit until the lease ends — but nobody released the lease, so the exporter hung forever in a zombie state.

Fixes #639
Fixes #640

## Root Cause

In `hooks.py`, `run_before_lease_hook()`'s error handlers for both `endLease` and `exit` had issues:

- **`endLease`**: Only reported status but never called `request_lease_release()` — the callback that frees the lease on the controller and unblocks the connection loop.
- **`exit`**: Called `shutdown(wait_for_lease_exit=True)` which deferred exit until the lease ends, but the lease was never released, creating a deadlock.

### `endLease` branch
- Set `skip_after_lease_hook = True` to prevent the after-lease hook from running
- Release the lease via `request_lease_release()` in a shielded `finally` block

### `exit` branch  
- Changed to `shutdown(wait_for_lease_exit=False)` for immediate exit
- The lease is **not** released — it stays active so the client can observe the failure status and expires naturally on the controller
- Set `skip_after_lease_hook = True`

### Structural improvements
- Extracted `_safe_release_lease()` helper for safe error-handling around lease release
- Moved lease release into the `finally` block with a `should_release` flag, wrapped in `CancelScope(shield=True)` to protect from task group cancellation — matching the pattern used by `run_after_lease_hook`

## Changes

| File | Change |
|---|---|
| `hooks.py` | Fix both `endLease` and `exit` branches; add `_safe_release_lease()` helper; shielded `finally` block |
| `hooks_test.py` | Update/add tests for both failure modes, including error handling paths |
| `hooks_test.go` | Add e2e tests B3 (lease release + recovery) and B4 (afterLease skipped) |
| `exporter-hooks-before-fail-endLease-with-after.yaml` | New e2e exporter config for B4 test |

## Testing

- 53 hooks unit tests pass
- 126 total exporter tests pass
- ruff lint clean
- New e2e tests cover the `endLease` regression
[...]

Issue #640, opened 2026-05-04
hooks enabled: creating a lease on exporter with beforeLease onFailure exit, exporter is set offline but does not exit.


evakhoni:
`Jumpstarter v0.8.2.dev113+gb19d6327f`

with an exporter with beforeLease set to fail, with onFailure exit, creating a lease does not exit the exporter until the lease is manually destroyed or fully expires. the lease is created successfully, exporter is marked as offline, does not exit, and an attempt to connect to such lease causes an Error: `Connection to exporter lost`

```
 uv run jmp create lease --duration 10m --selector hooks=enabled
NAME                                  SELECTOR       EXPIRES AT  REMAINING  CLIENT      EXPORTER
019df2c1-1499-7b9a-9994-0269cc292958  hooks=enabled                         crc-client       

  uv run jmp admin get exporter --namespace jumpstarter-lab  
NAME          STATUS   ENDPOINT                               DEVICES  AGE  
crc-exporter  Unknown  grpc.jumpstarter.apps-crc.testing:443  0        7d16h
hooks-verify  Offline  grpc.jumpstarter.apps-crc.testing:443  3        6d23h

 uv run jmp get leases
NAME                                  SELECTOR       EXPIRES AT           REMAINING  CLIENT      EXPORTER    
019df2c1-1499-7b9a-9994-0269cc292958  hooks=enabled  2026-05-04 14:40:33  9m         crc-client  hooks-verify

 uv run jmp shell --lease 019df2c1-1499-7b9a-9994-0269cc292958
[05/04/26 14:31:48] INFO     [jumpstarter.client.lease] Waiting for ready connection at /run/user/1000/jumpstarter-kjz_j_dd/socket                                
                    WARNING  Connection to exporter lost: exporter is offline                                                                                     
                    WARNING  Connection to exporter lost: exporter is offline                                                                                     
Error: Connection to exporter lost
```
in exporter log:
```
INFO:jumpstarter.exporter.exporter:Starting new lease: 019df2c1-1499-7b9a-9994-0269cc292958
INFO:jumpstarter.exporter.exporter:Currently leased by crc-client under 019df2c1-1499-7b9a-9994-0269cc292958
INFO:jumpstarter.exporter.exporter:Listening for incoming connection requests on lease 019df2c1-1499-7b9a-9994-0269cc292958
INFO:jumpstarter.exporter.exporter:Creating new session for lease
INFO:jumpstarter.exporter.session:Session server started on 2 ports
[...]

Issue #639, opened 2026-05-04
hooks enabled: jmp create lease, on an exporter with beforeLease set to onFailure: endLease, does not end the lease and stuck on BeforeLeaseHookFailed


evakhoni:
creating a lease to an exporter with beforeLease that set to cause an end lease, the lease is not ended and the exporter reports `BeforeLeaseHookFailed`. an attempt to connect the lease is failing with `Connection to exporter lost`.
```
 uv run jmp create lease --duration 10m --selector hooks=enabled
NAME                                  SELECTOR       EXPIRES AT  REMAINING  CLIENT      EXPORTER
019df2ae-342f-7cc5-904d-95a176b008b3  hooks=enabled                         crc-client          

 uv run jmp admin get exporter --namespace jumpstarter-lab      
NAME          STATUS                 ENDPOINT                               DEVICES  AGE  
crc-exporter  Unknown                grpc.jumpstarter.apps-crc.testing:443  0        7d16h
hooks-verify  BeforeLeaseHookFailed  grpc.jumpstarter.apps-crc.testing:443  3        6d23h

 uv run jmp get leases                                        
NAME                                  SELECTOR       EXPIRES AT           REMAINING  CLIENT      EXPORTER    
019df2ae-342f-7cc5-904d-95a176b008b3  hooks=enabled  2026-05-04 14:19:56  9m         crc-client  hooks-verify

 uv run jmp shell --lease 019df2ae-342f-7cc5-904d-95a176b008b3
[05/04/26 14:11:06] INFO     [jumpstarter.client.lease] Waiting for ready connection at /run/user/1000/jumpstarter-qf4x63cl/socket                                
                    WARNING  Connection to exporter lost: exporter beforeLease hook failed                                                                        
                    WARNING  Connection to exporter lost: exporter beforeLease hook failed                                                                        
Error: Connection to exporter lost

 uv run jmp get leases                                        
NAME                                  SELECTOR       EXPIRES AT           REMAINING  CLIENT      EXPORTER    
019df2ae-342f-7cc5-904d-95a176b008b3  hooks=enabled  2026-05-04 14:19:56  7m         crc-client  hooks-verify

 uv run jmp admin get exporter --namespace jumpstarter-lab    
NAME          STATUS                 ENDPOINT                               DEVICES  AGE  
crc-exporter  Unknown                grpc.jumpstarter.apps-crc.testing:443  0        7d16h
[...]

PR #827, opened 2026-06-23
fix(hooks): release lease when beforeLease hook fails with onFailure=exit

mangelajo:
**Superseded by #823**, which now fixes both #639 (endLease) and #640 (exit) in a single PR.

The `exit` fix was incorporated into #823 with the correct semantics: `shutdown(wait_for_lease_exit=False)` for immediate exit without releasing the lease (the lease stays active so the client can observe the failure status)."


mangelajo:
Not a bug, see issue comments

PR #828, opened 2026-06-23
fix: mark exporter offline when unregistering

bennyz:
instead of relying on the heartbeat

### DIFF
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -323,12 +323,14 @@ func (s *ControllerService) Unregister(
 	original := client.MergeFrom(exporter.DeepCopy())
 	exporter.Status.Devices = nil
 
+	applyUnregisterStatus(exporter, req.GetReason())
+
 	if err := s.Client.Status().Patch(ctx, exporter, original); err != nil {
 		logger.Error(err, "unable to update exporter status")
 		return nil, status.Errorf(codes.Internal, "unable to update exporter status: %s", err)
 	}
 
-	logger.Info("exporter unregistered, updated as unavailable")
+	logger.Info("exporter unregistered, marked offline")
 
 	return &pb.UnregisterResponse{}, nil
 }
@@ -456,6 +458,23 @@ func protoStatusToString(status pb.ExporterStatus) string {
 	default:
 		return jumpstarterdevv1alpha1.ExporterStatusUnspecified
 	}
+}
+
+func applyUnregisterStatus(exporter *jumpstarterdevv1alpha1.Exporter, reason string) {
+	exporter.Status.ExporterStatusValue = jumpstarterdevv1alpha1.ExporterStatusOffline
+	statusMessage := reason
+	if statusMessage == "" {
+		statusMessage = "Exporter unregistered"
+	}
+	exporter.Status.StatusMessage = statusMessage
+	syncOnlineConditionWithStatus(exporter)
+	meta.SetStatusCondition(&exporter.Status.Conditions, metav1.Condition{
+		Type:               string(jumpstarterdevv1alpha1.ExporterConditionTypeRegistered),
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: exporter.Generation,
+		Reason:             "Unregister",
+		Message:            "Exporter unregistered from the controller and became offline.",
+	})
 }
 
 // syncOnlineConditionWithStatus updates the Online condition based on ExporterStatusValue.
--- a/controller/internal/service/controller_service_test.go
+++ b/controller/internal/service/controller_service_test.go
@@ -214,6 +214,100 @@ func TestCheckExporterStatusForDriverCalls(t *testing.T) {
 	}
 }
 
+func TestApplyUnregisterStatus(t *testing.T) {
+	t.Run("sets offline status with default message", func(t *testing.T) {
+		exporter := &jumpstarterdevv1alpha1.Exporter{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:       "test-exporter",
+				Namespace:  "default",
+				Generation: 1,
+			},
+			Status: jumpstarterdevv1alpha1.ExporterStatus{
+				ExporterStatusValue: jumpstarterdevv1alpha1.ExporterStatusAvailable,
+			},
+		}
+
+		applyUnregisterStatus(exporter, "")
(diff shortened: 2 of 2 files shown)


bennyz:
@mangelajo @evakhoni let me know if it makes sense

mangelajo:
What is the advantage of this? I am trying to remember why we had Unregistered < Offline.

Unregistered is an indication that the exporter may come back soon, so when you have very few exporters of one type and an exporter is ending a lease, while it is restarting the sub-child, still the server will lease it.

The problem comes when the exporter goes offline longer than expected (image rotation), ... then the client fails to connect.

We want to keep the "Unregistered" as "Online" for a while for this reason.

You don't see it in our lab because we have plenty, but when you have 1-2... the user hits that problem and then the lease is instantly rejected because all exporters are "offline" sometimes.

mangelajo (changes requested):
See the comment on the main thread, that's the reason for the difference.

We should just do client retries, and eventual failure.

bennyz:
> What is the advantage of this? I am trying to remember why we had Unregistered < Offline.
> 
> Unregistered is an indication that the exporter may come back soon, so when you have very few exporters of one type and an exporter is ending a lease, while it is restarting the sub-child, still the server will lease it.
> 
> The problem comes when the exporter goes offline longer than expected (image rotation), ... then the client fails to connect.
> 
> We want to keep the "Unregistered" as "Online" for a while for this reason.
> 
> You don't see it in our lab because we have plenty, but when you have 1-2... the user hits that problem and then the lease is instantly rejected because all exporters are "offline" sometimes.

unless i misunderstood you, we already have that retry behavior with:
```
⠼ Waiting for lease: While there are 1 available exporters (i.e. ), none of them are online (0:00:01)
```
the lease is not rejected

mangelajo:
> > What is the advantage of this? I am trying to remember why we had Unregistered < Offline.
> > Unregistered is an indication that the exporter may come back soon, so when you have very few exporters of one type and an exporter is ending a lease, while it is restarting the sub-child, still the server will lease it.
> > The problem comes when the exporter goes offline longer than expected (image rotation), ... then the client fails to connect.
> > We want to keep the "Unregistered" as "Online" for a while for this reason.
> > You don't see it in our lab because we have plenty, but when you have 1-2... the user hits that problem and then the lease is instantly rejected because all exporters are "offline" sometimes.
> 
> unless i misunderstood you, we already have that retry behavior with:
> 
> ```
> ⠼ Waiting for lease: While there are 1 available exporters (i.e. ), none of them are online (0:00:01)
> ```
> 
> the lease is not rejected

Hmmmm, you're probably right. Ok, let me do a couple of tests tomorrow, but this would be complementary to the client-retry logic, while the window would be smaller, it'd still be possible.

Thanks for double checking @bennyz I was so sure...

bennyz:
> Hmmmm, you're probably right. Ok, let me do a couple of tests tomorrow, but this would be complementary to the client-retry logic, while the window would be smaller, it'd still be possible.

yeah we still need client retry because we still might have a race 

> > > What is the advantage of this? I am trying to remember why we had Unregistered < Offline.
> > > Unregistered is an indication that the exporter may come back soon, so when you have very few exporters of one type and an exporter is ending a lease, while it is restarting the sub-child, still the server will lease it.
> > > The problem comes when the exporter goes offline longer than expected (image rotation), ... then the client fails to connect.
> > > We want to keep the "Unregistered" as "Online" for a while for this reason.
> > > You don't see it in our lab because we have plenty, but when you have 1-2... the user hits that problem and then the lease is instantly rejected because all exporters are "offline" sometimes.
> > 
> > 
> > unless i misunderstood you, we already have that retry behavior with:
> > ```
> > ⠼ Waiting for lease: While there are 1 available exporters (i.e. ), none of them are online (0:00:01)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the lease is not rejected
> 
> Hmmmm, you're probably right. Ok, let me do a couple of tests tomorrow, but this would be complementary to the client-retry logic, while the window would be smaller, it'd still be possible.
> 
> Thanks for double checking @bennyz I was so sure...

yeah it does not replace client reconnection as there is still a chance for race

controller/internal/service/controller_service.go
```diff
@@ -323,12 +323,27 @@ func (s *ControllerService) Unregister(
...
+		exporter.Status.StatusMessage = reason
+	} else {
+		exporter.Status.StatusMessage = "Exporter unregistered"
+	}
+	syncOnlineConditionWithStatus(exporter)
+	meta.SetStatusCondition(&exporter.Status.Conditions, metav1.Condition{
+		Type:               string(jumpstarterdevv1alpha1.ExporterConditionTypeRegistered),
+		Status:             metav1.ConditionFalse,
+		ObservedGeneration: exporter.Generation,
+		Reason:             "Unregister",
```
bkhizgiy: Should we include the `Message` filed to make it consistent with how other statuses are being handled?

bkhizgiy (commented):
I think it's also worth adding some test coverage for this as part of the change.

evakhoni:
so this is complimentary to the client retry I'm working on :thinking: I'll take a look hopefully tomorrow to see how it would behave in conjunction with my changes

controller/internal/service/controller_service.go
mangelajo:
```suggestion
		Reason:             "Unregister",
		Message:            "Exporter unregistered from the controller and became offline."
```

mangelajo (changes requested):

mangelajo:
@bennyz rebase will also help catching up the new Github CI settings (less tests on top of the PR)

bkhizgiy (approved):

mangelajo:
See the tests ran by @evakhoni here https://github.com/jumpstarter-dev/jumpstarter/pull/829#issuecomment-4798553795
I would perhaps wait for 0.10.0 and see how this works, or backport it

mangelajo (changes requested):
Give me some time to learn about @evakhoni implementation, as this may break some of the improvements he did.

mangelajo:
after talking to benny, and verifying that his patch helps we are moving this to 0.10.0 then we can analyze it with more calm, backport if necessary.

mangelajo (approved):

PR #833, opened 2026-06-23
ci: add path-based filtering and rollup for container image builds

mangelajo:
## Problem

The `build-images` workflow builds all 8 container images on every merge queue run, even when most images' dependencies haven't changed. There's also no rollup job for use as a required status check.

### Path-based filtering per image

Added a `changes` job using `dorny/paths-filter` that detects which Containerfile dependencies were modified. Each matrix entry now has a `filter` key mapping to a path filter:

| Filter | Images | Trigger paths |
|--------|--------|---------------|
| `controller` | jumpstarter-controller | `controller/{Containerfile,go.mod,go.sum,cmd/,api/,internal/}` |
| `operator` | jumpstarter-operator | `controller/{Containerfile.operator,go.*,api/,internal/,deploy/operator/}` |
| `operator-bundle` | jumpstarter-operator-bundle | `controller/deploy/operator/{Containerfile.bundle,bundle/}` |
| `microshift` | microshift/bootc | `controller/deploy/microshift-bootc/`, `controller/deploy/operator/dist/` |
| `python` | jumpstarter, jumpstarter-devspace | `python/` |
| `python-utils` | jumpstarter-utils | `python/Containerfile.utils` |
| `python-dev` | jumpstarter-dev | `python/.devfile/Containerfile`, `.py-version` |

### Behavior by event

| Event | What gets built |
|-------|----------------|
| **push** (main/tags/release) | All images (always) |
| **workflow_dispatch** | All images (always) |
| **pull_request** (labeled) | Labeled images only (existing behavior preserved) |
| **merge_group** | Only images whose dependencies changed |

### Rollup job

Added an `image-builds` rollup job that can be set as a single required status check in branch protection. It passes when all builds succeed or are intentionally skipped (no changes), and fails on actual failures or cancellations.

## Required status check

Add `image-builds` to branch protection required checks. This single check covers all container image builds regardless of how many were skipped vs built.

### DIFF
--- a/.github/workflows/build-images.yaml
+++ b/.github/workflows/build-images.yaml
@@ -18,8 +18,66 @@ env:
   QUAY_ORG: quay.io/jumpstarter-dev
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      controller: ${{ steps.filter.outputs.controller }}
+      operator: ${{ steps.filter.outputs.operator }}
+      operator-bundle: ${{ steps.filter.outputs.operator-bundle }}
+      microshift: ${{ steps.filter.outputs.microshift }}
+      python: ${{ steps.filter.outputs.python }}
+      python-utils: ${{ steps.filter.outputs.python-utils }}
+      python-dev: ${{ steps.filter.outputs.python-dev }}
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          fetch-depth: 0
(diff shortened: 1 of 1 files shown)


bennyz (approved):

.github/workflows/build-images.yaml
```diff
@@ -18,8 +18,66 @@ env:
...
+              - 'controller/deploy/operator/dist/**'
+            python:
+              - 'python/Containerfile'
+              - 'python/.devfile/Containerfile.client'
+              - 'python/**'
+            python-utils:
+              - 'python/Containerfile.utils'
+            python-dev:
+              - 'python/.devfile/Containerfile'
+              - '.py-version'
```
raballew, line 72: .go-version
.uv-version

PR #836, opened 2026-06-24
test: mark PTY-dependent hooks tests as xfail on macOS (#821)

mangelajo:
## Problem

The alleged macOS PTY kernel buffer timing race condition continues to cause random test failures in CI despite multiple fix attempts:
- #560 — Added synchronous drain after async reader cancellation
- #733 — Added `select()`-based drain with timeout
- #826 — Increased `DRAIN_MAX_EMPTY_POLLS` from 1 to 10

**Latest failing CI run:** https://github.com/jumpstarter-dev/jumpstarter/actions/runs/28084374100/job/83146665529

Each fix reduced the frequency but did not eliminate the race. The CI runs parallel test suites with concurrent package installations, creating enough I/O contention for the macOS PTY buffer to miss the drain window. We haven't managed to reproduce it locally, but it happens on CI.

## Workaround

Mark all 19 tests that spawn real subprocesses via PTY and assert on captured logger output as `xfail(strict=False)` on macOS. This unblocks CI while the root cause is investigated.

Using `strict=False` means:
- If the test **passes** on macOS → reported as `xpass` (no failure)
- If the test **fails** on macOS → reported as `xfail` (no failure)
- On **Linux** → marker has no effect, tests run normally

### Tests marked with `@macos_pty_xfail`

**Output capture tests:**
- `test_hook_environment_variables`
- `test_real_time_output_logging`
- `test_post_lease_hook_execution_on_completion`
- `test_exec_bash`
- `test_exec_python3`
- `test_script_file_sh`
- `test_script_file_py_autodetects_python`
- `test_script_file_py_exec_override`
- `test_noninteractive_environment`
- `test_drain_captures_output_without_trailing_newline`

**Drain behavior tests (patched PTY):**
- `test_drain_reads_data_remaining_in_pty_buffer`
- `test_drain_select_oserror_exits_gracefully`
- `test_drain_select_valueerror_exits_gracefully`
- `test_drain_exits_when_deadline_exceeded_before_select`
- `test_drain_exception_is_suppressed`
- `test_drain_retries_empty_select_then_captures_data`
- `test_drain_terminates_after_max_empty_polls`
- `test_drain_empty_counter_resets_on_data`

**Regression tests:**
- `test_infrastructure_messages_at_debug_not_info`

## Related

- Re-opened #821 with updated description and moved to **0.10.0** milestone
- Added comments to #733 and #560 noting the issue persists

Handles #821 (targeted for proper fix in 0.10.0)

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
@@ -1,4 +1,5 @@
 import os
+import sys
 from contextlib import nullcontext
 from unittest.mock import AsyncMock, MagicMock, patch
 
@@ -17,6 +18,16 @@ from jumpstarter.exporter.hooks import (
 )
 
 pytestmark = pytest.mark.anyio
+
+# Tests that spawn real subprocesses via PTY and assert on captured logger
+# output are flaky on macOS due to a PTY kernel buffer timing race condition.
+# See https://github.com/jumpstarter-dev/jumpstarter/issues/821
+# Targeted for proper fix in 0.10.0.
+macos_pty_xfail = pytest.mark.xfail(
+    condition=sys.platform == "darwin",
+    reason="PTY output race condition on macOS (#821)",
+    strict=False,
+)
 
 
 class _PtyTracker:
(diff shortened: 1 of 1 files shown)


mangelajo:
Let's see if https://github.com/jumpstarter-dev/jumpstarter/pull/837 will fix it before merging this one

mangelajo:
> Let's see if #837 will fix it before merging this one

#837 made MacOS better, but it's creating some issues in E2E/linux which we don't want. As we talked about on the matrix channel xfail is acceptable for the hooks on macos until we investigate during 0.10.0

bennyz (approved):

PR #839, opened 2026-06-24
use 0.3.0 default for CLI

bennyz:
partial revert of https://github.com/jumpstarter-dev/jumpstarter/pull/838 due to defaulting to "" breaking flashing


mangelajo (commented):
oh right, the busybox image..

mangelajo (approved):

PR #846, opened 2026-06-28
fix: avoid reconciling ended leases

bennyz:
Otherwise this will spam the logs and can choke the server if there are invalid leases

### DIFF
--- a/controller/internal/controller/lease_controller.go
+++ b/controller/internal/controller/lease_controller.go
@@ -32,9 +32,12 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // LeaseReconciler reconciles a Lease object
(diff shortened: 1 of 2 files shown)


controller/internal/controller/lease_controller.go
```diff
@@ -77,6 +79,10 @@ func (r *LeaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		)
 	}
 
+	if lease.Status.Ended && isLeaseEnded(&lease) {
+		return ctrl.Result{}, nil
```
bkhizgiy: nit: I think this guard is dead code, the predicate already filters events with the ended label, so we only get here when the label is missing, and then isLeaseEnded returns false anyway.

controller/internal/controller/lease_controller.go
```diff
@@ -585,9 +591,38 @@ func filterOutOfflineExporters(approvedExporters []ApprovedExporter) []ApprovedE
...
+	if v, ok := obj.GetLabels()[string(jumpstarterdevv1alpha1.LeaseLabelEnded)]; ok {
+		return v == jumpstarterdevv1alpha1.LeaseLabelEndedValue
+	}
+	return false
+}
+
+// skipEndedPredicate returns a predicate that filters out leases carrying the
+// ended label. Leases without the label are admitted so the reconciler can
+// backfill it when Status.Ended is true but the label write was lost.
+func skipEndedPredicate() predicate.Funcs {
```
bkhizgiy, line 602: This predicate runs on all watched types, not just leases, which is fine for now since we only watch leases, but could be an issue if we later add more owns/watches,  maybe worth add a *Lease type check to be safe?

bkhizgiy (approved):
looks good from my side

raballew (approved):

evakhoni:
looks good from my side as well.

mangelajo (approved):

controller/internal/controller/lease_controller_test.go
```diff
@@ -1027,6 +1029,98 @@ var _ = Describe("orderApprovedExporters", func() {
...
+	It("should reject updates where new object has ended label", func() {
+		oldLease := &jumpstarterdevv1alpha1.Lease{}
+		newLease := &jumpstarterdevv1alpha1.Lease{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					string(jumpstarterdevv1alpha1.LeaseLabelEnded): jumpstarterdevv1alpha1.LeaseLabelEndedValue,
+				},
+			},
+		}
+		Expect(skipEnded.Update(event.UpdateEvent{ObjectOld: oldLease, ObjectNew: newLease})).To(BeFalse())
```
mangelajo, line 1091: I was reviewing a 2nd time because I was thinking about this.

Won't this prevent the controller from transitioning a lease to "ended"?

mangelajo (changes requested):

mangelajo:
ahh ok wait! t predicates control whether a reconcile is triggered, not whether an API write succeeds

mangelajo (approved):

PR #851, opened 2026-07-02
ride4: add support for fastboot erase

bennyz:
Support `fastboot erase` operation via the ride4 driver.
This allows erasing partitions the following way:

```shell
j storage erase recoveryinfo
```

### DIFF
--- a/python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/driver.py
+++ b/python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/driver.py
@@ -20,6 +20,7 @@ class RideSXDriver(Driver):
     decompression_timeout: int = field(default=15 * 60)  # 15 minutes
     flash_timeout: int = field(default=30 * 60)  # 30 minutes
     continue_timeout: int = field(default=20 * 60)  # 20 minutes
+    erase_timeout: int = field(default=120)  # 2 minutes
     storage_dir: str = field(default="/var/lib/jumpstarter/ridesx")
 
     # FLS configuration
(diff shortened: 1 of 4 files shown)


python/packages/jumpstarter-driver-ridesx/jumpstarter_driver_ridesx/driver.py
```diff
@@ -326,6 +327,40 @@ def flash_oci_image(
...
+        Args:
+            device_id: The fastboot device ID
+            partition: The partition name to erase (e.g., 'recoveryinfo')
+        """
+        if not partition or not partition.strip():
+            raise ValueError("Partition name cannot be empty")
+
+        self.logger.info(f"Erasing partition '{partition}' on device {device_id}")
+
+        cmd = ["fastboot", "-s", device_id, "erase", "--", partition]
```
mangelajo: why the -- between erase and partition?

bennyz: that's weird, i missed this hallucination but for some reason the operation worked correctly in the lab

mangelajo: yep, knowing you I suspected it was lab-tested but I was looking at the fastboot command and didn't seem to be needed.

bennyz: updated and retested, not sure why it worked though, an explanation could be that it tries to erase both "--" and "partition"

mangelajo (approved):

PR #853, opened 2026-07-02
feat(controller): track client token expiry in status

bennyz:
Add TokenExpiresAt timestamp and TokenExpiring condition to Client status, enabling downstream systems (Prometheus, alerting, kubectl) to detect tokens approaching expiry before CI pipelines break.


controller/deploy/operator/config/crd/bases/jumpstarter.dev_clients.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.18.0
+    controller-gen.kubebuilder.io/version: v0.16.3
```
mangelajo: I guess you need to update your kubebuilder.

bennyz: looks like makefile had this version, upgraded

mangelajo (approved):

PR #857, opened 2026-07-06
fix: unblock post-shell status probe when connection is lost

evakhoni:
## Summary

After the user exits `jmp shell`, a status probe is sent to detect exporter restarts before triggering the afterLease hook. If the exporter becomes unreachable without sending a TCP RST (e.g. network partition, firewall drop, suspended process), this probe blocks indefinitely — hanging the CLI after the shell exits.

The status monitor already handles this correctly: it detects stuck connections after 20 consecutive timeouts and sets `connection_lost = True`, logging progressive warnings to the user. However, the probe runs as an independent gRPC call and never observes the monitor's flag.

## Fix

Race the probe against a lightweight watcher task that polls `monitor.connection_lost` every second. When the monitor marks the connection as lost, the watcher cancels the probe's task group scope, unblocking the CLI. The probe succeeds normally in the healthy case and cancels the watcher itself.

This preserves the existing monitor behaviour — users still see the progressive timeout warnings before the CLI exits cleanly.

## Test

Added `test_probe_cancelled_when_connection_lost` to `TestRunShellWithLeaseAsync`:
- Verifies the probe stays blocked while `connection_lost` is False (5s negative phase)
- Verifies the probe is cancelled once `connection_lost` becomes True (positive phase)
- Guarded by `fail_after(10)` to fail hard instead of hanging if the fix regresses

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
@@ -263,6 +263,24 @@ async def _monitor_token_expiry(config, lease, cancel_scope, token_state=None) -
             return
 
 
+async def _cancel_if_connection_lost(monitor, coro):
+    """Run *coro* but cancel it as soon as monitor.connection_lost becomes True.
+
+    Returns the coroutine's result, or None if cancelled due to connection loss.
+    """
+    result = None
+    async with anyio.create_task_group() as tg:
+        async def _watcher():
+            while not monitor.connection_lost:
+                await anyio.sleep(1)
+            tg.cancel_scope.cancel()
+
+        tg.start_soon(_watcher)
+        result = await coro
+        tg.cancel_scope.cancel()
+    return result
+
+
 async def _run_shell_with_lease_async(lease, exporter_logs, config, command, cancel_scope):  # noqa: C901
     """Run shell with lease context managers and wait for afterLease hook if logs enabled.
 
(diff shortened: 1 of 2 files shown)


python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -352,14 +352,29 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
...
+                                        async with anyio.create_task_group() as probe_tg:
+
+                                            async def cancel_when_connection_lost():
+                                                while not monitor.connection_lost:
+                                                    await anyio.sleep(1)
+                                                probe_tg.cancel_scope.cancel()
+
+                                            probe_tg.start_soon(cancel_when_connection_lost)
+                                            probe_status = await client.get_status_async()
+                                            probe_tg.cancel_scope.cancel()
```
mangelajo: @evakhoni it looks good, but I have a small suggestion to improve reading: extracting the task-group + polling-watcher pattern into a small helper so the call site reads more clearly:

```python
async def _cancel_if_connection_lost(monitor, coro):
    """Run *coro* but cancel it as soon as monitor.connection_lost becomes True.

    Returns the coroutine's result, or None if cancelled due to connection loss.
    """
    result = None
    async with anyio.create_task_group() as tg:
        async def cancel_when_connection_lost():
            while not monitor.connection_lost:
                await anyio.sleep(1)
            tg.cancel_scope.cancel()

        tg.start_soon(cancel_when_connection_lost)
        result = await coro
        tg.cancel_scope.cancel()
    return result
```

Then we can call it like:

```python
probe_status = await _cancel_if_connection_lost(
    monitor, client.get_status_async()
)
```

Benefits:
- Makes the intent immediately clear at the call site
- Independently testable without the full `_run_shell_with_lease_async` machinery
- Reusable if the same race is needed elsewhere

mangelajo (changes requested):

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
evakhoni: makes sense. already too much nested logic in here..
pushed https://github.com/jumpstarter-dev/jumpstarter/pull/857/commits/58bf315c6ed8f7ad43c6ba12962bf82444b94541

mangelajo (approved):

Issue #41, opened 2025-10-30
RFE: On-Demand Virtual Target Provisioning


raballew:
based on discussion with @mangelajo 

**1. Problem Statement**

Jumpstarter currently excels at managing scarce, physical hardware targets. However, testing and development often require a mix of physical devices and scalable, virtual resources. Our current approach to virtual targets relies on pre-spawned, statically configured exporters (e.g., QEMU instances).

This model has several limitations:

- Scarcity: Virtual targets are treated as a finite pool, just like physical ones, which defeats their "virtually unlimited" potential.
- Inefficiency: Pre-spawned virtual resources consume cluster resources even when idle.
- Inflexibility: Users cannot dynamically request virtual targets with specific configurations (e.t., 1 core vs. 100 cores, specific memory/storage) on a per-lease basis.
- Orchestration Complexity: Users struggle to coordinate tests that require a mix of resources (e.g., "1 physical board and 3 virtual boards"), as they must compete for separate, static pools.

The core problem is that users have to think about "physical vs. virtual" and are constrained by a pre-configured supply of both.

**2. Proposal: A "Virtual Target Provider" (VTP) Subsystem**

We propose the creation of a "Virtual Target Provider" (VTP) subsystem designed to provision and manage virtual targets on-demand.

The guiding principle should be: "Get me a target that matches my requirements." The distinction between physical and virtual should be an implementation detail, not a primary concern for the user.

**Key Goals**

- Abstract the User Experience: A user should lease a target based on its properties (labels), not its physical or virtual nature.
- On-Demand Lifecycle: Virtual targets must be created when a lease is requested and destroyed when the lease is released.
- Dynamic Configuration: Leases must be able to specify resource requirements (CPU, memory, storage) or derive them from a known physical target's properties.
- Autoscaling: The system should be able to provision a virtually unlimited number of targets, constrained only by the underlying cluster's resources which is autoscalable.

**3 . Proposed Implementation & Architecture**

This RFE proposes a new set of controllers and CLI commands to handle the virtual target lifecycle.

**A. Architectural Vision (Long-Term)**

**VTP Controller(s):** A new, dedicated jumpstarter-virtual-provider-controller would be created. This controller would watch for Lease objects that have virtual requirements.

**Provider-Specific Logic:** This main VTP controller could delegate to provider-specific controllers (e.g., qemu-controller, corellium-controller), which would be deployed by the operator via new CRDs (e.g., JumpstarterQEMUProvider).

[...]

PR #863, opened 2026-07-07
feat: add enabled field to Exporter CRD and allowDisabled to Lease (JEP-0014 Phase 1)

mangelajo:
## Summary

Implements **Phase 1 of [JEP-0014: Virtual Scalable Exporters](https://github.com/jumpstarter-dev/jumpstarter/blob/main/python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md)** — the `enabled` field on the Exporter CRD.

This is the foundational building block for graceful scale-down of virtual exporter pools and is immediately useful for physical exporter maintenance workflows.

### Exporter CRD: `spec.enabled` field
- New boolean field (default: `true`) that controls whether the controller will assign new leases to this exporter
- When `false`, the exporter is excluded from selector-based lease matching and explicitly rejected for `exporterRef`-based leases
- Backward compatible: existing exporters with no `enabled` field behave as `enabled: true`
- New `Enabled` printer column visible in `kubectl get exporters`

### Lease CRD: `spec.allowDisabled` field
- New boolean field that permits leasing a disabled exporter when used with `exporterRef`
- Useful for investigating broken exporters that have been administratively disabled
- If a user targets a disabled exporter **without** `allowDisabled`, the lease gets an `Unsatisfiable` condition with a helpful error message:
  > Requested exporter X is disabled. To lease a disabled exporter, set spec.allowDisabled: true on the Lease, or use --allow-disabled with jmp create lease

### Controller Logic
- `filterOutDisabledExporters()`: filters disabled exporters from selector-based lease matching (before policy checks)
- `ExporterRef` path: checks if the targeted exporter is disabled, returns `Unsatisfiable("ExporterDisabled")` unless `allowDisabled` is set

### Protocol (gRPC)
- `Exporter` message: added `enabled` field (field 6, `OUTPUT_ONLY`)
- `Lease` message: added `allow_disabled` field (field 14, `IMMUTABLE`)
- `ToProtobuf()` methods updated to serialize both fields
- `buf.gen.yaml` updated to use local `protocol/proto` as source of truth (replaces external git repo reference)

### CLI
- `jmp create lease --allow-disabled`: new flag to permit leasing disabled exporters
- `jmp get exporters`: disabled exporters are **hidden by default**; use `--allow-disabled` to show them (adds an `ENABLED` column)

### Lab maintenance
```bash
# Admin takes an exporter offline for maintenance
kubectl patch exporter broken-board -p '{"spec":{"enabled":false}}'

# Leases no longer match this exporter
jmp lease -l board=rpi4  # skips the disabled exporter

# Listing hides disabled exporters by default
jmp get exporters         # broken-board not shown
jmp get exporters --allow-disabled  # shows all, with ENABLED column

# Investigator can still access it explicitly
jmp create lease -n broken-board --allow-disabled --duration 1h
```

### Graceful scale-down (JEP-0014)
```
1. ExporterSet controller sets enabled: false on the target exporter
2. Controller waits to confirm no lease was assigned
3. Controller deletes the Pod and Exporter CR
```

### Go (controller)
- **Unit test**: `filterOutDisabledExporters` with nil/true/false `Enabled` values, slice immutability
- **Integration test**: disabled exporter via `ExporterRef` → `ExporterDisabled` reason with helpful message
- **Integration test**: disabled exporter via `ExporterRef` + `AllowDisabled: true` → lease succeeds
- **Integration test**: selector match with one disabled exporter → lease goes to enabled one
- **Integration test**: selector match with all disabled → unsatisfiable

### Python (CLI)
- Updated `create_test.py` to pass `allow_disabled` parameter
- Updated `grpc_test.py` display tests — ENABLED column only shown with `show_disabled`
- New `TestWithDisabledOption` tests for conditional ENABLED column

## Backward Compatibility

- `spec.enabled` defaults to `true` — existing Exporters behave exactly as before
- `spec.allowDisabled` defaults to `false` — existing Leases behave exactly as before
- No changes to the existing gRPC protocol for physical exporters (fields are additive)

## References

- [JEP-0014: Virtual Scalable Exporters](https://github.com/jumpstarter-dev/jumpstarter/blob/main/python/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md) — Phase 1 deliverables
- [GitHub Issue #41: RFE: On-Demand Virtual Target Provisioning](https://github.com/jumpstarter-dev/jumpstarter/issues/41)


python/packages/jumpstarter/jumpstarter/client/grpc.py
```diff
@@ -28,6 +28,7 @@ def add_display_columns(table, options: WithOptions = None):
     if options is None:
         options = WithOptions()
     table.add_column("NAME")
+    table.add_column("ENABLED")
```
mangelajo: I believe that we should not print the enabled/disabled indicator, but instead, during listing, not print the disabled ones, unless --allow-disabled is included during the list request.

mangelajo: Good call. Updated in c72ad978:

- Disabled exporters are now **hidden** from `jmp get exporters` by default
- Added `--allow-disabled` flag to `jmp get exporters` to include them
- The `ENABLED` column only appears when `--allow-disabled` is used, so users can distinguish disabled from enabled exporters
- `ExporterList` now filters via `_visible_exporters()` for all output formats (table, JSON, YAML, name)

protocol/proto/jumpstarter/client/v1/client.proto
```diff
@@ -94,6 +94,9 @@ message Exporter {
   jumpstarter.v1.ExporterStatus status = 4 [(google.api.field_behavior) = OUTPUT_ONLY];
   // A human-readable message providing details about the exporter status.
   string status_message = 5 [(google.api.field_behavior) = OUTPUT_ONLY];
+  // Whether the exporter is enabled for lease assignment.
+  // When false, the controller will not assign new leases to this exporter.
+  bool enabled = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
```
bennyz: should this be `optional bool` for compat?

mangelajo: I believe everything is optional in grpc, specially new fields. 

Non provided field => ignored.
Unknown field => is ignored.

But double checking.

mangelajo: But it is true that in some areas we use optional everywhere...

investigating..

mangelajo: Claude AI analysis

@bennyz is right — this should be `optional bool` for backward compatibility.

In proto3, a non-optional `bool` has an implicit default of `false` with **no presence tracking**. When an **old server** (that doesn't know about the `enabled` field) sends an `Exporter` message to a **new client**, the field will be absent on the wire and the client deserializes it as `false` — making every exporter from an old server appear **disabled**.

The generated Go getter confirms this (`client.pb.go`):
```go
func (x *Exporter) GetEnabled() bool {
    if x != nil {
        return x.Enabled
    }
    return false  // wrong default for compat!
}
```

With `optional bool enabled = 6`, proto3 generates presence-tracking, so client code can distinguish "server didn't send this field" (old server -> treat as enabled) from "server explicitly sent `false`" (disabled).

The project already follows this pattern:
- `optional bool only_active = 5` in `ListLeasesRequest` (this file, line 207)
- `optional bool release_lease = 3` in `jumpstarter.proto`
- The K8s CRD side already uses `*bool` with nil-means-true in `exporter_helpers.go:15-16`

**`allow_disabled` (field 14) is fine as plain `bool`** — its default of `false` is the correct backward-compatible behavior.

Fix: change this line to:
```protobuf
optional bool enabled = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
```

Then update Go serialization in `exporter_helpers.go:44` to use a pointer:
```go
Enabled: proto.Bool(e.IsEnabled()),
```

And the Python client (`grpc.py:112`) to handle the optional:
```python
enabled=data.enabled if data.HasField("enabled") else True,
```

kirkbrauer: Yes, I believe that new fields will get their default Protobuf value, which is `false` in the case of a boolean. https://protobuf.dev/programming-guides/proto3/#updating

controller/internal/controller/lease_controller.go
```diff
@@ -248,14 +248,26 @@ func (r *LeaseReconciler) reconcileStatusExporterRef(
...
 			matchingExporters = []jumpstarterdevv1alpha1.Exporter{exporter}
 		} else {
 			// List all exporters matching selector
 			listed, err := r.ListMatchingExporters(ctx, lease, selector)
 			if err != nil {
 				return fmt.Errorf("reconcileStatusExporterRef: failed to list matching exporters: %w", err)
 			}
-			matchingExporters = listed.Items
+			// Filter out disabled exporters from selector-based listing
+			matchingExporters = filterOutDisabledExporters(listed.Items)
```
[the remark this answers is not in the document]
mangelajo: +1 we should provide a more precise response indicating that some are disabled.

controller/internal/controller/lease_controller_test.go
```diff
@@ -743,6 +743,96 @@ var _ = Describe("Lease Controller", func() {
...
+			Expect(k8sClient.Create(ctx, lease)).To(Succeed())
+			_ = reconcileLease(ctx, lease)
+
+			updatedLease := getLease(ctx, lease.Name)
+			Expect(updatedLease.Status.ExporterRef).To(BeNil())
+
+			// Restore for subsequent tests
+			setExporterEnabled(ctx, testExporter1DutA.Name, true)
+			setExporterEnabled(ctx, testExporter2DutA.Name, true)
+		})
```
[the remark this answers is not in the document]
mangelajo: +1

bennyz (approved):

kirkbrauer (approved):

PR #864, opened 2026-07-08
feat: bootstrap ExporterSet controller for JEP-0014 Phase 2

mangelajo:
## Summary

Scaffolds the exporter-set-controller infrastructure for virtual scalable exporters ([JEP-0014](https://github.com/jumpstarter-dev/jumpstarter/blob/main/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md) Phase 2). This PR provides the foundational CRDs, controller binary, container images, and CI integration — everything needed to start implementing the scaling logic.

Also updates JEP-0014 status to **Approved** and fixes the exporter sidecar image reference.

### New CRDs (`virtualtarget.jumpstarter.dev/v1alpha1`)

**VirtualTargetClass** — namespaced backend profile:
- `provisioner` — which exporter-set controller handles this class (e.g. `qemu.jumpstarter.dev`)
- `credentialsSecretRef` — optional Secret reference for API-backed provisioners
- `caBundleConfigMapRef` — optional ConfigMap reference for corporate CA injection
- `parameters` — nested schemaless YAML object for provisioner-specific config
- `bindingMode` — `Immediate` (warm pool) or `WaitForFirstConsumer` (on-demand)
- `reclaimPolicy` — `Delete` or `Retain`
- `scheduling` — nodeSelector, tolerations, resource requirements inherited by rendered Pods

**ExporterSet** — namespaced scaling resource (ReplicaSet + HPA analog):
- `minReplicas` / `maxReplicas` (default: 4) / `minAvailableReplicas` — familiar scaling vocabulary
- CEL validation: `minReplicas <= maxReplicas` when max is set
- `scaleDownCooldown` — configurable cooldown before scaling down excess replicas
- `recycleStrategy` — `ExitAndReplace` or `InPlaceReuse`
- `virtualTargetClassName` — same-namespace VirtualTargetClass reference
- `parameters` — optional overrides deep-merged over class parameters
- `selector` / `template` — label selector and embedded exporter template (Deployment idiom)
- **Scale subresource** — enables `kubectl scale` and HPA/KEDA interoperability
- Status: `replicas`, `readyReplicas`, `availableReplicas`, `leasedReplicas`, conditions

### New Controller Binary

`cmd/exporter-set-controller/main.go`:
- Standalone binary with `--provisioner` flag to select which provisioner to handle
- Single image, one Deployment per provisioner (JEP-0014 DD-2)
- Standard controller-runtime setup with health/ready probes, metrics, leader election
- `selectProvisioner()` switch for registering provisioner implementations

### Provisioner Architecture

Separate Go package (`internal/exporterset/`) with pluggable provisioner interface:

```
internal/exporterset/
├── provisioner.go              # Provisioner interface (RenderPod, Cleanup)
├── reconciler.go               # ExporterSetReconciler
└── provisioners/
    └── qemu/
        └── qemu.go             # qemu.jumpstarter.dev implementation
```

**Provisioner interface** (`RenderPod`, `Cleanup`) with documented future extension points for `RenderSecrets` / `RenderConfigMaps`.

**QEMU provisioner** renders the sidecar pattern (KEP-753):
- Exporter as native sidecar init container (`restartPolicy: Always`, image: `quay.io/jumpstarter-dev/jumpstarter:latest`)
- QEMU runtime as main container (image: `quay.io/jumpstarter-dev/virtual/qemu-runtime:latest`)
- Shared `emptyDir` volume at `/shared` for Unix socket communication (QMP, serial, launcher)
- `OwnerReferences` set on rendered Pods for garbage collection on ExporterSet deletion
- Scheduling from `VirtualTargetClass` applied (nodeSelector, tolerations, resources)

### Container Images

- `Containerfile.exporter-set-controller` — multi-stage Go build (UBI10 go-toolset builder, UBI9 ubi-micro runtime), matching existing controller pattern
- `Containerfile.qemu-runtime` — Fedora 44 base with `qemu-img`, `qemu-kvm`, `procps-ng`, `virtiofsd`; no ENTRYPOINT (exporter sidecar manages QEMU lifecycle via shared volume launcher)

### Build & CI

- `Makefile` — `build` target includes `bin/exporter-set-controller`, new `docker-build-exporter-set-controller` target
- `build-images.yaml` — CI matrix entries for `jumpstarter-exporterset-controller` and `virtual/qemu-runtime` images

### JEP-0014

- Status updated from `Draft` to `Approved`
- Fixed exporter sidecar image reference: `quay.io/jumpstarter-dev/exporter:latest` → `quay.io/jumpstarter-dev/jumpstarter:latest`
- Updated examples to use `virtualtarget.jumpstarter.dev/v1alpha1` API group

### Review feedback addressed

- **EmbeddedObjectMeta** — template `labels`/`annotations` no longer pruned by the apiserver
- **CEL validation** — `minReplicas <= maxReplicas` when `maxReplicas != 0`
- **Reconciler** — distinguish VTC NotFound; `.Owns(Pod)`; watch VTC changes
- **API group** — moved to `virtualtarget.jumpstarter.dev/v1alpha1`
- **Runtime images** — `quay.io/jumpstarter-dev/virtual/*` namespace (e.g. `virtual/qemu-runtime`)
- **CA injection** — `caBundleConfigMapRef` on VirtualTargetClass (pod mounting deferred to Phase 2)
- **Docs** — JEP links and index status updated

## What's Next (not in this PR)

- Scaling logic implementation (count owned Exporters, deep-merge parameters, scale up/down)
- Demand-driven scale-up (watch pending Leases matching selector)
- Graceful scale-down (set `enabled=false`, wait, delete)
- Exporter sidecar configuration (driver config, controller endpoint, credentials injection)
- QEMU runtime configuration from merged parameters (CPU, memory, firmware, machine type)
- CA bundle mounting in rendered Pods (API field added, wiring deferred)
- Operator integration (`Jumpstarter` CR `spec.exporterSets` section)
- Integration tests

## References

- [JEP-0014: Virtual Scalable Exporters](https://github.com/jumpstarter-dev/jumpstarter/blob/main/docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md) — Phase 2 deliverables
- [GitHub Issue #41: RFE: On-Demand Virtual Target Provisioning](https://github.com/jumpstarter-dev/jumpstarter/issues/41)
- Phase 1 (merged): #863 — `enabled` field on Exporter CRD


controller/api/v1alpha1/exporterset_types.go
```diff
@@ -0,0 +1,165 @@
...
+// ExporterSetSpec defines the desired state of ExporterSet.
+type ExporterSetSpec struct {
+	// MinReplicas is the minimum number of instances (floor).
+	// +kubebuilder:default=0
+	// +kubebuilder:validation:Minimum=0
+	MinReplicas int32 `json:"minReplicas,omitempty"`
+
+	// MaxReplicas is the maximum number of instances (ceiling).
+	// 0 means no upper bound.
+	// +kubebuilder:default=0
```
mangelajo: should we have a bounded default?

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,109 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-", exporterSet.Name),
+			Namespace:    exporterSet.Namespace,
+			Labels:       exporterSet.Spec.Template.Labels,
+		},
+		Spec: corev1.PodSpec{
+			// TODO: populate from VirtualTargetClass.scheduling
+			// TODO: render exporter sidecar + QEMU runtime containers
+			// TODO: add shared emptyDir volume
+			Containers: []corev1.Container{
```
mangelajo, line 133: This still does not reflect the sidecar pattern: https://jumpstarter.dev/main/contributing/jeps/JEP-0014-virtual-scalable-exporters.html#container-backed-targets-sidecar-pattern
(that reference needs correction quay.io/jumpstarter-dev/exporter:latest must be quay.io/jumpstarter-dev/jumpstarter:${VERSION}" matching the version of the controller.

where quay.io/jumpstarter-dev/qemu-runtime:latest is a new image that needs Containerfile as well, and it must be created from the ubi image installing the qemu tools (in our lab now we have a derivate image where we do: 
RUN dnf install --setopt=install_weak_deps=False -y qemu-img qemu-kvm procps-ng virtiofsd && \
      dnf clean all
)

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,109 @@
...
+//   - Exporter sidecar (native sidecar init container, restartPolicy: Always)
+//   - QEMU runtime container (main container)
+//   - Shared emptyDir volume for Unix socket communication (QMP, serial, launcher)
+//
+// TODO: This is a stub. Full implementation will:
+//   - Apply scheduling constraints from VirtualTargetClass
+//   - Configure QEMU runtime from merged parameters (CPU, memory, firmware, etc.)
+//   - Set up exporter sidecar with driver configuration from ExporterSet template
+//   - Mount firmware/OS images as OCI volume sources
+func (p *Provisioner) RenderPod(
```
mangelajo, line 95: We may want to have a RenderSecrets, RenderConfigmaps as well since some pods may need specific secrets or configmaps created to be accessed from the pod, perhaps it's enough to make a note now on the interface so that we can implement it later if it becomes necessary.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,109 @@
...
+	ctx context.Context,
+	exporterSet *jumpstarterdevv1alpha1.ExporterSet,
+	vtc *jumpstarterdevv1alpha1.VirtualTargetClass,
+	mergedParameters map[string]interface{},
+) (*corev1.Pod, error) {
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
```
mangelajo, line 105: we must annotate pods so that they become owned by the exporterset, if the exporter set gets deleted, the pod get deleted.

controller/deploy/operator/config/crd/bases/jumpstarter.dev_exportersets.yaml
```diff
@@ -58,12 +58,10 @@ spec:
...
               maxReplicas:
-                default: 0
-                description: |-
-                  MaxReplicas is the maximum number of instances (ceiling).
-                  0 means no upper bound.
+                default: 10
+                description: MaxReplicas is the maximum number of instances (ceiling).
                 format: int32
-                minimum: 0
+                minimum: 1
```
mangelajo: minimum must be 0, to let setting an unbound limit, and default I would use 4 to be on the cautious side.

controller/api/v1alpha1/exporterset_types.go
```diff
@@ -0,0 +1,164 @@
+/*
+Copyright 2024.
```
mangelajo:
```suggestion
Copyright 2026 The Jumpstarter Authors
```

mangelajo: or .. in the same way we do it in other files in this repo.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,183 @@
...
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(
+					exporterSet,
+					schema.GroupVersionKind{
+						Group:   "jumpstarter.dev",
+						Version: "v1alpha1",
+						Kind:    "ExporterSet",
+					},
+				),
+			},
```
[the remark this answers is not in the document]
mangelajo: Makes sense, it's simpler to let the common code set the references itself.

evakhoni:
small nit, the JEP-0014 links in the description are pointing to python/docs instead of /docs so they're 404

controller/api/virtualtarget/v1alpha1/exporterset_types.go
```diff
@@ -0,0 +1,165 @@
...
+
+	// Spec defines the exporter configuration.
+	Spec ExporterTemplateSpec `json:"spec,omitempty"`
+}
+
+// ExporterSetSpec defines the desired state of ExporterSet.
+type ExporterSetSpec struct {
+	// MinReplicas is the minimum number of instances (floor).
+	// +kubebuilder:default=0
+	// +kubebuilder:validation:Minimum=0
```
bennyz, line 80: do we want a CEL rule here for minReplicas<= maxReplicas ?

controller/internal/exporterset/reconciler.go
```diff
@@ -0,0 +1,115 @@
...
+	// 6. Update ExporterSet.status
+
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *ExporterSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&jumpstarterdevv1alpha1.ExporterSet{}).
+		Owns(&jumpstarterdevv1alpha1.Exporter{}).
```
bennyz, line 124: We need Owns for pods as well?
and watches for VTCs?

controller/internal/exporterset/reconciler.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	// Resolve the VirtualTargetClass
+	var vtc jumpstarterdevv1alpha1.VirtualTargetClass
+	vtcKey := client.ObjectKey{
+		Namespace: exporterSet.Namespace,
+		Name:      exporterSet.Spec.VirtualTargetClassName,
+	}
+	if err := r.Get(ctx, vtcKey, &vtc); err != nil {
+		logger.Error(err, "unable to get VirtualTargetClass",
```
bennyz: should we distinguish NotFound errors?

controller/api/virtualtarget/v1alpha1/exporterset_types.go
kirkbrauer: This might be a good idea, but perhaps too complex?

controller/api/v1alpha1/exporterset_types.go
kirkbrauer: Humm, it's hard to decide what a reasonable max should be, maybe base it on cluster size, or some default like 10?

controller/internal/exporterset/reconciler.go
kirkbrauer: That might be a good idea.

controller/internal/exporterset/reconciler.go
kirkbrauer: Yeah, from a usability perspective this is probably a good idea.

.github/workflows/build-images.yaml
```diff
@@ -34,6 +34,14 @@ jobs:
             label: jumpstarter-controller
             dockerfile: controller/Containerfile
             context: controller
+          - image_name: jumpstarter-dev/jumpstarter-exporterset-controller
+            label: jumpstarter-exporterset-controller
+            dockerfile: controller/Containerfile.exporter-set-controller
+            context: controller
+          - image_name: jumpstarter-dev/qemu-runtime
```
kirkbrauer: Maybe we create a sub-namespace for these runtimes for consistency?

controller/config/samples/v1alpha1_exporterset.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: jumpstarter.dev/v1alpha1
```
kirkbrauer: I think we should introduce API groups so this becomes `virtualtarget.jumpstarter.dev/v1alpha1` or something similar. This will help us keep the "core" API concerns separate from any pluggable components.

controller/Containerfile.qemu-runtime
```diff
@@ -0,0 +1,19 @@
+# QEMU runtime container for ExporterSet virtual targets.
+# Provides QEMU tools for running virtual machines as part of
+# the sidecar pattern (paired with the jumpstarter exporter).
+#
+# Uses Fedora as the base image because QEMU packages are not
+# available in UBI-minimal repositories.
+FROM registry.fedoraproject.org/fedora:44
+
+RUN dnf install --setopt=install_weak_deps=False -y \
```
kirkbrauer: We should support injecting corporate CAs here to allow network access from within QEMU to work as expected under all conditions.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,174 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// ProvisionerName is the provisioner identifier for
+	// QEMU-based virtual targets.
+	ProvisionerName = "qemu.jumpstarter.dev"
+
+	// DefaultExporterImage is the exporter sidecar image.
+	DefaultExporterImage = "quay.io/jumpstarter-dev/jumpstarter:latest"
```
kirkbrauer, line 41: Nice! I like the sidecar pattern here. We should be careful about shared interfaces though.

kirkbrauer (commented):
I think this is looking good, a few comments ;)

.github/workflows/build-images.yaml
mangelajo: 👍 

a few ideas:

jumpstarter-dev/virtualtarget/qemu-runtime
jumpstarter-dev/exporterset/qemu-runtime

controller/api/virtualtarget/v1alpha1/exporterset_types.go
mangelajo: I think it's probably fine, the API will check this automatically, let me try.

controller/config/samples/v1alpha1_exporterset.yaml
mangelajo: +1 I will check if this has any implications and if it can be handled on the same controller, I guess yes, but will ping back here.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: it was your idea :-D 

The only complication is spawning of processes, in theory we will need a thin wrapper to let us exec invocations on the other 2nd container. My plan is to use a unix socket over an emptyDir in between both, and a thin wrapper, perhaps a will be a good use case to start with rust and avoid any python runtimes on the 2nd container.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Done

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: This is done in the controller itself.

controller/api/v1alpha1/exporterset_types.go
```diff
@@ -0,0 +1,165 @@
...
+	// Drivers is the list of drivers to configure on each exporter instance.
+	// +optional
+	Drivers []DriverConfig `json:"drivers,omitempty"`
+}
+
+// ExporterSetTemplate defines the template for exporter instances created by this set.
+type ExporterSetTemplate struct {
+	// Metadata for created Exporter resources.
+	// +optional
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
bkhizgiy: spec.template.metadata is generated as a bare type: object, so the apiserver prunes nested fields like labels on apply. could we mark it schemaless, or explicitly declare labels/annotations? Otherwise the sample labels never stick.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,174 @@
...
+	if vtc.Spec.Scheduling != nil {
+		if vtc.Spec.Scheduling.NodeSelector != nil {
+			pod.Spec.NodeSelector = maps.Clone(vtc.Spec.Scheduling.NodeSelector)
+		}
+		if vtc.Spec.Scheduling.Tolerations != nil {
+			pod.Spec.Tolerations = append([]corev1.Toleration(nil), vtc.Spec.Scheduling.Tolerations...)
+		}
+		if vtc.Spec.Scheduling.Resources != nil {
+			// Apply resource requirements to target-runtime
+			pod.Spec.Containers[0].Resources = *vtc.Spec.Scheduling.Resources
```
bkhizgiy: I think some of the resources still shallow copy, so the request/limit maps stay shared with the cached vtc, maybe it's worth cloning those too.

controller/Containerfile.qemu-runtime
mangelajo: But QEMU itself may not need the CAs?, would that be the images themselves?

I think that runtime install of CAs could be well served by the runtime thin wrapper to allow exec calls from the sidecar (we need to extend the image to include this later anyway, I'll keep it in mind).

controller/go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	google.golang.org/grpc v1.80.0
 	google.golang.org/protobuf v1.36.11
 	k8s.io/api v0.33.0
+	k8s.io/apiextensions-apiserver v0.33.0
```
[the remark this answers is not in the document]
mangelajo: we have renovate handling our dependencies, and will be updated as soon as it's technically possible by the dependency tree.

controller/api/virtualtarget/v1alpha1/exporterset_types.go
mangelajo: Added a CEL rule: `self.maxReplicas == 0 || self.minReplicas <= self.maxReplicas`. The `== 0` escape handles the unbounded maxReplicas case.

controller/internal/exporterset/reconciler.go
mangelajo: Added `.Owns(&corev1.Pod{})` and a `.Watches()` for VirtualTargetClass that enqueues ExporterSets referencing the changed VTC.

controller/internal/exporterset/reconciler.go
mangelajo: Now distinguishing NotFound (log + no requeue) from transient errors (requeue). Added a TODO for setting a status condition when the class is missing.

.github/workflows/build-images.yaml
mangelajo: Moved to `quay.io/jumpstarter-dev/virtual/qemu-runtime` under the `virtual/*` namespace — future runtime images (corellium, etc.) will follow the same pattern.

controller/config/samples/v1alpha1_exporterset.yaml
mangelajo: Done — ExporterSet and VirtualTargetClass now live under `virtualtarget.jumpstarter.dev/v1alpha1`, keeping the core `jumpstarter.dev` API group clean.

controller/Containerfile.qemu-runtime
mangelajo: Added `caBundleConfigMapRef` to `VirtualTargetClassSpec` following the ConfigMap key selector pattern from the operator's JWT auth. The actual pod mounting will come in Phase 2 when we implement the full provisioner configuration.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Agreed — the thin launcher wrapper on the shared volume will be the contract between sidecar and runtime. Planning to spec that interface carefully, possibly as a small Rust binary to avoid pulling Python into the runtime container.

controller/api/v1alpha1/exporterset_types.go
mangelajo: Good catch. Replaced the embedded `metav1.ObjectMeta` with an explicit `EmbeddedObjectMeta` struct that only declares `Labels` and `Annotations` — this generates a proper CRD schema so the apiserver won't prune them on apply.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Fixed — using `ResourceRequirements.DeepCopy()` now so the `Limits` and `Requests` maps are independently copied from the cached VTC.

mangelajo:
@evakhoni Fixed the links in the JEP doc and PR description — they now point to `docs/source/contributing/jeps/` after the docs move in #880. Also updated the JEP index status to **Approved**.

mangelajo:
@kirkbrauer @bennyz @bkhizgiy should be ready for another round of reviews

controller/internal/exporterset/provisioners/qemu/qemu.go
kirkbrauer: The challenge is also interfaces, some will require a privileged container or some virtualization layer. For example, Kernel-level CAN interfaces in a container can behave _oddly_ sometimes. So we should think about off the interfaces we might want to support and look at how QEMU provides those.

controller/deploy/operator/config/rbac/role.yaml
```diff
@@ -2,22 +2,15 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: manager-role
+  name: jumpstarter-manager-role
```
raballew: The `role_binding.yaml` still references `name: manager-role`, creating a dangling reference.

controller/Containerfile.exporter-set-controller
```diff
@@ -0,0 +1,41 @@
+# Build the exporter-set-controller binary
+FROM registry.access.redhat.com/ubi10/go-toolset:1.26.3 AS builder
```
raballew: The builder stage is missing `--platform=$BUILDPLATFORM` in the FROM line.

controller/internal/exporterset/reconciler.go
```diff
@@ -0,0 +1,162 @@
...
+) []reconcile.Request {
+	vtc, ok := obj.(*virtualtargetv1alpha1.VirtualTargetClass)
+	if !ok {
+		return nil
+	}
+
+	var exporterSetList virtualtargetv1alpha1.ExporterSetList
+	if err := r.List(ctx, &exporterSetList, client.InNamespace(vtc.Namespace)); err != nil {
+		return nil
+	}
```
raballew, line 148: `findExporterSetsForVTC` silently swallows `List` errors without logging.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,175 @@
...
+	mergedParameters map[string]interface{},
+) (*corev1.Pod, error) {
+	restartAlways := corev1.ContainerRestartPolicyAlways
+
+	pod := &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-", exporterSet.Name),
+			Namespace:    exporterSet.Namespace,
+			Labels:       maps.Clone(exporterSet.Spec.Template.Metadata.Labels),
+		},
```
raballew, line 110: Users who set annotations in the ExporterSet template will have them silently discarded.

controller/api/virtualtarget/v1alpha1/exporterset_types.go
```diff
@@ -0,0 +1,177 @@
...
+	MaxReplicas int32 `json:"maxReplicas,omitempty"`
+
+	// MinAvailableReplicas is the warm buffer: ready and unleased instances.
+	// +kubebuilder:default=0
+	// +kubebuilder:validation:Minimum=0
+	MinAvailableReplicas int32 `json:"minAvailableReplicas,omitempty"`
+
+	// ScaleDownCooldown is how long to wait before scaling down excess replicas.
+	// +kubebuilder:default="5m"
```
raballew, line 95: The CEL validation rule does not check `minAvailableReplicas <= maxReplicas`. Add a second CEL rule: `self.maxReplicas == 0 || self.minAvailableReplicas <= self.maxReplicas` with an appropriate message.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,175 @@
...
+				},
+			},
+			Volumes: []corev1.Volume{
+				{
+					Name: sharedVolumeName,
+					VolumeSource: corev1.VolumeSource{
+						EmptyDir: &corev1.EmptyDirVolumeSource{},
+					},
+				},
+			},
```
raballew, line 156: The shared emptyDir volume has no `SizeLimit` set. A compromised or misbehaving container could write unbounded data to the emptyDir, consuming the node's ephemeral storage and potentially causing eviction of other pods on the same node.

raballew (changes requested):
in additon to my comments below, test coverage is completely absent for the reconciler and provisioner if I see this correctly.

controller/deploy/operator/config/rbac/role.yaml
mangelajo: wait, why is it changing all this? hmmmmmmmm

mangelajo: Good catch — and it was worse than a dangling name.

Root cause: `controller/Makefile` `make manifests` ran `controller-gen` with `rbac:roleName=jumpstarter-manager-role` writing into `deploy/operator/config/rbac/`, which **overwrote** the operator's `manager-role` ClusterRole (cert-manager, deployments, routes, Jumpstarter CR, etc.).

Fixed in d2e40ea0:
1. Restored `role.yaml` from `main` (`name: manager-role`)
2. Stopped the controller Makefile from emitting RBAC into the operator path (CRDs only; operator RBAC stays owned by `deploy/operator/Makefile`)

controller/Containerfile.exporter-set-controller
mangelajo: Fixed — added `--platform=$BUILDPLATFORM` to match `Containerfile` / `Containerfile.operator`.

controller/internal/exporterset/reconciler.go
mangelajo: Fixed — List errors are now logged before returning nil.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Fixed — template annotations are now cloned onto the rendered Pod ObjectMeta.

controller/api/virtualtarget/v1alpha1/exporterset_types.go
mangelajo: Fixed — added CEL: `self.maxReplicas == 0 || self.minAvailableReplicas <= self.maxReplicas` and regenerated the CRD.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Fixed — shared emptyDir now has `SizeLimit: 100Mi`.

mangelajo:
@raballew Also added basic unit tests in `d2e40ea0` for the reconciler stubs (VTC NotFound, provisioner mismatch, `findExporterSetsForVTC` filtering) and QEMU `RenderPod` (labels/annotations, SizeLimit, scheduling deep-copy). Fuller envtest coverage can follow once the scaling logic lands.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Agreed — we'll inventory the interfaces we want (serial, USB, CAN, network) and map them to QEMU device models plus required container capabilities/privileges. Captured as a Phase 2 TODO on `RenderPod` in a8052573 rather than blocking this scaffold PR. The thin launcher on the shared volume remains the sidecar↔runtime contract.

controller/api/virtualtarget/v1alpha1/exporterset_types.go
```diff
@@ -0,0 +1,178 @@
...
+const (
+	// ExporterSetConditionTypeHealthy indicates all desired replicas are ready.
+	ExporterSetConditionTypeHealthy ExporterSetConditionType = "Healthy"
+	// ExporterSetConditionTypeScalingLimited indicates scaling is constrained by limits.
+	ExporterSetConditionTypeScalingLimited ExporterSetConditionType = "ScalingLimited"
+)
+
+// +kubebuilder:object:root=true
+// +kubebuilder:subresource:status
+// +kubebuilder:subresource:scale:specpath=.spec.maxReplicas,statuspath=.status.replicas
```
raballew, line 150: shouldnt this point to a dedicated replica field instead of the ceiling? kubectl scale will behave incorrectly here.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,187 @@
...
+func (p *Provisioner) RenderPod(
+	ctx context.Context,
+	exporterSet *virtualtargetv1alpha1.ExporterSet,
+	vtc *virtualtargetv1alpha1.VirtualTargetClass,
+	mergedParameters map[string]interface{},
+) (*corev1.Pod, error) {
+	restartAlways := corev1.ContainerRestartPolicyAlways
+	sizeLimit := resource.MustParse(sharedVolumeSizeLimit)
+
+	pod := &corev1.Pod{
```
raballew, line 104: no securitycontext, no allowprivilege escalation flase, no cap drop, no readonlyrootfilesystem, ...

controller/api/virtualtarget/v1alpha1/exporterset_types.go
raballew: also no selectorpath is being used which will be a problem if we ever want to use HPA.

controller/Containerfile.qemu-runtime
```diff
@@ -0,0 +1,19 @@
+# QEMU runtime container for ExporterSet virtual targets.
+# Provides QEMU tools for running virtual machines as part of
+# the sidecar pattern (paired with the jumpstarter exporter).
+#
+# Uses Fedora as the base image because QEMU packages are not
+# available in UBI-minimal repositories.
+FROM registry.fedoraproject.org/fedora:44
```
raballew: there are fedora-minimal images too

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -0,0 +1,187 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: fmt.Sprintf("%s-", exporterSet.Name),
+			Namespace:    exporterSet.Namespace,
+			Labels:       maps.Clone(exporterSet.Spec.Template.Metadata.Labels),
```
raballew, line 108: returns nil if template metadata is empty. initialize an empty map when nil.

controller/api/virtualtarget/v1alpha1/virtualtargetclass_types.go
```diff
@@ -0,0 +1,140 @@
...
+	// +optional
+	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
+
+	// Tolerations are tolerations for the rendered Pods.
+	// +optional
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
+
+	// Resources defines resource requirements for the rendered Pods.
+	// +optional
+	Resources *corev1.ResourceRequirements `json:"resources,omitempty"`
```
raballew, line 57: Nodeselector and tolerations are pod level sched concerns, resources is container level. in qemu provisionier, resources is applied to containers[0] (the qemu runtime) - the sidecar get no resource config. a user setting scheduling resources to control resource alloc wont realize they are only ocnfiguring one of two containers. i think this either needs to move out of schedulingspec into a container specific field or has to become podoverrides (or similar term) to signal that it governs both sched and container attributes.

raballew (changes requested):

controller/api/virtualtarget/v1alpha1/exporterset_types.go
mangelajo: This can be resolved later, we are not going to use HPA (which has no concept of leases), just our own logic. We can either use the owner reference or later down the road if we need to add a selector we will add it.

controller/api/virtualtarget/v1alpha1/virtualtargetclass_types.go
mangelajo: We will resolve those details if necessary at implementation time of the qemu provider. I think it should be fine as it is.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: labels can be nil as well, I think it's ok. Will resolve this later if necessary.

controller/Containerfile.qemu-runtime
mangelajo: Probably a good idea, let me try.

mangelajo: The change is not massive but we save 40MB at least.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: This is to be resolved when we really implement the provisioner, this is only a scaffold

raballew (approved):

bennyz (approved):

kirkbrauer (approved):

mangelajo:
Thank you!! :)

evakhoni (approved):
nice and polished. I had a few findings here but Paul was quicker to submit them :) the only suggestion from my side may be the amount of changes here as a monolithic PR may been worth splitting in two, but that's just imo :)

PR #631, opened 2026-04-24
JEP-0013: Metrics, Tracing, and Log Observability

mangelajo:
## Summary

- Adds JEP-0013 proposing an optional, cross-component observability model for Jumpstarter.
- Covers lease context metadata, structured operational events (annotated logs), exporter/driver metrics, and standardized logging.
- Targets direct integration with Prometheus (scrape), Loki (log aggregation), and Perses (dashboards) without mandating OpenTelemetry.
- Introduces an optional in-cluster Jumpstarter Telemetry service so edge processes never need Loki or cluster-scrape credentials.
- Includes 10 design decisions (DD-1 through DD-10) covering metadata storage, event format, metrics path, log format, credential isolation, OTel stance, Telemetry deployment, HA, idempotency, and dashboarding.
image

## Test plan

- [x] Review JEP content for technical accuracy and completeness
- [x] Verify all DD cross-references resolve correctly
- [x] Confirm Mermaid diagrams render in GitHub markdown preview
- [x] Community discussion on design decisions

Made with [Cursor](https://cursor.com)

PR #865, opened 2026-07-08
feat: structured logging and lease context for observability (JEP-0013 Phase 1)

bkhizgiy:
## Summary

Implements Phase 1 of JEP-0013: Metrics, Tracing, and Log Observability, structured JSON logging with correlation fields and the `spec.context` lease metadata mechanism.

This is the foundation for cross-component log correlation and future metrics/telemetry phases. Operators and CI maintainers can now correlate logs across components filtering by pipeline, build ID, driver operation, or failure type from a single log query.

### Lease CRD: `spec.context` field

- New `map<string, string>` field for attaching run metadata (build ID, pipeline name, image digest, etc.) to a lease
- Validated at admission: max 8 entries, 32-char keys, 64-char values
- Immutable once set — context cannot be changed mid-lease
- Backward compatible: existing leases without `context` behave as before
- The controller propagates context to the exporter via `StatusResponse`, which binds it into all subsequent log lines

### Proto: `LogStreamResponse` enrichment

- Added optional `driver_type` (field 5), `operation` (field 6), `timestamp` (field 7), and `structured_fields` (field 8)
- Clients consuming the log stream get richer data without parsing message text
- Backward compatible: older clients ignore unknown fields, older exporters don't set them

### Proto: `StatusResponse.context`

- Added `map<string, string> context` (field 4) so the controller can propagate lease context to the exporter over the existing `Status` stream

### Proto: `AuditStream` removal

- Removed the unused `AuditStream` RPC and `AuditStreamRequest` message from `ControllerService`

### Python: structured JSON logging

- Exporter processes emit JSON logs with correlation fields when running as a service or with `--log-format json`
- Every log line includes: `lease_id`, `exporter`, `client`, `component`, `namespace`, and all `spec.context` entries
- Driver calls automatically emit `driver_type`, `operation`, `result`, and `error_type` (on failures)
- `LogHandler` enriches proto messages with first-class fields for the LogStream
- Human-readable text output remains the default for interactive use

### Python: `driver_type` on all drivers

- Every driver class now declares its category explicitly
- Categories: `power`, `storage`, `network`, `serial`, `console`, `video`, `composite`, `automotive`, `shell`, `gpio`, `bluetooth`, `debug`, `testing`

### CLI

- `jmp create lease --context key=value`: attach context metadata when creating a lease (repeatable, validated against CRD limits)
- `jmp --log-format json run`: emit structured JSON logs from the exporter

### Controller

- `lease_controller.go`: binds `spec.context` entries to the controller logger during reconciliation
- `controller_service.go`: populates `StatusResponse.context` from the active lease
- `cmd/main.go`, `cmd/router/main.go`: structured JSON logging setup

### CI pipeline correlation

```bash
# CI attaches its build metadata to the lease
jmp create lease -l board=rpi4 --duration 1h \
  --context build_id=autosd-nightly \
  --context pipeline=jmp-ci \
  --context image_digest=sha256:abc
```

### Structured log queries

```bash
# Run exporter with JSON output
JMP_GRPC_INSECURE=1 jmp --log-format json run --exporter-config config.yaml

# Output:
{"pipeline":"ci-main","build_id":"autosd-nightly","lease_id":"test-lease",
 "client":"ci-bot","driver_type":"power","operation":"power_on",
 "result":"success","level":"info","msg":"Operation completed"}
```

### Driver operation tracking

```bash
# When a client calls a driver method, the exporter logs:
# 1. "Operation started"  with driver_type + operation
# 2. "Operation completed" with result=success
#    or "Operation failed" with result=failure + error_type
```

### Go (controller)

- `lease_helpers_test.go`: `LeaseFromProtobuf`/`ToProtobuf` round-trip with context (nil and populated)
- `controller_service_test.go`: `StatusResponse` includes lease context, handles empty context and not-leased cases

### Python

- `exporter_test.py`: context propagation from `StatusResponse` → structlog context binding
- `logging_test.py`: `LogHandler` enrichment with `driver_type`, `operation`, `structured_fields`
- `session_test.py`: `LogStream` returns enriched messages to clients
- `grpc_test.py`: `CreateLease` passes context and `Lease.from_protobuf` deserializes it
- `create_test.py`: CLI `--context` validation (format, key length, value length, entry count)

### E2E (Go)

- Creates a lease with `--context build_id=nightly-42`, verifies it persists on the Kubernetes CRD via `kubectl get`

## Backward Compatibility

- `spec.context` defaults to empty — existing Leases behave exactly as before
- `LogStreamResponse` new fields are optional — older clients/exporters unaffected
- `StatusResponse.context` is empty when not leased — no change for idle exporters
- JSON logging only activates with `--log-format json` — default CLI output unchanged
- `driver_type` defaults to `"other"` in the base class — third-party drivers still work without changes

## References

- [JEP-0013: Metrics, Tracing, and Log Observability](https://github.com/jumpstarter-dev/jumpstarter/pull/631)


controller/deploy/operator/config/crd/bases/jumpstarter.dev_clients.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.18.0
+    controller-gen.kubebuilder.io/version: v0.16.3
```
mangelajo: you need to upgrade your kubebuilder :D

e2e/test/e2e_test.go
```diff
@@ -398,6 +398,45 @@ var _ = Describe("Core E2E Tests", Label("core"), Ordered, func() {
...
+				ns = "default"
+			}
+
+			kubectlOut := MustKubectl("get", "lease.jumpstarter.dev", leaseName,
+				"-n", ns, "-o", "jsonpath={.spec.context}")
+			Expect(kubectlOut).To(ContainSubstring("build_id"))
+			Expect(kubectlOut).To(ContainSubstring("nightly-42"))
+
+			MustJmp("delete", "leases", "--all")
+		})
```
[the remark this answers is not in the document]
bkhizgiy: fixed.

python/docs/source/reference/log-schema.json
```diff
@@ -0,0 +1,69 @@
...
+    "component": {
+      "type": "string",
+      "enum": ["controller", "router", "exporter", "cli", "telemetry"],
+      "description": "Service component that emitted this log line."
+    },
+    "logger": {
+      "type": "string",
+      "description": "Python logger name (e.g. 'jumpstarter.exporter.session'). Present in Python services only."
+    },
+    "lease_id": {
```
kirkbrauer, line 32: `lease_id` seems inconsistent with the following names: `exporter`, `client`, etc.

controller/deploy/operator/config/crd/bases/jumpstarter.dev_leases.yaml
```diff
@@ -77,6 +77,14 @@ spec:
                     type: string
                 type: object
                 x-kubernetes-map-type: atomic
+              context:
```
kirkbrauer, line 80: Could we use labels here, I guess those are not immutable? One thing to think about is provenance attestations and how we might be able to use signature to trace leases, tests, etc. which is a requirement for ISO 26262-8 I think.

python/packages/jumpstarter-driver-androidemulator/jumpstarter_driver_androidemulator/driver.py
```diff
@@ -26,6 +26,8 @@ class AndroidEmulator(Driver):
     - ``power``: Emulator lifecycle (on/off)
     """
 
+    driver_type = "composite"
```
kirkbrauer, line 29: Humm @mangelajo I wonder if this should be a static method or maybe a decorator rather than a "magic" field?

kirkbrauer (commented):
Here are some preliminary comments @bkhizgiy @mangelajo

python/docs/source/reference/log-schema.json
bkhizgiy: yeah I went with lease_id to follow observability conventions, correlation identifiers typically use the `_id` suffix (trace_id, span_id) since It's the key we'll grep for when tracing a session across services. exporter and client are entity names not correlation keys so the naming differs intentionally. but with that said I'm fine to rename to lease if you prefer for consistency.

python/packages/jumpstarter-driver-androidemulator/jumpstarter_driver_androidemulator/driver.py
bkhizgiy: yep I'm not thrilled with it either honestly. I looked at a couple alternatives and this ended up being the least bad option IMO... was thinking of:
- A central registry (drifts from actual drivers so it's another thing to keep in sync, easy to forget).
- Deriving from package/entry-point names (breaks when one package has multiple types or the name isn't consistent with the type).
Happy to switch to a decorator if you think that reads better, or open to other ideas!

controller/deploy/operator/config/crd/bases/jumpstarter.dev_leases.yaml
bkhizgiy: yeah the main issue as you mentioned, is labels are mutable and spec.context is immutable so it can't be tampered with after creation.
On provenance/attestation, totally agree. I see spec.context as the data layer (which build, which pipeline), but cryptographic traceability would need something on top I guess. we can talk a bit more about it and maybe post in a follow-up PR?

controller/api/v1alpha1/lease_types.go
```diff
@@ -23,6 +23,8 @@ import (
 
 // +kubebuilder:validation:XValidation:rule="((has(self.selector.matchLabels) && size(self.selector.matchLabels) > 0) || (has(self.selector.matchExpressions) && size(self.selector.matchExpressions) > 0)) || (has(self.exporterRef) && has(self.exporterRef.name) && size(self.exporterRef.name) > 0)",message="one of selector or exporterRef.name is required"
 // +kubebuilder:validation:XValidation:rule="!has(oldSelf.tags) || self.tags == oldSelf.tags",message="tags are immutable after creation"
+// +kubebuilder:validation:XValidation:rule="!has(oldSelf.context) || self.context == oldSelf.context",message="context is immutable after creation"
```
mangelajo, line 26: nice!

controller/cmd/main.go
```diff
@@ -121,13 +121,11 @@ func main() {
 		"If set the metrics endpoint is served securely")
 	flag.BoolVar(&enableHTTP2, "enable-http2", false,
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
-	opts := zap.Options{
-		Development: true,
-	}
+	opts := zap.Options{}
```
mangelajo, line 124: 👍

controller/deploy/operator/config/crd/bases/jumpstarter.dev_leases.yaml
mangelajo: Yes, that's content for a separate JEP even :)

python/docs/source/reference/log-schema.json
mangelajo: although in this case... it really depends if we are using the lease name, or the lease UUID. I believe it was the plan to use the name which is easier to trace, in this case I would keep just "lease" because it's not always an UUID necessarily. WDYT?

python/packages/jumpstarter-driver-androidemulator/jumpstarter_driver_androidemulator/driver.py
mangelajo: I would use constants defined somewhere, or even better (if that works) I would create multiple subclasses of Driver in jumpstarter.driver that overrides this field, perhaps expose it as a readable attribute function or anything similar to avoid writing. WDYT?

mangelajo:
**Minor note on `structured_fields` size limits (JEP-0013 compliance)**

The JEP specifies that `LogStreamResponse.structured_fields` should follow the same size limits as `LogEntry.extra_fields`: 16 entries max, 64-char keys, 256-char values. Currently `LogHandler.prepare()` (`exporter/logging.py`) collects extra record attributes into `structured_fields` without enforcing these caps.

This is low severity for Phase 1 since the limits matter more on the Telemetry ingest path (Phase 3, where `TelemetryService` validates `LogEntry.extra_fields` server-side). But it might be worth adding a TODO or tracking issue so it doesn't get lost when Phase 3 lands — or alternatively, adding a simple truncation/cap in `LogHandler.prepare()` now to be safe against unexpectedly large payloads on the LogStream.

mangelajo (changes requested):
some coments in a similar line to what kirk said, but excellent work :D

python/docs/source/reference/log-schema.json
kirkbrauer: I think maybe just `lease` like you mentioned might be best here since that is easier to trace.

python/packages/jumpstarter-driver-androidemulator/jumpstarter_driver_androidemulator/driver.py
kirkbrauer: Humm, sub-classing could be confusing as well, esp when mixed with "magic" fields. I think using a readable attribute function is probably the best way to handle this. I'm kind of biased towards decorators since those translate nicely across languages (if we introduce multi-language support in the future).

python/docs/source/reference/log-schema.json
bkhizgiy: Done, changed to lease.

python/packages/jumpstarter-driver-androidemulator/jumpstarter_driver_androidemulator/driver.py
bkhizgiy: Done, switched to a @driver_type(DriverType.COMPOSITE) decorator with constants defined in jumpstarter/driver/types.py. All driver types use it (POWER, STORAGE, NETWORK, SERIAL, AUTOMOTIVE, etc.). Interface classes that pass the type through inheritance use driver_type = DriverType.POWER directly.

bkhizgiy:
> The JEP specifies that `LogStreamResponse.structured_fields` should follow the same size limits as `LogEntry.extra_fields`: 16 entries max, 64-char keys, 256-char values. Currently `LogHandler.prepare()` (`exporter/logging.py`) collects extra record attributes into `structured_fields` without enforcing these caps.
> 
> This is low severity for Phase 1 since the limits matter more on the Telemetry ingest path (Phase 3, where `TelemetryService` validates `LogEntry.extra_fields` server-side). But it might be worth adding a TODO or tracking issue so it doesn't get lost when Phase 3 lands — or alternatively, adding a simple truncation/cap in `LogHandler.prepare()` now to be safe against unexpectedly large payloads on the LogStream.

good catch, done. LogHandler.prepare() truncates keys to 64 chars, values to 256 chars, and stops at 16 entries. Keeps us aligned with the JEP spec and safe against oversized payloads.

bkhizgiy:
@mangelajo @kirkbrauer could you please take another look when you have a moment, thanks!

controller/deploy/operator/config/crd/bases/jumpstarter.dev_leases.yaml
mangelajo: The reason why we use context here is that labels are used today to store the selector that was used for the lease as well among other things (ended lease). We could propagate the context to labels as well like context-$(key)=$(value) or similar, although the key space would be limited by the extra "context-" ... we could also copy as-is to labels when no collision exists to make them indexable.

mangelajo (approved):

mangelajo:
I am rebasing and fixing the conflict to merge it.

mangelajo:
Removed some unrelated files that sneaked in during rebase now it should be good.

python/packages/jumpstarter/jumpstarter/driver/types.py
```diff
@@ -0,0 +1,41 @@
...
+    TESTING = "testing"
+    OTHER = "other"
+
+
+def driver_type(type_name: str):
+    """Decorator to set the driver_type for observability categorization.
+
+    Usage::
+
+        @driver_type(DriverType.POWER)
```
kirkbrauer: I think this looks good, we might also want to think about this for interfaces!

python/packages/jumpstarter-driver-network/jumpstarter_driver_network/driver.py
```diff
@@ -75,6 +75,8 @@ def _format_address(host: str, port: int) -> str:
 
 
 class NetworkInterface(metaclass=ABCMeta):
+    driver_type = DriverType.NETWORK
```
kirkbrauer: Ah, nice!

kirkbrauer (approved):

mangelajo:
@bkhizgiy this is failing the MacOS tests on the merge queue

Error: Refusing to load formula renode/tap/renode-nightly from untrusted tap renode/tap.
Run `brew trust --formula renode/tap/renode-nightly` or `brew trust renode/tap` to trust it.
Error: Process completed with exit code 1.

We need to fix that on a separate pr.

mangelajo:
@bkhizgiy once this is merged it should work: https://github.com/jumpstarter-dev/jumpstarter/pull/906

bkhizgiy:
> @bkhizgiy once this is merged it should work: #906
thanks!

PR #867, opened 2026-07-09
feat: add --insecure-tls flags to config creation CLI

bennyz:
Currently --insecure-fls is missing from CLI commands such as:
```
  jmp config client create
```

Requiring manual editing of the client or usage of JMP_GRPC_INSECURE=1

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/config_client.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/config_client.py
@@ -7,6 +7,9 @@ from jumpstarter_cli_common.opt import (
     OutputMode,
     OutputType,
     PathOutputType,
+    confirm_insecure_tls,
+    opt_insecure_tls,
+    opt_nointeractive,
     opt_output_all,
     opt_output_path_only,
 )
@@ -14,6 +17,7 @@ from jumpstarter_cli_common.print import model_print
 
 from jumpstarter.config.client import ClientConfigV1Alpha1, ClientConfigV1Alpha1Drivers
 from jumpstarter.config.common import ObjectMeta
+from jumpstarter.config.tls import TLSConfigV1Alpha1
 from jumpstarter.config.user import UserConfigV1Alpha1
 
 
@@ -67,6 +71,8 @@ def config_client():
     default="",
 )
 @click.option("--unsafe", is_flag=True, help="Should all driver client packages be allowed to load (UNSAFE!).")
+@opt_insecure_tls
+@opt_nointeractive
 @opt_output_path_only
 @handle_exceptions
 def create_client_config(
@@ -77,6 +83,8 @@ def create_client_config(
     token: str,
     allow: str,
     unsafe: bool,
+    insecure_tls: bool,
+    nointeractive: bool,
     out: Optional[PathLike],
     output: PathOutputType,
 ):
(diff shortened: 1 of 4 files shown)


python/packages/jumpstarter-cli/jumpstarter_cli/config_client.py
```diff
@@ -90,6 +94,7 @@ def create_client_config(
         endpoint=endpoint,
         token=token,
         drivers=ClientConfigV1Alpha1Drivers(allow=allow.split(","), unsafe=unsafe),
+        tls=TLSConfigV1Alpha1(insecure=insecure_tls),
```
bkhizgiy, line 103: I see the login, create, and import commands all call confirm_insecure_tls(insecure_tls, nointeractive) to warn the user before proceeding in insecure mode. These two config create commands (here and in the exporter) skip that confirmation, so a user passing --insecure-tls gets no warning here, while other commands prompt them.

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
```diff
@@ -49,6 +52,7 @@ def create_exporter_config(alias, namespace, name, endpoint, token, output: Path
         metadata=ObjectMeta(namespace=namespace, name=name),
         endpoint=endpoint,
         token=token,
+        tls=TLSConfigV1Alpha1(insecure=insecure_tls),
```
bkhizgiy, line 62: same here

mangelajo (approved):

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter_test.py
```diff
@@ -97,19 +96,47 @@ def test_edit_passes_string_filename_to_click_edit():
...
+        ["create", "myexporter"],
+        input="default\nmyexporter\njumpstarter.my-lab.com:1443\ntest-token\n",
+    )
+    assert result.exit_code == 0, result.output
+
+    config = ExporterConfigV1Alpha1.load("myexporter")
+    assert config.tls.insecure is False
+
+
+def test_create_exporter_insecure_tls(monkeypatch: pytest.MonkeyPatch, tmp_path: Path):
```
bkhizgiy, line 129: The client tests cover both --insecure-tls and -k, but the exporter tests only cover --insecure-tls, should we add the -k here as well for consistency?

python/packages/jumpstarter-cli/jumpstarter_cli/config_client.py
bennyz: Good catch — aligned with login/admin. Both `config client create` and `config exporter create` now call `confirm_insecure_tls` (with `--nointeractive` to skip the prompt in scripts/CI).

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter.py
bennyz: Fixed here too — same `confirm_insecure_tls` + `--nointeractive` pattern as the client create path.

python/packages/jumpstarter-cli/jumpstarter_cli/config_exporter_test.py
bennyz: Agreed — added `test_create_exporter_insecure_tls_short_flag` covering `-k` for parity with the client tests.

bkhizgiy (approved):

PR #871, opened 2026-07-09
fix: transition exporter to AVAILABLE after afterLease hook failure

### LINKED ISSUE #870: Exporter stuck in AfterLeaseHookFailed status after hook failure with endLease
## Description

When an afterLease hook fails with `on_failure: endLease` or encounters an unexpected error, the exporter gets permanently stuck in `AfterLeaseHookFailed` status and cannot accept new leases.

## Reproduction

1. Configure an exporter with an afterLease hook that fails with `on_failure: endLease`:
```yaml
hooks:
  afterLease:
    script: "exit 1"
    timeout: 10
    onFailure: endLease
```

2. Create a lease and wait for it to complete
3. The afterLease hook fails (expected)
4. Check exporter status:
```bash
kubectl get exporters.jumpstarter.dev -n <namespace>
```

**Expected**: Exporter transitions `AfterLeaseHookFailed` → `Available`

**Actual**: Exporter stays in `AfterLeaseHookFailed` permanently. Subsequent lease attempts show "none are ready (still cleaning up previous lease)"

## Root Cause

The `run_after_lease_hook()` finally block calls `request_lease_release()`, which sends `AVAILABLE` status to the controller via gRPC but doesn't update the exporter's internal `_exporter_status` field. The serve() loop clears `_lease_context` correctly, but the stale internal status prevents new lease assignments.

The happy path and `on_failure: warn` paths work correctly because they call `report_status(AVAILABLE, ...)` in the try block. The `on_failure: endLease` and unexpected error paths only report `AFTER_LEASE_HOOK_FAILED` with no subsequent transition.

## Impact

- Exporters become unusable after first hook failure with `endLease`
- Requires exporter restart to recover
- `on_failure: exit` is not affected (correctly shuts down)
- `on_failure: warn` is not affected (reports AVAILABLE in try block)

evakhoni:
Fixes #870

## Problem

When an afterLease hook fails with `on_failure: endLease` or encounters an unexpected error, the exporter gets permanently stuck in `AfterLeaseHookFailed` status and cannot accept new leases.

**Root cause**: The `run_after_lease_hook()` finally block calls `request_lease_release()`, which sends `AVAILABLE` status to the controller via gRPC but doesn't update the exporter's internal `_exporter_status` field.

## Solution

Add `await report_status(ExporterStatus.AVAILABLE, ...)` in the finally block before calling `request_lease_release()`. This ensures both the internal state and controller are updated.

- Idempotent for happy/warn paths that already reported AVAILABLE in the try block
- Fixes endLease and unexpected-error paths that only reported AFTER_LEASE_HOOK_FAILED

## Testing

- **Unit tests**: 2 new test cases verify status transitions for endLease and unexpected errors
- **All tests pass**: 562 tests
- **Manual verification**: Exporter now transitions `AfterLeaseHookFailed` → `Available` and accepts subsequent leases

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/hooks.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/hooks.py
@@ -825,9 +825,8 @@ class HookExecutor:
                 shutdown_called = True
             else:
                 # on_failure='endLease' - report failure to the client, then release the lease.
-                # AFTER_LEASE_HOOK_FAILED is a transient status: the client sees the failure,
-                # the lease is released in the finally block, and the exporter's main loop
-                # clears the lease context and accepts new leases.
+                # AFTER_LEASE_HOOK_FAILED is transient: the finally block transitions to AVAILABLE
+                # and releases the lease, allowing the exporter to accept new leases.
                 logger.error("afterLease hook failed with on_failure='endLease': %s", e)
                 await report_status(
                     ExporterStatus.AFTER_LEASE_HOOK_FAILED,
@@ -836,8 +835,8 @@ class HookExecutor:
 
         except Exception as e:
             # Unexpected errors: report failure but do not shut down.
-            # Same transient status - the lease is released and the exporter
-            # accepts new leases after the finally block completes.
+            # AFTER_LEASE_HOOK_FAILED is transient: the finally block transitions to AVAILABLE
+            # and releases the lease, allowing the exporter to accept new leases.
             logger.error("afterLease hook failed with unexpected error: %s", e, exc_info=True)
             await report_status(
                 ExporterStatus.AFTER_LEASE_HOOK_FAILED,
@@ -851,6 +850,9 @@ class HookExecutor:
             # Don't release lease when exporter is shutting down - unregistration handles cleanup.
             # Releasing here would report AVAILABLE to the controller right before shutdown.
             if request_lease_release and not shutdown_called:
+                # Transition to AVAILABLE to clear AFTER_LEASE_HOOK_FAILED.
+                # Idempotent if already AVAILABLE from the happy/warn paths.
+                await report_status(ExporterStatus.AVAILABLE, "Available for new lease")
                 try:
                     await request_lease_release()
                 except Exception as e:
--- a/python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/hooks_test.py
@@ -1534,6 +1534,104 @@ class TestHookExecutorPRRegressions:
             f"Expected AVAILABLE message to start with '{HOOK_WARNING_PREFIX}', got: '{msg}'"
         )
 
+    async def test_after_hook_endlease_reports_failed_then_available(self, lease_scope) -> None:
+        """afterLease hook failure with on_failure=endLease must transition to AVAILABLE.
+
+        When afterLease hook fails with on_failure=endLease:
+        - AFTER_LEASE_HOOK_FAILED status must be reported
+        - AVAILABLE status must be reported in the finally block
+        - request_lease_release must be called (not shutdown)
+
+        This ensures the exporter can accept new leases after the failure.
+        """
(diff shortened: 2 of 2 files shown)


mangelajo:
I am not sure if this was intended behavior

if afterlease is failing, we may want the admin to look at it?
 @kirkbrauer

kirkbrauer:
@mangelajo I'm not sure we should really allow the `endLease` mode for `afterLease` hooks because it kind of is a no-op since the lease would end anyways after the lease. We should only force the admin to take a look at this failure in the case that we take the exporter offline due to a failed hook with `onFailure = exit`.

mangelajo:
> @mangelajo I'm not sure we should really allow the `endLease` mode for `afterLease` hooks because it kind of is a no-op since the lease would end anyways after the lease. We should only force the admin to take a look at this failure in the case that we take the exporter offline due to a failed hook with `onFailure = exit`.

May be this is something we can combine now with the "enabled" flag of the exporters, we could just flip the flag off and add a "disableExporter" mode or similar..

evakhoni:
interesting points. so basically what you @kirkbrauer saying is for being a no-op in this case, this should not even exists as an option? it's a good point, however I'm not sure if we should put a hard restriction on it, or just document it for being a no-op? after al it works as described, even though it's functionally useless, so it's a question of should it be left for the user judgement for not to use it, or restricted it while introducing an inconsistency with `beforeLease` to achieve that.

@mangelajo , seems like you suggesting to repurpose it into a different functionality :thinking: I'm not sure do we even benefit from keeping an exporter online but disabled for an admin to interact with, while we already have onFailure exit for that? do the admin even need that? if yes, do we also needs the other mode when it actually exits the process, or we can just set onfailure exit to always set offline or just disabled but keeping the process alive for debug? or if we do decide that we need two different modes, then what about `beforeLease` endlease should it act similarly?

python/packages/jumpstarter/jumpstarter/exporter/hooks.py
```diff
@@ -851,6 +850,9 @@ async def run_after_lease_hook(
             # Don't release lease when exporter is shutting down - unregistration handles cleanup.
             # Releasing here would report AVAILABLE to the controller right before shutdown.
             if request_lease_release and not shutdown_called:
+                # Transition to AVAILABLE to clear AFTER_LEASE_HOOK_FAILED.
+                # Idempotent if already AVAILABLE from the happy/warn paths.
+                await report_status(ExporterStatus.AVAILABLE, "Available for new lease")
```
mangelajo, line 855: sorry to be so late back to this PR.

I have another question. Why does not the exporter return "AVAILABLE" once it's shutdown/restarted, which would be a more natural place?, why do we need to do it here?

I feels to me like there is some bug in the sequence that makes the exporter exit and register the ready status once it's ready again.

Here we would be declaring ourselves ready even though we have not gone through exporter init on the restarted instance.

mangelajo (changes requested):
see the comment on setting available in hooks. Can you investigate that?

mangelajo:
> interesting points. so basically what you @kirkbrauer saying is for being a no-op in this case, this should not even exists as an option? it's a good point, however I'm not sure if we should put a hard restriction on it, or just document it for being a no-op? after al it works as described, even though it's functionally useless, so it's a question of should it be left for the user judgement for not to use it, or restricted it while introducing an inconsistency with `beforeLease` to achieve that.
> 
> @mangelajo , seems like you suggesting to repurpose it into a different functionality 🤔 I'm not sure do we even benefit from keeping an exporter online but disabled for an admin to interact with, while we already have onFailure exit for that? do the admin even need that? if yes, do we also needs the other mode when it actually exits the process, or we can just set onfailure exit to always set offline or just disabled but keeping the process alive for debug? or if we do decide that we need two different modes, then what about `beforeLease` endlease should it act similarly?

This is a fair point we should probably not repurpose it to another behavior, perhaps in the future add a more explicit one if that's what we want, i.e. "disableExporter?" ...

mangelajo:
@evakhoni we are not sure if this is still relevant, will close for now, but feel free to reopen.

PR #873, opened 2026-07-09
fix: improve Dial reliability for slow exporters

evakhoni:
## Problem

When connecting to slow hardware exporters (e.g., NVIDIA Jetson AGX Orin), `jmp shell` frequently failed or required multiple lease attempts before establishing a connection. Three compounding issues were identified:

1. **Spurious connection probe:** Before opening the shell, the client sent a test connection to the Unix socket to verify it was ready. This probe triggered the same connection handler as the real shell connection, spawning a second Dial attempt in the same task group. When the probe's connection was closed (having served its purpose), its handler failed — and because both handlers shared a task group, the failure cancelled the entire group, killing the real connection along with it.

2. **gRPC handshake timeout:** Each connection to the exporter requires a Dial call to the controller, which retries with exponential backoff while the exporter starts its Listen stream. The Dial retries and the gRPC HTTP/2 handshake share the same connection — while Dial is retrying, the gRPC client on the other end is waiting for an HTTP/2 SETTINGS frame. If Dial retries take longer than gRPC's internal handshake timeout (~18-20 seconds), the handshake times out and the connection is dropped. This affects any exporter where the Dial retry phase exceeds the handshake window, though slow exporters make it far more likely.

3. **Insufficient dial timeout:** The default Dial timeout of 30 seconds was too short for exporters that need more than 30 seconds to finish cleanup from a previous session and restart their Listen stream. When the timeout was exceeded, the client released the lease and acquired a new one, which forced the exporter to restart its setup from scratch — creating a churn cycle that made the problem worse with each iteration.

### Probe removal

The `_wait_for_ready_connection` method and its invocation in `serve_unix_async` are removed entirely. The probe was originally added to block until the Unix socket was accepting connections, but this is unnecessary — the socket is created by `TemporaryUnixListener` before `serve_unix_async` yields, so it is always ready by the time the shell connects. The `connect_unix` import that was only used by the probe is also removed.

### Pre-Dial readiness check

The connection handler (`handle_async`) previously did both Dial-with-retry and router stream setup in a single method. This meant the Dial retries and the gRPC handshake were coupled on the same connection, creating the timeout race.

The fix splits these concerns: `serve_unix_async` now calls a new `_dial_with_retry` method before accepting any connections. This readiness check ensures the exporter has its Listen stream active. The Dial response from this check is intentionally discarded — each Unix socket connection still gets its own per-connection Dial (since each connection needs a unique router tunnel), but since the exporter is already confirmed ready, these per-connection Dials succeed immediately (typically under 1 second), well within the gRPC handshake timeout.

`_dial_with_retry` is extracted from the old `handle_async` method. It uses exponential backoff (0.3s base, 2s max) bounded by `dial_timeout`, and raises `ExporterUnreachableError` on timeout instead of the raw `AioRpcError`. This allows the shell's outer retry loop to properly distinguish retryable exporter failures from other gRPC errors.

### Dial timeout increase and CLI flag

The default `dial_timeout` is raised from 30s to 60s to accommodate exporters that need more time to become ready. The field is also exposed for user configuration:

- **CLI flag:** `--dial-timeout` on `jmp shell` (e.g., `--dial-timeout 90s`, `--dial-timeout 2m`)
- **Environment variable:** `JMP_DIAL_TIMEOUT`
- **Config file:** `leases.dial_timeout` (previously excluded from serialization, now persisted)

This matches the existing pattern for `--retry-timeout` / `JMP_RETRY_TIMEOUT` and `--acquisition-timeout`.

## Test plan

- [x] `make pkg-test-jumpstarter` — 560 passed
- [x] `make lint-fix` — all checks passed
- [x] Manual: `jmp shell -n <exporter>` connects reliably with default 60s timeout
- [x] Manual: `jmp shell --dial-timeout 10` triggers retry/release as expected
- [x] Manual: back-to-back connections succeed without lease churn
- [x] Manual: commands work within an active shell session (each gets its own router tunnel)


python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -378,52 +381,45 @@ async def handle_async(self, stream):
...
+                stream,
+                self.tls_config,
+                self.grpc_options,
+            ):
+                pass
+
+        async with TemporaryUnixListener(_tunnel_handler) as path:
             logger.debug("Serving Unix socket at %s", path)
-            await self._wait_for_ready_connection(path)
             yield path
```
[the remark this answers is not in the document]
evakhoni: The per-connection Dial happens after the readiness check has confirmed the exporter is ready. At this point the exporter's Listen stream is active and the controller can route
  immediately. Each Dial should complete in under 1 second. If it fails here, it's not a transient "exporter not ready" situation — it's a genuine failure (exporter went offline,
  lease ended, etc.).

  Adding retries would reintroduce the bug we just fixed. The gRPC HTTP/2 handshake on the other end of the Unix socket has an ~18-20s internal timeout. If we retry Dial with
  backoff inside _tunnel_handler, we eat into that window and risk the same handshake timeout race that was the whole point of this PR.

python/packages/jumpstarter/jumpstarter/config/env.py
```diff
@@ -17,3 +17,4 @@
 JUMPSTARTER_GRPC_INSECURE = "JUMPSTARTER_GRPC_INSECURE"
 JMP_GRPC_PASSPHRASE = "JMP_GRPC_PASSPHRASE"
 JMP_RETRY_TIMEOUT = "JMP_RETRY_TIMEOUT"
+JMP_DIAL_TIMEOUT = "JMP_DIAL_TIMEOUT"
```
mangelajo, line 21: this needs to be updated in the docs/website

evakhoni: ack

evakhoni: ok [done](https://github.com/jumpstarter-dev/jumpstarter/pull/873/changes/c993232d5ab0a75fe7603d24c3af6e0a709b33ef). added this one, and a few other gaps it found there as we already touching that.

on a side note, not sure why it launched e2e-tests and pytest-matrix for a docs-only commit :thinking:

evakhoni:
had a merge conflict, just solved

docs/source/getting-started/configuration/files.md
```diff
@@ -46,6 +46,10 @@ token: "******************" # An authentication token
 drivers:
   allow: ["jumpstarter_drivers_*", "vendorpackage.*"] # Driver packages the client can dynamically load
   unsafe: false # Allow any driver package to load dynamically
+leases:
+  acquisition_timeout: 7200 # Timeout in seconds for lease acquisition (default: 7200)
+  dial_timeout: 60 # Timeout in seconds for connecting to the exporter (default: 60)
+  retry_timeout: 300 # Timeout in seconds for retrying unreachable exporters (default: 300, 0 to disable)
```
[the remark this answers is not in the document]
evakhoni: well, I rephrased a bit. ptal

mangelajo (approved):

mangelajo (changes requested):
hold on, I need to verify something, I remember we were blocking the unix socket until beforeLease hooks end for different reasons.

python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -378,52 +381,45 @@ async def handle_async(self, stream):
...
+                jumpstarter_pb2.DialRequest(lease_name=self.name)
+            )
+            async with connect_router_stream(
+                response.router_endpoint,
+                response.router_token,
+                stream,
+                self.tls_config,
+                self.grpc_options,
+            ):
+                pass
```
mangelajo, line 423: **No error handling in `_tunnel_handler` — letting it crash is correct, but the exception type should be wrapped.**

I traced the full flow and catching the exception here would not help: `client_from_path` creates a **single** gRPC channel (one HTTP/2 connection over the Unix socket → one `_tunnel_handler` → one Dial → one router tunnel). All shell commands are multiplexed over this one tunnel. If this Dial fails, the entire session is broken — catching and logging would just leave a half-dead shell where gRPC's transport-level reconnect might open a new socket connection (new handler, new Dial), but the `DriverClient` stubs still reference driver UUIDs from the original `GetReport` call. The session needs a full teardown and re-initialization.

So letting it crash is the right behavior. However, there's a gap in the exception propagation: the raw `AioRpcError` bubbles up as `BaseExceptionGroup(AioRpcError)`, which doesn't match `ExporterUnreachableError` in the shell's retry handler (shell.py:508-511). It falls through to the outer catch-all at shell.py:551 and becomes `ExporterOfflineError("Connection to exporter lost")` — **the retry loop is never engaged**.

If the intent is for post-readiness-check failures to trigger a reconnect attempt, wrapping the error would fix this:

```python
async def _tunnel_handler(stream):
    try:
        response = await self.controller.Dial(
            jumpstarter_pb2.DialRequest(lease_name=self.name)
        )
    except AioRpcError as e:
        raise ExporterUnreachableError(
            f"Per-connection Dial failed for {self.exporter_name}: {e.details()}"
        ) from e
    async with connect_router_stream(...):
        pass
```

This way the exception propagates, tears down the session (correct), AND engages the retry loop (shell.py:514-527) instead of dying with a generic error.

python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -325,20 +324,20 @@ def __contextmanager__(self) -> Generator[Self]:
...
-        # Uses time-based retry bounded by dial_timeout instead of fixed retry count.
+    async def _dial_with_retry(self):
+        """Dial the controller with exponential backoff, waiting for the exporter to be ready.
+
+        Returns DialResponse on success.
+        Raises ExporterUnreachableError on timeout or unrecoverable error.
+        """
+        logger.debug("Dialing controller for lease %s", self.name)
         base_delay = 0.3
         max_delay = 2.0
```
mangelajo, line 336: **Default mismatch with config:** The `Lease` dataclass still defaults `dial_timeout` to `30.0` (line 95), but `ClientConfigV1Alpha1Lease` now defaults to `60.0`. In practice this is harmless because the config layer always passes an explicit value, but it's misleading — a developer constructing a `Lease` directly (e.g. in tests or automation) would get the old 30s default instead of the documented 60s.

Consider updating the dataclass default to `60.0` for consistency.

python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -378,52 +381,45 @@ async def handle_async(self, stream):
...
-            response.router_endpoint, response.router_token, stream, self.tls_config, self.grpc_options
-        ):
-            pass
+                    raise ExporterUnreachableError(
+                        f"Lease {self.name} transferred to another client"
+                    ) from e
+                logger.warning("Connection to exporter lost: %s", e.details())
+                raise ExporterUnreachableError(
+                    f"Connection to exporter {self.exporter_name} lost: {e.details()}"
+                ) from e
```
mangelajo, line 398: **Behavioral change: permission denied now raises instead of returning silently.** The old code did a `return` (graceful exit). The new code raises `ExporterUnreachableError`, which triggers the shell's retry loop.

For the lease-transferred case, the retry loop (shell.py:514-527) will release the old lease and try to re-acquire. Since `lease_transferred` is set on the old lease object (replaced on the next iteration), the retry effectively gets a fresh start. This seems intentional and correct.

python/packages/jumpstarter/jumpstarter/config/client.py
```diff
@@ -344,6 +344,11 @@ async def lease_async(
                 if retry_timeout is not None
                 else float(os.environ.get(JMP_RETRY_TIMEOUT, self.leases.retry_timeout))
             )
+            dial_timeout_seconds = (  # pragma: no cover
+                dial_timeout.total_seconds()
```
mangelajo, line 354: **Nit: Unnecessary `# pragma: no cover`:**

python/packages/jumpstarter/jumpstarter/client/lease_test.py
```diff
@@ -626,10 +625,124 @@ async def mock_dial(request):
...
+        # Verify per-connection Dial was called
+        assert dial_call_count == 1
+
+        # Verify connect_router_stream was called with correct args
+        assert len(router_stream_calls) == 1
+        endpoint, token, tls_config, grpc_options = router_stream_calls[0]
+        assert endpoint == "test-endpoint"
+        assert token == "test-token"
+        assert tls_config is lease.tls_config
+        assert grpc_options is lease.grpc_options
```
mangelajo, line 838: **Test coverage gap:** No test for what happens when `_tunnel_handler`'s per-connection `controller.Dial` fails after the readiness check succeeds. Worth adding a test to pin down the intended behavior.

docs/source/getting-started/configuration/files.md
```diff
@@ -59,6 +63,16 @@ drivers:
...
+- `JMP_GRPC_PASSPHRASE` - Shared passphrase for authenticating against passphrase-protected exporters
+
+**Shell Session Variables** (automatically set by `jmp shell`):
+
+- `JMP_LEASE` - Active lease name (enables reconnection via `JMP_LEASE=<name> jmp shell`)
+- `JMP_EXPORTER` - Name of the connected exporter
+- `JMP_EXPORTER_LABELS` - Connected exporter's labels as comma-separated `key=value` pairs
 
 **CLI Commands**:
 ```{code-block}  console
```
mangelajo, line 78: Nit: The docs additions for `JMP_OIDC_CALLBACK_PORT`, `JMP_GRPC_PASSPHRASE`, shell session variables, and `JMP_DISABLE_COMPRESSION` are great to have documented but are unrelated to the dial reliability fix. Consider splitting into a separate commit/PR for cleaner scope. Not a blocker.

python/packages/jumpstarter/jumpstarter/client/lease.py
mangelajo: @evakhoni after some back and forth of analysis I found this, what do you think? ^

mangelajo (changes requested):
Some human+AI review, let me know what you think

python/packages/jumpstarter/jumpstarter/config/env.py
evakhoni: Done in 7787ef7 — added JMP_DIAL_TIMEOUT to docs along with other undocumented env vars.

python/packages/jumpstarter/jumpstarter/client/lease.py
evakhoni: Fixed in 7787ef7 — updated Lease.dial_timeout default to 60.0 to match ClientConfigV1Alpha1Lease.

python/packages/jumpstarter/jumpstarter/config/client.py
evakhoni: Fixed in 7787ef7 — removed the pragma to match retry_timeout_seconds above it.

python/packages/jumpstarter/jumpstarter/client/lease_test.py
evakhoni: Fixed in 7787ef7 — added test_serve_unix_async_per_connection_dial_failure_wrapped that verifies AioRpcError is wrapped as ExporterUnreachableError.

python/packages/jumpstarter/jumpstarter/client/lease.py
evakhoni: Fixed in 7787ef7 — wrapped per-connection Dial in try/except to raise ExporterUnreachableError instead of raw AioRpcError. This ensures the retry loop is properly engaged when transient failures occur after the readiness check.

docs/source/getting-started/configuration/files.md
evakhoni: ah just a misc doc addition, not worth a separate pr imo.. just shipped alongside, that's intentional

mangelajo:
@evakhoni it's looking good now, but ran into a conflict, can you have an eye?

python/packages/jumpstarter/jumpstarter/client/lease.py
evakhoni: that's actually a valid concern. i'll check what happens during lease transfer

evakhoni: ok addressed in https://github.com/jumpstarter-dev/jumpstarter/pull/873/changes/51bc183801cea12054537d48e41b1792e2d9fb72

evakhoni:
@mangelajo fixed the conflict, and fixed an issue with lease transfer. CI passed well, so it's all ready now. 
thanks!

bennyz (approved):

docs/source/getting-started/configuration/files.md
mangelajo: sorry this was automated and I thought I had deleted it.

mangelajo (approved):

PR #877, opened 2026-07-10
fix: replace deprecated asyncio event loop patterns

mangelajo:
## Summary

Audited the entire codebase for deprecated asyncio event loop patterns that emit `DeprecationWarning` on Python 3.12/3.13 and will break on Python 3.14. Found and fixed three locations (the SNMP driver had the same issue and was already fixed in #775).

## Compatibility matrix

| Deprecated pattern | 3.12 | 3.13 | 3.14 | Replacement (since 3.7) |
|---|---|---|---|---|
| `get_event_loop()` (no running loop) | DeprecationWarning | DeprecationWarning | **RuntimeError** | `get_running_loop()` |
| `set_event_loop()` | DeprecationWarning | DeprecationWarning | **No-op** | `asyncio.run()` |
| `asyncio.Event()` (no running loop) | DeprecationWarning | DeprecationWarning | **RuntimeError** | Create inside `async def` |
| `ensure_future()` | DeprecationWarning | DeprecationWarning | DeprecationWarning | `create_task()` |

### 1. TFTP driver (HIGH — will crash on 3.14)

`jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py`

`_start_server()` ran in a thread using `new_event_loop()` + `set_event_loop()` + `run_until_complete()`, and constructed `TftpServer` (which creates `asyncio.Event()` objects) before the loop was running.

Replaced with `asyncio.run()` and moved `TftpServer` construction into an async method so `asyncio.Event()` objects are created with a running loop.

### 2. Shell driver (MEDIUM — deprecation warnings on 3.12/3.13)

`jumpstarter-driver-shell/jumpstarter_driver_shell/driver.py`

Replaced `asyncio.get_event_loop().time()` with `asyncio.get_running_loop().time()` (called from async context, so trivial swap).

### 3. mitmproxy example (LOW — deprecated but functional)

`jumpstarter-driver-mitmproxy/examples/addons/data_stream_websocket.py`

Replaced `asyncio.ensure_future()` with `asyncio.create_task()`.

## Audit scope

Searched all Python packages for `get_event_loop`, `new_event_loop`, `set_event_loop`, `run_until_complete`, `ensure_future`, and `asyncio.Event()`/`Queue()`/`Lock()` created outside async contexts. The rest of the codebase already uses the correct modern APIs.

### DIFF
--- a/python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
+++ b/python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
@@ -67,27 +67,31 @@ class Tftp(Driver):
         return "jumpstarter_driver_tftp.client.TftpServerClient"
 
     def _start_server(self):
-        self._loop = asyncio.new_event_loop()
-        asyncio.set_event_loop(self._loop)
+        try:
+            asyncio.run(self._run_server_lifecycle())
+        except Exception as e:
+            self.logger.error(f"Error running TFTP server: {e}")
+        finally:
+            self.logger.info("TFTP server thread completed")
+
+    async def _run_server_lifecycle(self):
+        """Set up and run the TFTP server within a proper async context.
+
+        This ensures asyncio.Event() objects inside TftpServer are created
+        with a running event loop, as required by Python 3.14+.
+        """
+        self._loop = asyncio.get_running_loop()
         self.server = TftpServer(
             host=self.host,
             port=self.port,
             operator=self.children["storage"]._operator,
             logger=self.logger,
         )
+        self._loop_ready.set()
         try:
-            self._loop_ready.set()
-            self._loop.run_until_complete(self._run_server())
-        except Exception as e:
-            self.logger.error(f"Error running TFTP server: {e}")
+            await self._run_server()
         finally:
-            try:
-                self._loop.run_until_complete(self._loop.shutdown_asyncgens())
-                self._loop.close()
-            except Exception as e:
-                self.logger.error(f"Error during event loop cleanup: {e}")
             self._loop = None
-            self.logger.info("TFTP server thread completed")
 
     async def _run_server(self):
         try:
--- a/python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
+++ b/python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
@@ -45,6 +45,7 @@ class Tftp(Driver):
     _shutdown_event: threading.Event = field(init=False, default_factory=threading.Event)
     _loop_ready: threading.Event = field(init=False, default_factory=threading.Event)
     _loop: Optional[asyncio.AbstractEventLoop] = field(init=False, default=None)
+    _startup_error: Optional[BaseException] = field(init=False, default=None)
 
     def __post_init__(self):
         if hasattr(super(), "__post_init__"):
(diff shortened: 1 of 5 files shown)


python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
```diff
@@ -67,27 +67,31 @@ def client(cls) -> str:
...
+        except Exception as e:
+            self.logger.error(f"Error running TFTP server: {e}")
+        finally:
+            self.logger.info("TFTP server thread completed")
+
+    async def _run_server_lifecycle(self):
+        """Set up and run the TFTP server within a proper async context.
+
+        This ensures asyncio.Event() objects inside TftpServer are created
+        with a running event loop, as required by Python 3.14+.
```
mmahut: Dumb question maybe, but is this true? The docs say the loop parameter was removed from asyncio.Event in 3.10 and I thought the loop gets picked up lazily on first await since then, so creating it without a running loop should be OK even on 3.14? https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event

mangelajo: Thanks for the review, reading the docs.

mangelajo: You're right! I verified from the CPython 3.14 source:

- `Event.__init__()` only does `self._waiters = collections.deque()` and `self._value = False` — no loop access
- The `_LoopBoundMixin._get_loop()` (which calls `events._get_running_loop()`) is only invoked lazily inside `Event.wait()` when `self._get_loop().create_future()` is called
- The `loop` parameter was removed in 3.10, and since then `Event()` does lazy loop binding on first `await`

So creating `asyncio.Event()` outside a running loop is fine on 3.14+. The actual motivation for this refactor is replacing the deprecated `new_event_loop()` + `set_event_loop()` + `run_until_complete()` pattern with `asyncio.run()` — `get_event_loop()` raises `RuntimeError` on 3.14 when no loop is running, and the old manual lifecycle code is replaced by `asyncio.run()` which handles `shutdown_asyncgens()` and `loop.close()` automatically.

Updated the docstrings to reflect the correct motivation in the latest commit.

python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
```diff
@@ -67,28 +68,41 @@ def client(cls) -> str:
...
+        Uses asyncio.run() instead of the deprecated new_event_loop() +
+        set_event_loop() + run_until_complete() pattern, which emits
+        DeprecationWarning on Python 3.12/3.13 and breaks on 3.14
+        (get_event_loop() raises RuntimeError when no loop is running).
+        """
+        try:
+            self._loop = asyncio.get_running_loop()
+            self.server = TftpServer(
+                host=self.host,
+                port=self.port,
```
mangelajo, line 90: Addressed CodeRabbit's suggestion: setup code is now wrapped in `try/except/finally` so `_loop_ready.set()` always fires. If `TftpServer()` construction fails, the error is captured in `_startup_error` and `start()` re-raises it as a `TftpError` with the real cause chained — no more waiting 5s for a generic timeout.

Added `test_tftp_start_surfaces_startup_error` to cover this path.

python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
```diff
@@ -67,28 +68,41 @@ def client(cls) -> str:
...
-            self._loop = None
             self.logger.info("TFTP server thread completed")
 
+    async def _run_server_lifecycle(self):
+        """Set up and run the TFTP server within a proper async context.
+
+        Uses asyncio.run() instead of the deprecated new_event_loop() +
+        set_event_loop() + run_until_complete() pattern, which emits
+        DeprecationWarning on Python 3.12/3.13 and breaks on 3.14
+        (get_event_loop() raises RuntimeError when no loop is running).
```
mangelajo, line 84: Addressed mmahut's review: the previous docstring incorrectly claimed `asyncio.Event()` needs a running loop on Python 3.14+. Verified from CPython 3.14 source that `Event.__init__()` does no loop access — the loop is bound lazily via `_LoopBoundMixin._get_loop()` on first `await`.

Updated the docstring to state the correct motivation: replacing the deprecated `new_event_loop()` + `set_event_loop()` + `run_until_complete()` pattern with `asyncio.run()`.

mangelajo (commented):
Addressed all review comments:

1. **CodeRabbit — startup error surfacing**: Setup code in `_run_server_lifecycle` is now wrapped in `try/except/finally` so `_loop_ready.set()` always fires. Errors are captured in `_startup_error` and `start()` re-raises them as `TftpError` with the real cause chained. Added `test_tftp_start_surfaces_startup_error` to cover this.

2. **mmahut — asyncio.Event() docstring correction**: Verified from CPython 3.14 source that `Event.__init__()` does no loop access (lazy binding via `_LoopBoundMixin`). Updated docstrings to reflect the correct motivation: replacing deprecated `new_event_loop()` + `set_event_loop()` + `run_until_complete()` patterns.

3. **Rebased** onto current `main` (was 17 commits behind).

All 25 TFTP tests pass, lint clean.

python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
mangelajo: fixed ':D

python/packages/jumpstarter-driver-tftp/jumpstarter_driver_tftp/driver.py
mmahut: Thank you for confirming, this was just a nitpick comment really!

mmahut (approved):

bennyz (approved):

PR #881, opened 2026-07-10
feat: motd for exporter shells

### LINKED ISSUE #51: Provide a motd message when users get a shell into a exporter
The motd should contain any relevant documentation the administrator seems valuable for the users, as well as any extra details the hooks want to put into motd (like previous reports for the exporters, etc...)

> jmp shell -l board=ti-123

You have been assigned to exporter ti-123-001:

You can use images from 

upstream: https://.....
downstream: https://....

You can flash the device with:

j storage flash ......

and access the serial console with 

j serial start-console

$ 

This is related to jumpstarter-dev/jumpstarter#65 and could take some of the output of the hooks to add it to the motd message.

Also reports from jumpstarter-dev/jumpstarter#52 could be listed in the motd

mmahut:
Print an admin-configured message when entering a shell on an exporter.

- `motd:` field in the exporter config, returned to the client via GetReport
- shown before interactive shells and skipped for `-- <cmd>`
- _beforeLease_ hooks can append dynamic content via _$JMP_MOTD_FILE_

Closes #51


python/packages/jumpstarter/jumpstarter/client/base.py
```diff
@@ -40,6 +40,9 @@ class DriverClient(AsyncDriverClient):
     methods_description: dict[str, str] = field(default_factory=dict)
     """Map of method names to their help descriptions from GetReport()"""
 
+    motd: str | None = None
```
mangelajo: This is the only part of the PR that doesn't feel correct, can you look for a different way that doesn't involve extending the driver client with motd? (it feels really unrelated to the client driver itself)

python/packages/jumpstarter/jumpstarter/exporter/hooks.py
```diff
@@ -180,9 +185,52 @@ async def _execute_hook(
             hook_env.get("CLIENT_NAME", "NOT_SET"),
         )
 
-        return await self._execute_hook_process(
-            hook_config, lease_scope, log_source, hook_env, lease_scope.session, hook_type
-        )
+        # $JMP_MOTD_FILE is beforeLease-only; clear any inherited value first
```
mangelajo, line 188: nice

mangelajo (changes requested):
Just one comment about motd in DriverClient, other than that looks awesome!, 

Thank you so much

python/packages/jumpstarter/jumpstarter/client/base.py
mmahut: Hmm, you are right, the driver client is probably the wrong place for it. I have rewrote it as a small helper.

mangelajo:
argh, sorry, there is a conflict now, I rebased and fixed the conflict.

mangelajo (approved):

PR #883, opened 2026-07-12
fix: add capped backoff for requeueing resources

bennyz:
Reconciling pending leases every second can strain the system, leading to delays in servicing requests.

Instead, back off exponentially until a cap of 30s

### DIFF
--- a/controller/internal/controller/lease_controller.go
+++ b/controller/internal/controller/lease_controller.go
@@ -36,8 +36,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/log"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
 // LeaseReconciler reconciles a Lease object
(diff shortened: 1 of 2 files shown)


controller/internal/controller/lease_controller.go
```diff
@@ -647,9 +649,112 @@ func skipEndedPredicate() predicate.Funcs {
...
+			return true
+		},
+	}
+}
+
+func (r *LeaseReconciler) mapExporterToLeases(ctx context.Context, obj client.Object) []reconcile.Request {
+	exporter := obj.(*jumpstarterdevv1alpha1.Exporter)
+
+	var leaseList jumpstarterdevv1alpha1.LeaseList
+	if err := r.List(ctx, &leaseList, client.InNamespace(exporter.Namespace)); err != nil {
```
mangelajo: We should list only the non-ended ones using the label filter.

mangelajo: Now that list that we don't cleanup can be considerable size.

mangelajo: I also wonder if we should keep an in-memory list of waiting leases based on what we receive through events instead of querying the server every time, perhaps this has some caching?

bennyz: I think controller-runtime already has caching for this at the framework level, but i'll verify

mangelajo (approved):

controller/internal/controller/lease_controller.go
bennyz: https://github.com/kubernetes-sigs/controller-runtime/blob/71f7db556ca57ce7ea6563f77d739f0d2a54233a/pkg/client/client.go#L82-L88

looks like it is cached by default

added label filtering

PR #889, opened 2026-07-13
fix: force close the socket when cleaning up http thread

### LINKED ISSUE #888: bug: stale port 8080 binding on exporter prevents storage flash after aborted operations
## Summary

After aborted or failed `storage flash` operations, the exporter's flasher service leaves port 8080 bound. Subsequent flash attempts fail with:

```
Error: DriverCall 'start' failed with gRPC UNKNOWN: [Errno 98] error while attempting to bind on address ('10.26.28.116', 8080): [errno 98] address already in use
```

## Reproduction

1. Start a `storage flash` operation via `jmp shell` or MCP
2. Abort it (timeout, lease expiration, client disconnect, or Ctrl+C)
3. Acquire a new lease for the same exporter
4. Attempt `storage flash` again → fails with address already in use

## Environment

- Jumpstarter 0.8.1 (client + drivers)
- Controller: ROSA cluster (`grpc.jumpstarter-lab.apps.rosa.auto-devcluster.bzdx.p3.openshiftapps.com`)
- Affected exporters: R-Car S4 boards (exporter hosts at 10.26.28.116, 10.26.28.118, and others)
- Not all exporters affected — one that had no aborted flash operations works fine

## Observed behavior

- Exporter reports `ONLINE: yes` and `AVAILABLE` in lease status
- Lease acquisition succeeds
- `power read`, `tcp address`, `ssh` all work through the tunnel
- Only `storage flash` fails — the flasher's HTTP/TFTP service can't bind port 8080
- The stale binding persists across lease cycles (release + re-acquire doesn't fix it)
- Multiple independent exporter hosts affected simultaneously after a series of aborted fleet test runs

## Expected behavior

- The flasher service should release port 8080 when the lease ends, the client disconnects, or the flash operation is aborted
- If port 8080 is already bound from a previous session, the exporter should clean it up before starting a new flash
- Alternatively, the exporter's health/availability status should reflect that it cannot flash

## Workaround

Manually kill the stale process holding port 8080 on each affected exporter host:
```bash
ss -tlnp | grep 8080
# Kill the stale process
```

bennyz:
fixes #888

reproducer:
1. Start flashing with the flasher driver, kill/ctrl+c mid flash
2. delete the lease
3. create a shell with a new lease and try to flash again, which will fail on address already in use

### DIFF
--- a/python/packages/jumpstarter-driver-http/jumpstarter_driver_http/driver.py
+++ b/python/packages/jumpstarter-driver-http/jumpstarter_driver_http/driver.py
@@ -137,12 +137,29 @@ class HttpServer(Driver):
         if self.runner:
             try:
                 anyio.from_thread.run(self._async_cleanup)
-            except Exception as e:
-                self.logger.warning(f"HTTP server cleanup failed: {e}")
+            except Exception:
+                self._force_close_sockets()
             finally:
                 self.runner = None
                 self._bound_port = 0
         super().close()
+
+    def _force_close_sockets(self):
+        """Force-close server sockets when async cleanup is unavailable.
+
+        When close() is called from the event loop thread (e.g. during
+        Session teardown), anyio.from_thread.run() cannot dispatch the
+        async cleanup. Closing the underlying transport sockets directly
+        ensures the port is released.
+        """
+        try:
+            if self.runner:
+                for site in self.runner.sites:
+                    if hasattr(site, "_server") and site._server:
+                        site._server.close()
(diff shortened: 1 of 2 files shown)


mangelajo (approved):

mangelajo:
I think it's good, in theory the finally: self.runner = None should have triggered GC, but it doesn't seem to be triggering, then looking at the internals of AppRunner to tap and close the server is quite OK IMHO.

mangelajo:
ahh, backport workflow does not work with own branches.

PR #892, opened 2026-07-16
feat: strip hidden labels from exporter display output

bennyz:
Labels prefixed with hidden.jumpstarter.dev/ are excluded from jmp get exporters. Pass --show-hidden
to reveal them. Filtering happens at the data boundary in list_exporters() so all display paths get consistent behavior.

This is useful when there is data we need but don't want to pollute the label output too much


bennyz:
@mangelajo i am not set on the prefix, we could also use `_` or something

mangelajo:
> @mangelajo i am not set on the prefix, we could also use `_` or something

"_" would be invalid for k8s I believe

I was thinking, what if we added an annotation to exporters to list hidden labels, would be very helpful to deprecate labels. May be even later a deprecated annotation to warn users when they are using a phased out label. i.e.


annotations:
     jumpstarter.dev/hidden-labels=abc,def,ghi
     jumpstarter.dev/deprecated-labels=cde,efg,hij

(but the deprecated is another feature tb...)

Or a better option could be (and that could make more sense) to provide a list in the controller config + operator:

labels:
   hiden:
      - abc
      - def
      - ghi
   deprecated:
     - cde
     - efg
     - hij

WDYT? 

I know you're trying to avoid controller update in this case, but we can backport this and bring it to the lab quickly

bennyz:
> > @mangelajo i am not set on the prefix, we could also use `_` or something
> 
> "_" would be invalid for k8s I believe

it is legal https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

> I was thinking, what if we added an annotation to exporters to list hidden labels, would be very helpful to deprecate labels. May be even later a deprecated annotation to warn users when they are using a phased out label. i.e.
> 
> annotations: jumpstarter.dev/hidden-labels=abc,def,ghi jumpstarter.dev/deprecated-labels=cde,efg,hij
> 
> (but the deprecated is another feature tb...)
> 
> Or a better option could be (and that could make more sense) to provide a list in the controller config + operator:
> 
> labels: hiden: - abc - def - ghi deprecated: - cde - efg - hij
> 
> WDYT?
> 
> I know you're trying to avoid controller update in this case, but we can backport this and bring it to the lab quickly

yeah i was trying to avoid controller changes for something like this, i'll have a look

protocol/proto/jumpstarter/client/v1/client.proto
```diff
@@ -156,6 +156,8 @@ message GetExporterRequest {
     (google.api.field_behavior) = REQUIRED,
     (google.api.resource_reference) = {type: "jumpstarter.dev/Exporter"}
   ];
+  // When true, include labels hidden by controller configuration.
+  bool show_hidden = 2 [(google.api.field_behavior) = OPTIONAL];
```
mangelajo:
```suggestion
  bool show_hidden_labels = 2 [(google.api.field_behavior) = OPTIONAL];
```
Otherwise could be mistaken by show hidden "exporters" which would be my first understanding outside of the context of this pr.

protocol/proto/jumpstarter/client/v1/client.proto
```diff
@@ -171,6 +173,8 @@ message ListExportersRequest {
   string page_token = 3 [(google.api.field_behavior) = OPTIONAL];
   // Filter expression for the results.
   string filter = 4 [(google.api.field_behavior) = OPTIONAL];
+  // When true, include labels hidden by controller configuration.
+  bool show_hidden = 5 [(google.api.field_behavior) = OPTIONAL];
```
mangelajo:
```suggestion
  bool show_hidden_labels = 5 [(google.api.field_behavior) = OPTIONAL];
```

python/packages/jumpstarter-cli/jumpstarter_cli/get.py
```diff
@@ -31,8 +31,16 @@ def get():
     default=False,
     help="Include disabled exporters in the listing",
 )
+@click.option(
+    "--show-hidden",
```
mangelajo: for the same reasons:

```suggestion
    "--show-hidden-labels",
```

it's also a parameter that wouldn't be used so often, so it's ok to make it a little bit longer IMO.

python/packages/jumpstarter-cli/jumpstarter_cli/get.py
```diff
@@ -31,8 +31,16 @@ def get():
...
+    is_flag=True,
+    default=False,
+    help="Show labels hidden by controller config",
+)
 @handle_exceptions_with_reauthentication(relogin_client)
-def get_exporters(config, selector: str | None, output: OutputType, with_options: list[str], allow_disabled: bool):
+def get_exporters(
+    config, selector: str | None, output: OutputType, with_options: list[str], allow_disabled: bool, show_hidden: bool
```
mangelajo:
```suggestion
    config, selector: str | None, output: OutputType, with_options: list[str], allow_disabled: bool, show_hidden_labels: bool
```

protocol/proto/jumpstarter/client/v1/client.proto
bennyz: done

protocol/proto/jumpstarter/client/v1/client.proto
bennyz: done

python/packages/jumpstarter-cli/jumpstarter_cli/get.py
bennyz: done

python/packages/jumpstarter-cli/jumpstarter_cli/get.py
bennyz: done

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -183,6 +183,18 @@ type JumpstarterSpec struct {
 	// Lease policy configuration for controlling lease behavior.
 	// +kubebuilder:default={}
 	LeasePolicy LeasePolicyConfig `json:"leasePolicy,omitempty"`
+
+	// Hidden labels configuration for hiding specific label keys from exporter listings.
+	// +optional
+	HiddenLabels HiddenLabelsConfig `json:"hiddenLabels,omitempty"`
+}
```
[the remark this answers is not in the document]
mangelajo: Yes, but also LeasePolicy governs how the controller behaves in terms of leases. It could work also on the ControllerConfig.

This is general usage of the overall controller/leasing/etc. I think it's natural to keep here, although I had the same question.

mangelajo (approved):

PR #893, opened 2026-07-16
feat: add --exit-on-lease-end flag for single-lease exporter mode

bennyz:
When enabled via CLI flag or YAML config (exitOnLeaseEnd: true), the exporter exits cleanly after serving one lease instead of looping for the next. Uses the existing _stop_requested mechanism, guarded by previous_leased to prevent premature exit on startup before any lease is served.

### DIFF
--- a/e2e/test/exit_on_lease_end_test.go
+++ b/e2e/test/exit_on_lease_end_test.go
@@ -108,33 +108,23 @@ var _ = Describe("Exit On Lease End E2E Tests", Label("exit-on-lease-end"), Orde
 			"exporter should remain running before any lease is served")
 	})
 
-	It("exporter serves exactly one lease then exits", func() {
-		// Start the exporter and complete two lease attempts:
-		// the first should succeed, then the exporter should exit,
-		// making the second lease fail.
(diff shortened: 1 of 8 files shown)


mangelajo (approved):
Looks good, an E2E test would be positive and cover this if we migrate to rust.

mangelajo:
I will add an E2E test commit on top of yours, hold on :)

e2e/test/exit_on_lease_end_test.go
```diff
@@ -0,0 +1,130 @@
...
+			"--selector", "example.com/board=exit-on-lease-end", "j", "power", "on")
+		Expect(err).NotTo(HaveOccurred(), out)
+
+		// The exporter should exit after the lease ends.
+		Eventually(func() bool {
+			return tracker.IsProcessRunning()
+		}, 60*time.Second, 1*time.Second).Should(BeFalse(),
+			"exporter process should have exited after lease ended")
+
+		// Verify the controller reflects the exporter as offline.
```
mangelajo: This is basically the same as the first test, it's a duplicate

PR #894, opened 2026-07-17
cli: add --page-size flag to get commands and reduce e2e pagination test load

mangelajo:
## Motivation

The e2e pagination tests for leases and exporters create **101 elements** each, solely to exceed the hardcoded client-side page size of 100 and force multi-page pagination. This is unnecessarily heavy on CI:

- 101 sequential `create` gRPC calls per test
- 101 sequential `delete` calls for cleanup
- Significant cluster resource consumption

The gRPC server already fully supports arbitrary `page_size` values — it passes the value directly to the Kubernetes API `Limit` parameter. The Python client config layer (`list_leases` / `list_exporters`) also accepts `page_size` as a parameter. The only missing link was the CLI, which hardcoded the default of 100 by never passing it through.

### 1. CLI: Add `--page-size` flag (`python/packages/jumpstarter-cli/jumpstarter_cli/get.py`)

- Add `--page-size` option to both `jmp get exporters` and `jmp get leases` commands
- Default remains **100** (no behavior change for existing users)
- The value is passed through to `config.list_exporters(page_size=...)` and `config.list_leases(page_size=...)`

### 2. E2E tests: Reduce pagination test load (`e2e/test/e2e_test.go`)

- **Lease pagination test**: 101 → 10 leases, uses `--page-size 5` (forces 2 pages)
- **Exporter pagination test**: 101 → 10 exporters, uses `--page-size 5` (forces 2 pages)

This is a **10x reduction** in resources created and API calls per test while still validating the full pagination loop (multiple pages, token forwarding, result aggregation).

### What doesn't change

- Default page size for users remains 100
- Python unit tests (`client_config_test.py`) — mocked tests that validate the pagination loop logic at the library level, unaffected
- Server-side Go code — already supports arbitrary `PageSize`
- Proto definitions — `page_size` is already an optional field

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/get.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/get.py
@@ -37,6 +37,7 @@ def get():
     default=False,
     help="Show labels hidden by controller config",
 )
+@click.option("--page-size", type=click.IntRange(min=1), default=100, help="Number of results per page for pagination")
 @handle_exceptions_with_reauthentication(relogin_client)
 def get_exporters(
     config,
@@ -45,6 +46,7 @@ def get_exporters(
     with_options: list[str],
     allow_disabled: bool,
     show_hidden_labels: bool,
+    page_size: int,
 ):
     """
     Display one or many exporters
@@ -60,6 +62,7 @@ def get_exporters(
         include_status=include_status,
         include_disabled=allow_disabled,
         show_hidden_labels=show_hidden_labels,
+        page_size=page_size,
     )
 
     model_print(exporters, output)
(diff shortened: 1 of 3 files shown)


python/packages/jumpstarter-cli/jumpstarter_cli/get.py
```diff
@@ -31,8 +31,11 @@ def get():
     default=False,
     help="Include disabled exporters in the listing",
 )
+@click.option("--page-size", type=int, default=100, help="Number of results per page for pagination")
```
raballew: No validation of `page_size` zero or negative values.

python/packages/jumpstarter-cli/jumpstarter_cli/get_test.py
```diff
@@ -242,9 +243,10 @@ def test_get_leases_calls_list_leases(self):
         with patch("jumpstarter_cli.get.model_print"):
             get_leases.callback.__wrapped__.__wrapped__(
                 config=config, selector=None, output="text", show_all=False, all_clients=False, tag_filter=None,
+                page_size=100,
```
raballew, line 247: I think we should also test non default page sizes

raballew (changes requested):

python/packages/jumpstarter-cli/jumpstarter_cli/get_test.py
mangelajo: test_get_exporters_passes_custom_page_size now handles different page size than default.

bennyz (approved):

raballew (approved):

PR #905, opened 2026-07-21
feat: warn users when selector or exporter labels are deprecated

bennyz:
Adds a `deprecatedLabels` config section (map of label key to message) that lets operators mark label keys as deprecated. When clients create leases, list leases, or list exporters using deprecated labels, they receive warnings with the configured message.

```
deprecatedLabels:
    keys:
      legacy-board: Use board label instead
      old-pool: migrate to env label
```

```
uv run jmp create lease -l old-pool=fsd --client kind-test --duration 1h
Warning: selector label 'old-pool' is deprecated: Removed in v2.0
NAME                                  SELECTOR      EXPIRES AT  REMAINING  CLIENT       EXPORTER  TAGS
019f85bb-22ca-74a0-a2bc-07dcd16f136c  old-pool=fsd                         test-client
```


controller/internal/config/config.go
```diff
@@ -51,20 +51,20 @@ func LoadConfiguration(
 	key client.ObjectKey,
 	signer *oidc.Signer,
 	certificateAuthority string,
-) (authenticator.Token, string, Router, []grpc.ServerOption, *Provisioning, *LeasePolicy, *HiddenLabels, error) {
+) (authenticator.Token, string, Router, []grpc.ServerOption, *Provisioning, *LeasePolicy, *HiddenLabels, *DeprecatedLabels, error) {
```
mangelajo, line 54: Ok, at some point we probably must return a structure here, may be not on this patch if we plan to backport, but, the return list is too big :D

protocol/proto/jumpstarter/client/v1/client.proto
```diff
@@ -97,6 +97,9 @@ message Exporter {
   // Whether the exporter is enabled for lease assignment.
   // When false, the controller will not assign new leases to this exporter.
   optional bool enabled = 6 [(google.api.field_behavior) = OUTPUT_ONLY];
+  // Label keys on this exporter that the controller has marked as deprecated.
+  // Keys are the deprecated label names; values are optional human-readable messages.
+  map<string, string> deprecated_labels = 7 [(google.api.field_behavior) = OUTPUT_ONLY];
```
mangelajo, line 102: if we want to backport this one, we probably should backport enabled/disabled as well. (has field 6)

protocol/proto/jumpstarter/client/v1/client.proto
```diff
@@ -147,6 +150,9 @@ message Lease {
   // Allow leasing a disabled exporter. Only effective when exporter_name is set.
   // Useful for investigating broken exporters that have been administratively disabled.
   bool allow_disabled = 14 [(google.api.field_behavior) = IMMUTABLE];
+  // Selector keys on this lease that the controller has marked as deprecated.
+  // Keys are the deprecated label names; values are optional human-readable messages.
+  map<string, string> deprecated_labels = 15 [(google.api.field_behavior) = OUTPUT_ONLY];
```
mangelajo: same for this one + allow_disabled

python/packages/jumpstarter-cli/jumpstarter_cli/create.py
```diff
@@ -106,4 +106,13 @@ def create_lease(
         allow_disabled=allow_disabled,
     )
 
+    for label_key, message in lease.deprecated_labels.items():
+        warning = f"selector label '{label_key}' is deprecated"
```
mangelajo, line 148: I wonder if we may want to be more strict here, for creation, you cannot create deprecated labels... anyway, it's still the admin, they can do it in other ways.

mangelajo (commented):
Some comments for thought after reading. No need to change anything IMHO

controller/internal/config/config.go
bennyz: yeah, this is very prominent in the tests 😅

python/packages/jumpstarter-cli/jumpstarter_cli/create.py
bennyz: for the user it's ok, since they may already be using the deprecated label and this gives them time to transition

protocol/proto/jumpstarter/client/v1/client.proto
bennyz: i assumed this was already going to be backported 👍🏻

bennyz: now context sneaked in...

mangelajo (approved):

Issue #907, opened 2026-07-22
bug: j CLI via JUMPSTARTER_HOST socket ignores client config driver settings


dustinblack:
## Problem

When `j` runs via `JUMPSTARTER_HOST` (Unix socket path, e.g., inside `jmp shell` or `jmp mcp serve`), it does not read the client config's `drivers.unsafe` and `drivers.allow` settings. Instead, `env_async()` in `jumpstarter/utils/env.py` creates a fresh `ClientConfigV1Alpha1Drivers()` with defaults (`unsafe=False`, `allow=[]`).

This means every driver class import is rejected:
```
ImportError: jumpstarter_driver_snmp.client.SNMPServerClient doesn't match any of the allowed patterns
```

## Impact

Boards using SNMP power control (e.g., QC8775/SA8775P) are completely unusable via `j` CLI — all commands fail. The driver is installed but the import check blocks it.

R-Car S4 boards are not affected because their drivers happen to not trigger the check (different power driver).

## Root Cause

```python
# jumpstarter/utils/env.py line ~58
drivers = ClientConfigV1Alpha1Drivers()  # defaults: unsafe=False, allow=[]
```

The client config at `~/.config/jumpstarter/clients/perf-ci.yaml` has `drivers.unsafe: true` but `env_async()` never reads it.

## Workaround

Set `JMP_DRIVERS_UNSAFE=true` environment variable, which the `ClientConfigV1Alpha1Drivers` pydantic-settings model picks up via its `env_prefix="JMP_DRIVERS_"` config.

## Expected Behavior

`env_async()` should read the active client config's driver settings, or at minimum honor the env vars that `ClientConfigV1Alpha1Drivers` already supports.

PR #909, opened 2026-07-22
fix: read client config driver settings in env_async()

bennyz:
env_async() created a fresh ClientConfigV1Alpha1Drivers() that only read JMP_DRIVERS_* env vars. When JUMPSTARTER_HOST was set without those env vars, all driver imports were rejected.

Add _resolve_drivers_config() that falls back to the active client config's drivers section from ~/.config/jumpstarter/clients/*.yaml when env vars are absent.


bennyz:
fixes #907

mangelajo (approved):

PR #904, opened 2026-07-21
feat: implement status reporting on ExporterSet resources

bennyz:
Add deployment-style status conditions and replica counters to ExporterSet CRs, providing visibility into pod health, exporter availability, and scaling state on every reconcile loop.

depends on https://github.com/jumpstarter-dev/jumpstarter/pull/864

tested with:
```shell
  go build -o bin/exporter-set-controller cmd/exporter-set-controller/main.go
 ./bin/exporter-set-controller --provisioner=qemu.jumpstarter.dev --health-probe-bind-address=:8082

kubectl apply -f - <<'EOF'
  apiVersion: virtualtarget.jumpstarter.dev/v1alpha1
  kind: VirtualTargetClass
  metadata:
    name: qemu-test
    namespace: jumpstarter
  spec:
    provisioner: qemu.jumpstarter.dev
    bindingMode: Immediate
    reclaimPolicy: Delete
  ---
  apiVersion: virtualtarget.jumpstarter.dev/v1alpha1
  kind: ExporterSet
  metadata:
    name: test-set
    namespace: jumpstarter
  spec:
    minReplicas: 2
    maxReplicas: 4
    minAvailableReplicas: 1
    virtualTargetClassName: qemu-test
    selector:
      matchLabels:
        exporterset: test-set
    template:
      metadata:
        labels:
          exporterset: test-set
  EOF

kubectl apply -f - <<EOF
  apiVersion: jumpstarter.dev/v1alpha1
  kind: Exporter
  metadata:
    name: exp-1
    namespace: jumpstarter
    labels:
      exporterset: test-set
    ownerReferences:
    - apiVersion: virtualtarget.jumpstarter.dev/v1alpha1
      kind: ExporterSet
      name: test-set
      uid: "$ES_UID"
      controller: true
      blockOwnerDeletion: true
  spec:
    enabled: true
  ---
  apiVersion: jumpstarter.dev/v1alpha1
  kind: Exporter
  metadata:
    name: exp-2
    namespace: jumpstarter
    labels:
      exporterset: test-set
    ownerReferences:
    - apiVersion: virtualtarget.jumpstarter.dev/v1alpha1
      kind: ExporterSet
      name: test-set
      uid: "$ES_UID"
      controller: true
      blockOwnerDeletion: true
  spec:
    enabled: true
  ---
  apiVersion: jumpstarter.dev/v1alpha1
  kind: Exporter
  metadata:
    name: exp-3
    namespace: jumpstarter
    labels:
      exporterset: test-set
[...]

PR #910, opened 2026-07-22
feat: QEMU ExporterSet sidecar pattern with jumpstarter-exec (PITCREW-487)

mangelajo:
## Summary

Implements the QEMU ExporterSet sidecar pattern for JEP-0014, enabling the jumpstarter exporter to remotely execute commands on a co-located QEMU runtime container via a shared `emptyDir` volume.

- **`jumpstarter-exec` Rust binary** (`rust/jumpstarter-exec/`): A lightweight command execution bridge using NDJSON over Unix sockets. Runs in `serve` mode on the runtime container (launched from the shared volume) and `exec` mode from the exporter sidecar, transparently proxying stdin/stdout/stderr and exit codes. Embeds a Git-derived version via `build.rs` / `GIT_VERSION`.
- **Cargo workspace** (`rust/Cargo.toml`): Sets up the Rust monorepo root aligned with the `rust-core-rewrite` branch structure.
- **Container images**: UBI 10 multi-stage Rust builder added to `python/Containerfile`; `Containerfile.qemu-runtime` updated with a `wait-for-binary.sh` entrypoint that polls for the binary on the shared volume.
- **Pod rendering** (`qemu.go`): Init container copies `jumpstarter-exec` to the shared volume and sets `JUMPSTARTER_LAUNCHER_SOCKET` env var; uses `IfNotPresent` pull policy for local kind iteration.
- **QEMU driver** (`driver.py`): Sidecar mode via `launcher_socket` config -- wraps commands with `jumpstarter-exec exec`, adjusts paths to `/shared/`, uses Unix socket for serial console.
- **Deployment**: Manual deployment YAML into `jumpstarter-lab` (same namespace as controller/router), x86_64 KVM sample + kind-friendly no-KVM sample, and Makefile targets for image builds and controller deployment.

## Test plan

- [x] `make test-rust` passes all tests (protocol serde, server integration, e2e binary)
- [x] `make fmt` and `make lint` pass in `rust/`
- [x] Build exporter / qemu-runtime / exporter-set-controller images locally and load into kind
- [x] Deploy into `jumpstarter-lab` with `make -C controller deploy-exporterset deploy-exporterset-sample-kind`
- [x] Manually validated sidecar handoff: binary copy, `jumpstarter-exec serve` as PID 1, remote `exec` of `qemu-system-x86_64 --version` from exporter container
- [ ] Verify QEMU VM boots and serial console is accessible via the exporter (blocked on reconciler Pod creation)
- [ ] CI workflows (`rust-tests.yaml`, `lint.yaml`) run successfully on this PR

Relates to: [PITCREW-487](https://redhat.atlassian.net/browse/PITCREW-487)


controller/hack/sample-x86_64.yaml
```diff
@@ -0,0 +1,69 @@
...
+        arch: x86_64
+        virtual: "true"
+    spec:
+      drivers:
+        - type: jumpstarter_driver_qemu.driver.Qemu
+          config:
+            arch: x86_64
+            smp: 2
+            mem: 2G
+            launcher_socket: /shared/launcher.sock
```
mangelajo: This should become an injected detail, anybody constructing the exporterset config should not need to write this.

controller/hack/sample-x86_64.yaml
```diff
@@ -0,0 +1,69 @@
...
+        board: x86-64-virtual
+        arch: x86_64
+        virtual: "true"
+    spec:
+      drivers:
+        - type: jumpstarter_driver_qemu.driver.Qemu
+          config:
+            arch: x86_64
+            smp: 2
+            mem: 2G
```
mangelajo, line 69: this should already be inferred by the virtualclass parameters, here we could override details

controller/scripts/wait-for-binary.sh
```diff
@@ -0,0 +1,22 @@
...
+# Usage: wait-for-binary.sh <binary-path> [args...]
+
+set -eu
+
+BINARY="$1"
+shift
+
+echo "wait-for-binary: waiting for ${BINARY}..." >&2
+
+while [ ! -x "$BINARY" ]; do
```
bennyz: need timeout?

rust/jumpstarter-exec/Cargo.lock
bennyz, line 1: the root lock file will override this i think

rust/jumpstarter-exec/src/log.rs
```diff
@@ -0,0 +1,299 @@
...
+}
+
+fn write_json(
+    level: &str,
+    ts: &str,
+    msg: &str,
+    context: &BTreeMap<String, String>,
+    fields: &[(&str, Value)],
+) {
+    let mut map = Map::new();
```
bennyz, line 104: let mut map = json!({"level": level, "ts": ts, "msg": msg});
maybe?

rust/jumpstarter-exec/src/protocol.rs
```diff
@@ -0,0 +1,177 @@
...
+                assert_eq!(argv, vec!["ls"]);
+                assert!(env.is_empty());
+                assert!(cwd.is_none());
+            }
+            _ => panic!("expected Exec"),
+        }
+    }
+
+    #[test]
+    fn client_stdin_roundtrip() {
```
bennyz: this seems to just test serde

rust/jumpstarter-exec/src/protocol.rs
```diff
@@ -0,0 +1,177 @@
...
+        let json = serde_json::to_string(&msg).unwrap();
+        let parsed: ClientMessage = serde_json::from_str(&json).unwrap();
+        match parsed {
+            ClientMessage::Stdin { data } => assert_eq!(data, "aGVsbG8="),
+            _ => panic!("expected Stdin"),
+        }
+    }
+
+    #[test]
+    fn client_stdin_close_roundtrip() {
```
bennyz: also seems to just test serde

rust/jumpstarter-exec/src/protocol.rs
```diff
@@ -0,0 +1,177 @@
...
+        let json = serde_json::to_string(&ClientMessage::StdinClose).unwrap();
+        assert_eq!(json, r#"{"type":"StdinClose"}"#);
+    }
+
+    #[test]
+    fn client_signal_roundtrip() {
```
bennyz: same

rust/jumpstarter-exec/src/protocol.rs
```diff
@@ -0,0 +1,177 @@
...
+        let msg = ClientMessage::Signal { signal: 15 };
+        let json = serde_json::to_string(&msg).unwrap();
+        assert!(json.contains("\"signal\":15"));
+    }
+
+    #[test]
+    fn server_started_roundtrip() {
```
bennyz: same

rust/jumpstarter-exec/src/protocol.rs
```diff
@@ -0,0 +1,177 @@
...
+        let msg = ServerMessage::Started { pid: 42 };
+        let json = serde_json::to_string(&msg).unwrap();
+        let parsed: ServerMessage = serde_json::from_str(&json).unwrap();
+        assert_eq!(parsed, ServerMessage::Started { pid: 42 });
+    }
+
+    #[test]
+    fn server_exit_with_code() {
```
bennyz: same

rust/jumpstarter-exec/src/protocol.rs
```diff
@@ -0,0 +1,177 @@
...
+        let msg = ServerMessage::Exit { code: Some(0) };
+        let json = serde_json::to_string(&msg).unwrap();
+        let parsed: ServerMessage = serde_json::from_str(&json).unwrap();
+        assert_eq!(parsed, ServerMessage::Exit { code: Some(0) });
+    }
+
+    #[test]
+    fn server_exit_signal_killed() {
```
bennyz: same

rust/jumpstarter-exec/src/server.rs
```diff
@@ -0,0 +1,269 @@
...
+        .stderr(Stdio::piped());
+
+    for (k, v) in &env {
+        cmd.env(k, v);
+    }
+    if let Some(ref dir) = cwd {
+        cmd.current_dir(dir);
+    }
+
+    let mut child = match cmd.spawn() {
```
bennyz: can be simplified to:
```rust
let mut child = cmd.spawn().map_err(|e| {
      let msg = format!("spawn failed: {e}");
      log.error("spawn failed", &[("argv", json!(argv)), ("error", json!(e.to_string()))]);
      let _ = send(&writer, &ServerMessage::Error { message: msg.clone() });
      io_err(&msg)
  })?;
```

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -109,37 +135,73 @@ func (p *Provisioner) RenderPod(
...
 				{
-					Name:          "exporter",
-					Image:         DefaultExporterImage,
-					RestartPolicy: &restartAlways,
+					// One-shot init: stage jumpstarter-exec onto the
+					// shared volume so target-runtime can start it.
+					// No restartPolicy — must exit before later
+					// containers are allowed to start.
+					Name:            "copy-jumpstarter-exec",
+					Image:           DefaultExporterImage,
```
mangelajo: This should instead be auto-calculated based on the current exporterset controller version. Otherwise it will default to latest. It's basically the same problem as in https://github.com/jumpstarter-dev/jumpstarter/pull/844

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -109,37 +135,73 @@ func (p *Provisioner) RenderPod(
...
+							MountPath: sharedMountPath,
+						},
+					},
+				},
+				{
+					// Native sidecar (KEP-753): starts after the
+					// copy init completes and drains after main
+					// containers.
+					Name:            "exporter",
+					Image:           DefaultExporterImage,
```
mangelajo: same

controller/hack/sample-x86_64.yaml
mangelajo: we should also support "ref" entries.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -109,37 +135,73 @@ func (p *Provisioner) RenderPod(
...
 			InitContainers: []corev1.Container{
 				{
-					Name:          "exporter",
-					Image:         DefaultExporterImage,
-					RestartPolicy: &restartAlways,
+					// One-shot init: stage jumpstarter-exec onto the
+					// shared volume so target-runtime can start it.
+					// No restartPolicy — must exit before later
+					// containers are allowed to start.
+					Name:            "copy-jumpstarter-exec",
```
bennyz, line 195: once we require k8s 1.36+ we can use an image volume instead of this https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/

this would else remove the need for wait-for-binary

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -175,6 +242,142 @@ func (p *Provisioner) RenderPod(
...
+// - Auto-injects tcp wrapper driver entry
+func (p *Provisioner) EnrichExporterExport(
+	drivers []virtualtargetv1alpha1.DriverConfig,
+	mergedParameters map[string]interface{},
+) []virtualtargetv1alpha1.DriverConfig {
+	result := make([]virtualtargetv1alpha1.DriverConfig, 0, len(drivers)+1)
+	hasTCP := false
+
+	for _, d := range drivers {
+		if d.Type == tcpDriverType {
```
mangelajo, line 289: may be this is not the best detection. It could be that we have another TCP redirection configured, and we would stop providing the ssh redirect.

we should probably always add it unless it's overriden with the exact same name.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -175,6 +242,142 @@ func (p *Provisioner) RenderPod(
...
+	d.Config = &apiextensionsv1.JSON{Raw: raw}
+	return d
+}
+
+// defaultPartitionsForArch returns the firmware partition paths for the given architecture.
+func defaultPartitionsForArch(arch string) map[string]string {
+	switch arch {
+	case "aarch64":
+		return map[string]string{
+			"OVMF_CODE.fd": "/usr/share/edk2/aarch64/QEMU_EFI-pflash.raw",
```
mangelajo: The correct aarch64 partitions are:

```yaml
            OVMF_CODE.fd: /usr/share/AAVMF/AAVMF_CODE.fd
            OVMF_VARS.fd: /usr/share/AAVMF/AAVMF_VARS.fd
```

mangelajo: or at least this is what we use in our internal lab.

controller/internal/exporterset/reconciler.go
```diff
@@ -316,39 +338,113 @@ func (r *ExporterSetReconciler) scaleUp(
...
+		}
+
+		if err := ctrl.SetControllerReference(exporter, configSecret, r.Scheme); err != nil {
+			return created, fmt.Errorf("unable to set owner reference on config Secret: %w", err)
+		}
+
+		if err := r.Create(ctx, configSecret); err != nil {
+			if !apierrors.IsAlreadyExists(err) {
+				return created, fmt.Errorf("unable to create config Secret for %s: %w", exporter.Name, err)
+			}
```
bkhizgiy: I don't think token rotation works as expected here. Once the secret already exists `AlreadyExists` is ignored and we never update it. The ExporterReconciler periodically rotates the JWT in the credentials secret, but the config secret will continue serving the old token indefinitely. The Pod won't pick up the new token until it's deleted and recreated.

controller/internal/exporterset/exporterconfig.go
```diff
@@ -0,0 +1,194 @@
...
+			Name:      exporter.Name,
+			Namespace: exporter.Namespace,
+		},
+		Endpoint: exporter.Status.Endpoint,
+		TLS: exporterConfigTLS{
+			CA: caBase64,
+		},
+		Token:          token,
+		Export:         exportMap,
+		ExitOnLeaseEnd: true,
```
bkhizgiy: Do we want to have the exit strategy hardcoded here? I think it would be better if we took this value from the spec and let the user decide on what they wished to do as part of the exporter set.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: ohhh they finally implemented that! :) nice

mangelajo: Talking of volumes, this provider probably requires creating a volume for the storage under a specific storage class for holding the qcow2, otherwise I guess it runs on fully local storage, may be we can declare an emptydir of some size tho make sure it's accounted by the host scheduling disk space?

Gemini says we can use this:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: large-emptydir-pod
spec:
  containers:
  - name: my-container
    image: nginx
    resources:
      requests:
        # THIS handles the scheduling. The scheduler will only place 
        # the pod on a node with at least 40Gi of free ephemeral storage.
        ephemeral-storage: "40Gi" 
      limits:
        ephemeral-storage: "45Gi"
    volumeMounts:
    - name: huge-temp-space
      mountPath: /data
  volumes:
  - name: huge-temp-space
    emptyDir:
      # Omitting the 'medium' field (or setting it to "") defaults to disk.
      # Do NOT set 'medium: Memory'.
      sizeLimit: "40Gi" # THIS handles eviction if the pod writes too much.
```
time to double check.

Perhaps it won't be enough for some types of nodes if the available local space is not big enough.

controller/internal/exporterset/exporterconfig.go
```diff
@@ -0,0 +1,194 @@
...
+		Namespace: exporter.Namespace,
+		Name:      caConfigMapName,
+	}, caCM); err != nil {
+		return nil, fmt.Errorf("unable to get CA ConfigMap %s: %w", caConfigMapName, err)
+	}
+
+	caPEM := caCM.Data[caConfigMapKey]
+	if caPEM == "" {
+		return nil, fmt.Errorf("CA ConfigMap %s has no '%s' key", caConfigMapName, caConfigMapKey)
+	}
```
bkhizgiy: This will permanently block pod creation on any cluster that doesn't have jumpstarter ca cert. so in a testing/dev cluster or any non ocp environmens it will always fail.

maybe it's worth checking VirtualTargetClass.Spec.CABundleConfigMapRef first or add some sort of insecure flow for this.

rust/jumpstarter-exec/src/server.rs
```diff
@@ -0,0 +1,269 @@
...
+
+fn send(writer: &Mutex<UnixStream>, msg: &ServerMessage) -> std::io::Result<()> {
+    let mut buf = serde_json::to_vec(msg)?;
+    buf.push(b'\n');
+    writer.lock().unwrap().write_all(&buf)
+}
+
+fn io_err(msg: &str) -> std::io::Error {
+    std::io::Error::other(msg)
+}
```
bkhizgiy, line 284: Maybe it's worth adding some integration tests for the full flow? even a basic one just to test simple flow functionality.

controller/internal/exporterset/provisioners/qemu/qemu.go
bennyz: sounds like we want storageClassName for the VTC?

rust/jumpstarter-exec/src/server.rs
bennyz: see integration.rs

controller/internal/exporterset/reconciler.go
bkhizgiy, line 926: This isn't directly related to your changes, but I think it should be fixed (I've already addressed it in my other PR).

after my scaling PR the ownership changes so this check is looking at the wrong owner. pods are now owned by Exporter, not ExporterSet, so es.UID will never match. As a result, PodsPending, PodsRunning, and PodsFailed will always stay 0, we proboly should check for `isOwnedByKind` instead.

This is a leftover from #904 and #914 being developed in parallel with different assumptions about the ownership chain :/

rust/jumpstarter-exec/src/server.rs
bkhizgiy: ahh you right, missed this file it was collapsed as part of the review XD

controller/internal/exporterset/exporterconfig.go
mangelajo: When no-ca we are expected to use the system certificates. But it should work at least.

I agree we should provide the ability to supply a custom CA or insecure mode

controller/internal/exporterset/reconciler.go
mangelajo: Yeah, and the controller tests were failing on this, I will extract missing things from your PR and add you as co-author, thanks bella.

controller/internal/exporterset/reconciler.go
```diff
@@ -316,39 +336,117 @@ func (r *ExporterSetReconciler) scaleUp(
...
+	return created, nil
+}
+
+// exporterHasPod checks whether the given exporter owns a Pod.
+func (r *ExporterSetReconciler) exporterHasPod(
+	ctx context.Context,
+	exporter *jumpstarterdevv1alpha1.Exporter,
+) (bool, error) {
+	var podList corev1.PodList
+	if err := r.List(ctx, &podList, client.InNamespace(exporter.Namespace)); err != nil {
```
maboras-rh: it list all ods in the namespace and linearly scans owner references. In a namespace with hundreds of Pods, this is called once per Exporter per reconcile in the `ensureExporterPods` loop.

considering `MatchingLabels` will filter the only pods belongs to particular ExporterSet


```suggestion
	if err := r.List(ctx, &podList,
	   client.InNamespace(exporter.Namespace),
	   client.MatchingLabels{labelExporterSetName: esName},
	); err != nil {
```

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -175,6 +249,142 @@ func (p *Provisioner) RenderPod(
...
+	}
+
+	return result
+}
+
+// enrichQemuDriver applies QEMU-specific defaults to a driver config entry.
+func enrichQemuDriver(d virtualtargetv1alpha1.DriverConfig, params map[string]interface{}) virtualtargetv1alpha1.DriverConfig {
+	config := make(map[string]interface{})
+	if d.Config != nil && d.Config.Raw != nil {
+		_ = json.Unmarshal(d.Config.Raw, &config)
```
bennyz: is it ok to ignore the error here?

rust/jumpstarter-exec/src/client.rs
```diff
@@ -0,0 +1,86 @@
...
+        &ClientMessage::Exec {
+            argv,
+            env: vec![],
+            cwd: None,
+        },
+    )?;
+
+    // Forward local stdin to the remote child in a background thread.
+    let w = Arc::clone(&writer);
+    let _stdin_thread = thread::spawn(move || {
```
maboras-rh, line 28: nit: the stdin thread `_stdin_thread` is never joined - when the server sends `Exit`, the function returns while the thread may still be blocked on `stdin.read()` this is fine for a short-lived CLI process (OS cleans up on exit), and joining would actually be worse since it'd block until stdin EOF. 
I think it worth be documented somewhere

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -52,55 +56,100 @@ const (
...
+			Value: fmt.Sprintf(
+				"component=exporter,exporter=%s,namespace=%s",
+				exporter.Name, exporter.Namespace,
+			),
+		})
+	}
+
 	pod := &corev1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			GenerateName: fmt.Sprintf("%s-", exporterSet.Name),
```
bennyz: Considering it's 1:1 exporter:qemu instance, might make sense to have a stable name and add a check for existence?

rust/jumpstarter-exec/Cargo.lock
maboras-rh: Cargo Workspace Architecture: In a Rust workspace - all member crates share a single, unified `Cargo.lock` file located at the workspace root (`rust/Cargo.lock`). Cargo relies on this root lockfile to ensure that dependencies are resolved consistently across all crates in the workspace.
https://doc.rust-lang.org/cargo/reference/workspaces.html#the-workspace-section

controller/hack/sample-x86_64.yaml
mangelajo: will handle on separate PR

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: yes, I am fixing this to make sure they at least match exporter name == pod name.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: benny if that's Ok i will handle the storage detail in a follow up PR, this has grown too big. :) and this won't go too far anyway without the storage support.

bennyz: of course :)

controller/hack/sample-x86_64.yaml
mangelajo: This is now handled by `EnrichExporterExport` — the QEMU provisioner auto-injects `launcher_socket`, `default_partitions`, and `hostfwd` into the driver config, so users don't need to specify these in the ExporterSet YAML.

.github/workflows/lint.yaml
```diff
@@ -102,6 +106,27 @@ jobs:
         working-directory: python
         run: make ty
 
+  lint-rust:
+    needs: changes
+    if: needs.changes.outputs.rust == 'true'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
```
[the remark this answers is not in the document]
mangelajo: Done — added `persist-credentials: false` to the checkout step.

.github/workflows/lint.yaml
```diff
@@ -102,6 +106,27 @@ jobs:
...
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy, rustfmt
```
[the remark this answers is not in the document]
mangelajo: Done — SHA-pinned to `dtolnay/rust-toolchain@2c7215f132e9ebf062739d9130488b56d53c060c` (master 2026-07-16) with explicit `toolchain: stable` input.

controller/scripts/wait-for-binary.sh
```diff
@@ -0,0 +1,22 @@
...
+    sleep 0.1
+done
+
+echo "wait-for-binary: found ${BINARY}, starting" >&2
+exec "$BINARY" "$@"
```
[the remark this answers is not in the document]
mangelajo: Done — added configurable timeout via `WAIT_FOR_BINARY_TIMEOUT` env var (default 60s). Exits with error if the binary doesn't appear within the timeout.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -261,7 +261,9 @@ async def on(self) -> None:  # noqa: C901
             "-vga",
             "none",
             "-serial",
-            "pty",
+            f"unix:{self.parent._serial_socket},server=on,wait=off"
+            if self.parent.launcher_socket
+            else "pty",
```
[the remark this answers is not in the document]
mangelajo: Driver test and README updates for sidecar mode will be tracked in a follow-up PR — this PR is already quite large.

rust/jumpstarter-exec/src/client.rs
```diff
@@ -0,0 +1,86 @@
...
+            ServerMessage::Stderr { data } => {
+                if let Ok(bytes) = STANDARD.decode(&data) {
+                    std::io::stderr().write_all(&bytes)?;
+                    std::io::stderr().flush()?;
+                }
+            }
+            ServerMessage::Exit { code } => {
+                exit_code = code.unwrap_or(1);
+                break;
+            }
```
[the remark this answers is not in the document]
mangelajo: Good point about signal-killed vs real failure. For now this is acceptable since the QEMU driver doesn't inspect the exit code for signal distinction. We can extend the protocol with `signal` field in `Exit` and map to `128+signum` in a follow-up if needed.

rust/jumpstarter-exec/src/main.rs
```diff
@@ -0,0 +1,139 @@
...
+            let separator = args.iter().position(|a| a == "--").unwrap_or_else(|| {
+                eprintln!("Usage: jumpstarter-exec exec [--socket <path>] -- <command> [args...]");
+                process::exit(1);
+            });
+
+            let argv: Vec<String> = args[separator + 1..].to_vec();
+            if argv.is_empty() {
+                eprintln!("jumpstarter-exec exec: no command specified after '--'");
+                process::exit(1);
+            }
```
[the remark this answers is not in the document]
mangelajo: Fixed — `exec` subcommand now parses `--socket` only from `args[2..separator]` (before the `--` separator), preventing it from consuming command arguments.

rust/jumpstarter-exec/src/server.rs
```diff
@@ -0,0 +1,269 @@
...
+/// Listen on `socket_path` with the given options.
+pub fn serve_with(socket_path: &str, opts: ServeOptions) -> std::io::Result<()> {
+    let log = Arc::new(Logger::new(opts.log_format, opts.debug, opts.log_fields));
+
+    let _ = std::fs::remove_file(socket_path);
+    let listener = UnixListener::bind(socket_path)?;
+    log.info(
+        "listening",
+        &[("socket", json!(socket_path)), ("debug", json!(opts.debug))],
+    );
```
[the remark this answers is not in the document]
mangelajo: Done — added stale-socket detection (checks if an existing socket is actively serving before removing it) and sets `0o660` permissions on new sockets.

rust/jumpstarter-exec/src/server.rs
```diff
@@ -0,0 +1,269 @@
...
+                    log_in.debug("signal", &[("pid", json!(pid)), ("signal", json!(signal))]);
+                    unsafe { kill(pid as i32, signal) };
+                }
+                _ => {}
+            }
+        }
+        unsafe { kill(pid as i32, 15) }; // SIGTERM on client disconnect
+    });
+
+    let status = child.wait()?;
```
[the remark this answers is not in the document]
mangelajo: Fixed — added an `AtomicBool` reaped flag that is set after `child.wait()` returns. The reader thread checks this flag before sending signals, preventing delivery to recycled PIDs.

controller/scripts/wait-for-binary.sh
mangelajo: Done — added configurable timeout (default 60s via `WAIT_FOR_BINARY_TIMEOUT` env var).

rust/jumpstarter-exec/Cargo.lock
mangelajo: Correct — removed the duplicate `rust/jumpstarter-exec/Cargo.lock`. The workspace-level `rust/Cargo.lock` is the single source of truth.

rust/jumpstarter-exec/src/log.rs
mangelajo: The current approach uses `Map::insert` so we can add context fields from the `BTreeMap` alongside event fields. Using `json!({...})` would produce a `Value::Object` that we'd then need to merge keys into, which isn't simpler. Keeping it as-is for consistency.

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated — merged all the individual serde roundtrip tests into 2 parameterized tests (`client_variants_roundtrip` and `server_variants_roundtrip`). The `Exec` roundtrip and `client_exec_defaults` (deserialization from raw JSON) are kept separate since they test meaningful behavior beyond pure serialization.

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated into `server_variants_roundtrip` (see above).

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated (see above).

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated (see above).

rust/jumpstarter-exec/src/server.rs
mangelajo: Done — simplified to `map_err` pattern as suggested.

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated (see above).

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated (see above).

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Done — image version auto-calculation is now implemented. The controller passes its build-time version to the provisioner, which uses `resolveImage` to replace `:latest` with the specific version tag. Dirty git versions (containing `-g`) are treated as dev and preserve `:latest`.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Done (same approach — both exporter and runtime images go through `resolveImage`).

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Good point — the current logic injects `hostfwd.ssh` only if `hostfwd["ssh"]` is not already present, regardless of other `hostfwd` entries (e.g. `hostfwd.vnc`). So having another TCP redirection won't prevent SSH injection. Users can override the SSH hostfwd by specifying their own `hostfwd.ssh` in the driver config.

controller/internal/exporterset/exporterconfig.go
```diff
@@ -0,0 +1,194 @@
...
+		var config map[string]interface{}
+		if d.Config != nil && d.Config.Raw != nil {
+			_ = json.Unmarshal(d.Config.Raw, &config)
+		}
+
+		exportMap[name] = exporterConfigDriver{
+			Type:   d.Type,
+			Config: config,
+		}
+	}
```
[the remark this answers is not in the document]
mangelajo: Done — `buildExportMap` now returns an error on duplicate keys. Also documented in JEP-0014.

controller/internal/exporterset/exporterconfig.go
```diff
@@ -0,0 +1,194 @@
...
+	for _, d := range drivers {
+		name := d.Name
+		if name == "" {
+			name = deriveDriverName(d.Type)
+		}
+
+		var config map[string]interface{}
+		if d.Config != nil && d.Config.Raw != nil {
+			_ = json.Unmarshal(d.Config.Raw, &config)
+		}
```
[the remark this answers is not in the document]
mangelajo: Fixed — `buildExportMap` now returns the `json.Unmarshal` error (already done in a previous push). `enrichQemuDriver` now also returns errors instead of discarding them.

controller/internal/exporterset/reconciler.go
```diff
@@ -316,39 +338,113 @@ func (r *ExporterSetReconciler) scaleUp(
...
+		} else if hasPod {
+			continue
+		}
+
+		// Build and create the ExporterConfig Secret.
+		configSecret, err := r.buildExporterConfigSecret(ctx, es, exporter, mergedParameters)
+		if err != nil {
+			logger.Error(err, "unable to build ExporterConfig Secret", "exporter", exporter.Name)
+			continue
+		}
```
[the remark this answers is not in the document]
mangelajo: The errors from `syncConfigSecret` are now properly propagated (not just logged). The `ensureExporterPods` function returns them directly, causing a retry.

controller/internal/exporterset/reconciler.go
```diff
@@ -316,39 +338,113 @@ func (r *ExporterSetReconciler) scaleUp(
...
+			return created, fmt.Errorf("unable to set owner reference on Pod: %w", err)
 		}
 
 		if err := r.Create(ctx, pod); err != nil {
-			if delErr := r.Delete(ctx, exporter); delErr != nil {
-				logger.Error(delErr, "failed to roll back Exporter after Pod creation failure", "exporter", exporter.Name)
-			}
-			return fmt.Errorf("unable to create Pod for Exporter %s: %w", exporter.Name, err)
+			return created, fmt.Errorf("unable to create Pod for Exporter %s: %w", exporter.Name, err)
 		}
```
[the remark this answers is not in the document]
mangelajo: Fixed — `createExporterPod` now tolerates `AlreadyExists` errors, and Pod naming uses `exporter.Name` (deterministic, not `GenerateName`), which eliminates duplicate Pod creation.

docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -1290,6 +1300,57 @@ Provisioning → Ready (warm pool) → Leased → Ready
...
+
+**Provisioner enrichment (`EnrichExporterExport`):** Before persisting the
+config, the provisioner may inject or override driver entries. For example,
+the `qemu.jumpstarter.dev` provisioner:
+
+- Injects `launcher_socket` pointing to the shared-volume Unix socket path.
+- Injects architecture-appropriate `default_partitions` (EDK2 firmware paths)
+  unless the user already specified them.
+- Injects `hostfwd` settings for SSH access.
+- Auto-injects `tcp` and `ssh` wrapper drivers if not already present.
```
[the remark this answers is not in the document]
mangelajo: Done — JEP-0014 now documents driver-name collision rejection, Pod naming conventions, and `exitOnLeaseEnd` derivation.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Updated to use `/usr/share/AAVMF/AAVMF_CODE.fd` and `/usr/share/AAVMF/AAVMF_VARS.fd` — matching the lab configuration. Both paths are confirmed present in the Fedora `edk2-aarch64` package.

controller/internal/exporterset/reconciler.go
mangelajo: Token rotation works correctly now (leaning on your PR) — we used `controllerutil.CreateOrUpdate` in `syncConfigSecret` which updates the Secret data on every reconcile loop, so rotated tokens are picked up. The Pod does read the token at startup from the mounted Secret volume, but Kubernetes automatically updates mounted Secret volumes (kubelet sync period, typically ~1 minute).

controller/internal/exporterset/exporterconfig.go
mangelajo: Done — `exitOnLeaseEnd` is now derived from `ExporterSet.spec.recycleStrategy`: `ExitAndReplace` → `true`, `InPlaceReuse` → `false`.

controller/internal/exporterset/reconciler.go
mangelajo: `listPodsByExporter` already uses `MatchingLabels{labelExporterSetName: es.Name}` which filters server-side. The returned list only contains Pods belonging to this ExporterSet, which is typically a small set. The O(n) scan over owner references is on this filtered subset, not all namespace Pods.

controller/internal/exporterset/reconciler.go
```diff
@@ -316,41 +330,288 @@ func (r *ExporterSetReconciler) scaleUp(
...
+	}
+	pod.Labels[labelExporterSetName] = es.Name
+
+	if err := ctrl.SetControllerReference(exp, pod, r.Scheme); err != nil {
+		return fmt.Errorf("set owner on Pod for %s: %w", exp.Name, err)
+	}
+
+	if err := r.Create(ctx, pod); err != nil {
+		return fmt.Errorf("create Pod for %s: %w", exp.Name, err)
+	}
```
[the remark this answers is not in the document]
mangelajo: Fixed — Pods now use `exporter.Name` (deterministic name, not `GenerateName`) and `createExporterPod` tolerates `AlreadyExists`. This eliminates the race-based duplicate Pod risk.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Fixed — `enrichQemuDriver` now returns errors instead of ignoring them. `EnrichExporterExport` propagates these errors to the caller.

rust/jumpstarter-exec/src/client.rs
mangelajo: Added a comment documenting the intentional non-join. The thread exits naturally when the socket closes or EOF is read; joining would block the exit path when stdin remains open (e.g. piped input).

rust/jumpstarter-exec/src/server.rs
mangelajo: Integration tests already exist in `rust/jumpstarter-exec/tests/integration.rs` — they test the full server/client flow including exec, stdin forwarding, signals, stderr, concurrent connections, and large payloads.

controller/internal/exporterset/exporterconfig.go
```diff
@@ -0,0 +1,199 @@
...
+	parts := strings.Split(d.Type, ".")
+	if last := strings.ToLower(parts[len(parts)-1]); last != "" {
+		return last
+	}
+	return fmt.Sprintf("driver%d", idx)
+}
+
+// deriveDriverName extracts a short name from a fully qualified Python class name.
+// e.g. "jumpstarter_driver_qemu.driver.Qemu" -> "Qemu"
+// Deprecated: use driverKey instead for new code.
```
bennyz: already depracted? 😅

bkhizgiy (commented):
Did another review cycle, and I think the PR looks good now. I don't see any blockers, just need to fix the linter issue arround ExitOnLeaseEnd.

rust/jumpstarter-exec/src/client.rs
maboras-rh: clear, thanks!

controller/internal/exporterset/reconciler.go
maboras-rh: Thanks for clarification :)

controller/internal/exporterset/exporterconfig.go
mangelajo: lol LLMs good catch

controller/internal/exporterset/exporterconfig.go
```diff
@@ -0,0 +1,199 @@
...
+func deriveDriverName(fqn string) string {
+	parts := strings.Split(fqn, ".")
+	if len(parts) == 0 {
+		return fqn
+	}
+	return parts[len(parts)-1]
+}
```
mangelajo: this one must be removed

controller/api/virtualtarget/v1alpha1/exporterset_types.go
```diff
@@ -33,6 +33,11 @@ const (
 
 // DriverConfig defines a single driver entry in an exporter template.
 type DriverConfig struct {
+	// Name is the key used in the ExporterConfig export map.
+	// If omitted, derived from the Type (last segment after the last dot).
```
mangelajo:
```suggestion
	// Name is the key used in the ExporterConfig export map.
```

The driver name derivation should not happen. Name should be mandatory instead.

controller/scripts/wait-for-binary.sh
mangelajo: We probably don't even need a wait for binary at this point, since the first init container, is not a sidecar and must execute first, populating the shared volume even before the other containers have a chance to start.

mangelajo: Let's remove the use of wait-for-binary.sh and the script at all.

controller/Containerfile.qemu-runtime
```diff
@@ -2,18 +2,33 @@
...
         qemu-img \
         qemu-kvm \
+        qemu-system-aarch64 \
+        edk2-ovmf \
+        edk2-aarch64 \
         procps-ng \
         virtiofsd && \
     dnf clean all
 
+COPY scripts/wait-for-binary.sh /usr/local/bin/wait-for-binary.sh
```
mangelajo: not needed anymore.
```suggestion
```

controller/Containerfile.qemu-runtime
```diff
@@ -2,18 +2,33 @@
...
+
 USER 65532:65532
 
-# No ENTRYPOINT — the exporter sidecar will manage QEMU
-# lifecycle via a launcher/wrapper on the shared volume.
+ENTRYPOINT ["/usr/local/bin/wait-for-binary.sh"]
```
mangelajo:
```suggestion
```

not needed anymore

docs/source/contributing/jeps/JEP-0014-virtual-scalable-exporters.md
```diff
@@ -190,8 +195,10 @@ spec:
         virtual: "true"
     spec:
       drivers:
-        - type: jumpstarter_driver_power.driver.QemuPower
-        - type: jumpstarter_driver_network.driver.TcpNetwork
+        - name: qemu                   # optional; defaults to last segment of type
```
mangelajo:
```suggestion
        - name: qemu
```

controller/scripts/wait-for-binary.sh
mangelajo: Addressed: `wait-for-binary.sh` has been removed entirely. The `qemu-runtime` Containerfile now uses `jumpstarter-exec` directly as its entrypoint (`ENTRYPOINT ["/shared/jumpstarter-exec"]`). The exporter init container copies the binary to the shared volume before the runtime container starts.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Good callout on image volumes! For now, `wait-for-binary.sh` has been removed entirely — the Containerfile now uses `ENTRYPOINT ["/shared/jumpstarter-exec"]` directly since the init container copies the binary before the runtime container starts. We can revisit with image volumes once k8s 1.36+ is the baseline.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Good point. Updated the SSH hostfwd injection to always add it unless a `hostfwd` entry with the exact same `name: "ssh"` key already exists, rather than checking for any existing TCP driver. This way users can still have additional TCP redirections without losing the default SSH forwarding.

rust/jumpstarter-exec/src/protocol.rs
mangelajo: Consolidated: the 7 individual serde roundtrip tests have been merged into 2 parameterized tests (`client_variants_roundtrip` and `server_variants_roundtrip`) in commit d4dc011c.

rust/jumpstarter-exec/src/server.rs
mangelajo: Done in d4dc011c — simplified to use `map_err` as suggested.

rust/jumpstarter-exec/Cargo.lock
mangelajo: Done — the duplicate `rust/jumpstarter-exec/Cargo.lock` has been deleted. The workspace-level `rust/Cargo.lock` is sufficient.

rust/jumpstarter-exec/src/log.rs
mangelajo: Good suggestion. The BTreeMap approach was chosen for consistent field ordering in the JSON output (alphabetical keys), but `json!()` would be cleaner for a fixed set of fields. Will consider for a follow-up cleanup.

controller/internal/exporterset/exporterconfig.go
mangelajo: Done — `deriveDriverName` has been removed along with `driverKey`, since `DriverConfig.Name` is now mandatory. The `buildExportMap` function uses `d.Name` directly.

controller/hack/sample-x86_64.yaml
mangelajo: Agreed — supporting `ref` entries in driver configs is a good idea for reusable configs. Will track as a follow-up.

bennyz (approved):

bkhizgiy (approved):

mangelajo:
Thank you @bennyz, @bkhizgiy and @maboras-rh  :)

controller/hack/sample-x86_64.yaml
mangelajo: Tracked: https://github.com/jumpstarter-dev/jumpstarter/issues/923 / [PITCREW-500](https://redhat.atlassian.net/browse/PITCREW-500) (infer from VTC params + `ref` entries).

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Tracked: https://github.com/jumpstarter-dev/jumpstarter/issues/924 / [PITCREW-501](https://redhat.atlassian.net/browse/PITCREW-501) (QEMU ExporterSet storage/disk support; image volumes called out as related later work).

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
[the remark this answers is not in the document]
mangelajo: Tracked: https://github.com/jumpstarter-dev/jumpstarter/issues/925 / [PITCREW-502](https://redhat.atlassian.net/browse/PITCREW-502) (sidecar-mode driver tests + README).

rust/jumpstarter-exec/src/client.rs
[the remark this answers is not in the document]
mangelajo: Tracked: https://github.com/jumpstarter-dev/jumpstarter/issues/926 / [PITCREW-503](https://redhat.atlassian.net/browse/PITCREW-503) (`Exit.signal` / `128+signum` protocol extension).

PR #911, opened 2026-07-22
fix: retry transient gRPC failures in status reports

### LINKED ISSUE #890: Exporter stuck in AfterLeaseHook status permanently if gRPC status update fails
## Summary

If the exporter fails to update its status from `AfterLeaseHook` → `Available` on the controller (e.g. due to a transient gRPC timeout), the error is silently swallowed with no retry. The controller permanently sees the exporter as stuck in `AfterLeaseHook`, and it is never assigned new leases — effectively bricked until manually restarted.

## How it was discovered

An E2E test (`C1: afterLease onFailure=warn keeps exporter available`) timed out after 300s waiting for the exporter to return to `Available` status. The exporter log showed:

```
ERROR:jumpstarter.exporter.exporter:Failed to update status: Timeout connecting to grpc.jumpstarter.10.1.0.185.nip.io:8082 (Caused by: )
```

The afterLease hook completed successfully, but the subsequent status update to the controller failed due to a transient network issue. The exporter proceeded to close the session internally, but the controller was left seeing `AfterLeaseHook` as the last reported state indefinitely.

CI run: https://github.com/jumpstarter-dev/jumpstarter/actions/runs/29279289621/job/86916394262?pr=889

## Impact

In production, any transient network disruption (timeout, controller restart, DNS blip) during the `AfterLeaseHook` → `Available` status update will permanently remove the exporter from the available pool. The exporter is internally idle and ready, but the controller thinks it is still running hooks.

This could affect any status transition, but `AfterLeaseHook` → `Available` is the most damaging because it is the transition that returns the exporter to the lease pool.

## Expected behavior

The exporter should retry failed status updates with backoff, or re-register with the controller when it detects the status stream has gone stale — similar to how the exporter already recovers after an `onFailure=exit` shutdown/restart cycle via full re-registration.

## Affected component

`jumpstarter.exporter.exporter` — status update path after afterLease hook completion.

### LINKED ISSUE #912: Bound exporter ReportStatus RPC attempts with a deadline
## Summary

Add a bounded per-attempt deadline to exporter `ReportStatus` RPC calls so a stalled controller call cannot block status reporting indefinitely.

## Rationale

`Exporter._send_report_status_rpc()` retries transient gRPC failures, including `DEADLINE_EXCEEDED`, but its `controller.ReportStatus(request)` invocation has no timeout. A call that hangs without completing cannot produce a deadline error or advance the retry/backoff flow.

## Affected area

- `python/packages/jumpstarter/jumpstarter/exporter/exporter.py`
- `Exporter._send_report_status_rpc()`

## Required changes

- Pass a short, explicitly configured per-attempt timeout/deadline to `ReportStatus`.
- Preserve the existing retry policy: retry `grpc.StatusCode.DEADLINE_EXCEEDED` and `grpc.StatusCode.UNAVAILABLE` while attempts remain; fail after retry exhaustion.
- Add focused coverage for a per-attempt timeout and the resulting retry behavior.

## Acceptance criteria

- A non-responsive `ReportStatus` attempt completes within the configured deadline.
- Deadline failures participate in the existing transient-error retry and exponential-backoff behavior.
- A terminal failure is logged and returns failure after retries are exhausted.
- Tests cover timeout-triggered retry and success/failure outcomes.

## Backlinks

- Follow-up from https://github.com/jumpstarter-dev/jumpstarter/pull/911
- Originating review discussion: https://github.com/jumpstarter-dev/jumpstarter/pull/911#discussion_r3633052810
- Requested by @evakhoni

evakhoni:
## Summary

Fixes #890, closes #912.

Adds retry logic with exponential backoff for transient gRPC failures in exporter status reporting and lease release, with per-RPC deadlines to prevent hanging. Ensures exporters recover from controller outages instead of becoming permanently stuck.

## Problem

When the afterLease hook completes, the exporter calls `_report_status(AVAILABLE)` to tell the controller it's ready for new leases. If that gRPC call fails transiently (network blip, controller momentarily unreachable), the error is silently swallowed. The controller permanently sees `AFTER_LEASE_HOOK`, and the exporter is never assigned new leases — even though it locally believes it's `AVAILABLE`. Only a pod restart recovers.

Additionally, `_request_lease_release` had no retry logic and no per-RPC deadline, so a single transient failure or TCP hang could permanently block lease cleanup.

### Shared retry helper (`_retry_rpc`)

Extracted a reusable retry helper used by both `ReportStatus` and `ReleaseLease` RPCs:

- **Max retries:** 20 (~18 min worst case: 21 attempts × 30s timeout + backoff sum ~8 min)
- **Backoff:** Exponential with cap — 1s, 2s, 4s, ..., 30s, 30s, ...
- **Per-RPC timeout:** 30s deadline on each gRPC call (prevents hanging on TCP/keepalive timeout)
- **Retryable:** `UNAVAILABLE`, `DEADLINE_EXCEEDED`
- **Non-retryable:** Configurable per call site (e.g., `UNIMPLEMENTED` for ReportStatus)

The extended retry window is safe because the exporter has nothing useful to do while the controller is unreachable — even a restart just fails at `_register_with_controller` (no retry there).

### Non-blocking status reports during `serve()`

`_report_status` was fire-and-forget before this PR. Adding retry made it blocking, which could stall critical paths like LEASE_READY updates that gate client connections. Fixed with a background drain task using "latest wins" semantics: `_report_status` enqueues the update and returns immediately; the drain task sends the most recent status with full retry.

### ReleaseLease RPC with exporter auth

Added exporter authentication support to the `ReleaseLease` RPC (Go controller), using metadata-based routing (`jumpstarter-kind: Exporter`) instead of try-both-auth. The Python exporter uses `ReleaseLease` with backward-compat fallback to `ReportStatus(release_lease=true)` for older controllers.

Controller-side error codes use `FailedPrecondition`/`NotFound`/`Internal` (not `PermissionDenied`, which would wrongly trigger the UNSUPPORTED fallback path).

## Testing

- 13 new Python tests covering retry, timeout, backoff cap, non-blocking drain, lease release (success, retry, exhaustion, fallback, compat)
- 7 new Go integration tests for `handleExporterLeaseRelease` and `releaseLeaseAsExporter`
- Go auth package at 100% coverage
- All 637 Python tests pass, all Go tests pass
- No lint or type errors

## Test plan

- [x] Unit tests pass (`make -C python pkg-test-jumpstarter`)
- [x] Go tests pass (`make -C controller test`)
- [x] Lint clean (`make -C python lint-fix`)
- [x] Type check clean (`make -C python pkg-ty-jumpstarter`)

### DIFF
--- a/controller/internal/service/auth/auth.go
+++ b/controller/internal/service/auth/auth.go
@@ -8,6 +8,7 @@ import (
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/authorization"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/oidc"
 	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
 	"google.golang.org/grpc/status"
 	"k8s.io/apiserver/pkg/authorization/authorizer"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -92,6 +93,17 @@ func (s *Auth) VerifyExporter(ctx context.Context) (*jumpstarterdevv1alpha1.Expo
 	}
 
 	return jexporter, nil
+}
+
+// IsExporter checks the jumpstarter-kind metadata to determine if the caller
+// is an exporter without performing full authentication.
+func (s *Auth) IsExporter(ctx context.Context) bool {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return false
+	}
+	kinds := md.Get("jumpstarter-kind")
+	return len(kinds) == 1 && kinds[0] == "Exporter"
 }
 
 func (s *Auth) AuthExporter(ctx context.Context, namespace string) (*jumpstarterdevv1alpha1.Exporter, error) {
--- a/controller/internal/service/auth/auth.go
+++ b/controller/internal/service/auth/auth.go
@@ -96,7 +96,7 @@ func (s *Auth) VerifyExporter(ctx context.Context) (*jumpstarterdevv1alpha1.Expo
 }
 
 // IsExporter checks the jumpstarter-kind metadata to determine if the caller
-// is an exporter without performing full authentication.
+// is an exporter.
 func (s *Auth) IsExporter(ctx context.Context) bool {
 	md, ok := metadata.FromIncomingContext(ctx)
 	if !ok {
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -54,6 +54,7 @@ import (
 	"google.golang.org/protobuf/types/known/durationpb"
 	"google.golang.org/protobuf/types/known/timestamppb"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/fields"
@@ -1051,11 +1052,15 @@ func (s *ControllerService) releaseLeaseAsExporter(
 		Namespace: exporter.Namespace,
 		Name:      req.Name,
 	}, &lease); err != nil {
-		return nil, fmt.Errorf("failed to get lease: %w", err)
+		if apierrors.IsNotFound(err) {
+			return nil, status.Errorf(codes.NotFound, "lease %s not found", req.Name)
+		}
+		return nil, status.Errorf(codes.Internal, "failed to get lease: %s", err)
 	}
 
 	if lease.Status.ExporterRef == nil || lease.Status.ExporterRef.Name != exporter.Name {
-		return nil, fmt.Errorf("ReleaseLease permission denied")
+		return nil, status.Errorf(codes.FailedPrecondition,
+			"lease %s is not held by exporter %s", req.Name, exporter.Name)
 	}
 
 	// Idempotent: already ended or marked for release
@@ -1067,7 +1072,7 @@ func (s *ControllerService) releaseLeaseAsExporter(
 	lease.Spec.Release = true
 
 	if err := s.Client.Patch(ctx, &lease, original); err != nil {
-		return nil, fmt.Errorf("failed to mark lease for release: %w", err)
+		return nil, status.Errorf(codes.Internal, "failed to mark lease for release: %s", err)
 	}
 
 	return &pb.ReleaseLeaseResponse{}, nil
@@ -1077,8 +1082,12 @@ func (s *ControllerService) ReleaseLease(
 	ctx context.Context,
 	req *pb.ReleaseLeaseRequest,
 ) (*pb.ReleaseLeaseResponse, error) {
-	// Try exporter auth first, then client auth
-	if exporter, err := s.authenticateExporter(ctx); err == nil {
+	// Route based on jumpstarter-kind metadata to avoid spurious auth failures
+	if s.getAuth().IsExporter(ctx) {
+		exporter, err := s.authenticateExporter(ctx)
+		if err != nil {
+			return nil, err
+		}
 		return s.releaseLeaseAsExporter(ctx, exporter, req)
 	}
 
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -1041,10 +1041,47 @@ func (s *ControllerService) RequestLease(
 	}, nil
 }
 
+func (s *ControllerService) releaseLeaseAsExporter(
+	ctx context.Context,
+	exporter *jumpstarterdevv1alpha1.Exporter,
+	req *pb.ReleaseLeaseRequest,
+) (*pb.ReleaseLeaseResponse, error) {
+	var lease jumpstarterdevv1alpha1.Lease
+	if err := s.Client.Get(ctx, types.NamespacedName{
+		Namespace: exporter.Namespace,
+		Name:      req.Name,
+	}, &lease); err != nil {
+		return nil, fmt.Errorf("failed to get lease: %w", err)
+	}
+
+	if lease.Status.ExporterRef == nil || lease.Status.ExporterRef.Name != exporter.Name {
+		return nil, fmt.Errorf("ReleaseLease permission denied")
+	}
+
+	// Idempotent: already ended or marked for release
+	if lease.Status.Ended || lease.Spec.Release {
+		return &pb.ReleaseLeaseResponse{}, nil
+	}
+
+	original := client.MergeFrom(lease.DeepCopy())
+	lease.Spec.Release = true
+
+	if err := s.Client.Patch(ctx, &lease, original); err != nil {
+		return nil, fmt.Errorf("failed to mark lease for release: %w", err)
+	}
+
+	return &pb.ReleaseLeaseResponse{}, nil
+}
+
 func (s *ControllerService) ReleaseLease(
 	ctx context.Context,
 	req *pb.ReleaseLeaseRequest,
 ) (*pb.ReleaseLeaseResponse, error) {
+	// Try exporter auth first, then client auth
+	if exporter, err := s.authenticateExporter(ctx); err == nil {
+		return s.releaseLeaseAsExporter(ctx, exporter, req)
+	}
+
 	jclient, err := s.authenticateClient(ctx)
 	if err != nil {
 		return nil, err
@@ -1060,6 +1097,11 @@ func (s *ControllerService) ReleaseLease(
 
 	if lease.Spec.ClientRef.Name != jclient.Name {
 		return nil, fmt.Errorf("ReleaseLease permission denied")
+	}
+
+	// Idempotent: already ended or marked for release
+	if lease.Status.Ended || lease.Spec.Release {
+		return &pb.ReleaseLeaseResponse{}, nil
 	}
 
 	original := client.MergeFrom(lease.DeepCopy())
--- a/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
@@ -36,6 +36,11 @@ if TYPE_CHECKING:
 
 logger = logging.getLogger(__name__)
 
+# ReportStatus retry configuration
+_TRANSIENT_GRPC_CODES = frozenset({grpc.StatusCode.UNAVAILABLE, grpc.StatusCode.DEADLINE_EXCEEDED})
+_REPORT_STATUS_MAX_RETRIES = 3
+_REPORT_STATUS_BACKOFF_BASE = 1.0
+
 
 async def _standalone_shutdown_waiter():
     """Wait forever; used so serve_standalone_tcp can be cancelled by stop()."""
@@ -348,6 +353,42 @@ class Exporter(AsyncContextManagerMixin, Metadata):
         if self._lease_context is None:
             await self._report_status(ExporterStatus.AVAILABLE, "Exporter registered and available")
 
+    async def _send_report_status_rpc(self, request: jumpstarter_pb2.ReportStatusRequest) -> bool:
+        """Send ReportStatus RPC to the controller with retry on transient errors.
+
+        Args:
+            request: The ReportStatusRequest protobuf message to send
+
+        Returns:
+            True if the RPC succeeded (or controller doesn't support it), False otherwise
+        """
+        for attempt in range(_REPORT_STATUS_MAX_RETRIES + 1):
+            try:
+                async with self._controller_stub() as controller:
+                    await controller.ReportStatus(request)
+                return True
+            except grpc.aio.AioRpcError as e:
+                if e.code() == grpc.StatusCode.UNIMPLEMENTED:
+                    logger.warning("ReportStatus not supported by controller, status updates will be skipped")
+                    return False
+                if e.code() in _TRANSIENT_GRPC_CODES and attempt < _REPORT_STATUS_MAX_RETRIES:
+                    backoff = _REPORT_STATUS_BACKOFF_BASE * (2**attempt)
+                    logger.warning(
+                        "Transient error reporting status (attempt %d/%d), retrying in %.1fs: %s",
+                        attempt + 1,
+                        _REPORT_STATUS_MAX_RETRIES + 1,
+                        backoff,
+                        e,
+                    )
+                    await anyio.sleep(backoff)
+                    continue
+                logger.error("Failed to update status: %s", e)
+                return False
+            except Exception as e:
+                logger.error("Failed to update status: %s", e)
+                return False
+        return False
+
     async def _report_status(self, status: ExporterStatus, message: str = ""):
         """Report the exporter status with the controller and session."""
         self._exporter_status = status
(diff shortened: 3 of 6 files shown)


python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -348,6 +353,42 @@ async def _register_with_controller(self, local_channel: grpc.aio.Channel):
...
+                    backoff = _REPORT_STATUS_BACKOFF_BASE * (2**attempt)
+                    logger.warning(
+                        "Transient error reporting status (attempt %d/%d), retrying in %.1fs: %s",
+                        attempt + 1,
+                        _REPORT_STATUS_MAX_RETRIES + 1,
+                        backoff,
+                        e,
+                    )
+                    await anyio.sleep(backoff)
+                    continue
```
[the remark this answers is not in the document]
evakhoni: seems like an existing behavior that predates my patch. is it something worth addressing? probably. but is it in the scope of the current pr? probably not.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -400,20 +433,17 @@ async def _request_lease_release(self):
...
-                        release_lease=True,
-                    )
-                )
+        success = await self._send_report_status_rpc(
+            jumpstarter_pb2.ReportStatusRequest(
+                status=ExporterStatus.AVAILABLE.to_proto(),
+                message="Lease released after afterLease hook",
+                release_lease=True,
+            )
+        )
```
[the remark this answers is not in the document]
evakhoni: makes no sense. if the controller haven't received the first release_lease=True (so we had to retry) why should it already try to lease the exporter to someone else?

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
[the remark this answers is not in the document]
evakhoni: yes please

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -348,6 +353,42 @@ async def _register_with_controller(self, local_channel: grpc.aio.Channel):
...
+        """
+        for attempt in range(_REPORT_STATUS_MAX_RETRIES + 1):
+            try:
+                async with self._controller_stub() as controller:
+                    await controller.ReportStatus(request)
+                return True
+            except grpc.aio.AioRpcError as e:
+                if e.code() == grpc.StatusCode.UNIMPLEMENTED:
+                    logger.warning("ReportStatus not supported by controller, status updates will be skipped")
+                    return False
```
mangelajo: Your comment says that it does return true if succeeded, or controller doesn't support it, but here on unsupported it returns False.

I am not sure if it's the comment or the code what needs to be fixed.

mangelajo (changes requested):

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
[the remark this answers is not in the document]
evakhoni: hmm... yeah that's a concern actually. I have just pushed a fix for it - ptal

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
evakhoni: ah my bad. yeah and the whole controller doesn't support it thing is legacy at this point, so I marked it as such so one day we can hopefully give it a cleanup. 

updated this one, and a few more changes addressing bot reviewers :white_check_mark:

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -400,6 +434,11 @@ async def _request_lease_release(self):
             self._lease_context.lease_ended.set()
             return
 
+        # Phase 1: attempt release_lease=true once (non-idempotent, not safe to retry)
+        # If the controller processed the request but response was lost, retrying could
```
mangelajo: This is false, every lease has a different name, and ended leases are not really released, so it's really idempotent.

mangelajo: We can, and should retry end lease too.

evakhoni: hmm.. the thing is, we will have a race condition here. 
afaict, exporter sends release_lease + status AVAILABLE, but it never sends the lease name to the controller in the same call, so controller just sees it as "release whatever lease is assigned at the moment"

think about this:
1. exporter sends release + available. 
2. controller releases lease 1, replies to the exporter
3. the **reply from the controller to the exporter** gets lost in the network
4. the controller unaware that the reply was dropped, already started to **assigns lease 2** 
5. meanwhile the exporter which haven't seen a reply is resending release + available
6. controller receives this, sees lease 2 and releases it.

bennyz: So if we split this too actually ReleaseLease which is named + ReportStatus(release_lease=False) it should simplify things and eliminate the race and makes them both idempotent?

and in general having ReportStatus(release_lease=True) is not great since it creates a side effect in aread request

evakhoni: yep. evaluating stuff around this one for a while. iiuc ReleaseLease is actually a client-controller communication rather than exporter-controller, while ReportStatus from the exporter doesn't have lease name (fetches from exporter.Status.LeaseRef at the controller so it is prone to lease 2 race). one solution would be extending it so the exporter sends lease name alongside release, but that would introduce probably breaking RPC change, so I was thinking about another solution: 

at first, I was thinking of sending (and retrying) release separately, then status available, however seems like there are a few cases where a controller would see a previous available just prior to us sending release_lease, so a proper solution seems like:
1. sending some temporary status different than available, alongside release_lease. retrying as needed
2. as looping 1 is completed, sending available separately retrying as well.
I'm just checking why in the first place we have that situation where the controller sees the exporter as briefly available prior to running afterlease, which seems like a bug in the hooks code to me, but i want to double-check that before I patch it.

bennyz: But ReportStatus effectively is ReleaseLease for the exporter, so the only gap for using ReleaseLease in the exporter I've seen is adding auth in the backend, so I am still in favor of avoiding the side effect unless there is another issue with it

evakhoni: @bennyz ok I have checked it, your solution is indeed feasible, but it'll require quite a larger change as it seems, on both the controller side, as well as the exporter, and would need some code for backward compatibility.

I have found an easy one with just separating the two: sending release_lease first and status available second, retrying each as needed. 

I have checked, the only case where the status prior to release_lease was "available" (which is something we don't want at the controller to prevent a race condition) is set right after successful AFTER_LEASE_HOOK (that was a workaround by Kirk) so we can send AFTER_LEASE_HOOK alongside release_lease in this specific case which is a very minimal and harmless side effect

all in all, I"m up to whichever fix is actually better, rather than whichever is easier, but I do think if we end up moving too many things around we may just over-complicate our code :thinking:

evakhoni: as for what you mentioned about `ReportStatus(release_lease=True)` being smelly, this was actually the behavior well predating my code, it's the existing one for a while now. so one argument in favor of your solution would be if we really want to completely get rid of passing release_lease as part of status, which may make sense semantically, but arguably may be as well out of scope of the current PR. honestly, idk I'm good with whichever :man_shrugging:

bennyz: i was not aware of the side effect in ReportStatus previously so imo it's a good opportunity to change it

evakhoni: ok, I checked what we need for that one, it seems we cannot fully remove the side effect from ReportStatus without breaking backward compat, but we can deprecate it and mark it as such, so the real cleanup would be only possible 2 versions into the future. if we're all fine with that, I can implement it this way. as for the compat code needed, it should either contain my fix from above (separate release_lease from status) or the unpatched code with the original bug. wdyt?

bennyz: yeah i did not mean to remove it, but rather not use + deprecate, so that sounds good to me

controller/internal/service/controller_service.go
```diff
@@ -1041,10 +1041,47 @@ func (s *ControllerService) RequestLease(
...
+
+	return &pb.ReleaseLeaseResponse{}, nil
+}
+
 func (s *ControllerService) ReleaseLease(
 	ctx context.Context,
 	req *pb.ReleaseLeaseRequest,
 ) (*pb.ReleaseLeaseResponse, error) {
+	// Try exporter auth first, then client auth
+	if exporter, err := s.authenticateExporter(ctx); err == nil {
```
bennyz: hmm this might spam the logs with errors in the common flow where the client does the releasing?

bennyz: a check like this might work, not super pretty because it returns []string:
```go
  func (s *Auth) IsExporter(ctx context.Context) bool {
      md, ok := metadata.FromIncomingContext(ctx)
      if !ok { return false }
      kinds := md.Get("jumpstarter-kind")
      return len(kinds) == 1 && kinds[0] == "Exporter"
  }
```

evakhoni: good idea. I actually already asked claude about that one before, and it convinced me that no error is expected :facepalm: 
anyway, implemented in https://github.com/jumpstarter-dev/jumpstarter/pull/911/commits/834fde27abf9768ec52f66265a09100171ed8784

controller/internal/service/controller_service.go
```diff
@@ -1041,10 +1041,47 @@ func (s *ControllerService) RequestLease(
...
+	// Idempotent: already ended or marked for release
+	if lease.Status.Ended || lease.Spec.Release {
+		return &pb.ReleaseLeaseResponse{}, nil
+	}
+
+	original := client.MergeFrom(lease.DeepCopy())
+	lease.Spec.Release = true
+
+	if err := s.Client.Patch(ctx, &lease, original); err != nil {
+		return nil, fmt.Errorf("failed to mark lease for release: %w", err)
```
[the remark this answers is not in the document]
evakhoni: permission denied is the wrong one to use in this case, as the exporter won't know to differentiate it from a legacy controller with no exporter-facing ReleaseLease support. implemented a proper one instead ptal https://github.com/jumpstarter-dev/jumpstarter/commit/834fde27abf9768ec52f66265a09100171ed8784

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -348,6 +371,90 @@ async def _register_with_controller(self, local_channel: grpc.aio.Channel):
...
+    async def _send_report_status_rpc(self, request: jumpstarter_pb2.ReportStatusRequest) -> bool:
+        """Send ReportStatus RPC to the controller with retry on transient errors.
+
+        Args:
+            request: The ReportStatusRequest protobuf message to send
+
+        Returns:
+            True if the RPC succeeded, False if the controller doesn't support it or retries were exhausted
+        """
+        for attempt in range(_REPORT_STATUS_MAX_RETRIES + 1):
```
bennyz: can we push the retries into a helper?
the code in _send_release_lease_rpc looks almost identical

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -400,20 +502,60 @@ async def _request_lease_release(self):
...
-                await controller.ReportStatus(
-                    jumpstarter_pb2.ReportStatusRequest(
-                        status=ExporterStatus.AVAILABLE.to_proto(),
-                        message="Lease released after afterLease hook",
-                        release_lease=True,
-                    )
+        lease_name = self._lease_context.lease_name
+
+        # Try ReleaseLease RPC unless we've already discovered it's unsupported
+        use_compat_fallback = self._release_lease_unsupported
```
bennyz: consider
```python
	if self._release_lease_unsupported:
	      await self._send_compat_release(lease_name)
	  else:
	      ok, code = await _retry_grpc(...) # if my suggestion to move retries to a helper is accepted
	
	      if ok:
	          logger.info("Released lease %s via ReleaseLease", lease_name)
	      elif code in _UNSUPPORTED_CODES:
	          self._release_lease_unsupported = True
	          await self._send_compat_release(lease_name)
	      else:
	          logger.warning("Failed to release lease %s", lease_name)
	
	  await self._report_status(AVAILABLE, ...)
	  if self._lease_context and not self._lease_context.lease_ended.is_set():
	      self._lease_context.lease_ended.set()
        use_compat_fallback = self._release_lease_unsupported
```

evakhoni:
@bennyz good ideas. [addressed](https://github.com/jumpstarter-dev/jumpstarter/pull/911/commits/8d2093f4b3045b6fc76a193b4a834b7e390026a7) both, and pushed [another one](https://github.com/jumpstarter-dev/jumpstarter/pull/911/commits/8ea6ca2cd9f72928fd958cd2c6db98a052066f68) to improve the retry process and to prevent clients from being blocked. ptal tnx!

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -348,6 +392,88 @@ async def _register_with_controller(self, local_channel: grpc.aio.Channel):
...
+            await self._status_rpc_event.wait()
+            self._status_rpc_event = Event()  # Reset for next update
+            request = self._pending_status_request
+            if request and await self._send_report_status_rpc(request):
+                logger.info(
+                    "Updated status to %s: %s",
+                    ExporterStatus.from_proto(request.status),
+                    request.message,
+                )
+
```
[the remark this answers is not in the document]
evakhoni: 1. The status stream dies first. _retry_stream for the Status stream has only 5 retries with shorter backoff. If the controller is unreachable for long enough to exhaust 20 drain retries (~18 min), the status stream has already died, crashed the task group, and killed the exporter. The drain task gets cancelled along with everything else.
2. Even if it survived, the next state transition (lease end, new lease, etc.) triggers a _report_status call which replaces _pending_status_request and re-arms the event. The exporter is rarely idle for 18+ minutes without a status change.
3. For UNIMPLEMENTED (legacy controller), returning False immediately and going idle is the correct behavior — the controller doesn't support ReportStatus at all.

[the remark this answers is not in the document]
evakhoni: also, Potential issues with re-arming -  UNIMPLEMENTED infinite loop — This is the main risk. If the controller is legacy and returns UNIMPLEMENTED, _send_report_status_rpc returns False immediately. If we naively re-arm the event, the drain enters a tight loop: try → UNIMPLEMENTED → re-arm → try → UNIMPLEMENTED → ... forever, spamming logs.

controller/internal/service/auth/auth.go
```diff
@@ -94,6 +95,17 @@ func (s *Auth) VerifyExporter(ctx context.Context) (*jumpstarterdevv1alpha1.Expo
 	return jexporter, nil
 }
 
+// IsExporter checks the jumpstarter-kind metadata to determine if the caller
+// is an exporter without performing full authentication.
```
mangelajo:
```suggestion
// IsExporter checks the jumpstarter-kind metadata to determine if the caller
// is an exporter.
```

Remove the "without performing full authentication" , I was worried for a second that we would not be authenticating the request if it was a exporter, but I see you auth the exporter in that case. :)

bennyz (approved):

mangelajo (approved):
only a tiny nit, if you can fix the comment before merging :)

Thanks for looking at this.

mangelajo:
I applied the fix as it was just a tiny tweak on the comment itself, then we can just leave it to auto-merge

evakhoni:
merge failure seems to be a CI flake. just readded to queue :crossed_fingers:

PR #927, opened 2026-07-29
feat: kubevirt provisioner

bennyz:
The KubeVirt provisioner enables Jumpstarter to create and manage virtual machines on a remote KubeVirt or OpenShift Virtualization cluster while running lightweight exporter pods on a management cluster.

A VirtualTargetClass defines the KubeVirt connection, VM namespace, disk image, compute resources, storage, cloud-init configuration, SSH access, and optional port forwarding. An ExporterSet then maintains the requested pool of virtual machines through the normal Jumpstarter leasing lifecycle.

1. The provisioner reads a kubeconfig from a management-cluster Secret.
2. It connects to the remote KubeVirt cluster.
3. It creates a VirtualMachine resource for each exporter.
4. It optionally creates DataVolume, SSH, and port-forwarding Services.
5. It starts a lightweight exporter pod in the management cluster.
6. The exporter pod connects to the VM using the injected remote kubeconfig.
7. When the exporter is reclaimed, the provisioner removes the VM and associated resources according to the reclaim policy.


mangelajo:
haha surprise provisioner, but honestly probably not a bad idea since now it's very easy to make, we can have a comparison point or alternative if the qemu is problematic at some point and the users don't need any of the special stuff like can or future bluetooth emulation.

bennyz:
> haha surprise provisioner, but honestly probably not a bad idea since now it's very easy to make, we can have a comparison point or alternative if the qemu is problematic at some point and the users don't need any of the special stuff like can or future bluetooth emulation.

i wanted to play around with cuttlefish and it's easy getting kubevirt instances internally 😅 

it also help find some issues in the general provisioner and also offer some insights into the storage handling

raballew:
@bennyz what are we going to do about this one?

bennyz:
> @bennyz what are we going to do about this one?

i want to get this in, i'll rebase soon

PR #928, opened 2026-07-29
feat: add guest disk storage for QEMU ExporterSet pods

### LINKED ISSUE #924: Add storage/disk support for QEMU ExporterSet virtual targets
## Summary

QEMU ExporterSet pods need proper guest disk / storage support so lessees can flash and boot images. Without this, the sidecar pattern is limited for real workflows.

Deferred from PR #910 (scope already large).

## Background

From [PR #910](https://github.com/jumpstarter-dev/jumpstarter/pull/910#discussion_r3659038447):

> …I will handle the storage detail in a follow up PR, this has grown too big. :) and this won't go too far anyway without the storage support.

Related discussion on staging binaries via Kubernetes [image volumes](https://kubernetes.io/docs/tasks/configure-pod-container/image-volumes/) once the cluster baseline is 1.36+ ([comment](https://github.com/jumpstarter-dev/jumpstarter/pull/910#discussion_r3655668698)) — that is a related optimization for the `copy-jumpstarter-exec` init container, but the primary gap is guest disk storage for flash/boot.

## What needs to be done

1. Define how guest disks are provisioned for ExporterSet QEMU pods (emptyDir, PVC, image volume, OCI artifact, etc.)
2. Wire storage into the QEMU provisioner `RenderPod` + ExporterConfig enrichment so the QEMU driver can flash/boot

## Acceptance criteria

- [ ] ExporterSet QEMU pods expose flashable storage usable via existing QEMU driver APIs
- [ ] Sample manifests / docs updated
- [ ] Basic e2e or documented manual test: lease → flash → boot

## Tracking

- Jira: https://redhat.atlassian.net/browse/PITCREW-501
- PR: https://github.com/jumpstarter-dev/jumpstarter/pull/910
- Follow-up comment: https://github.com/jumpstarter-dev/jumpstarter/pull/910#discussion_r3659038447

mangelajo:
## Summary
- Guest disk lives at `/disk`, sized from `parameters.resources.storage`. Optional `parameters.storage.storageClassName` / `accessModes` select the Kubernetes backend (deep-merged class → set).
- No StorageClass → sized `emptyDir` plus `ephemeral-storage` requests/limits. StorageClass set → generic ephemeral `volumeClaimTemplate` so the claim dies with the Pod on ExitAndReplace.
- QEMU driver flashes/boots from `/disk`; sockets and cidata stay on `/shared`. Alpine flash/boot e2e is no longer skipped.

Closes #924
Relates to: [PITCREW-501](https://redhat.atlassian.net/browse/PITCREW-501)

## Test plan
- [x] Unit: `go test ./internal/exporterset/disk/ ./internal/exporterset/provisioners/qemu/` and exporterset reconciler tests (envtest skipped locally)
- [ ] Apply kind sample without `parameters.storage.storageClassName`; confirm Pod has `disk` emptyDir + ephemeral-storage requests/limits
- [ ] Apply with `parameters.storage.storageClassName` set; confirm ephemeral PVC mounted at `/disk`
- [ ] Confirm ExporterSet `parameters.storage.storageClassName: ""` forces emptyDir over the class
- [x] `make e2e-exporterset-qemu` — lease → flash Alpine → boot marker
- [x] Confirm ExitAndReplace deletes Pod + ephemeral claim together


bennyz:
@mangelajo maybe group it all under `parameters`?
so it's
```
parameters:
  storage:
    storageClassName: ...
    resources: ...
```

mangelajo:
> @mangelajo maybe group it all under `parameters`? so it's
> 
> ```
> parameters:
>   storage:
>     storageClassName: ...
>     resources: ...
> ```

Yes, I think that makes more sense, I thought about it when I was working on it, and thought the feature could be useful to other exporterset controllers, but we can make all the behavior reusable by common functions, and it's just a parameter.

We don't need any special checking at API level, and many exportersets don't even care about the storage classes at all.

👍 

Thanks benny

e2e/test/exporterset_qemu_test.go
```diff
@@ -159,24 +159,6 @@ var _ = Describe("ExporterSet QEMU E2E Tests", Label("exporterset-qemu"), Ordere
...
-		}, 2*time.Minute, qemuPollPeriod).ShouldNot(BeEmpty())
-
-		sizeLimit := KubectlQuery("-n", ns, "get", "pod",
-			"-l", guest.Selector,
-			"--field-selector=status.phase=Running",
-			"-o", "jsonpath={.items[0].spec.volumes[?(@.name==\"shared\")].emptyDir.sizeLimit}")
-		// Without the storage follow-up (#924), SizeLimit stays at 100Mi and
-		// flashing Alpine evicts the Pod. Skip until capacity is available.
-		if sizeLimit == "" || sizeLimit == "100Mi" {
-			Skip(fmt.Sprintf("shared emptyDir SizeLimit=%q is too small for Alpine flash; needs #924 storage work", sizeLimit))
```
mangelajo, line 177: the conditional skip is removed

controller/internal/exporterset/disk/disk.go
```diff
@@ -0,0 +1,206 @@
...
+		sc := spec.StorageClassName
+		vol.VolumeSource = corev1.VolumeSource{
+			Ephemeral: &corev1.EphemeralVolumeSource{
+				VolumeClaimTemplate: &corev1.PersistentVolumeClaimTemplate{
+					Spec: corev1.PersistentVolumeClaimSpec{
+						AccessModes:      spec.AccessModes,
+						StorageClassName: &sc,
+						Resources: corev1.VolumeResourceRequirements{
+							Requests: corev1.ResourceList{
+								corev1.ResourceStorage: spec.Size,
```
bennyz: since the driver checks for free space on the FS
```
                available = shutil.disk_usage(root.parent).free
```

do we need to account for FS overhead?

hubeadmin:
General question, non-blocking for the PR.
Have we considered using something like https://github.com/knadh/koanf for data input validation?
There's some benefit to using it, cutting out some number of lines for repetitive type assertions and value validation that I see in this PR, so it just had me wondering whether there cost of adding it would be worth it for cleaner code.

Again, just theoretical questions, nothing strictly related to code changes here.

controller/internal/exporterset/disk/disk_test.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if !spec.UsePVC() {
+		t.Error("expected PVC when storageClassName is set")
+	}
+	if len(spec.AccessModes) != 2 {
+		t.Fatalf("accessModes = %v", spec.AccessModes)
+	}
+}
+
+func TestFromParameters_emptyStorageClassForcesEmptyDir(t *testing.T) {
```
maboras-rh, line 71: nit: `TestFromParameters_emptyStorageClassForcesEmptyDir` covers the parser, but the PR test plan also calls out ExporterSet overriding a class-level `storageClassName` with `""`. A `deepMergeParameters` + `FromParameters` test (class=`gp3`, set=`""`) would lock in that merge semantics end-to-end

controller/internal/exporterset/disk/disk.go
```diff
@@ -0,0 +1,206 @@
...
+	out := make([]corev1.PersistentVolumeAccessMode, 0, len(items))
+	for _, item := range items {
+		s, ok := item.(string)
+		if !ok || s == "" {
+			return nil, fmt.Errorf("parameters.storage.accessModes must be a list of strings, got %T", item)
+		}
+		out = append(out, corev1.PersistentVolumeAccessMode(s))
+	}
+	return out, nil
+}
```
maboras-rh, line 283: No validation against known Kubernetes access modes (`ReadWriteOnce`, `ReadOnlyMany`, etc.). A typo like `ReadWriteEverywhere` is silently accepted and only fails at PVC creation time, making it harder to diagnose.
I think it worth validating against the known set. WDYT?

controller/internal/exporterset/provisioners/qemu/qemu_test.go
maboras-rh, line 1: `assertRenderPodSharedVolume` and `TestRenderPod_diskEmptyDirUsesParamSize` access the volume by index `pod.Spec.Volumes[1]`, coupling to append order. The ephemeral PVC test already does this correctly by searching by name. Suggest making the other tests consistent.

something like this example :-
```
var diskVol *corev1.Volume
for i := range pod.Spec.Volumes {
    if pod.Spec.Volumes[i].Name == disk.VolumeName {
        diskVol = &pod.Spec.Volumes[i]
        break
    }
}
```

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -482,13 +484,26 @@ def __post_init__(self):
...
 
+    @property
+    def _disk_dir(self) -> str:
+        """Directory for flashable guest disk images (root, bios, …)."""
+        if self.launcher_socket:
+            work = Path(self._work_dir)
+            # Production sidecar: sockets on /shared, guest disk on /disk.
+            # Tests use a tmp shared dir — keep disks next to sockets.
+            if work == Path("/shared"):
+                return "/disk"
```
maboras-rh, line 503: Hardcoded "/disk" and Path("/shared") in Python are implicitly coupled with Go constants `disk.MountPath` and `sharedMountPath`. If either side changes, the other breaks silently. A comment cross-referencing the Go source would help.

controller/internal/exporterset/disk/disk.go
maboras-rh: +1 
The QEMU driver checks free space with `shutil.disk_usage(root.parent).free` against the requested image size, but the provisioner only sets `emptyDir.sizeLimit` (and matching `ephemeral-storage` requests). Those are related but not identical:

- `sizeLimit` is enforced by the kubelet on the volume
- `disk_usage().free` is whatever the node filesystem actually has under `/disk` at flash time
- No inode/metadata headroom is reserved

For e2e (10Gi Alpine tiny) this is fine. For production, worth documenting whether operators should pad `parameters.resources.storage` above the largest expected image, or whether we want the driver/provisioner to align on a single notion of “available disk”.

maboras-rh (approved):
LGTM, with some minor nitpicks

controller/internal/exporterset/disk/disk.go
mangelajo: Agree a typo is annoying, but I would rather defer strict validation here: Kubernetes already rejects unknown access modes at PVC creation with a clear API error, and hardcoding the set in our controller means we need to update it when Kubernetes adds modes. Added a comment in parseAccessModes noting that K8s validates at claim time.

mangelajo:
Re koanf: good idea to keep in mind. For this PR the parameter parsing is localized to the new disk package and mostly type/shape checks, so I would not pull in koanf yet. If we end up with several provisioners sharing richer parameter schemas, revisiting a small validation library would be worth a separate discussion.

controller/internal/exporterset/disk/disk.go
```diff
@@ -0,0 +1,206 @@
...
+		spec.AccessModes = modes
+	}
+	return spec, nil
+}
+
+// SizeFromParameters reads parameters.resources.storage, defaulting to DefaultSize.
+func SizeFromParameters(params map[string]interface{}) (resource.Quantity, error) {
+	raw := DefaultSize
+	if params != nil {
+		if resources, ok := params["resources"].(map[string]interface{}); ok {
```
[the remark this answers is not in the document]
mangelajo: Fixed in caeec2e9: size now comes from `parameters.storage.size`, so malformed `parameters.resources` no longer affects disk sizing silently.

mangelajo:
Done in caeec2e9: guest disk config is grouped under `parameters.storage` — `size`, `fsOverhead`, `storageClassName`, and `accessModes`. CPU/memory stay under `parameters.resources` for scheduling/QEMU defaults.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -311,6 +326,23 @@ func (p *Provisioner) RenderPod(
...
+	if diskSpec.UsePVC() {
+		// fsGroup so the non-root exporter can write the ephemeral claim.
+		if pod.Spec.SecurityContext == nil {
+			pod.Spec.SecurityContext = &corev1.PodSecurityContext{}
+		}
+		pod.Spec.SecurityContext.FSGroup = &runAsExporter
+	} else {
+		// emptyDir guest disks consume node ephemeral storage — ensure the
+		// scheduler and kubelet account for it on containers that mount /disk.
+		disk.SetEphemeralStorage(&pod.Spec.Containers[0].Resources, diskSpec.Size)
```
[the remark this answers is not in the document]
mangelajo: Fixed in caeec2e9: `SetEphemeralStorage` is applied on the runtime sidecar only. The exporter no longer reserves guest-disk ephemeral-storage (scheduler was double-counting).

controller/internal/exporterset/disk/disk.go
mangelajo: Yes — added `parameters.storage.fsOverhead` (default `"10%"`, set `"0%"` to disable). Logical `storage.size` stays the user-facing guest disk size; volume/PVC/emptyDir requests use size × (1 + overhead).

controller/internal/exporterset/disk/disk_test.go
mangelajo: Added `TestFromParameters_mergedEmptyStorageClassForcesEmptyDir` in caeec2e9: deep-merge class `storageClassName: gp3` with set `storageClassName: ""`, then `FromParameters` → emptyDir.

controller/internal/exporterset/provisioners/qemu/qemu_test.go
mangelajo: Done in caeec2e9: tests now find the disk volume by `disk.VolumeName` via `findVolumeByName`.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
mangelajo: Added cross-reference comments in caeec2e9 pointing at `disk.MountPath` and `sharedMountPath` in the Go provisioner sources.

bennyz (approved):

maboras-rh (approved):

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
maboras-rh: ack, thanks!

mangelajo:
oops linter

PR #929, opened 2026-07-29
feat: exporter-as-main QEMU pods with ExitAndReplace recycle e2e

mangelajo:
## Summary

Make QEMU ExporterSet pods follow an exporter-as-main / runtime-as-sidecar model so Pod lifecycle, logs, and ExitAndReplace recycling behave correctly, and cover that path in the main e2e suite.

### Pod layout (`qemu.jumpstarter.dev`)
- **exporter** is the main container (`jmp run`, UID 65532) — default `kubectl logs`, and exporter exit completes the Pod (`restartPolicy: Never`).
- **target-runtime** is a native sidecar (`restartPolicy: Always`, runs as root) — starts first so `launcher.sock` is ready; runs QEMU via `jumpstarter-exec`.
- **copy-jumpstarter-exec** init container stages the binary from the **exporter** image onto `/shared` (not baked into each runtime image).
- Shared `emptyDir` holds Unix sockets (QMP, serial, launcher), disk, and cloud-init cidata (world-traversable so the runtime can mount vvfat across the UID boundary).
- Runtime creates sockets with permissive mode so the non-root exporter can connect.
- Normalize Kubernetes quantity suffixes (`10Gi` → `10G`) for QEMU driver config.

### Lifecycle / recycle
- Add `jumpstarter-exec shutdown` and best-effort runtime teardown from the exporter on `exitOnLeaseEnd`.
- ExitAndReplace: delete unleased exporters whose Pods are `Succeeded`/`Failed`, then refill `minAvailableReplicas` (avoids Completed/Offline zombies and `maxReplicas` stalls).

### E2E / CI
- Fold ExporterSet QEMU coverage into the main `e2e-tests` job; add `make e2e-exporterset-qemu` for focused local runs.
- Build/load exporter + qemu-runtime images in CI; prefetch Alpine guest image during `e2e-setup` (CI and local).
- Specs: Online/Ready Pod; power cycle + ExitAndReplace recycle stays responsive.
- Flash/boot remains skipped while shared emptyDir SizeLimit is `100Mi` (needs #924).
- Share OIDC/legacy client–exporter helpers in `e2e/test/utils.go`; cover `shutdown_runtime_sidecar` paths for diff-cover.
- Update JEP-0014 to match the exporter-main / runtime-sidecar model.

## Test plan
- [x] `go test ./internal/exporterset/...` (ExitAndReplace cleanup + RenderPod layout)
- [x] `make e2e-exporterset-qemu` — Online/Ready + power cycle/recycle; flash/boot skipped on 100Mi SizeLimit
- [x] `make -C python pkg-test-jumpstarter` — includes `shutdown_runtime_sidecar` coverage
- [x] CI `e2e-tests` on this PR
- [x] Manual: apply `controller/hack/sample-x86_64-kind.yaml`, lease/release, confirm Completed Pod is deleted and a new Ready instance appears

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -227,8 +227,9 @@ jobs:
         uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
         with:
           path: /tmp/qemu-runtime-image.tar
-          # Separate from exporter cache: only QEMU runtime image build inputs.
-          key: qemu-runtime-image-${{ matrix.arch }}-${{ hashFiles('controller/Makefile', 'controller/Containerfile.qemu-runtime') }}
+          # Bust when QEMU runtime inputs change: Containerfile embeds
+          # jumpstarter-exec from rust/, and the Makefile selects build context.
+          key: qemu-runtime-image-${{ matrix.arch }}-${{ hashFiles('controller/Makefile', 'controller/Containerfile.qemu-runtime', 'rust/jumpstarter-exec/**') }}
 
       - name: Build qemu-runtime image
         if: steps.cache.outputs.cache-hit != 'true'
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -172,6 +172,74 @@ jobs:
           path: /tmp/exporterset-controller-image.tar
           retention-days: 1
 
+  build-exporter-image:
+    needs: changes
+    if: needs.changes.outputs.should_run == 'true' || github.event_name == 'workflow_dispatch'
+    strategy:
+      matrix:
+        include: ${{ fromJson(needs.changes.outputs.e2e-matrix) }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 45
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          fetch-depth: 0
+
+      - name: Cache exporter image
+        id: cache
+        uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
+        with:
+          path: /tmp/exporter-image.tar
+          key: exporter-image-${{ matrix.arch }}-${{ hashFiles('python/Containerfile', 'python/**/pyproject.toml', 'python/uv.lock', 'rust/**', 'controller/Makefile') }}
+
+      - name: Build exporter image
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          make -C controller docker-build-exporter
+          docker save quay.io/jumpstarter-dev/jumpstarter:latest -o /tmp/exporter-image.tar
+
+      - name: Upload exporter image
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: exporter-image-${{ matrix.arch }}
+          path: /tmp/exporter-image.tar
+          retention-days: 1
+
+  build-qemu-runtime-image:
+    needs: changes
+    if: needs.changes.outputs.should_run == 'true' || github.event_name == 'workflow_dispatch'
+    strategy:
+      matrix:
+        include: ${{ fromJson(needs.changes.outputs.e2e-matrix) }}
+    runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+        with:
+          fetch-depth: 0
+
+      - name: Cache qemu-runtime image
+        id: cache
+        uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
+        with:
+          path: /tmp/qemu-runtime-image.tar
+          key: qemu-runtime-image-${{ matrix.arch }}-${{ hashFiles('controller/Makefile', 'controller/Containerfile.qemu-runtime') }}
+
+      - name: Build qemu-runtime image
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          make -C controller docker-build-qemu-runtime
+          docker save quay.io/jumpstarter-dev/virtual/qemu-runtime:latest -o /tmp/qemu-runtime-image.tar
+
+      - name: Upload qemu-runtime image
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
+        with:
+          name: qemu-runtime-image-${{ matrix.arch }}
+          path: /tmp/qemu-runtime-image.tar
+          retention-days: 1
+
   build-python-wheels:
     needs: changes
     if: needs.changes.outputs.should_run == 'true' || github.event_name == 'workflow_dispatch'
@@ -215,12 +283,13 @@ jobs:
   # ===========================================================================
 
   e2e-tests:
-    needs: [changes, build-controller-image, build-operator-image, build-exporterset-controller-image, build-python-wheels]
+    needs: [changes, build-controller-image, build-operator-image, build-exporterset-controller-image, build-exporter-image, build-qemu-runtime-image, build-python-wheels]
     strategy:
       matrix:
         include: ${{ fromJson(needs.changes.outputs.e2e-matrix) }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 60
+    # Includes ExporterSet QEMU coverage (TCG); flash/boot still skipped until #924.
+    timeout-minutes: 90
     steps:
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
@@ -253,6 +322,9 @@ jobs:
           for mod in veth bridge nf_nat nf_conntrack nft_masq nft_nat nft_chain_nat; do
             sudo modprobe "$mod" 2>/dev/null || true
           done
+
+      - name: Prefetch Alpine guest image for ExporterSet QEMU
+        run: bash e2e/scripts/ensure-qemu-guest-image.sh
 
       - name: Setup e2e test environment
         run: make e2e-setup
@@ -320,7 +392,7 @@ jobs:
           artifact-name: e2e-logs-compat-old-controller
 
   e2e-compat-old-client:
-    needs: [changes, build-controller-image, build-operator-image, build-exporterset-controller-image, build-python-wheels]
+    needs: [changes, build-controller-image, build-operator-image, build-exporterset-controller-image, build-exporter-image, build-qemu-runtime-image, build-python-wheels]
     # Skip on PRs — compat tests run in merge queue and workflow_dispatch.
     if: >-
       github.event_name != 'pull_request'
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -327,9 +327,6 @@ jobs:
             sudo modprobe "$mod" 2>/dev/null || true
           done
 
-      - name: Prefetch Alpine guest image for ExporterSet QEMU
-        run: bash e2e/scripts/ensure-qemu-guest-image.sh
-
       - name: Setup e2e test environment
         run: make e2e-setup
         env:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -227,9 +227,8 @@ jobs:
         uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
         with:
           path: /tmp/qemu-runtime-image.tar
-          # Bust when QEMU runtime inputs change: Containerfile embeds
-          # jumpstarter-exec from rust/, and the Makefile selects build context.
-          key: qemu-runtime-image-${{ matrix.arch }}-${{ hashFiles('controller/Makefile', 'controller/Containerfile.qemu-runtime', 'rust/jumpstarter-exec/**') }}
+          # Bust when QEMU runtime inputs change.
+          key: qemu-runtime-image-${{ matrix.arch }}-${{ hashFiles('controller/Makefile', 'controller/Containerfile.qemu-runtime') }}
 
       - name: Build qemu-runtime image
         if: steps.cache.outputs.cache-hit != 'true'
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -191,7 +191,9 @@ jobs:
         uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
         with:
           path: /tmp/exporter-image.tar
-          key: exporter-image-${{ matrix.arch }}-${{ hashFiles('python/Containerfile', 'python/**/pyproject.toml', 'python/uv.lock', 'rust/**', 'controller/Makefile') }}
+          # Bust when exporter image inputs change: Containerfile, Python sources,
+          # lockfiles, or embedded jumpstarter-exec (rust/).
+          key: exporter-image-${{ matrix.arch }}-${{ hashFiles('python/Containerfile', 'python/packages/**/*.py', 'python/**/pyproject.toml', 'python/uv.lock', 'rust/jumpstarter-exec/**', 'controller/Makefile') }}
 
       - name: Build exporter image
         if: steps.cache.outputs.cache-hit != 'true'
@@ -225,6 +227,7 @@ jobs:
         uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
         with:
           path: /tmp/qemu-runtime-image.tar
+          # Separate from exporter cache: only QEMU runtime image build inputs.
           key: qemu-runtime-image-${{ matrix.arch }}-${{ hashFiles('controller/Makefile', 'controller/Containerfile.qemu-runtime') }}
 
       - name: Build qemu-runtime image
@@ -289,7 +292,8 @@ jobs:
         include: ${{ fromJson(needs.changes.outputs.e2e-matrix) }}
     runs-on: ${{ matrix.os }}
     # Includes ExporterSet QEMU coverage (TCG); flash/boot still skipped until #924.
-    timeout-minutes: 90
+    # Job > Ginkgo suite timeout (60m in e2e/lib/common.sh) for setup/log upload.
+    timeout-minutes: 70
     steps:
       - name: Checkout repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
--- a/controller/hack/sample-x86_64-kind.yaml
+++ b/controller/hack/sample-x86_64-kind.yaml
@@ -72,4 +72,4 @@ spec:
             arch: x86_64
             smp: 1
             mem: 1G
-            disk_size: 10G
+            disk_size: 2G
--- a/controller/internal/exporterset/exporterconfig.go
+++ b/controller/internal/exporterset/exporterconfig.go
@@ -44,8 +44,12 @@ const (
 	// configVolumeName is the volume name for the ExporterConfig Secret.
 	configVolumeName = "exporter-config"
 
-	// configMountPath is where the ExporterConfig Secret is mounted.
-	configMountPath = "/etc/jumpstarter/exporters"
+	// ExporterConfigMountPath is where the ExporterConfig Secret is mounted
+	// inside the exporter container (QEMU provisioner and injectConfigVolume).
+	ExporterConfigMountPath = "/etc/jumpstarter/exporters"
+
+	// configMountPath is the unexported alias used within this package.
+	configMountPath = ExporterConfigMountPath
 
 	// exporterContainerName is the main container that runs jmp run.
 	exporterContainerName = "exporter"
--- a/controller/internal/exporterset/exporterconfig.go
+++ b/controller/internal/exporterset/exporterconfig.go
@@ -47,7 +47,7 @@ const (
 	// configMountPath is where the ExporterConfig Secret is mounted.
 	configMountPath = "/etc/jumpstarter/exporters"
 
-	// exporterContainerName is the init-container name in the sidecar Pod.
+	// exporterContainerName is the main container that runs jmp run.
 	exporterContainerName = "exporter"
 )
 
--- a/controller/internal/exporterset/provisioners/qemu/qemu.go
+++ b/controller/internal/exporterset/provisioners/qemu/qemu.go
@@ -135,11 +135,15 @@ func (p *Provisioner) resolveImageSpec(spec *virtualtargetv1alpha1.ImageSpec, de
 //
 //   - copy-jumpstarter-exec (regular init container) copies the
 //     jumpstarter-exec binary onto the shared volume and exits.
-//   - Exporter sidecar (init container with restartPolicy: Always)
-//     starts next and drains last; registers with the controller.
-//   - QEMU runtime (main container) runs the virtual machine.
-//   - Shared emptyDir volume for Unix socket communication
-//     (QMP, serial console, launcher socket).
+//   - target-runtime (native sidecar, restartPolicy: Always) starts
+//     next so launcher.sock is ready before the exporter; runs
+//     jumpstarter-exec serve / QEMU.
+//   - exporter (main container) runs `jmp run` — default kubectl logs
+//     target; when it exits (exitOnLeaseEnd / ExitAndReplace),
+//     Kubernetes terminates sidecars and the Pod completes.
+//     Pod restartPolicy is Never so a clean exporter exit is not
+//     restarted in-place (ExporterSet replaces the instance instead).
+//   - Shared emptyDir for Unix sockets (QMP, serial, launcher) and disk.
 //
 // The caller (reconciler) is responsible for setting
 // OwnerReferences on the Pod and injecting the config volume.
@@ -190,6 +194,9 @@ func (p *Provisioner) RenderPod(
 	pod := &corev1.Pod{
 		ObjectMeta: podMeta,
 		Spec: corev1.PodSpec{
+			// Never: ExitAndReplace relies on exporter (main) exit completing
+			// the Pod. Always would restart jmp run in-place and skip recycle.
+			RestartPolicy: corev1.RestartPolicyNever,
 			InitContainers: []corev1.Container{
 				{
 					Name:            "copy-jumpstarter-exec",
@@ -208,17 +215,14 @@ func (p *Provisioner) RenderPod(
 					},
 				},
 				{
-					Name:            "exporter",
-					Image:           exporterImage,
-					ImagePullPolicy: exporterPullPolicy,
+					// Native sidecar: starts before the main exporter so
+					// launcher.sock exists when jmp run begins. Torn down
+					// automatically when the exporter (main) container exits.
+					Name:            "target-runtime",
+					Image:           runtimeImage,
+					ImagePullPolicy: runtimePullPolicy,
 					RestartPolicy:   &restartAlways,
-					Command:         []string{"jmp", "run", "--exporter-config", configMountPath + "/config.yaml"},
-					Env: []corev1.EnvVar{
-						{
-							Name:  "JUMPSTARTER_LAUNCHER_SOCKET",
-							Value: launcherSocketPath,
-						},
-					},
+					Env:             runtimeEnv,
 					VolumeMounts: []corev1.VolumeMount{
 						{
 							Name:      sharedVolumeName,
@@ -229,10 +233,16 @@ func (p *Provisioner) RenderPod(
 			},
 			Containers: []corev1.Container{
 				{
-					Name:            "target-runtime",
-					Image:           runtimeImage,
-					ImagePullPolicy: runtimePullPolicy,
-					Env:             runtimeEnv,
+					Name:            "exporter",
+					Image:           exporterImage,
+					ImagePullPolicy: exporterPullPolicy,
+					Command:         []string{"jmp", "run", "--exporter-config", configMountPath + "/config.yaml"},
+					Env: []corev1.EnvVar{
+						{
+							Name:  "JUMPSTARTER_LAUNCHER_SOCKET",
+							Value: launcherSocketPath,
+						},
+					},
 					VolumeMounts: []corev1.VolumeMount{
 						{
 							Name:      sharedVolumeName,
@@ -264,8 +274,13 @@ func (p *Provisioner) RenderPod(
 			pod.Spec.Tolerations = append([]corev1.Toleration(nil), vtc.Spec.Scheduling.Tolerations...)
 		}
 		if vtc.Spec.Scheduling.Resources != nil {
-			// Apply resource requirements to target-runtime
-			pod.Spec.Containers[0].Resources = *vtc.Spec.Scheduling.Resources.DeepCopy()
+			// CPU/memory belong on the runtime sidecar (where QEMU runs).
+			for i := range pod.Spec.InitContainers {
+				if pod.Spec.InitContainers[i].Name == "target-runtime" {
+					pod.Spec.InitContainers[i].Resources = *vtc.Spec.Scheduling.Resources.DeepCopy()
+					break
+				}
+			}
 		}
 	}
 
@@ -392,7 +407,30 @@ func setDefault(config map[string]interface{}, key string, params map[string]int
 	}
 
 	if val != nil {
+		// Kubernetes resource quantities use binary suffixes (Gi, Mi);
+		// the QEMU driver expects qemu-img style sizes (G, M).
+		if key == "disk_size" || key == "mem" {
+			val = normalizeQemuSize(val)
+		}
 		config[key] = val
+	}
+}
+
+// normalizeQemuSize converts Kubernetes binary quantity strings (e.g. "10Gi")
+// to the form expected by the QEMU driver / qemu-img (e.g. "10G").
+func normalizeQemuSize(v interface{}) interface{} {
+	s, ok := v.(string)
+	if !ok || len(s) < 2 {
+		return v
+	}
+	if s[len(s)-1] != 'i' {
+		return v
+	}
+	switch s[len(s)-2] {
+	case 'K', 'M', 'G', 'T', 'k', 'm', 'g', 't':
+		return s[:len(s)-1]
+	default:
+		return v
 	}
 }
 
--- a/controller/internal/exporterset/provisioners/qemu/qemu.go
+++ b/controller/internal/exporterset/provisioners/qemu/qemu.go
@@ -66,9 +66,15 @@ const (
 	// the QEMU runtime container.
 	launcherSocketPath = "/shared/launcher.sock"
 
-	// configMountPath is where the ExporterConfig Secret is mounted
-	// inside the exporter sidecar.
-	configMountPath = "/etc/jumpstarter/exporters"
+	// exporterNonRootUID is the UID for the exporter main container.
+	// The runtime sidecar runs as root so it can read exporter-created
+	// paths on the shared volume without world-writable permissions.
+	exporterNonRootUID int64 = 65532
+
+	// exporterConfigPath is the jmp run config path. Must match
+	// exporterset.ExporterConfigMountPath + "/" + exporterConfigKey
+	// (cannot import the parent package — test import cycle).
+	exporterConfigPath = "/etc/jumpstarter/exporters/config.yaml"
 
 	// QEMU driver type for identification during enrichment.
 	qemuDriverType = "jumpstarter_driver_qemu.driver.Qemu"
@@ -157,6 +163,9 @@ func (p *Provisioner) RenderPod(
 ) (*corev1.Pod, error) {
 	restartAlways := corev1.ContainerRestartPolicyAlways
 	sizeLimit := resource.MustParse(sharedVolumeSizeLimit)
+	runAsRoot := int64(0)
+	runAsExporter := exporterNonRootUID
+	exporterNonRoot := true
 
 	var exporterSpec, runtimeSpec *virtualtargetv1alpha1.ImageSpec
 	if images != nil {
@@ -218,11 +227,17 @@ func (p *Provisioner) RenderPod(
 					// Native sidecar: starts before the main exporter so
 					// launcher.sock exists when jmp run begins. Torn down
 					// automatically when the exporter (main) container exits.
+					// Runs as root so QEMU can use KVM devices and read
+					// exporter-owned paths on the shared volume.
 					Name:            "target-runtime",
 					Image:           runtimeImage,
 					ImagePullPolicy: runtimePullPolicy,
 					RestartPolicy:   &restartAlways,
 					Env:             runtimeEnv,
+					SecurityContext: &corev1.SecurityContext{
+						RunAsUser:    &runAsRoot,
+						RunAsNonRoot: boolPtr(false),
+					},
 					VolumeMounts: []corev1.VolumeMount{
 						{
 							Name:      sharedVolumeName,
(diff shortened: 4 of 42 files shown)


e2e/scripts/qemu_flash_boot.py
```diff
@@ -0,0 +1,109 @@
...
+        try:
+            with PexpectAdapter(client=qemu.console) as p:
+                p.logfile = sys.stdout.buffer
+                idx = p.expect(markers + [pexpect.TIMEOUT, pexpect.EOF], timeout=args.timeout)
+                if idx >= len(markers):
+                    print("FAILED: timed out / EOF waiting for Alpine boot marker", flush=True)
+                    try:
+                        qemu.power.off()
+                    except Exception as exc:  # noqa: BLE001 - best-effort cleanup
+                        print(f"power off after failure also failed: {exc}", flush=True)
```
mangelajo: poweroff should work, no best effort...

e2e/manifests/exporterset-qemu-kind-aarch64.yaml
```diff
@@ -0,0 +1,64 @@
...
+      storage: 10Gi
+---
+apiVersion: virtualtarget.jumpstarter.dev/v1alpha1
+kind: ExporterSet
+metadata:
+  name: x86-64-virtual-e2e
+  namespace: jumpstarter-lab
+spec:
+  minReplicas: 0
+  maxReplicas: 1
```
bkhizgiy, line 43: I think this works well for testing the replacement mechanism on exit but doesn't cover the warm pool behavior, specifically maintaining min replicas while instances are leased and the scale-up and scale-down mechanism. I know it's all covered by unit testing but maybe it's worth adding it also as an e2e test with minReplicas: 2 / maxReplicas: 3 to verify the scaling engine works correctly against a real cluster. (can be done also as a followup)

e2e/manifests/exporterset-qemu-kind-aarch64.yaml
```diff
@@ -0,0 +1,64 @@
...
+  minAvailableReplicas: 1
+  scaleDownCooldown: 5m
+  recycleStrategy: ExitAndReplace
```
bkhizgiy, line 46: do we also want to test the `recycleStrategy: InPlaceReuse` flow?

mangelajo: Yes, and many other things. That's a good topic for a follow up PR if that sounds good.

e2e/manifests/exporterset-qemu-kind-aarch64.yaml
mangelajo: Yes thanks!,  I wanted to establish the base and then continue adding items like this. Also I will have to check how feasible it is without making the tests to way too heavy.

Adding this one to the list as well.

.github/workflows/e2e.yaml
```diff
@@ -172,6 +172,74 @@ jobs:
...
+        uses: actions/cache@caa296126883cff596d87d8935842f9db880ef25 # v5
+        with:
+          path: /tmp/exporter-image.tar
+          key: exporter-image-${{ matrix.arch }}-${{ hashFiles('python/Containerfile', 'python/**/pyproject.toml', 'python/uv.lock', 'rust/**', 'controller/Makefile') }}
+
+      - name: Build exporter image
+        if: steps.cache.outputs.cache-hit != 'true'
+        run: |
+          make -C controller docker-build-exporter
+          docker save quay.io/jumpstarter-dev/jumpstarter:latest -o /tmp/exporter-image.tar
```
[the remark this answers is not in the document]
mangelajo: We probably should separate different types of cache hits depending of which parts of the code changed (rust jumpstarter-exec, python, qemu runtime containerfile, ... etc.)

.github/workflows/e2e.yaml
```diff
@@ -215,12 +283,13 @@ jobs:
...
   e2e-tests:
-    needs: [changes, build-controller-image, build-operator-image, build-exporterset-controller-image, build-python-wheels]
+    needs: [changes, build-controller-image, build-operator-image, build-exporterset-controller-image, build-exporter-image, build-qemu-runtime-image, build-python-wheels]
     strategy:
       matrix:
         include: ${{ fromJson(needs.changes.outputs.e2e-matrix) }}
     runs-on: ${{ matrix.os }}
-    timeout-minutes: 60
+    # Includes ExporterSet QEMU coverage (TCG); flash/boot still skipped until #924.
+    timeout-minutes: 90
```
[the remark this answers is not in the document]
mangelajo: 90 mins is probably way too high, 60min maximum should be good yet at this point.

controller/hack/sample-x86_64-kind.yaml
```diff
@@ -72,3 +72,4 @@ spec:
             arch: x86_64
             smp: 1
             mem: 1G
+            disk_size: 10G
```
mangelajo:
```suggestion
            disk_size: 2G
```

May be more conservative like this, to make sure CI is ok.

controller/internal/exporterset/reconciler.go
```diff
@@ -653,6 +676,99 @@ func (r *ExporterSetReconciler) cleanupDisabledExporters(
 	return deleted, nil
 }
 
+// cleanupTerminalExporters deletes unleased exporters whose Pod has reached a
+// terminal phase (Succeeded or Failed). This is the ExitAndReplace recycle
+// path: exitOnLeaseEnd completes the Pod, then the controller deletes the
+// Exporter (cascading the Pod) so scale-up can refill minAvailableReplicas.
+// Without this, Offline/Succeeded instances inflate replicas, block warm-buffer
+// refill at maxReplicas, and leave Completed Pods behind.
+func (r *ExporterSetReconciler) cleanupTerminalExporters(
```
mangelajo, line 699: **Follow-up (not blocking this PR):** disable-before-delete grace for ExitAndReplace.

Today we delete as soon as Pods are terminal and `LeaseRef` is nil. That can race with the lease controller (separate process): after release the exporter may still look Online/`Available` long enough to accept a new lease while we delete it (worse if unregister/Offline lags).

Plan (tracked in #935):
1. On terminal + unleased + still enabled → set `enabled=false`, stamp a recycle annotation, `RequeueAfter: 10s`.
2. Delete only from **`cleanupTerminalExporters`** once Offline **or** disabled ≥ 10s (and still unleased).
3. **`cleanupDisabledExporters`** must skip recycle-stamped exporters so scale-down cleanup does not bypass the grace window.

Keeping that out of this PR to avoid growing the patch.

e2e/manifests/exporterset-qemu-kind-aarch64.yaml
mangelajo: Yes, I makes total sense, I will work on a follow up to not make this bigger.

I want to verify how does it behave in terms of pressure for the CI, it probably shouldn't be too bad if we don't start (power on) all the qemus at once.

e2e/manifests/exporterset-qemu-kind-aarch64.yaml
mangelajo: Yes, will do an an independent PR.

e2e/scripts/ensure-qemu-guest-image.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
+# Ensure the Alpine UEFI tiny guest image is available for exporterset-qemu e2e.
+#
+# Resolution order:
+#   1. JUMPSTARTER_E2E_QEMU_IMAGE (absolute path or URL — printed and exited)
+#   2. Existing file at e2e/testdata/<default name>
+#   3. Copy from python/packages/jumpstarter-driver-qemu/images/ if present
+#   4. Download from ALPINE_IMAGE_URL
```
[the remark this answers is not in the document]
mangelajo: let's just correct the comment.

e2e/scripts/qemu_flash_boot.py
```diff
@@ -0,0 +1,109 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Flash a guest image to a leased ExporterSet QEMU target and wait for boot.
+
+Intended to run under `jmp shell`, which sets JUMPSTARTER_HOST so env() works:
+
+    jmp shell --client … --selector board=… -- \\
+        python3 e2e/scripts/qemu_flash_boot.py /path/to/image.qcow2
```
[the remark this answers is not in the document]
mangelajo: This is just an script for internal E2E usage, not intended for users.

e2e/scripts/qemu_flash_boot.py
```diff
@@ -0,0 +1,109 @@
...
+    parser.add_argument(
+        "--disk-size",
+        default=os.environ.get("JUMPSTARTER_E2E_DISK_SIZE", "10G"),
+        help="Disk size passed to qemu.set_disk_size before power on",
+    )
+    parser.add_argument(
+        "--skip-flash",
+        action="store_true",
+        help="Skip flashing (use an already-written root disk)",
+    )
```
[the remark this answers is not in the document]
mangelajo: it is ok, we are doing this to temporarily skip the flashing phase until we know it's fully functional, but will always pass the image, this is internal to our E2E.

e2e/scripts/qemu_flash_boot.py
```diff
@@ -0,0 +1,109 @@
...
+    parser.add_argument("image", help="Path or URL to the guest disk image to flash")
+    parser.add_argument(
+        "--timeout",
+        type=int,
+        default=int(os.environ.get("JUMPSTARTER_E2E_CONSOLE_TIMEOUT", "900")),
+        help="Seconds to wait for a console boot marker (default: 900)",
+    )
+    parser.add_argument(
+        "--disk-size",
+        default=os.environ.get("JUMPSTARTER_E2E_DISK_SIZE", "10G"),
```
mangelajo:
```suggestion
        default=os.environ.get("JUMPSTARTER_E2E_DISK_SIZE", "2G"),
```

or consistent with any previous comments.

e2e/scripts/qemu_flash_boot.py
```diff
@@ -0,0 +1,109 @@
...
+                qemu.power.off()
+            except Exception as exc:  # noqa: BLE001 - best-effort cleanup
+                print(f"power off failed: {exc}", flush=True)
+
+    print("done", flush=True)
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main())
```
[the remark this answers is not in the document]
mangelajo: No this is e2e specific, equivalent to an .sh No need to put everything under python.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
```diff
@@ -538,9 +538,14 @@ def validate_partition(
...
-        tmp = TemporaryDirectory()
-
+        # In sidecar mode QEMU runs in the runtime container and can only
+        # see paths on the shared volume — never the exporter's /tmp.
+        # Mode must be world-traversable: TemporaryDirectory defaults to 0o700,
+        # but the runtime sidecar runs as UID 65532 while the exporter is root.
+        tmp = TemporaryDirectory(dir=self._work_dir)
         path = Path(tmp.name)
+        path.chmod(0o755)
+
```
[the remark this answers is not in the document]
mangelajo: Our runtime sidecar should probably run as root, while the main exporter probably should not need full root permissions. Can we turn it around?

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -36,23 +36,72 @@
...
+    cmd = [str(exec_bin), "shutdown", "--socket", sock]
+    logger.info("Shutting down runtime sidecar via %s", " ".join(cmd))
+    try:
+        result = subprocess.run(cmd, check=False, capture_output=True, text=True, timeout=timeout)
+    except FileNotFoundError:
+        logger.warning("jumpstarter-exec not executable at %s", exec_bin)
+        return False
+    except subprocess.TimeoutExpired:
+        logger.warning("jumpstarter-exec shutdown timed out after %ss", timeout)
+        return False
```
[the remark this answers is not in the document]
mangelajo: let's catch the other exceptions, for example, there not being able to execute or anything similar.

rust/jumpstarter-exec/tests/integration.rs
```diff
@@ -439,6 +439,25 @@ fn e2e_echo() {
...
+        .expect("failed to run jumpstarter-exec shutdown");
+    assert!(status.success(), "shutdown should exit 0");
+
+    let wait = server
+        .wait()
+        .expect("failed to wait for serve process after shutdown");
+    assert!(
+        wait.success(),
+        "serve should exit 0 after shutdown, got {wait}"
+    );
```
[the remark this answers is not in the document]
mangelajo: I like the timeout, yes. Let's add it.

rust/jumpstarter-exec/src/server.rs
```diff
@@ -91,11 +96,30 @@ fn handle_connection(stream: UnixStream, log: Arc<Logger>) -> std::io::Result<()
...
+    match msg {
+        ClientMessage::Shutdown => {
+            log.info("shutdown requested", &[("socket", json!(socket_path))]);
+            // Acknowledge before exiting so the client can observe success.
+            send(&writer, &ServerMessage::Exit { code: Some(0) })?;
+            // Drop the writer to flush the ack, remove the socket, then exit
+            // so the container's PID 1 terminates and Kubernetes replaces the Pod.
+            drop(writer);
+            let _ = std::fs::remove_file(socket_path);
+            std::process::exit(0);
```
raballew: `std::process::exit(0)` immediately terminates the process from a worker thread. Since `serve_with` spawns a new thread per connection, a Shutdown arriving while another connection is running an Exec session kills all child processes and I/O threads without signaling or draining. 

Thus, if my brain does not trick me here Rust destructors are skipped, and since jumpstarter-exec runs as PID 1 in the container, orphaned children become **zombies** - I have seen entire clusters going doing from a misconfigured daemonset creating zombies ...

controller/internal/exporterset/exporterconfig.go
```diff
@@ -47,7 +47,7 @@ const (
 	// configMountPath is where the ExporterConfig Secret is mounted.
 	configMountPath = "/etc/jumpstarter/exporters"
```
raballew: Consider exporting the constant from the `exporterset` package (e.g. `ExporterConfigMountPath`) and importing it in the QEMU provisioner.

controller/internal/exporterset/reconciler.go
```diff
@@ -653,6 +676,99 @@ func (r *ExporterSetReconciler) cleanupDisabledExporters(
...
+				"Deleted Exporter %s after terminal Pod", exp.Name)
+		}
+		deleted = true
+	}
+
+	return deleted, nil
+}
+
+// listPodsGroupedByExporter maps Exporter names to their owned Pods.
+func (r *ExporterSetReconciler) listPodsGroupedByExporter(
```
raballew: This new `listPodsGroupedByExporter` function (returns `map[string][]corev1.Pod`) issues the same List call with the same label selector and ownerReference filter as the existing `listPodsByExporter` (which returns `map[string]struct{}`). Both queries hit the API server or informer cache with identical parameters in a single reconcile cycle.

Keeping only `listPodsGroupedByExporter` and deriving the existence check at the call site would remove the duplication.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1093,13 +1160,19 @@ async def serve(self):  # noqa: C901
 
                     if self.exit_on_lease_end and previous_leased:
                         logger.info("Exporter configured to exit after lease, shutting down")
+                        shutdown_runtime_sidecar()
```
raballew: Isnt this blocking for up to 10s?

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1195,14 +1197,16 @@ async def test_serve_not_triggered_on_startup(self):
...
-            await anyio.sleep(0.1)
-            # Check that _stop_requested was NOT set
-            assert exporter._stop_requested is False
-            # Clean exit
-            tg.cancel_scope.cancel()
+        with patch("jumpstarter.exporter.exporter.shutdown_runtime_sidecar"):
+            async with create_task_group() as tg:
+                tg.start_soon(exporter.serve)
+                # Give serve time to process the status
+                await anyio.sleep(0.1)
```
raballew: Using an event or condition variable to synchronize (e.g. wire `_report_status` to set an event after processing, then await that event before asserting) would make the test deterministic.

e2e/test/exit_on_lease_end_test.go
```diff
@@ -64,10 +58,7 @@ var _ = Describe("Exit On Lease End E2E Tests", Label("exit-on-lease-end"), Orde
...
 
 	AfterEach(func() {
-		if CurrentSpecReport().Failed() {
-			tracker.DumpLogs(250)
-			DumpControllerLogs(250)
-		}
+		DumpOnFailure(250, tracker.DumpLogs)
 		// Stop any running exporter between tests
 		tracker.StopAll()
 		time.Sleep(time.Second)
```
raballew: Polling `tracker.IsProcessRunning()` with a short timeout via `Eventually` instead of a fixed sleep would be faster and more reliable.

raballew (changes requested):

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: exporter image cache key now hashes `python/packages/**/*.py`, lockfiles, and `rust/jumpstarter-exec/**`. QEMU runtime cache stays on Makefile + `Containerfile.qemu-runtime` only (exec binary comes from the exporter image, not the runtime image).

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: Ginkgo full-suite timeout is **60m** (`e2e/lib/common.sh`); the e2e job is **70m** so setup/log upload still fit after the suite limit.

e2e/scripts/ensure-qemu-guest-image.sh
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: header now documents an absolute path only (no URL claim).

e2e/test/exporterset_qemu_test.go
```diff
@@ -0,0 +1,270 @@
...
+		}, 2*time.Minute, 5*time.Second).ShouldNot(BeEmpty())
+
+		sizeLimit, _ := Kubectl("-n", ns, "get", "pod",
+			"-l", exporterSetQemuSelector,
+			"-o", "jsonpath={.items[0].spec.volumes[?(@.name==\"shared\")].emptyDir.sizeLimit}")
+		// Without the storage follow-up (#924), SizeLimit stays at 100Mi and
+		// flashing Alpine evicts the Pod. Skip until capacity is available.
+		if sizeLimit == "" || sizeLimit == "100Mi" {
+			Skip(fmt.Sprintf("shared emptyDir SizeLimit=%q is too small for Alpine flash; needs #924 storage work", sizeLimit))
+		}
```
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: wait for a Running pod, then Skip when emptyDir sizeLimit is missing/too small — no longer asserting capacity via Eventually first.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver_test.py
```diff
@@ -205,6 +205,35 @@ def test_set_memory_size_invalid():
...
+    driver = Qemu(launcher_socket=str(shared / "launcher.sock"))
+    with patch.object(Qemu, "_work_dir", property(lambda self: str(shared))):
+        cidata = driver.cidata()
+        try:
+            assert Path(cidata.name).is_relative_to(shared)
+            assert oct(Path(cidata.name).stat().st_mode & 0o777) == "0o755"
+            assert (Path(cidata.name) / "meta-data").is_file()
+            assert (Path(cidata.name) / "user-data").is_file()
+        finally:
+            cidata.cleanup()
```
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: cidata test constructs the driver with `launcher_socket` under the shared dir and no longer patches `_work_dir`.

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: dropped the unconditional `chmod(0o755)`. QEMU ExporterSet pods now run `target-runtime` as root and the exporter as non-root (`65532`), so the runtime can read exporter-created `0o700` cidata on the shared volume.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -36,23 +36,72 @@
...
+        logger.warning(
+            "jumpstarter-exec shutdown exited %s: stdout=%r stderr=%r",
+            result.returncode,
+            result.stdout,
+            result.stderr,
+        )
+        return False
+
+    logger.info("Runtime sidecar shutdown acknowledged")
+    return True
```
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: call sites use `await anyio.to_thread.run_sync(shutdown_runtime_sidecar)` so the subprocess work stays off the event loop.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: `shutdown_runtime_sidecar` now catches `OSError` (plus `TimeoutExpired`), logs, and returns `False`.

rust/jumpstarter-exec/tests/integration.rs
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: `e2e_shutdown_exits_serve` polls `try_wait` with a 5s deadline and kills + fails on timeout.

e2e/scripts/qemu_flash_boot.py
mangelajo: Done in 45d69a67: success-path `power.off()` is no longer best-effort — failures propagate. Best-effort remains only on the failure/timeout cleanup branch.

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1214,14 +1218,126 @@ async def test_serve_continues_when_disabled(self):
...
+        binary.write_text("#!/bin/sh\nexit 0\n")
+        binary.chmod(0o755)
+
+        monkeypatch.setenv("JUMPSTARTER_LAUNCHER_SOCKET", str(sock))
+
+        with patch("subprocess.run") as run:
+            run.return_value = MagicMock(returncode=0, stdout="", stderr="")
+            # No binary= arg → default /shared/jumpstarter-exec missing → fall back.
+            assert shutdown_runtime_sidecar() is True
+            assert run.call_args.args[0][0] == str(binary)
```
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: fallback test patches `Path.is_file` so `/shared/jumpstarter-exec` presence cannot flake the assertion.

controller/hack/sample-x86_64-kind.yaml
mangelajo: Done in 45d69a67: sample uses `disk_size: 2G`.

e2e/scripts/qemu_flash_boot.py
mangelajo: Done in 45d69a67: default is now `2G` (and the e2e flash invocation passes `--disk-size 2G`).

python/packages/jumpstarter-driver-qemu/jumpstarter_driver_qemu/driver.py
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: turned around — `target-runtime` runs as root; exporter main is non-root (`65532` + `runAsNonRoot`). Dropped the cidata `chmod(0o755)` since root runtime can read exporter-owned `0o700` dirs.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: catching `OSError` broadly (covers PermissionError / not-executable / etc.) and returning `False`.

rust/jumpstarter-exec/tests/integration.rs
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: bounded `try_wait` loop (5s) with kill + fail on timeout.

rust/jumpstarter-exec/src/server.rs
mangelajo: Done in 45d69a67: Shutdown no longer calls `process::exit`. It acks the client, sets a shared flag, SIGTERMs tracked Exec children, and lets `serve_with` return cleanly so handlers can unwind.

controller/internal/exporterset/exporterconfig.go
mangelajo: Exported `ExporterConfigMountPath` from `exporterconfig.go` in 45d69a67. The QEMU provisioner still uses a local `exporterConfigPath` constant (same value) — importing the parent `exporterset` package from `provisioners/qemu` creates a test import cycle (`exporterset` tests import qemu). Left a comment pointing at the shared constant.

controller/internal/exporterset/reconciler.go
mangelajo: Done in 45d69a67: `listPodsByExporter` now derives from `listPodsGroupedByExporter` (single List implementation).

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
mangelajo: Yes — fixed in 45d69a67 by running it via `await anyio.to_thread.run_sync(shutdown_runtime_sidecar)` at the call sites.

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
mangelajo: Done in 45d69a67: exit-on-lease tests synchronize with an `anyio.Event` after statuses are processed instead of fixed sleeps.

e2e/test/exit_on_lease_end_test.go
mangelajo: Done in 45d69a67: `AfterEach` uses `Eventually` until `!tracker.IsProcessRunning()` after `StopAll()`.

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: exporter cache keyed on Python packages + rust exec + Containerfile/lockfiles; qemu-runtime cache remains separate (Makefile + Containerfile only).

.github/workflows/e2e.yaml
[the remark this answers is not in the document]
mangelajo: Done in 45d69a67: suite capped at 60m; job at 70m for buffer.

.github/workflows/e2e.yaml
```diff
@@ -254,6 +327,9 @@ jobs:
             sudo modprobe "$mod" 2>/dev/null || true
           done
 
+      - name: Prefetch Alpine guest image for ExporterSet QEMU
+        run: bash e2e/scripts/ensure-qemu-guest-image.sh
```
mangelajo: this must run from the setup-e2e make target, not be part of github ci yaml.

e2e/manifests/exporterset-qemu-kind-aarch64.yaml
bkhizgiy: ack, sounds good, we can do it as a follow up then.

controller/internal/exporterset/reconciler.go
```diff
@@ -653,6 +695,73 @@ func (r *ExporterSetReconciler) cleanupDisabledExporters(
...
+	if len(pods) == 0 {
+		return false
+	}
+	for i := range pods {
+		if !isTerminalPodPhase(pods[i].Status.Phase) {
+			return false
+		}
+	}
+	return true
+}
```
bkhizgiy, line 754: with the new pod layout, the exporter can exit while the native sidecar is still running, so k8s keeps the pod in the running phase. As a result `allPodsTerminal()` wont returns true, the pod isn't cleaned up, and warm buffer scaling is blocked because it still counts toward the replica total.

mangelajo: That was a problem on the initial version, but I re-organized the containers, now the main container is the exporter, so when it exits, the other sidecar is torn down:

When the main exporter exits, kubelet terminates the sidecar (SIGTERM → SIGKILL after grace). Sidecars do not block Pod completion. The Pod becomes Succeeded/Failed, then allPodsTerminal() can return true. The QEMU recycle e2e already exercises that path.

We also added a shutdown signal from the exporter to the jumpstarter-exec to make it happen more quickly.

mangelajo:
I am handling a few of the comments I had around and will push a newer version.

controller/internal/exporterset/reconciler.go
bkhizgiy: ack, thanks for clarifying.

e2e/test/exit_on_lease_end_test.go
```diff
@@ -64,13 +58,13 @@ var _ = Describe("Exit On Lease End E2E Tests", Label("exit-on-lease-end"), Orde
...
-		}
-		// Stop any running exporter between tests
+		DumpOnFailure(250, tracker.DumpLogs)
+		// Stop any running exporter between tests and wait until it is gone.
 		tracker.StopAll()
-		time.Sleep(time.Second)
+		Eventually(func() bool {
+			return tracker.IsProcessRunning()
+		}, 10*time.Second, 100*time.Millisecond).Should(BeFalse(),
+			"exporter process should stop after StopAll")
```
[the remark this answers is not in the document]
mangelajo: Fixed: `StopAll` now waits until tracked PIDs are gone before clearing the list, and `AfterEach` snapshots `TrackedPIDs()` then polls `AnyPIDAlive(pids)` so the assertion observes real termination rather than an emptied tracker.

rust/jumpstarter-exec/src/server.rs
```diff
@@ -91,11 +131,40 @@ fn handle_connection(stream: UnixStream, log: Arc<Logger>) -> std::io::Result<()
...
+            send(&writer, &ServerMessage::Exit { code: Some(0) })?;
+            drop(writer);
+            let _ = std::fs::remove_file(socket_path);
+            let pids: Vec<u32> = state.children.lock().unwrap().iter().copied().collect();
+            for pid in pids {
+                unsafe {
+                    kill(pid as i32, 15);
+                }
+            }
+            state.shutdown.store(true, Ordering::SeqCst);
```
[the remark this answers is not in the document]
mangelajo: Fixed: shutdown flag and child PIDs now share one `Mutex` lifecycle. `Shutdown` sets `shutdown=true` and snapshots PIDs under that lock; `Exec` after spawn only registers if shutdown is not active, otherwise it SIGTERMs/reaps the new child. Added `shutdown_sigterms_in_flight_exec` integration coverage.

bkhizgiy (approved):

controller/internal/exporterset/reconciler.go
```diff
@@ -653,6 +695,73 @@ func (r *ExporterSetReconciler) cleanupDisabledExporters(
...
+
+		if r.Recorder != nil {
+			r.Recorder.Eventf(es, corev1.EventTypeNormal, "Recycle",
+				"Deleted Exporter %s after terminal Pod", exp.Name)
+		}
+		deleted = true
+	}
+
+	return deleted, nil
+}
```
raballew, line 742: `cleanupTerminalExporters` deletes Exporter CRs without checking `RecycleStrategy`. An ExporterSet using `InPlaceReuse` whose pod fails for an external reason (OOM, node eviction) will have `allPodsTerminal` return `true` and its Exporter CR permanently deleted. No test ever checks this behaviour.

controller/internal/exporterset/provisioners/qemu/qemu.go
```diff
@@ -264,8 +297,13 @@ func (p *Provisioner) RenderPod(
 			pod.Spec.Tolerations = append([]corev1.Toleration(nil), vtc.Spec.Scheduling.Tolerations...)
 		}
 		if vtc.Spec.Scheduling.Resources != nil {
-			// Apply resource requirements to target-runtime
-			pod.Spec.Containers[0].Resources = *vtc.Spec.Scheduling.Resources.DeepCopy()
+			// CPU/memory belong on the runtime sidecar (where QEMU runs).
+			for i := range pod.Spec.InitContainers {
+				if pod.Spec.InitContainers[i].Name == "target-runtime" {
```
raballew: `"target-runtime"` appears as a raw literal in two places. Add `const runtimeContainerName = "target-runtime"`.

controller/internal/exporterset/provisioners/qemu/qemu_test.go
```diff
@@ -111,16 +111,31 @@ func TestRenderPod_copiesMetadataAndAppliesDefaults(t *testing.T) {
...
+		*exporter.SecurityContext.RunAsUser != exporterNonRootUID {
+		t.Errorf("exporter RunAsUser = %v, want %d", exporter.SecurityContext, exporterNonRootUID)
+	}
+	if exporter.SecurityContext.RunAsNonRoot == nil || !*exporter.SecurityContext.RunAsNonRoot {
+		t.Errorf("exporter RunAsNonRoot = %v, want true", exporter.SecurityContext.RunAsNonRoot)
+	}
+	if pod.Spec.RestartPolicy != corev1.RestartPolicyNever {
+		t.Errorf("RestartPolicy = %q, want Never (ExitAndReplace)", pod.Spec.RestartPolicy)
+	}
 }
```
raballew, line 166: `TestRenderPod_copiesMetadataAndAppliesDefaults` checks container names, restart policies, and security contexts but does not verify `VolumeMount` entries.

controller/internal/exporterset/reconciler.go
```diff
@@ -653,6 +695,73 @@ func (r *ExporterSetReconciler) cleanupDisabledExporters(
...
+// Without this, Offline/Succeeded instances inflate replicas, block warm-buffer
+// refill at maxReplicas, and leave Completed Pods behind.
+func (r *ExporterSetReconciler) cleanupTerminalExporters(
+	ctx context.Context,
+	es *virtualtargetv1alpha1.ExporterSet,
+	exporters []jumpstarterdevv1alpha1.Exporter,
+) (bool, error) {
+	logger := log.FromContext(ctx)
+
+	podsByExporter, err := r.listPodsGroupedByExporter(ctx, es)
```
raballew: `listPodsGroupedByExporter` is called twice per steady-state reconcile cycle: once inside `cleanupTerminalExporters` and once inside `ensureExporterPods` via `listPodsByExporter`. The comment states otherwise.

controller/internal/exporterset/reconciler_test.go
```diff
@@ -580,6 +583,131 @@ func TestReconcile_scaleDown_doesNotDeleteLeasedDisabledExporter(t *testing.T) {
...
+	// Second reconcile: warm buffer scale-up creates a replacement.
+	reconcileOnce(t, r)
+	exporters := listExporters(t, c)
+	if len(exporters) != 1 {
+		t.Fatalf("after refill: expected 1 exporter, got %d", len(exporters))
+	}
+	if exporters[0].Name == "exp-dead" {
+		t.Fatal("replacement should be a new Exporter, not exp-dead")
+	}
+}
```
raballew, line 763: The five `TestReconcile_exitAndReplace_*` tests all supply at least one pod. The `allPodsTerminal` function returns `false` when the pod slice is empty (line 750-751), protecting freshly-created exporters from deletion. If that guard were inverted or removed, all five tests would still pass while new exporters would be silently deleted on every reconcile.

e2e/scripts/qemu_flash_boot.py
```diff
@@ -0,0 +1,106 @@
...
+                    print("FAILED: timed out / EOF waiting for Alpine boot marker", flush=True)
+                    try:
+                        qemu.power.off()
+                    except Exception as exc:  # noqa: BLE001 - best-effort cleanup
+                        print(f"power off after failure also failed: {exc}", flush=True)
+                    return 1
+                print(f"OK: matched marker {markers[idx]!r}", flush=True)
+        finally:
+            print("power off...", flush=True)
+            qemu.power.off()
```
raballew, line 97: Remove the `qemu.power.off()` call from inside the timeout branch and rely entirely on the `finally` block. Wrap the `finally` call in a try/except so transport errors are logged rather than propagated.

e2e/scripts/ensure-qemu-guest-image.sh
```diff
@@ -0,0 +1,46 @@
...
+  echo "Copying guest image from ${QEMU_PKG_IMAGE}" >&2
+  cp -f "${QEMU_PKG_IMAGE}" "${DEST}"
+  echo "${DEST}"
+  exit 0
+fi
+
+echo "Downloading Alpine guest image from ${ALPINE_IMAGE_URL}" >&2
+tmp="${DEST}.partial"
+curl -fL --retry 3 --retry-delay 2 -o "${tmp}" "${ALPINE_IMAGE_URL}"
+mv "${tmp}" "${DEST}"
```
raballew, line 51: The `.partial` temp file created before the `curl` download has no cleanup trap. On SIGINT or a `curl` failure that causes a non-local exit, the partial file is left on disk. A subsequent run may then skip re-downloading if the stale partial overlaps with a target path check.

raballew (dismissed):

controller/internal/exporterset/reconciler.go
mangelajo: Fixed: `cleanupTerminalExporters` now no-ops when `RecycleStrategy == InPlaceReuse`. Added `TestReconcile_inPlaceReuse_keepsFailedUnleasedExporter`.

controller/internal/exporterset/provisioners/qemu/qemu.go
mangelajo: Fixed: added `runtimeContainerName` and use it for the sidecar name + scheduling Resources loop.

controller/internal/exporterset/provisioners/qemu/qemu_test.go
mangelajo: Fixed: `TestRenderPod_copiesMetadataAndAppliesDefaults` now asserts shared → `/shared` VolumeMounts on copy-jumpstarter-exec, target-runtime, and exporter.

controller/internal/exporterset/reconciler.go
mangelajo: Fixed: Reconcile lists pods once via `listPodsGroupedByExporter` and passes the map into `cleanupTerminalExporters` and `ensureExporterPods` (removed the second List / unused `listPodsByExporter`).

controller/internal/exporterset/reconciler_test.go
mangelajo: Fixed: added `TestReconcile_exitAndReplace_keepsExporterWithNoPods` and `TestAllPodsTerminal_emptyReturnsFalse` so the empty-slice guard is covered.

e2e/scripts/qemu_flash_boot.py
mangelajo: Fixed: removed the timeout-branch `power.off()`; `finally` is the single power-off path. Left `finally` without try/except so success-path failures still propagate (earlier review: poweroff should not be best-effort on success).

e2e/scripts/ensure-qemu-guest-image.sh
mangelajo: Fixed: trap removes `${DEST}.partial` on EXIT (cleared after successful `mv`). Note: a leftover `.partial` would not skip download today — the existence check is on `DEST`, not `.partial` — but the trap still avoids littering interrupted downloads.

mangelajo:
Added another commit to make sure it works on arm64 E2E tests:

https://github.com/jumpstarter-dev/jumpstarter/pull/929/changes/d723764b3650671d2fe1b4d82759d6ef28bf0931

mangelajo:
rebased.

PR #930, opened 2026-07-30
feat: structured log gathering from exporters (JEP-0013)

bkhizgiy:
Exporters now push structured logs to an optional jumpstarter-telemetry service, giving operators a single place to query what happened during any lease , which pipeline ran it, what driver operation failed, and with what result.

**What changed:**
New proto RPCs - GetServiceEndpoints lets exporters discover the telemetry service after registration. TelemetryService.PushLogs receives batched structured log entries.

Controller config - A new telemetry section in the controller ConfigMap (endpoint, certificate, min_severity) tells authenticated exporters where to send logs.

jumpstarter-telemetry service - A standalone Go binary that receives log entries and writes them to structured stdout. Ready to forward to Loki in a future phase without any protocol changes.

TelemetryLogHandler (Python) - A logging.Handler attached to the root logger after registration. It pulls correlation fields from structlog contextvars (lease_id, client, exporter) and spec.context fields (build_id, pipeline, etc.), batches them, and flushes every 500ms via PushLogs. Telemetry failures are silently dropped — device operations are never blocked.

**What you can do with it:**
A single log query like {lease="telemetry-test-4"} returns every event from that lease's full lifecycle. {pipeline="ci-main"} finds all leases across any run tagged with that pipeline. The correlation data answers questions like "which pipeline was running on this lease?" and "what driver operation failed?" without joining multiple sources.


bkhizgiy:
@coderabbitai review

bkhizgiy:
@coderabbitai review

bkhizgiy:
@coderabbitai review

protocol/proto/jumpstarter/v1/telemetry.proto
```diff
@@ -0,0 +1,42 @@
+// Copyright 2024 The Jumpstarter Authors
```
RoddieKieley: Am guessing that if this is a new protobuf definition file then this should be 2026?

python/packages/jumpstarter-protocol/jumpstarter_protocol/jumpstarter/v1/telemetry_pb2.pyi
```diff
@@ -0,0 +1,143 @@
+"""
+@generated by mypy-protobuf.  Do not edit manually!
+isort:skip_file
+Copyright 2024 The Jumpstarter Authors"""
```
RoddieKieley: s/2024/2026 from the proto before generation perhaps?

python/packages/jumpstarter-protocol/jumpstarter_protocol/jumpstarter/v1/telemetry_pb2_grpc.pyi
```diff
@@ -0,0 +1,88 @@
+"""
+@generated by mypy-protobuf.  Do not edit manually!
+isort:skip_file
+Copyright 2024 The Jumpstarter Authors"""
```
RoddieKieley: s/2024/2026 from the proto generation perhaps?

controller/internal/config/config.go
```diff
@@ -51,20 +51,20 @@ func LoadConfiguration(
 	key client.ObjectKey,
 	signer *oidc.Signer,
 	certificateAuthority string,
-) (authenticator.Token, string, Router, []grpc.ServerOption, *Provisioning, *LeasePolicy, *HiddenLabels, error) {
+) (authenticator.Token, string, Router, []grpc.ServerOption, *Provisioning, *LeasePolicy, *HiddenLabels, *Telemetry, error) {
 	var configmap corev1.ConfigMap
 	if err := client.Get(ctx, key, &configmap); err != nil {
-		return nil, "", nil, nil, nil, nil, nil, err
+		return nil, "", nil, nil, nil, nil, nil, nil, err
```
RoddieKieley: I've noticed that the LLM's don't mind not being dry, an in this case there's quite a few of this particular return value repeated. However prior to this code change it was only 1 nil different so for the moment it's probably fine without the refactor. This is probably one of those areas where adopting some clean code type rules to guide the coding agent up front could help.

controller/cmd/telemetry/main.go
```diff
@@ -0,0 +1,88 @@
+/*
+Copyright 2024.
```
RoddieKieley: s/2024/2026?

controller/internal/service/telemetry_service.go
```diff
@@ -0,0 +1,111 @@
+/*
+Copyright 2024.
```
RoddieKieley: s/2024/2026?

controller/internal/service/telemetry_service_test.go
```diff
@@ -0,0 +1,200 @@
+/*
+Copyright 2024.
```
RoddieKieley: s/2024/2026?

python/packages/jumpstarter/jumpstarter/exporter/telemetry.py
```diff
@@ -0,0 +1,133 @@
...
+from datetime import datetime, timezone
+
+import structlog.contextvars
+from anyio import sleep
+from google.protobuf.timestamp_pb2 import Timestamp
+from jumpstarter_protocol import telemetry_pb2, telemetry_pb2_grpc
+
+_BATCH_SIZE = 50
+_FLUSH_INTERVAL = 0.5  # seconds between automatic flushes
+_MAX_QUEUE_SIZE = 10_000  # entries; oldest are dropped when full
```
RoddieKieley: Any particular reason these magic numbers were chosen? Could be useful to leave a comment relating to the why if there were any particular reason they were chosen. Otherwise the existing two comments are helpful.

python/packages/jumpstarter/jumpstarter/exporter/telemetry.py
```diff
@@ -0,0 +1,133 @@
...
+_STDLIB_RECORD_ATTRS = frozenset(vars(logging.LogRecord("", 0, "", 0, "", (), None)).keys())
+
+# Map structlog context var names to proto field names where they differ.
+_CONTEXT_KEY_MAP = {"lease_id": "lease"}
+# Context keys that map to first-class proto fields (skip from extra_fields).
+_CONTEXT_PROTO_KEYS = _WELL_KNOWN_KEYS | frozenset(_CONTEXT_KEY_MAP.keys())
+
+_MAX_EXTRA_FIELDS = 16
+_MAX_KEY_LEN = 64
+_MAX_VAL_LEN = 256
```
RoddieKieley, line 38: Same as above regarding number choices. The LLM can likely connect the dots but a new human to the codebase might wonder.

controller/internal/config/types.go
```diff
@@ -14,6 +14,24 @@ type Config struct {
...
+	// Endpoint is the jumpstarter-telemetry gRPC address (e.g. "telemetry.jumpstarter.svc:9093").
+	// Leave empty to disable telemetry service discovery.
+	Endpoint string `json:"endpoint,omitempty" yaml:"endpoint,omitempty"`
+
+	// Certificate is an optional PEM-encoded CA certificate to trust for the telemetry endpoint.
+	// Leave empty to use the system certificate pool.
+	Certificate string `json:"certificate,omitempty" yaml:"certificate,omitempty"`
+
+	// MinSeverity is the minimum log severity to forward to the telemetry service.
+	// Accepted values: debug, info, warning, error.  Defaults to "info" when empty.
```
RoddieKieley: Is this correct? Is there also a critical that is mentioned elsewhere in the code?
"
"""Normalise Python logging level names to JEP-0013 severity strings."""
    return {
        "DEBUG": "debug",
        "INFO": "info",
        "WARNING": "warning",
        "ERROR": "error",
        "CRITICAL": "critical",
"

RoddieKieley (commented):
Mostly nit-pick type comments on minor details for the moment. May have more detailed comments after building the branch and testing it in-cluster.

RoddieKieley:
@bkhizgiy I ran up your exporter-logs branch on top of my local setup here and confirm local exporter logging output:
Screenshot From 2026-07-31 08-42-05

matches the log output as viewed via LogQL inside the local openshift sno cluster:
Screenshot From 2026-07-31 08-42-15

I did have to manually configure and expose a Service to connect to the jumpstarter-telemetry pods endpoint but I think that's expected at this point of the phased implementation of JEP-0013.

controller/internal/config/config.go
mangelajo: Yes, a few PRs ago we had the same conversation with HiddenLabels.

It could be worth to return all this as a config structure, and then just pass the structure to the controller later in https://github.com/jumpstarter-dev/jumpstarter/pull/930/changes#diff-032eafab48326b32dad354dd24c3da532106bde78b3589793bd8fdc47be0bb65R283 as a single thing, we could also return "nil" config :) single nil

controller/internal/config/types.go
mangelajo: def _severity_to_level(severity: str) -> int:

controller/deploy/operator/config/crd/bases/virtualtarget.jumpstarter.dev_exportersets.yaml
```diff
@@ -218,15 +218,15 @@ spec:
                               description: Config holds driver-specific configuration.
                               x-kubernetes-preserve-unknown-fields: true
                             name:
-                              description: |-
-                                Name is the key used in the ExporterConfig export map.
-                                If omitted, derived from the Type (last segment after the last dot).
+                              description: Name is the key used in the ExporterConfig
+                                export map.
```
mangelajo: this change sneaked in? :)

controller/internal/service/telemetry_service.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// BindAddr is the TCP address to listen on (e.g. ":9093").
+	BindAddr string
+}
+
+// PushLogs receives a batch of structured log entries and writes them via the
+// controller-runtime logger (structured JSON to stdout).
+// Future phase: forward to Loki push API.
+func (s *TelemetryService) PushLogs(ctx context.Context, req *pb.PushLogsRequest) (*pb.PushLogsResponse, error) {
+	logger := ctrl.Log.WithName("telemetry")
```
mangelajo, line 102: we need to authenticate the exporter from the call.

controller/internal/config/types.go
```diff
@@ -14,6 +14,24 @@ type Config struct {
 	Grpc           Grpc           `json:"grpc" yaml:"grpc"`
 	LeasePolicy    LeasePolicy    `json:"leasePolicy,omitempty" yaml:"leasePolicy,omitempty"`
 	HiddenLabels   HiddenLabels   `json:"hiddenLabels,omitempty" yaml:"hiddenLabels,omitempty"`
+	Telemetry      *Telemetry     `json:"telemetry,omitempty" yaml:"telemetry,omitempty"`
+}
```
mangelajo, line 22: The proposed JEP configuration is quite different: https://jumpstarter.dev/main/contributing/jeps/JEP-0013-observability-telemetry-logs.html#operator-configuration


At least the logging related should match.

Also, the JEP was missing endpoint config, but we should re-use the structs from the router/controller/etc. And make it auto-calculate the endpoint name as we do with the others.

enabled/disabled must be controlled by the telemetry.enabled flag.

protocol/proto/jumpstarter/v1/telemetry.proto
bkhizgiy: Good catch, fixed.

python/packages/jumpstarter/jumpstarter/exporter/telemetry.py
bkhizgiy: That was taken from the JEP configuration, added some comments to clarify.

python/packages/jumpstarter/jumpstarter/exporter/telemetry.py
bkhizgiy: same

protocol/proto/jumpstarter/v1/telemetry.proto
```diff
@@ -0,0 +1,43 @@
...
+  string message = 3;      // Human-readable log message.
+  string component = 4;    // Loki stream label: cli, exporter, controller, router, telemetry.
+  string exporter = 5;     // Loki stream label: exporter CRD name (bounded by cluster size).
+  string lease = 6;        // Log body only (high cardinality): active lease name.
+  string client = 7;       // Log body only (high cardinality): client CRD name.
+  string operation = 8;    // Log body: operation name (flash, power, etc.).
+  string result = 9;       // Log body: operation outcome (success, failure, etc.).
+  string driver_type = 10; // Log body: driver category (storage, power, network, etc.).
+  map<string, string> extra_fields = 11; // Additional structured key-value fields.
+                                         // Capped at 16 entries, 64-char keys, 256-char values.
```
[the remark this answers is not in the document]
bkhizgiy: wasnt this fixed in the last commit?

controller/deploy/operator/config/crd/bases/virtualtarget.jumpstarter.dev_exportersets.yaml
bkhizgiy: haha yep, good eye was removed.

controller/internal/service/telemetry_service.go
bkhizgiy: Done, `PushLogs` now validates the bearer token and rejects any token whose subject is not in `exporter:namespace:name:uid` form (client tokens and other validly signed tokens are rejected with `PermissionDenied`). Entries whose exporter or namespace fields don't match the token's identity are counted as dropped rather than failing the whole batch.

controller/internal/config/types.go
bkhizgiy: good point, the telemetry.logging.filter structure now matches the JEP.

Endpoint auto-derivation was implemented when the endpoint is empty and `enabled: true`, the controller derives` jumpstarter-telemetry.<namespace>:9093` (the same pattern as the router and controller services). The Certificate field is reserved for Phase 2 TLS support and is now clearly documented as such.

And regarding enabled/disabled, `GetServiceEndpoints` only advertises the telemetry endpoint when `telemetry.enabled: true`.

controller/internal/config/config.go
bkhizgiy: Done, added `LoadConfiguration` now returns (*LoadedConfig, error), a single struct that carries all the previously separate return values. All error paths return a single nil, err. The call site in main.go was updated accordingly.

mangelajo:
@bkhizgiy sorry, some conflicts, I am reviewing in the meanwhile.

controller/cmd/main.go
```diff
@@ -217,8 +217,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	authenticator, prefix, router, option, provisioning, leasePolicy,
-		hiddenLabels, deprecatedLabels, err := config.LoadConfiguration(
+	cfg, err := config.LoadConfiguration(
```
mangelajo, line 220: Thank you!, it keep being tangential to every PR touching this, but in the end, there it goes!

controller/internal/config/types.go
```diff
@@ -15,6 +18,60 @@ type Config struct {
...
 	HiddenLabels     HiddenLabels     `json:"hiddenLabels,omitempty" yaml:"hiddenLabels,omitempty"`
 	DeprecatedLabels DeprecatedLabels `json:"deprecatedLabels,omitempty" yaml:"deprecatedLabels,omitempty"`
+	Telemetry        *Telemetry       `json:"telemetry,omitempty" yaml:"telemetry,omitempty"`
+}
+
+// Telemetry configures the optional jumpstarter-telemetry service.
+// When Enabled is true the controller advertises the telemetry endpoint to
+// exporters and clients via GetServiceEndpoints so they can push logs without
+// holding cluster credentials.
+type Telemetry struct {
```
mangelajo, line 28: oh, I confused this with the operator config.  :-/

This is the internal config we pass to the jumpstarter-controller, but anyway I think it may be ok, the controller needs to know if Telemetry is enabled/disabled, where is the endpoint, what will be the certificate, so it can announce it on the "Get telemetry endpoint" call. 
:-D

phew :) I thought my confusion made it bad, it looks good actually, thanks Bella!

controller/cmd/telemetry/main.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	svc := &service.TelemetryService{
+		BindAddr: bindAddr,
+		Signer:   signer,
+	}
+
+	errCh := make(chan error, 1)
+	go func() {
+		errCh <- svc.Start(ctx)
```
mangelajo, line 84: I was investigating how the main controller configured TLS and the endpoint name  (not sure if it's needed here really.., may be for the gRPC server), the telemetry service should follow the same established pattern for consistency, but can be done in follow up:

The existing controller and router services receive their TLS certificates via **Secret volume mounts** (not inline ConfigMap strings), and their endpoint addresses via **environment variables** on the Deployment pod spec. 

1. **TLS**: The operator should create (or reference) a TLS Secret for the telemetry service (either via cert-manager `Certificate` CR or a user-provided `spec.telemetry.tls.certSecret`), mount it at `/tls/`, and set `EXTERNAL_CERT_PEM=/tls/tls.crt` and `EXTERNAL_KEY_PEM=/tls/tls.key` env vars — exactly like `createControllerDeployment` and `createRouterDeployment` do today. Then `Start()` should read those files and configure `grpc.Creds(credentials.NewTLS(...))` on the server, falling back to a self-signed cert when the env vars are absent.
  (let's not do plaintext in any case, that can be dangerous since the wire always carries authentication tokens, now for exporters, but later will be clients as well pushing logs).

2. **Endpoint address**: The telemetry endpoint should be passed to the **controller** pod as an env var (e.g. `GRPC_TELEMETRY_ENDPOINT`) rather than auto-derived from the namespace in the ConfigMap loader. The operator already knows the address because it creates the Service/Route/Ingress for it — same as it does for `GRPC_ENDPOINT` and `GRPC_ROUTER_ENDPOINT`.


(AI made, human touch afterwards :D )

mangelajo (approved):
Only comments for follow up, specially the TLS and config-passing to the jumpstarter-telemetry container/deployment.

bkhizgiy:
@mangelajo Thanks for the review! I’ll merge this one and follow up with a separate PR for the TLS setup and passing the telemetry endpoint through the deployment config/env vars.

PR #933, opened 2026-07-30
feat: router Prometheus /metrics endpoint (JEP-0013 Phase 2)

RoddieKieley:
- Add HTTP `GET /metrics` on the Jumpstarter router (Prometheus/OpenMetrics text) via `--metrics-bind-address` (or equivalent).
- Operator wires Router Deployment metrics bind address and container port.
- Tests assert `/metrics` returns 200 and parseable exposition; no undocumented `jumpstarter_*` router series invented in this PR.

### DIFF
--- a/controller/cmd/router/main.go
+++ b/controller/cmd/router/main.go
@@ -45,6 +45,10 @@ func main() {
 	opts := zap.Options{}
 	opts.BindFlags(flag.CommandLine)
 
+	var metricsAddr string
+	flag.StringVar(&metricsAddr, "metrics-bind-address", "0",
+		"The address the metric endpoint binds to. Use :8080 to enable. Set to 0 to disable.")
+
 	flag.Parse()
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)).WithValues("component", "router"))
@@ -57,6 +61,13 @@ func main() {
 		"gitCommit", gitCommit,
 		"buildDate", buildDate,
 	)
+
+	if listenAddr, err := startMetricsServer(metricsAddr); err != nil {
+		logger.Error(err, "failed to start metrics server", "bindAddress", metricsAddr)
+		os.Exit(1)
+	} else if listenAddr != "" {
+		logger.Info("Serving metrics server", "bindAddress", listenAddr)
+	}
 
 	cfg := ctrl.GetConfigOrDie()
 	client, err := kclient.New(cfg, kclient.Options{})
--- a/controller/cmd/router/main.go
+++ b/controller/cmd/router/main.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	ctrl "sigs.k8s.io/controller-runtime"
 	kclient "sigs.k8s.io/controller-runtime/pkg/client"
@@ -62,10 +63,12 @@ func main() {
 		"buildDate", buildDate,
 	)
 
-	if listenAddr, err := startMetricsServer(metricsAddr); err != nil {
+	var shutdownMetrics func(context.Context) error
+	if listenAddr, shutdown, err := startMetricsServer(metricsAddr); err != nil {
 		logger.Error(err, "failed to start metrics server", "bindAddress", metricsAddr)
 		os.Exit(1)
 	} else if listenAddr != "" {
+		shutdownMetrics = shutdown
 		logger.Info("Serving metrics server", "bindAddress", listenAddr)
 	}
 
@@ -99,4 +102,12 @@ func main() {
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 	sig := <-sigs
 	logger.Info("received signal, exiting", "signal", sig)
+
+	if shutdownMetrics != nil {
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		if err := shutdownMetrics(shutdownCtx); err != nil {
+			logger.Error(err, "failed to shut down metrics server")
+		}
+	}
 }
--- a/controller/cmd/router/metrics.go
+++ b/controller/cmd/router/metrics.go
@@ -19,6 +19,7 @@ package main
 import (
 	"net"
 	"net/http"
+	"time"
 
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
@@ -40,7 +41,15 @@ func startMetricsServer(addr string) (string, error) {
 	mux := http.NewServeMux()
 	mux.Handle("/metrics", promhttp.Handler())
 
-	srv := &http.Server{Handler: mux}
+	srv := &http.Server{
+		Handler:           mux,
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      30 * time.Second,
+		// IdleTimeout is generous so Prometheus scrape keepalives survive
+		// typical scrape intervals without churning connections.
+		IdleTimeout: 5 * time.Minute,
+	}
 	go func() {
 		_ = srv.Serve(ln)
 	}()
--- a/controller/cmd/router/metrics.go
+++ b/controller/cmd/router/metrics.go
@@ -1,0 +1,49 @@
+/*
+Copyright 2026. The Jumpstarter Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"net"
+	"net/http"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// startMetricsServer starts an HTTP server exposing GET /metrics.
+// addr "0" or empty disables the server and returns ("", nil).
+// addr ending with ":0" binds an ephemeral port; the returned listen address
+// is host:port suitable for http.Get.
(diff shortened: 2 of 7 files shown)


controller/cmd/router/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+
+	srv := &http.Server{Handler: mux}
+	go func() {
+		_ = srv.Serve(ln)
+	}()
+
+	return ln.Addr().String(), nil
+}
```
raballew: `startMetricsServer` returns only `(string, error)` but creates an `http.Server` and launches `srv.Serve(ln)` in a detached goroutine. The caller has no way to call `srv.Shutdown(ctx)` on SIGINT/SIGTERM, so the server never gets a graceful shutdown.

controller/cmd/router/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+		return "", err
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+
+	srv := &http.Server{Handler: mux}
+	go func() {
+		_ = srv.Serve(ln)
+	}()
```
raballew, line 60:
```go
go func() {
    if err := srv.Serve(ln); err != nil && err != http.ErrServerClosed {
        ctrl.Log.WithName("metrics").Error(err, "metrics server stopped unexpectedly")
    }
}()
```

controller/cmd/router/metrics.go
```diff
@@ -0,0 +1,49 @@
...
+
+	ln, err := net.Listen("tcp", addr)
+	if err != nil {
+		return "", err
+	}
+
+	mux := http.NewServeMux()
+	mux.Handle("/metrics", promhttp.Handler())
+
+	srv := &http.Server{Handler: mux}
```
raballew: Set at least `ReadHeaderTimeout` and `IdleTimeout`:

```go
srv := &http.Server{
    Handler:           mux,
    ReadHeaderTimeout: 10 * time.Second,
    IdleTimeout:       60 * time.Second,
}
```

controller/cmd/router/metrics_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if !strings.Contains(ct, "text/plain") && !strings.Contains(ct, "openmetrics") {
+		t.Fatalf("unexpected Content-Type %q", ct)
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		t.Fatalf("read body: %v", err)
+	}
+	text := string(body)
+	// Default Go process metrics should appear; do not require undocumented jumpstarter_* series.
+	if !strings.Contains(text, "go_") && !strings.Contains(text, "process_") && !strings.Contains(text, "promhttp_") {
```
raballew: If `go_*` metrics were removed but `promhttp_*` remained, the test would still pass. Assert each required family independently.

raballew (changes requested):

controller/cmd/router/metrics.go
mangelajo: The idle timeout may need to be higher, i.e. it really depends on prometheus doing keepalive & the periodicity of checks,  I'd give it 5 minutes perhaps.

controller/cmd/router/metrics.go
RoddieKieley: Suggested go func now on [#L56-60](https://github.com/RoddieKieley/jumpstarter/blob/193d8e108ef9b57f33808d30c8ecfb667254c1f0/controller/cmd/router/metrics.go#L56-L60)

raballew (approved):
I think this looks good.

bkhizgiy (approved):
looks good from my side

controller/cmd/router/main.go
```diff
@@ -88,4 +102,12 @@ func main() {
 	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 	sig := <-sigs
 	logger.Info("received signal, exiting", "signal", sig)
```
mangelajo: Nit: the router gracefully shuts down the metrics server on signal, but the `RouterService` started at line 95 is never explicitly stopped. The `ctx` passed to `svc.Start(ctx)` is derived from `context.Background()` and is never cancelled. If `RouterService.Start` spawns goroutines or holds connections, they'll be abandoned on exit.

Consider cancelling the context on signal so the router service can also clean up:

```go
ctx, cancel := context.WithCancel(logr.NewContext(context.Background(), logger))
// ...
sig := <-sigs
logger.Info("received signal, exiting", "signal", sig)
cancel() // signal RouterService to stop
```

This is pre-existing behavior, not introduced by this PR, so just a suggestion for a follow-up.

---

controller/cmd/router/metrics.go
```diff
@@ -0,0 +1,63 @@
...
+
+package main
+
+import (
+	"context"
+	"net"
+	"net/http"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/promhttp"
```
mangelajo, line 25: Good: the server uses `promhttp.Handler()` which serves from the default `prometheus.DefaultGatherer`. This means any future code that registers custom `jumpstarter_*` metrics on the default registry will automatically appear here. The test at `metrics_test.go:88` guards against accidental/undocumented series — that's a nice contract.

One thing to be aware of: if a future dependency (e.g., a gRPC middleware) auto-registers Prometheus collectors on the default registry, those metrics will surface here too. 

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
mangelajo, line 1: Now we are ready to add the a new ServiceMonitor :-)

controller/cmd/router/metrics.go
mangelajo, line 1: `prometheus/client_golang` is imported directly by this file (via `promhttp`), but it's listed as `// indirect` in `controller/go.mod` (line 85). Running `go mod tidy` should promote it to a direct dependency. While this doesn't affect builds, it misrepresents the dependency relationship and can confuse tooling.

---

mangelajo (commented):
## Review Summary — PR #933: Router Prometheus /metrics endpoint (JEP-0013 Phase 2)

Overall this is a clean, well-scoped PR. The metrics server implementation is solid with proper timeouts, graceful shutdown, and good test coverage. The previous review feedback from both CodeRabbit and the human reviewer has been addressed.

### What looks good

- **Server hardening**: `ReadHeaderTimeout`, `ReadTimeout`, `WriteTimeout`, and `IdleTimeout` are all set with sensible values. The `IdleTimeout` at 5 minutes is appropriate for Prometheus scrape keepalives.
- **Graceful shutdown**: The shutdown function is properly returned from `startMetricsServer` and called with a timeout on signal. Clean lifecycle management.
- **Disable-by-default**: The `"0"` default for `--metrics-bind-address` matches the operator's existing convention and keeps backward compatibility.
- **Test quality**: Tests cover the happy path, disabled state, and shutdown behavior. The exposition format is validated with a proper Prometheus parser rather than substring heuristics. The `jumpstarter_*` guard is a nice contract enforcement.
- **Build changes**: Switching from `cmd/router/main.go` to `./cmd/router` in the Containerfile and Makefile is necessary now that the package has multiple `.go` files.

### Findings

1. **`go.mod` indirect marker** (minor): `prometheus/client_golang` is now directly imported but still marked `// indirect` in `controller/go.mod`. A `go mod tidy` should fix this.

2. **No `ServiceMonitor` for the router** (question): The metrics port is exposed in the container spec, but no `ServiceMonitor` is created. Prometheus won't auto-discover the router's metrics without one. Is this planned for a follow-up?

3. **Default registry exposure** (informational): Using `promhttp.Handler()` with the default gatherer means any future dependency that auto-registers collectors will surface metrics here. The test guards against undocumented `jumpstarter_*` series, which is good. A dedicated registry could provide stricter isolation if needed later.

4. **RouterService shutdown** (pre-existing, not introduced by this PR): The router service context is never cancelled on signal — only the metrics server is cleanly shut down. Noted as a follow-up opportunity.

No blocking issues found. The comments are suggestions and questions for the authors.

---

mangelajo (changes requested):
looks great, I'd only would run go mod tidy before merging.

controller/cmd/router/metrics.go
RoddieKieley: After I rebased the // indirect was already gone when I ran the go mod tidy.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
RoddieKieley: Yes, was trying to keep the scope here manageable while also adhering to the outline in the actual JEP-0013. As per the agent description:
>...only says the Phase 2 HTTP /metrics path is compatible with Prometheus Operator ServiceMonitor; it does not require generating those CRs now. The operator flag spec.telemetry.metrics.serviceMonitor lives in the end-state CR, which Phase 5 implements.

controller/cmd/router/metrics.go
RoddieKieley: > Correct: promhttp.Handler() serves prometheus.DefaultGatherer, so a later MustRegister of a jumpstarter_* collector on the default registry would appear on /metrics automatically.

> That is intentional for Phase 2. This PR does not register any router-owned jumpstarter_* series (the JEP catalog has none for the router). TestMetricsEndpointServesPrometheusText fails if any jumpstarter_* family is present, so an accidental or undocumented series cannot land without updating the test.

> The middleware note is also fair: the test is a Jumpstarter-name contract, not a full scrape-surface allowlist, so a dependency that auto-registers on the default registry (e.g. grpc_*) would show up and still pass. A private prometheus.NewRegistry() with explicit Go/process collectors is the right hardening when the router grows first-party metrics or when a dependency starts polluting DefaultGatherer. Out of scope for this PR.

___

RoddieKieley:
What else is needed here @mangelajo ? I took a look and didn't see anything obvious and also ran a review that pointed out that go mod tidy was completed, the go.mod entry is direct, the build of ./cmd/router includes required go files. Outside of that, need CI to run and reviews to be completed against the latest code?

mangelajo (approved):

PR #934, opened 2026-07-30
feat: exporter prometheus_client /metrics (JEP-0013 Phase 2)

RoddieKieley:
- Add exporter-local `prometheus_client` registry with JEP-named series: `jumpstarter_operations_total`, `jumpstarter_operation_duration_seconds`, `jumpstarter_operation_errors_total`, `jumpstarter_stream_bytes_total`, `jumpstarter_active_sessions`, plus exemplars (`client`, `lease_id`).
- Expose HTTP `GET /metrics` on the exporter process for lab/dev scrape (same registry Phase 3 will later reverse-scrape via MetricsStream).
- Minimal core-path wiring so series increment under test; full per-driver telemetry architecture remains Phase 4.

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
@@ -15,6 +15,11 @@ from jumpstarter.metrics import start_metrics_server
 
 logger = logging.getLogger(__name__)
 
+# Phase 2 interim: always expose local HTTP /metrics on ephemeral loopback.
+# Phase 3 replaces this with Telemetry reverse-scrape (unix/memory or in-process);
+# bind address is intentionally not a user-facing CLI option.
+_METRICS_BIND_ADDRESS = ":0"
+
 
 def _parse_listener_bind(value: str) -> tuple[str, int]:
     """Parse '[host:]port' into (host, port). Default host is 0.0.0.0."""
@@ -79,7 +84,6 @@ def _handle_child(  # noqa: C901
     tls_cert=None,
     tls_key=None,
     passphrase=None,
-    metrics_bind_address=":0",
 ):
     """Handle child process with graceful shutdown."""
     async def serve_with_graceful_shutdown():  # noqa: C901
@@ -109,9 +113,8 @@ def _handle_child(  # noqa: C901
             # Start signal handler immediately
             signal_tg.start_soon(signal_handler)
 
-            listen_addr, shutdown_metrics = start_metrics_server(metrics_bind_address)
-            if listen_addr:
-                logger.info("Serving metrics server at http://%s/metrics", listen_addr)
+            listen_addr, shutdown_metrics = start_metrics_server(_METRICS_BIND_ADDRESS)
+            logger.info("Serving metrics server at http://%s/metrics", listen_addr)
 
             try:
                 if parsed_bind is not None:
@@ -228,7 +231,6 @@ def _serve_with_exc_handling(
     tls_cert=None,
     tls_key=None,
     passphrase=None,
-    metrics_bind_address=":0",
 ):
     max_rapid_failures = config.failure_detection.max_rapid_failures
     rapid_failure_window = config.failure_detection.rapid_failure_window
@@ -284,7 +286,6 @@ def _serve_with_exc_handling(
                 tls_cert,
                 tls_key,
                 passphrase,
-                metrics_bind_address,
             )
             sys.exit(1) # should never happen
 
@@ -326,16 +327,6 @@ def _serve_with_exc_handling(
     default=False,
     help="Exit after the current lease ends instead of waiting for a new one.",
 )
-@click.option(
-    "--metrics-bind-address",
-    "metrics_bind_address",
-    default=":0",
-    show_default=True,
-    help=(
-        "Address for HTTP GET /metrics (Prometheus/OpenMetrics). "
-        "Default :0 binds an ephemeral loopback port. Use 0 to disable."
-    ),
-)
 @handle_exceptions
 def run(
     config,
@@ -345,7 +336,6 @@ def run(
     tls_key,
     passphrase,
     exit_on_lease_end,
-    metrics_bind_address,
 ):
     """Run an exporter locally."""
     if listener_bind is not None and config is None:
@@ -371,5 +361,4 @@ def run(
         tls_cert,
         tls_key,
         passphrase,
-        metrics_bind_address,
     )
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
@@ -70,7 +70,15 @@ def _reap_zombie_processes(capture_child=None):
         logger.warning(f"PARENT: Error during zombie reaping: {e}")
 
 
-def _handle_child(config, parsed_bind=None, tls_insecure=False, tls_cert=None, tls_key=None, passphrase=None):  # noqa: C901
+def _handle_child(  # noqa: C901
+    config,
+    parsed_bind=None,
+    tls_insecure=False,
+    tls_cert=None,
+    tls_key=None,
+    passphrase=None,
+    metrics_bind_address=":8080",
+):
     """Handle child process with graceful shutdown."""
     async def serve_with_graceful_shutdown():  # noqa: C901
         received_signal = 0
@@ -98,6 +106,12 @@ def _handle_child(config, parsed_bind=None, tls_insecure=False, tls_cert=None, t
 
             # Start signal handler immediately
             signal_tg.start_soon(signal_handler)
+
+            from jumpstarter.metrics import start_metrics_server
+
+            listen_addr = start_metrics_server(metrics_bind_address)
+            if listen_addr:
+                logger.info("Serving metrics server at http://%s/metrics", listen_addr)
 
             if parsed_bind is not None:
                 host, port = parsed_bind
@@ -204,7 +218,13 @@ def _handle_parent(pid):
 
 
 def _serve_with_exc_handling(
-    config, parsed_bind=None, tls_insecure=False, tls_cert=None, tls_key=None, passphrase=None
+    config,
+    parsed_bind=None,
+    tls_insecure=False,
+    tls_cert=None,
+    tls_key=None,
+    passphrase=None,
+    metrics_bind_address=":8080",
 ):
     max_rapid_failures = config.failure_detection.max_rapid_failures
     rapid_failure_window = config.failure_detection.rapid_failure_window
@@ -253,7 +273,15 @@ def _serve_with_exc_handling(
                 rapid_failure_count = 0
         else:
             os.setsid() # Become group leader so all spawned subprocesses are reached by parent's signals
-            _handle_child(config, parsed_bind, tls_insecure, tls_cert, tls_key, passphrase)
+            _handle_child(
+                config,
+                parsed_bind,
+                tls_insecure,
+                tls_cert,
+                tls_key,
+                passphrase,
+                metrics_bind_address,
+            )
             sys.exit(1) # should never happen
 
 
@@ -294,8 +322,24 @@ def _serve_with_exc_handling(
     default=False,
     help="Exit after the current lease ends instead of waiting for a new one.",
 )
+@click.option(
+    "--metrics-bind-address",
+    "metrics_bind_address",
+    default=":8080",
+    show_default=True,
+    help="Address for HTTP GET /metrics (Prometheus/OpenMetrics). Use 0 to disable.",
+)
 @handle_exceptions
-def run(config, listener_bind, tls_insecure, tls_cert, tls_key, passphrase, exit_on_lease_end):
+def run(
+    config,
+    listener_bind,
+    tls_insecure,
+    tls_cert,
+    tls_key,
+    passphrase,
+    exit_on_lease_end,
+    metrics_bind_address,
+):
     """Run an exporter locally."""
     if listener_bind is not None and config is None:
         raise click.UsageError("--exporter-config (or --exporter) is required when using --tls-grpc-listener")
@@ -313,4 +357,12 @@ def run(config, listener_bind, tls_insecure, tls_cert, tls_key, passphrase, exit
     if exit_on_lease_end:
         config.exit_on_lease_end = True
     parsed_bind = _parse_listener_bind(listener_bind) if listener_bind is not None else None
-    return _serve_with_exc_handling(config, parsed_bind, tls_insecure, tls_cert, tls_key, passphrase)
+    return _serve_with_exc_handling(
+        config,
+        parsed_bind,
+        tls_insecure,
+        tls_cert,
+        tls_key,
+        passphrase,
+        metrics_bind_address,
+    )
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
@@ -11,6 +11,8 @@ from anyio import create_task_group, open_signal_receiver
 from jumpstarter_cli_common.config import opt_config
 from jumpstarter_cli_common.exceptions import handle_exceptions
 
+from jumpstarter.metrics import start_metrics_server
+
 logger = logging.getLogger(__name__)
 
 
@@ -106,8 +108,6 @@ def _handle_child(  # noqa: C901
 
             # Start signal handler immediately
             signal_tg.start_soon(signal_handler)
-
-            from jumpstarter.metrics import start_metrics_server
 
             listen_addr, shutdown_metrics = start_metrics_server(metrics_bind_address)
             if listen_addr:
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
@@ -109,50 +109,54 @@ def _handle_child(  # noqa: C901
 
             from jumpstarter.metrics import start_metrics_server
 
-            listen_addr = start_metrics_server(metrics_bind_address)
+            listen_addr, shutdown_metrics = start_metrics_server(metrics_bind_address)
             if listen_addr:
                 logger.info("Serving metrics server at http://%s/metrics", listen_addr)
 
-            if parsed_bind is not None:
-                host, port = parsed_bind
-                tls_credentials = None
-                if tls_insecure:
+            try:
+                if parsed_bind is not None:
+                    host, port = parsed_bind
+                    tls_credentials = None
+                    if tls_insecure:
+                        if passphrase:
+                            click.echo(
+                                "WARNING: --passphrase has no effect without TLS; "
+                                "the passphrase will be transmitted in plaintext",
+                                err=True,
+                            )
+                    elif tls_cert and tls_key:
+                        tls_credentials = _tls_server_credentials(tls_cert, tls_key)
+
+                    interceptors = None
                     if passphrase:
-                        click.echo(
-                            "WARNING: --passphrase has no effect without TLS; "
-                            "the passphrase will be transmitted in plaintext",
-                            err=True,
-                        )
-                elif tls_cert and tls_key:
-                    tls_credentials = _tls_server_credentials(tls_cert, tls_key)
+                        from jumpstarter.exporter.auth import PassphraseInterceptor
+                        interceptors = [PassphraseInterceptor(passphrase)]
 
-                interceptors = None
-                if passphrase:
-                    from jumpstarter.exporter.auth import PassphraseInterceptor
-                    interceptors = [PassphraseInterceptor(passphrase)]
+                    exporter_exit_code = None
+                    async with config.create_exporter(standalone=True) as exporter:
+                        try:
+                            await exporter.serve_standalone_tcp(
+                                host, port,
+                                tls_credentials=tls_credentials,
+                                interceptors=interceptors,
+                            )
+                        except* Exception as excgroup:
+                            _handle_exporter_exceptions(excgroup)
+                        exporter_exit_code = exporter.exit_code
+                else:
+                    # Create exporter and run it (controller mode)
+                    exporter_exit_code = None
+                    async with config.create_exporter() as exporter:
+                        try:
+                            await exporter.serve()
+                        except* Exception as excgroup:
+                            _handle_exporter_exceptions(excgroup)
 
-                exporter_exit_code = None
-                async with config.create_exporter(standalone=True) as exporter:
-                    try:
-                        await exporter.serve_standalone_tcp(
-                            host, port,
-                            tls_credentials=tls_credentials,
-                            interceptors=interceptors,
-                        )
-                    except* Exception as excgroup:
-                        _handle_exporter_exceptions(excgroup)
-                    exporter_exit_code = exporter.exit_code
-            else:
-                # Create exporter and run it (controller mode)
-                exporter_exit_code = None
-                async with config.create_exporter() as exporter:
-                    try:
-                        await exporter.serve()
-                    except* Exception as excgroup:
-                        _handle_exporter_exceptions(excgroup)
-
-                    # Check if exporter set an exit code (e.g., from hook failure with on_failure='exit')
-                    exporter_exit_code = exporter.exit_code
+                        # Check if exporter set an exit code (e.g., from hook failure with on_failure='exit')
+                        exporter_exit_code = exporter.exit_code
+            finally:
+                if shutdown_metrics is not None:
+                    shutdown_metrics()
 
             # Cancel the signal handler after exporter completes
             signal_tg.cancel_scope.cancel()
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/run.py
@@ -77,7 +77,7 @@ def _handle_child(  # noqa: C901
     tls_cert=None,
     tls_key=None,
     passphrase=None,
-    metrics_bind_address=":8080",
+    metrics_bind_address=":0",
 ):
     """Handle child process with graceful shutdown."""
     async def serve_with_graceful_shutdown():  # noqa: C901
@@ -224,7 +224,7 @@ def _serve_with_exc_handling(
     tls_cert=None,
     tls_key=None,
     passphrase=None,
-    metrics_bind_address=":8080",
+    metrics_bind_address=":0",
 ):
     max_rapid_failures = config.failure_detection.max_rapid_failures
     rapid_failure_window = config.failure_detection.rapid_failure_window
@@ -325,9 +325,12 @@ def _serve_with_exc_handling(
 @click.option(
     "--metrics-bind-address",
     "metrics_bind_address",
-    default=":8080",
+    default=":0",
     show_default=True,
-    help="Address for HTTP GET /metrics (Prometheus/OpenMetrics). Use 0 to disable.",
+    help=(
+        "Address for HTTP GET /metrics (Prometheus/OpenMetrics). "
+        "Default :0 binds an ephemeral loopback port. Use 0 to disable."
+    ),
 )
 @handle_exceptions
 def run(
--- a/python/packages/jumpstarter/jumpstarter/driver/base.py
+++ b/python/packages/jumpstarter/jumpstarter/driver/base.py
@@ -128,15 +128,29 @@ class Driver(
         duration_seconds: float,
         error_type: str | None = None,
     ) -> None:
-        get_registry().record_operation(
-            exporter=exporter_from_log_context(default=self.name if hasattr(self, "name") else "unknown"),
-            operation=operation,
-            result=result,
-            driver_type=self.driver_type,
-            duration_seconds=duration_seconds,
-            exemplars=exemplars_from_log_context(),
-            error_type=error_type,
-        )
+        # Metrics must never discard a computed gRPC response or change the
+        # abort status: keep recording failures isolated from the RPC path.
+        try:
+            get_registry().record_operation(
+                exporter=exporter_from_log_context(default=self.name if hasattr(self, "name") else "unknown"),
+                operation=operation,
+                result=result,
+                driver_type=self.driver_type,
+                duration_seconds=duration_seconds,
+                exemplars=exemplars_from_log_context(),
+                error_type=error_type,
+            )
+        except Exception:
+            self.logger.warning(
+                "Failed to record operation metrics",
+                extra={
+                    "operation": operation,
+                    "driver_type": self.driver_type,
+                    "result": result,
+                    "error_type": error_type,
+                },
+                exc_info=True,
+            )
 
     async def DriverCall(self, request, context):
         """
--- a/python/packages/jumpstarter/jumpstarter/driver/base.py
+++ b/python/packages/jumpstarter/jumpstarter/driver/base.py
@@ -6,6 +6,7 @@ from __future__ import annotations
 
 import logging
 import os
+import time
 from abc import ABCMeta, abstractmethod
 from contextlib import asynccontextmanager
 from dataclasses import field
@@ -113,11 +114,36 @@ class Driver(
     def extra_labels(self) -> dict[str, str]:
         return {}
 
+    def _record_operation_metrics(
+        self,
+        *,
+        operation: str,
+        result: str,
+        duration_seconds: float,
+        error_type: str | None = None,
+    ) -> None:
+        from jumpstarter.metrics.registry import (
+            exemplars_from_log_context,
+            exporter_from_log_context,
+            get_registry,
+        )
+
+        get_registry().record_operation(
+            exporter=exporter_from_log_context(default=self.name if hasattr(self, "name") else "unknown"),
+            operation=operation,
+            result=result,
+            driver_type=self.driver_type,
+            duration_seconds=duration_seconds,
+            exemplars=exemplars_from_log_context(),
+            error_type=error_type,
+        )
+
     async def DriverCall(self, request, context):
         """
         :meta private:
         """
         op = request.method
+        started = time.perf_counter()
         self.logger.info(
             "Operation started",
             extra={"operation": op, "driver_type": self.driver_type},
@@ -132,6 +158,11 @@ class Driver(
             else:
                 result = await to_thread.run_sync(method, *args)
 
+            self._record_operation_metrics(
+                operation=op,
+                result="success",
+                duration_seconds=time.perf_counter() - started,
+            )
             self.logger.info(
                 "Operation completed",
                 extra={"operation": op, "driver_type": self.driver_type, "result": "success"},
@@ -141,6 +172,12 @@ class Driver(
                 result=encode_value(result),
             )
         except NotImplementedError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="not_implemented",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -148,6 +185,12 @@ class Driver(
             )
             await context.abort(StatusCode.UNIMPLEMENTED, str(e))
         except ValueError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="validation_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -155,6 +198,12 @@ class Driver(
             )
             await context.abort(StatusCode.INVALID_ARGUMENT, str(e))
         except TimeoutError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="timeout",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -162,6 +211,12 @@ class Driver(
             )
             await context.abort(StatusCode.DEADLINE_EXCEEDED, str(e))
         except ConnectionError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="connection_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -169,6 +224,12 @@ class Driver(
             )
             await context.abort(StatusCode.UNAVAILABLE, str(e))
         except OSError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="device_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -176,6 +237,12 @@ class Driver(
             )
             await context.abort(StatusCode.INTERNAL, str(e))
         except Exception as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="internal_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -188,6 +255,7 @@ class Driver(
         :meta private:
         """
         op = request.method
+        started = time.perf_counter()
         self.logger.info(
             "Operation started",
             extra={"operation": op, "driver_type": self.driver_type},
@@ -209,11 +277,22 @@ class Driver(
                         uuid=str(uuid4()),
                         result=encode_value(result),
                     )
+            self._record_operation_metrics(
+                operation=op,
+                result="success",
+                duration_seconds=time.perf_counter() - started,
+            )
             self.logger.info(
                 "Operation completed",
                 extra={"operation": op, "driver_type": self.driver_type, "result": "success"},
             )
         except NotImplementedError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="not_implemented",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -221,6 +300,12 @@ class Driver(
             )
             await context.abort(StatusCode.UNIMPLEMENTED, str(e))
         except ValueError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="validation_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -228,6 +313,12 @@ class Driver(
             )
             await context.abort(StatusCode.INVALID_ARGUMENT, str(e))
         except TimeoutError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="timeout",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -235,6 +326,12 @@ class Driver(
             )
             await context.abort(StatusCode.DEADLINE_EXCEEDED, str(e))
         except ConnectionError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="connection_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -242,6 +339,12 @@ class Driver(
             )
             await context.abort(StatusCode.UNAVAILABLE, str(e))
         except OSError as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="device_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
@@ -249,6 +352,12 @@ class Driver(
             )
             await context.abort(StatusCode.INTERNAL, str(e))
         except Exception as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="internal_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
--- a/python/packages/jumpstarter/jumpstarter/driver/base.py
+++ b/python/packages/jumpstarter/jumpstarter/driver/base.py
@@ -41,6 +41,8 @@ from jumpstarter.common.streams import (
 from jumpstarter.config.env import JMP_DISABLE_COMPRESSION
 from jumpstarter.exporter.logging import get_logger
 from jumpstarter.metrics.registry import (
+    ErrorType,
+    OperationResult,
     exemplars_from_log_context,
     exporter_from_log_context,
     get_registry,
@@ -52,7 +54,7 @@ from jumpstarter.streams.metadata import MetadataStream
 from jumpstarter.streams.progress import ProgressStream
 
 # Ordered most-specific first: ConnectionError is an OSError subclass.
-_DRIVER_CALL_ERRORS: tuple[tuple[type[BaseException], str, StatusCode], ...] = (
+_DRIVER_CALL_ERRORS: tuple[tuple[type[BaseException], ErrorType, StatusCode], ...] = (
     (NotImplementedError, "not_implemented", StatusCode.UNIMPLEMENTED),
     (ValueError, "validation_error", StatusCode.INVALID_ARGUMENT),
     (TimeoutError, "timeout", StatusCode.DEADLINE_EXCEEDED),
@@ -133,9 +135,9 @@ class Driver(
         self,
         *,
         operation: str,
-        result: str,
+        result: OperationResult,
         duration_seconds: float,
-        error_type: str | None = None,
+        error_type: ErrorType | None = None,
     ) -> None:
         # Metrics must never discard a computed gRPC response or change the
         # abort status: keep recording failures isolated from the RPC path.
@@ -161,7 +163,13 @@ class Driver(
                 exc_info=True,
             )
 
-    async def _handle_driver_exception(self, exc: BaseException, op: str, started: float, context) -> None:
+    async def _handle_driver_exception(
+        self,
+        exc: BaseException,
+        op: str,
+        started: float,
+        context: grpc.aio.ServicerContext,
+    ) -> None:
         for exc_type, error_type, status in _DRIVER_CALL_ERRORS:
             if isinstance(exc, exc_type):
                 self._record_operation_metrics(
--- a/python/packages/jumpstarter/jumpstarter/driver/base.py
+++ b/python/packages/jumpstarter/jumpstarter/driver/base.py
@@ -158,6 +158,12 @@ class Driver(
             else:
                 result = await to_thread.run_sync(method, *args)
 
+            # Encode before recording success so serialization failures are
+            # counted only as failures (not success then failure).
+            response = jumpstarter_pb2.DriverCallResponse(
+                uuid=str(uuid4()),
+                result=encode_value(result),
+            )
             self._record_operation_metrics(
                 operation=op,
                 result="success",
@@ -167,10 +173,7 @@ class Driver(
                 "Operation completed",
                 extra={"operation": op, "driver_type": self.driver_type, "result": "success"},
             )
-            return jumpstarter_pb2.DriverCallResponse(
-                uuid=str(uuid4()),
-                result=encode_value(result),
-            )
+            return response
         except NotImplementedError as e:
             self._record_operation_metrics(
                 operation=op,
--- a/python/packages/jumpstarter/jumpstarter/driver/base.py
+++ b/python/packages/jumpstarter/jumpstarter/driver/base.py
@@ -45,6 +45,15 @@ from jumpstarter.metrics.registry import (
     exporter_from_log_context,
     get_registry,
 )
+
+# Ordered most-specific first: ConnectionError is an OSError subclass.
+_DRIVER_CALL_ERRORS: tuple[tuple[type[BaseException], str, StatusCode], ...] = (
+    (NotImplementedError, "not_implemented", StatusCode.UNIMPLEMENTED),
+    (ValueError, "validation_error", StatusCode.INVALID_ARGUMENT),
+    (TimeoutError, "timeout", StatusCode.DEADLINE_EXCEEDED),
+    (ConnectionError, "connection_error", StatusCode.UNAVAILABLE),
+    (OSError, "device_error", StatusCode.INTERNAL),
+)
 from jumpstarter.streams.aiohttp import AiohttpStreamReaderStream
 from jumpstarter.streams.common import create_memory_stream
 from jumpstarter.streams.encoding import Compression, compress_stream
@@ -152,6 +161,43 @@ class Driver(
                 exc_info=True,
             )
 
+    async def _handle_driver_exception(self, exc: BaseException, op: str, started: float, context) -> None:
+        for exc_type, error_type, status in _DRIVER_CALL_ERRORS:
+            if isinstance(exc, exc_type):
+                self._record_operation_metrics(
+                    operation=op,
+                    result="failure",
+                    duration_seconds=time.perf_counter() - started,
+                    error_type=error_type,
+                )
+                self.logger.warning(
+                    "Operation failed",
+                    extra={
+                        "operation": op,
+                        "driver_type": self.driver_type,
+                        "result": "failure",
+                        "error_type": error_type,
+                    },
+                )
+                await context.abort(status, str(exc))
+                return
+        self._record_operation_metrics(
+            operation=op,
+            result="failure",
+            duration_seconds=time.perf_counter() - started,
+            error_type="internal_error",
+        )
+        self.logger.warning(
+            "Operation failed",
+            extra={
+                "operation": op,
+                "driver_type": self.driver_type,
+                "result": "failure",
+                "error_type": "internal_error",
+            },
+        )
+        await context.abort(StatusCode.UNKNOWN, str(exc))
+
     async def DriverCall(self, request, context):
         """
         :meta private:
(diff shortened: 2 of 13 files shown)


python/packages/jumpstarter/jumpstarter/metrics/__init__.py
```diff
@@ -0,0 +1,17 @@
+"""Exporter-local Prometheus metrics (JEP-0013 Phase 2)."""
```
bennyz: I see we do use top level comments in some places, but i'd drop the JEP part at least, from the others as well

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -141,41 +172,77 @@ async def DriverCall(self, request, context):
...
         except Exception as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="internal_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
```
raballew, line 252: Consider adding `except grpc.aio.AbortError: raise` immediately before `except Exception` in both methods to let abort errors propagate without distorting metrics or status codes.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @property
+    def collector_registry(self) -> CollectorRegistry:
+        return self._registry
+
+    def record_operation(
+        self,
+        *,
+        exporter: str,
+        operation: str,
```
raballew, line 95: Using plain `str` means a typo at any call site silently creates a new Prometheus time series. Consider using `Literal` types for all bounded label parameters, e.g. `result: Literal["success", "failure"]`.

python/packages/jumpstarter/jumpstarter/driver/base.py
raballew, line 252: Consider adding test drivers whose `@export` methods raise each exception type (e.g., `TimeoutError`), then assert the correct `error_type` label appears in the metrics output.

python/packages/jumpstarter/jumpstarter/exporter/session.py
```diff
@@ -57,11 +57,20 @@ class Session(
...
         self.root_device.reset()
+        set_log_context(exporter=self.name)
+        get_registry().inc_active_sessions(exporter=self.name, delta=1.0)
         try:
             yield self
         finally:
+            try:
+                get_registry().inc_active_sessions(exporter=self.name, delta=-1.0)
+            except Exception:
+                pass
```
raballew: Log exception at WARNING or DEBUG level before suppressing it.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -113,11 +114,36 @@ def client(cls) -> str:
...
+
     async def DriverCall(self, request, context):
         """
         :meta private:
         """
         op = request.method
+        started = time.perf_counter()
         self.logger.info(
             "Operation started",
             extra={"operation": op, "driver_type": self.driver_type},
```
raballew, line 149: The `operation` label passed to `_record_operation_metrics` comes from `request.method`, a client-controlled gRPC field. When `__lookup_drivercall` rejects an unknown method, the resulting AbortError falls through to the catch-all handler, which records metrics with the untrusted value as a Prometheus label. Each unique method name creates a new time series, allowing an authenticated client to exhaust exporter memory by sending random method names. Fixing the AbortError handling (catching it before the generic handler) would largely mitigate this.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,61 @@
...
+
+def _parse_bind_addr(addr: str) -> tuple[str, int]:
+    """Parse bind address forms: ':8080', '127.0.0.1:0', '8080'."""
+    if ":" in addr:
+        host, _, port_s = addr.rpartition(":")
+        if host == "":
+            host = "0.0.0.0"
+    else:
+        host = "0.0.0.0"
+        port_s = addr
```
raballew, line 26: Metrics endpoints in Kubernetes are typically scraped via sidecar or PodMonitor which access localhost. Defaulting to `127.0.0.1` instead of `0.0.0.0` might be reasonable.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,144 @@
...
+    """Build default exemplars from the current structlog contextvars."""
+    ctx = structlog.contextvars.get_contextvars()
+    return filter_exemplars({key: str(ctx[key]) for key in DEFAULT_EXEMPLAR_KEYS if key in ctx})
+
+
+def exporter_from_log_context(default: str = "unknown") -> str:
+    ctx = structlog.contextvars.get_contextvars()
+    value = ctx.get("exporter")
+    return str(value) if value else default
+
```
raballew, line 49: `filter_exemplars`, `exemplars_from_log_context`, and `exporter_from_log_context` contain non-trivial logic (filtering empty values, reading from structlog contextvars, defaulting to "unknown") but have no direct test coverage.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -0,0 +1,133 @@
...
+    from jumpstarter.common.utils import serve
+
+    with serve(MockPower()) as client:
+        client.on()
+        body = get_registry().generate_latest().decode()
+        assert "jumpstarter_active_sessions" in body
+        assert "jumpstarter_operations_total" in body
+        assert 'driver_type="power"' in body
+        assert 'result="success"' in body
+        assert 'operation="on"' in body or 'operation="On"' in body
```
raballew, line 260: Consider parsing the exposition output or using `get_sample_value()` to verify `jumpstarter_operations_total` equals 1.0 for the specific label combination.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -113,11 +114,36 @@ def client(cls) -> str:
...
+
+        get_registry().record_operation(
+            exporter=exporter_from_log_context(default=self.name if hasattr(self, "name") else "unknown"),
+            operation=operation,
+            result=result,
+            driver_type=self.driver_type,
+            duration_seconds=duration_seconds,
+            exemplars=exemplars_from_log_context(),
+            error_type=error_type,
+        )
```
raballew, line 139: Consider moving the imports to module level or caching the resolved functions. Capture the exporter name and exemplars once before the stream loop.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,144 @@
...
+        )
+        self._duration = Histogram(
+            "jumpstarter_operation_duration_seconds",
+            "Duration of each operation.",
+            ["exporter", "operation", "result", "driver_type"],
+            registry=self._registry,
+        )
+        self._errors = Counter(
+            "jumpstarter_operation_errors_total",
+            "Errors by class (timeout, device, …).",
```
raballew: Consider replacing ellipsis with `...`

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,144 @@
...
+        nbytes: int,
+        exemplars: dict[str, str] | None = None,
+    ) -> None:
+        if nbytes <= 0:
+            return
+        self._stream_bytes.labels(
+            exporter=exporter,
+            driver_type=driver_type,
+            direction=direction,
+        ).inc(nbytes, exemplar=filter_exemplars(exemplars))
```
raballew, line 135: What happens to the metric increment for zero or negative byte counts?

raballew (changes requested):

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: In theory this should not happen, according to the guided review of the functionality anyway, as ..."the program deliberately does not record zero/negative amounts into the counter."

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Done — [base.py#L239-L242](https://github.com/RoddieKieley/jumpstarter/blob/1c23b951754c434999ee4ccba1e9d6a4bf1f8e01/python/packages/jumpstarter/jumpstarter/driver/base.py#L239-L242).

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: Done — [registry.py#L13-L22](https://github.com/RoddieKieley/jumpstarter/blob/1c23b951754c434999ee4ccba1e9d6a4bf1f8e01/python/packages/jumpstarter/jumpstarter/metrics/registry.py#L13-L22)

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Done - [metrics_test.py#L221-L230](https://github.com/RoddieKieley/jumpstarter/blob/1c23b951754c434999ee4ccba1e9d6a4bf1f8e01/python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py#L221-L230)

python/packages/jumpstarter/jumpstarter/exporter/session.py
RoddieKieley: Done — [session.py#L70-L77](https://github.com/RoddieKieley/jumpstarter/blob/1c23b951754c434999ee4ccba1e9d6a4bf1f8e01/python/packages/jumpstarter/jumpstarter/exporter/session.py#L70-L77)

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Addressed above https://github.com/jumpstarter-dev/jumpstarter/pull/934#discussion_r3762229789

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Done — now defaults to 127.0.0.1 in [_parse_bind_addr](https://github.com/RoddieKieley/jumpstarter/blob/1c23b951754c434999ee4ccba1e9d6a4bf1f8e01/python/packages/jumpstarter/jumpstarter/metrics/server.py#L13-L26); will require an explicit 0.0.0.0 host ip when remote scraping is required. Covered by test_parse_bind_addr_defaults_to_loopback.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: Done — direct unit coverage for filter_exemplars, exemplars_from_log_context, and exporter_from_log_context in metrics_test.py.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
RoddieKieley: Done — tests now use _sample_value() against specific label sets (e.g. jumpstarter_operations_total == 1.0) rather than substring-only checks.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: AI driven re-review of posted comment gives more detail prior to resolving:
> Zero/negative nbytes are deliberately not observed: [add_stream_bytes returns early](https://github.com/RoddieKieley/jumpstarter/blob/1c23b951754c434999ee4ccba1e9d6a4bf1f8e01/python/packages/jumpstarter/jumpstarter/metrics/registry.py#L128-L130) so Prometheus counters are never incremented with a non-positive amount (counters are monotonic). Call sites typically pass len(v) (≥ 0); zero can appear for non-bytes stream values. Covered by test_add_stream_bytes_ignores_non_positive.

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: copy_stream now captures exporter, exemplars, and registry once before the loop: [common.py#L31-L44](https://github.com/RoddieKieley/jumpstarter/blob/0f9aa41f173c4ecfcd9fe64c47c16c3c6432dd67/python/packages/jumpstarter/jumpstarter/streams/common.py#L31-L44)

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -209,46 +284,91 @@ async def StreamingDriverCall(self, request, context):
...
         except Exception as e:
+            self._record_operation_metrics(
+                operation=op,
+                result="failure",
+                duration_seconds=time.perf_counter() - started,
+                error_type="internal_error",
+            )
             self.logger.warning(
                 "Operation failed",
                 extra={"operation": op, "driver_type": self.driver_type,
```
raballew: Every `_record_operation_metrics()` call in both `DriverCall` and `StreamingDriverCall` is unguarded inside the outer `try` block. If that call raises the already-computed gRPC response is discarded and `context.abort()` is called with the wrong status code.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -132,62 +158,111 @@ async def DriverCall(self, request, context):
...
                        "result": "failure", "error_type": "internal_error"},
             )
             await context.abort(StatusCode.UNKNOWN, str(e))
```
raballew: Nit, extract a shared `async def _handle_driver_exception(self, exc, op, started, context)` coroutine or a module-level `_ERROR_MAP: dict[type, tuple[str, StatusCode]]` to centralize the branch mapping.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,82 @@
...
+        return ""
+
+    thread = Thread(target=server.serve_forever, name="jumpstarter-metrics", daemon=True)
+    thread.start()
+
+    bound_host, bound_port = server.server_address[:2]
+    # Prefer loopback-friendly host for ephemeral binds used in tests.
+    if bound_host in ("0.0.0.0", "::", ""):
+        bound_host = "127.0.0.1"
+    return f"{bound_host}:{bound_port}"
```
raballew: `start_metrics_server` discards the `ThreadingHTTPServer` instance and returns only the address string. Every test that calls it creates a live server that is never shut down.

python/packages/jumpstarter/jumpstarter/metrics/__init__.py
```diff
@@ -0,0 +1,17 @@
...
+    get_registry,
+    reset_registry_for_tests,
+)
+from .server import start_metrics_server
+
+__all__ = [
+    "DEFAULT_EXEMPLAR_KEYS",
+    "MetricsRegistry",
+    "get_registry",
+    "reset_registry_for_tests",
```
raballew: Remove `reset_registry_for_tests` from `__all__` and from `__init__.py`. Move it to `jumpstarter/metrics/_testing.py` (underscore-prefixed, not re-exported) and update test imports accordingly.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,82 @@
...
+    try:
+        server = ThreadingHTTPServer((host, port), Handler)
+    except OSError as e:
+        logger.warning(
+            "Failed to bind metrics server at %s:%s (%s); continuing without /metrics",
+            host,
+            port,
+            e,
+        )
+        return ""
```
raballew: `_parse_bind_addr` calls `int(port_s)` without guarding against non-numeric input. Suggested fix: move the `_parse_bind_addr` call inside the `try` block and widen the except clause to `except (OSError, ValueError) as e:`.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,82 @@
...
+                self.send_error(404)
+                return
+            body = reg.generate_latest()
+            self.send_response(200)
+            self.send_header("Content-Type", CONTENT_TYPE_LATEST)
+            self.send_header("Content-Length", str(len(body)))
+            self.end_headers()
+            self.wfile.write(body)
+
+        def log_message(self, fmt: str, *args) -> None:
```
raballew: Rename the parameter back to `format` (add `# noqa: A002` to suppress the builtin-shadow warning if needed): `def log_message(self, format: str, *args: Any) -> None: ...`

python/packages/jumpstarter/jumpstarter/exporter/session.py
```diff
@@ -57,11 +57,24 @@ class Session(
 
     @contextmanager
     def __contextmanager__(self) -> Generator[Self]:
+        from jumpstarter.logging import set_log_context
+        from jumpstarter.metrics import get_registry
```
raballew: Move both imports to the module-level import block. If a circular import exists, resolve it structurally rather than deferring.

python/packages/jumpstarter/jumpstarter/exporter/session.py
```diff
@@ -57,11 +57,24 @@ class Session(
...
+                get_registry().inc_active_sessions(exporter=self.name, delta=-1.0)
+            except Exception:
+                logger.warning(
+                    "Failed to decrement active sessions metric for exporter %s",
+                    self.name,
+                    exc_info=True,
+                )
             try:
                 self.root_device.close()
             except Exception as e:
```
raballew, line 80: Add `unbind_log_context("exporter")` in the `finally` block after the metrics decrement.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -0,0 +1,349 @@
...
+        operation="on",
+        result="success",
+        driver_type="power",
+        duration_seconds=0.01,
+        exemplars={"client": "ci-bot", "lease_id": "lease-xyz"},
+    )
+    body = reg.generate_latest().decode()
+    assert re.search(r'client="ci-bot"', body)
+    assert re.search(r'lease_id="lease-xyz"', body)
+    assert "# {" in body or " # {" in body
```
raballew: Use a regex: `assert re.search(r'jumpstarter_operations_total.*# \{.*client="ci-bot".*lease_id="lease-xyz"', body)` instead.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,82 @@
...
+            return
+
+    try:
+        server = ThreadingHTTPServer((host, port), Handler)
```
raballew, line 82: ThreadingHTTPServer has no request timeout and no connection limit.

python/packages/jumpstarter/jumpstarter/streams/common.py
```diff
@@ -37,11 +63,33 @@ async def copy_stream(dst: AnyByteStream, src: AnyByteStream):
 
 
 @asynccontextmanager
-async def forward_stream(a, b):
+async def forward_stream(a, b, *, metrics_driver_type: str | None = None):
+    from functools import partial
+
```
raballew: Move the import to the module-level import block.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def generate_latest(self) -> bytes:
+        return generate_latest_openmetrics(self._registry)
+
+
+def get_registry() -> MetricsRegistry:
+    global _REGISTRY
+    if _REGISTRY is None:
+        _REGISTRY = MetricsRegistry()
+    return _REGISTRY
```
raballew, line 152: The asyncio event loop and the `ThreadingHTTPServer` handler thread could both observe `_REGISTRY is None` simultaneously, each construct a `MetricsRegistry`, and the second assignment would silently discard any metrics already recorded by the first.

Initialize `_REGISTRY = MetricsRegistry()` eagerly at module import time, or guard the lazy init with a `threading.Lock`.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,158 @@
...
+            driver_type=driver_type,
+            direction=direction,
+        ).inc(nbytes, exemplar=filter_exemplars(exemplars))
+
+    def set_active_sessions(self, *, exporter: str, value: float) -> None:
+        self._active_sessions.labels(exporter=exporter).set(value)
+
+    def inc_active_sessions(self, *, exporter: str, delta: float = 1.0) -> None:
+        self._active_sessions.labels(exporter=exporter).inc(delta)
+
```
raballew: `inc_active_sessions` is named "increment" but is called with `delta=-1.0` at teardown time in `session.py`.

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Addressed via [21dd3501](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/21dd3501e6b71d046ecfd4cf090492f84df8253f)

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Addressed via [1806f85a](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/1806f85a3f3f36a1a192bf2668b05d72a4e8bb92)

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Addressed via [cb702784](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/cb7027847dadb8624e060275656d9b7b9ca79dc2)

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: Addressed via [5cd03059](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/5cd030599944d445119f480f463b53f575a6ff68)

python/packages/jumpstarter/jumpstarter/exporter/session.py
RoddieKieley: Addressed via [bf97c97b](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/bf97c97bc4506c4a11768efbcc2dc3561ff8597a)

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/metrics/__init__.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/exporter/session.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/streams/common.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: Addressed via [58df72b4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/commits/58df72b4ed64b086a1f86b2d8b677576d61ec730)

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,152 @@
...
+    def collector_registry(self) -> CollectorRegistry:
+        return self._registry
+
+    def record_operation(
+        self,
+        *,
+        exporter: str,
+        operation: str,
+        result: OperationResult,
+        driver_type: str,
```
mangelajo, line 95: This comment can be ignored, the plan is to do it metrics server side, and anyway I believe the bigger problem is going to be the operations. (previous comment L93)


**JEP-0013 conformance — `driver_type` is unbounded**

The JEP specifies a bounded `driverTypeEnum` allowlist (`storage`, `power`, `network`, `serial`, `console`, `video`, `composite`) with unmapped values remapped to `"other"`. However `driver_type` here is accepted as a plain `str` with no validation.

In practice, the codebase already has drivers using `automotive`, `bluetooth`, `gpio`, `debug`, `shell`, and `testing` — all outside the JEP's predefined set. There's no enforcement at the registry level, so each unique `driver_type` string creates new Prometheus label combinations, growing series cardinality.

The JEP envisions server-side remapping in `spec.telemetry.metrics.driverTypeEnum`, but that's a Phase 5 concern. For Phase 2, consider at least:
1. Documenting that `driver_type` cardinality is bounded by the number of loaded drivers (finite per process), or
2. Adding a `_KNOWN_DRIVER_TYPES` set and remapping unknown values to `"other"` in `record_operation()`.

Not a blocker — just flagging a gap between the JEP's design and the current implementation.

---

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,152 @@
...
+
+    @property
+    def collector_registry(self) -> CollectorRegistry:
+        return self._registry
+
+    def record_operation(
+        self,
+        *,
+        exporter: str,
+        operation: str,
```
mangelajo, line 93: **JEP-0013 conformance — `operation` label is also unbounded**

The `operation` label comes from `request.method` in `DriverCall`, which is the driver method name (e.g., `"on"`, `"off"`, `"flash"`). While the `grpc.aio.AbortError` catch prevents recording metrics for unknown methods rejected by `__lookup_drivercall`, the set of *valid* operations is still unbounded across all driver types.

The JEP mentions `operation` in the label set but doesn't explicitly constrain it. In practice this is bounded by the number of `@export` methods across all loaded drivers per process, so it's unlikely to cause cardinality explosion. But it's worth documenting this assumption.

---

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -38,12 +40,26 @@
...
+    get_registry,
+)
 from jumpstarter.streams.aiohttp import AiohttpStreamReaderStream
 from jumpstarter.streams.common import create_memory_stream
 from jumpstarter.streams.encoding import Compression, compress_stream
 from jumpstarter.streams.metadata import MetadataStream
 from jumpstarter.streams.progress import ProgressStream
 
+# Ordered most-specific first: ConnectionError is an OSError subclass.
+_DRIVER_CALL_ERRORS: tuple[tuple[type[BaseException], str, StatusCode], ...] = (
```
mangelajo: The `_DRIVER_CALL_ERRORS` table is clean and the ordering (most-specific first) is correct — `ConnectionError` before `OSError` since the former is a subclass.

One observation: the `error_type` values here (`"not_implemented"`, `"validation_error"`, `"timeout"`, `"connection_error"`, `"device_error"`, `"internal_error"`) match the `ErrorType` Literal in `registry.py`, which is good for type safety. But there's no compile-time enforcement that these strings match — if someone adds a new entry to this table with a typo, the `Literal` type won't catch it at the call site since `_record_operation_metrics` accepts `str | None` for `error_type`, not `ErrorType | None`.

Consider using the `ErrorType` literal directly in `_record_operation_metrics`'s signature and in this table's type annotation.

---

python/packages/jumpstarter/jumpstarter/streams/common.py
```diff
@@ -11,12 +12,38 @@
...
+):
     try:
+        # Capture once per copy; context should be stable for the stream lifetime.
+        if metrics_direction is not None:
+            metrics_exporter = exporter_from_log_context()
+            metrics_exemplars = exemplars_from_log_context()
+            metrics_registry = get_registry()
         async for v in src:
+            if metrics_direction is not None:
+                metrics_registry.add_stream_bytes(
```
mangelajo, line 40: Good that the metrics context (exporter, exemplars, registry) is captured once before the loop rather than per-chunk. This avoids per-iteration structlog lookups in the hot path.

However, `add_stream_bytes` is called for every chunk in the stream. For high-throughput streams (e.g., storage flashing), this means one `Counter.inc()` call per chunk. The `prometheus_client` Counter is thread-safe (uses a lock internally), so this could introduce contention if tx and rx run concurrently in the task group.

For Phase 2 this is likely fine (the contention is on different label sets so different `_ValueClass` instances). Just something to monitor if stream throughput becomes a concern — batching byte counts and flushing periodically would reduce overhead.

---

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,152 @@
...
+
+from __future__ import annotations
+
+from typing import Literal
+
+import structlog
+from prometheus_client import CollectorRegistry, Counter, Gauge, Histogram
+from prometheus_client.openmetrics.exposition import generate_latest as generate_latest_openmetrics
+
+DEFAULT_EXEMPLAR_KEYS = ("client", "lease_id")
```
mangelajo, line 11: **JEP-0013 conformance — exemplar keys**

The JEP lists exemplar keys as `client`, `lease_id`, and `trace_id` (when present). This implementation only includes `client` and `lease_id`. The `trace_id` is presumably a Phase 3+ concern (telemetry service), but it's worth noting the gap for tracking purposes.

---

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,104 @@
...
+
+        def log_message(self, format: str, *args) -> None:  # noqa: A002
+            return
+
+    try:
+        host, port = _parse_bind_addr(addr)
+        server = ThreadingHTTPServer((host, port), Handler)
+        server.request_queue_size = _METRICS_REQUEST_QUEUE_SIZE
+        server.timeout = _METRICS_REQUEST_TIMEOUT_S
+    except (OSError, ValueError) as e:
```
mangelajo: I think this should be fatal, if we want metrics this should become a more evident error to the administrator. Probably if we exit, a new random port will be tried, this risk could be avoided by listening on a unix socket or memory stream if that's possible with the prometheus stack.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -132,62 +218,35 @@ async def DriverCall(self, request, context):
...
-                extra={"operation": op, "driver_type": self.driver_type,
-                       "result": "failure", "error_type": "timeout"},
-            )
-            await context.abort(StatusCode.DEADLINE_EXCEEDED, str(e))
-        except ConnectionError as e:
-            self.logger.warning(
-                "Operation failed",
-                extra={"operation": op, "driver_type": self.driver_type,
-                       "result": "failure", "error_type": "connection_error"},
+            self._record_operation_metrics(
```
mangelajo, line 235: The `try/except` wrapping around `_record_operation_metrics` is critical and well-designed — a metrics failure must never discard an already-computed gRPC response. The test `test_driver_call_succeeds_when_metrics_recording_raises` validates this, which is excellent.

One subtlety: this line falls through to `_record_operation_metrics` which itself has a `try/except`. So there's double protection — the method-level wrapper and the internal wrapper. That's fine for defense-in-depth, but the outer wrapper here might mask bugs in the metrics code during development. Consider logging at DEBUG rather than WARNING for the inner catch if both levels are active.

---

python/packages/jumpstarter/jumpstarter/exporter/session.py
```diff
@@ -59,9 +61,20 @@ class Session(
     def __contextmanager__(self) -> Generator[Self]:
         logging.getLogger().addHandler(self._logging_handler)
         self.root_device.reset()
+        set_log_context(exporter=self.name)
+        get_registry().adjust_active_sessions(exporter=self.name, delta=1.0)
```
mangelajo: The session lifecycle correctly increments on entry and decrements in `finally`. The `unbind_log_context("exporter")` after the decrement ensures the exporter correlation field doesn't leak to subsequent contexts.

One edge case: if the process serves multiple exporters concurrently (multiple sessions), the structlog contextvars are per-async-context, so each session's `set_log_context(exporter=self.name)` should be isolated. This is correct as long as sessions run in separate async tasks, which they do via gRPC server handlers.

---

python/packages/jumpstarter-cli/jumpstarter_cli/run.py
```diff
@@ -294,8 +326,27 @@ def _serve_with_exc_handling(
...
+    default=":0",
+    show_default=True,
+    help=(
+        "Address for HTTP GET /metrics (Prometheus/OpenMetrics). "
+        "Default :0 binds an ephemeral loopback port. Use 0 to disable."
+    ),
+)
 @handle_exceptions
-def run(config, listener_bind, tls_insecure, tls_cert, tls_key, passphrase, exit_on_lease_end):
+def run(
```
mangelajo: The default `":0"` (ephemeral loopback port) means metrics are always enabled with a random port. This is useful for lab/dev scrape scenarios, but it means the exporter logs "Serving metrics server at http://127.0.0.1:XXXXX/metrics" every start and there's no way for an external scraper to know the port without reading logs.

For controller-mode exporters (no `--tls-grpc-listener`), the port is only reachable from localhost — useful for debugging but not for Prometheus scraping in a k8s pod unless combined with a sidecar or the Phase 3 reverse-scrape. 

Is the intent that controller-mode users would explicitly pass `--metrics-bind-address 0` to disable, or is the ephemeral default specifically designed for the Phase 3 reverse-scrape to connect to?

---

mangelajo (commented):
## Review Summary — PR #934: Exporter Prometheus metrics (JEP-0013 Phase 2)

This PR adds exporter-local Prometheus metrics with a `prometheus_client` registry, an HTTP `/metrics` endpoint, and instrumentation of the `DriverCall`/`StreamingDriverCall`/`Stream` code paths. Previous review feedback (26 threads from raballew and CodeRabbit) has been thoroughly addressed.

### JEP-0013 Compliance Assessment

**What Phase 2 should deliver per the JEP:**
1. `/metrics` scrape endpoints on Controller and Router _(covered by PR #933)_
2. Exporter-local `prometheus_client` counters/histograms/gauges with `driver_type`
3. Prometheus exemplars for `client`, `lease_id`, `trace_id` (when present)

**Metric series implemented vs. JEP specification:**

| JEP Series | Implemented | Notes |
|---|---|---|
| `jumpstarter_operations_total` | Yes | Labels: `exporter`, `operation`, `result`, `driver_type` — matches JEP |
| `jumpstarter_operation_duration_seconds` | Yes | Histogram with default buckets — matches JEP |
| `jumpstarter_operation_errors_total` | Yes | Labels include `error_type` — matches JEP |
| `jumpstarter_stream_bytes_total` | Yes | Labels: `exporter`, `driver_type`, `direction` — matches JEP |
| `jumpstarter_active_sessions` | Yes | Gauge with `exporter` label — matches JEP |
| `jumpstarter_lease_acquisitions_total` | No | Controller-side metric — not in scope for this PR |
| `jumpstarter_telemetry_dropped_total` | No | Phase 3 (Telemetry service) — correctly deferred |
| `jumpstarter_scrape_timeouts_total` | No | Phase 3 (Telemetry service) — correctly deferred |

**Gaps / deviations from JEP:**
1. **`trace_id` exemplar**: JEP lists `trace_id` as an exemplar key. This PR only implements `client` and `lease_id`. Likely a Phase 3+ concern.
2. **`driver_type` unbounded**: JEP specifies a `driverTypeEnum` allowlist with remap to `"other"`. The implementation accepts any string. The codebase already has 6 driver types outside the JEP's predefined set (`automotive`, `bluetooth`, `gpio`, `debug`, `shell`, `testing`). No remapping is done.
3. **`operation` unbounded**: The label comes from `@export` method names. Bounded per-process by loaded drivers but not validated against an allowlist.

All three gaps are documented in inline comments. None are blockers for Phase 2.

### What looks good

- **Isolation from RPC path**: `_record_operation_metrics` wraps all metrics calls in try/except, ensuring metrics failures never discard gRPC responses or change abort status codes. This is tested by `test_driver_call_succeeds_when_metrics_recording_raises`.
- **AbortError bypass**: The `grpc.aio.AbortError` catch before the generic `except Exception` prevents cardinality explosion from client-controlled invalid method names. Tested by `test_unknown_driver_method_does_not_record_operation_metric`.
- **Encode-before-record**: Success metrics are recorded only after the response is encoded, so serialization failures correctly count as failures.
- **Dedicated registry**: Uses a separate `CollectorRegistry` (not `prometheus_client.REGISTRY`), avoiding pollution from/to default Go-style process metrics. This is the right choice for the exporter.
- **Non-fatal server binding**: Metrics server bind failures log a warning and return `("", None)`, so the exporter continues operating without metrics. 
- **Error type consolidation**: The `_DRIVER_CALL_ERRORS` table replaces duplicated exception handling blocks in `DriverCall` and `StreamingDriverCall`, reducing code from ~90 lines to ~35 — a nice refactor.
- **Stream metrics**: `copy_stream` captures metrics context once before the loop, not per-chunk, avoiding repeated structlog lookups in the hot path.
- **Test coverage**: 400+ lines of tests covering the registry, server lifecycle, driver error mapping, exemplars, cardinality protection, and graceful degradation.

### Findings (see inline comments)

1. **`driver_type` unbounded** — No remapping of unknown driver types to `"other"` as the JEP specifies. Not a blocker but worth tracking.
2. **`operation` label unbounded** — Bounded in practice by loaded drivers but not documented/validated.
3. **`error_type` type safety** — `_record_operation_metrics` accepts `str | None` rather than `ErrorType | None`, so the Literal types can't catch typos.
4. **Stream metrics hot-path** — Per-chunk `Counter.inc()` could add overhead for high-throughput streams; monitor and batch if needed.
5. **Ephemeral port default** — `--metrics-bind-address` defaults to `:0` (always enabled, random port). Question about intent for controller-mode exporters.
6. **`trace_id` exemplar missing** — JEP includes it; presumably Phase 3.

No blocking issues. The implementation is well-aligned with JEP-0013 Phase 2 scope.

---

python/packages/jumpstarter-cli/jumpstarter_cli/run.py
```diff
@@ -99,46 +107,56 @@ async def signal_handler():
...
-                    exporter_exit_code = exporter.exit_code
-            else:
-                # Create exporter and run it (controller mode)
-                exporter_exit_code = None
-                async with config.create_exporter() as exporter:
-                    try:
-                        await exporter.serve()
-                    except* Exception as excgroup:
-                        _handle_exporter_exceptions(excgroup)
+            from jumpstarter.metrics import start_metrics_server
```
mangelajo: why do we import here and not at the top?

python/packages/jumpstarter-cli/jumpstarter_cli/run.py
```diff
@@ -294,8 +326,27 @@ def _serve_with_exc_handling(
     default=False,
     help="Exit after the current lease ends instead of waiting for a new one.",
 )
+@click.option(
```
mangelajo: more than an option itself, this should be something internal, that the exporter decides to start or not based on the availability of a metrics endpoint in jumpstarter.

What I mean is that this is an internal detail, the exporter should recognize if the endpoint should be created, probably we don't even need to listen on a TCP port, since the exporter itself will funnel the scrapping requests down to the metrics server. Could be a unix socket, or a memory stream.

Probably until we add a mechanism to look at the server availability of a metrics endpoint we can start it always by default in the main branch.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
mangelajo: This is true... We may want to keep an eye on this and move them to exemplars if necessary later. I would make a note of this on the JEP somewhere.

May be after the Unresolved questions, add a Risks section with a note on evaluating how this works? https://jumpstarter.dev/main/contributing/jeps/JEP-0013-observability-telemetry-logs.html#unresolved-questions

python/packages/jumpstarter/jumpstarter/streams/common.py
mangelajo: Can we add this to risks as well?, if we notice degradation in flashing performance we may want to batch the increments.

mangelajo (changes requested):

python/packages/jumpstarter/jumpstarter/streams/common.py
```diff
@@ -11,12 +12,38 @@
...
+                    exporter=metrics_exporter,
+                    driver_type=metrics_driver_type,
+                    direction=metrics_direction,
+                    nbytes=len(v) if isinstance(v, (bytes, bytearray, memoryview)) else 0,
+                    exemplars=metrics_exemplars,
+                )
             await dst.send(v)
         with suppress(
             AttributeError,
```
raballew: [HIGH] There is no test that exercises the `copy_stream` metrics path end-to-end. All stream-bytes tests call `reg.add_stream_bytes()` directly, bypassing the actual wiring. The `if metrics_direction is not None` branch inside the loop is never exercised by any test. Consider adding an anyio test that wires two in-memory streams through `copy_stream(dst, src, metrics_direction="tx", metrics_driver_type="serial")` with a known payload and asserts `_sample_value(reg, "jumpstarter_stream_bytes_total", {...}) == len(payload)`.

python/packages/jumpstarter/jumpstarter/exporter/session.py
```diff
@@ -59,9 +61,20 @@ class Session(
...
+            except Exception:
+                logger.warning(
+                    "Failed to decrement active sessions metric for exporter %s",
+                    self.name,
+                    exc_info=True,
+                )
+            unbind_log_context("exporter")
             try:
                 self.root_device.close()
             except Exception as e:
```
raballew: [HIGH] The `active_sessions` gauge decrement and log context cleanup on exception exit paths are untested. `session_test.py` has no test that raises inside a session body and verifies the gauge returned to 0.0 and the exporter log context was cleared. Consider adding a test that raises inside `with Session(...)`, catches it outside, and asserts both `_sample_value(get_registry(), "jumpstarter_active_sessions", {"exporter": name}) == 0.0` and `"exporter" not in structlog.contextvars.get_contextvars()`.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -0,0 +1,407 @@
+"""Exporter-local Prometheus metrics tests."""
```
raballew, line 1: There is no test covering client-initiated gRPC cancellation. Consider adding an integration test that starts a slow driver call, cancels it from the client side, and asserts `jumpstarter_operation_errors_total` is not incremented.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -113,11 +129,81 @@ def client(cls) -> str:
...
+        self.logger.warning(
+            "Operation failed",
+            extra={
+                "operation": op,
+                "driver_type": self.driver_type,
+                "result": "failure",
+                "error_type": "internal_error",
+            },
+        )
+        await context.abort(StatusCode.UNKNOWN, str(exc))
```
raballew: [HIGH] `_handle_driver_exception` carries three orthogonal concerns: exception classification, metrics recording, and gRPC abort. The record+log+abort triple is duplicated verbatim for the generic fallback case, so any future change must be applied in two places. Consider extracting `_classify_exception(exc) -> tuple[str, StatusCode]` that walks `_DRIVER_CALL_ERRORS`, then having `_handle_driver_exception` call classify, record metrics, and abort, eliminating the duplicated blocks.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -113,11 +129,81 @@ def client(cls) -> str:
     def extra_labels(self) -> dict[str, str]:
         return {}
 
+    def _record_operation_metrics(
+        self,
+        *,
+        operation: str,
+        result: str,
+        duration_seconds: float,
+        error_type: str | None = None,
```
raballew: `_record_operation_metrics` declares `result: str` and `error_type: str | None`, but passes them to `MetricsRegistry.record_operation` which expects `result: OperationResult = Literal["success", "failure"]` and `error_type: ErrorType | None`. Consider importing `OperationResult` and `ErrorType` from `jumpstarter.metrics.registry` and narrowing the signature accordingly.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,104 @@
...
+            self.end_headers()
+            self.wfile.write(body)
+
+        def log_message(self, format: str, *args) -> None:  # noqa: A002
+            return
+
+    try:
+        host, port = _parse_bind_addr(addr)
+        server = ThreadingHTTPServer((host, port), Handler)
+        server.request_queue_size = _METRICS_REQUEST_QUEUE_SIZE
```
raballew: `server.request_queue_size = _METRICS_REQUEST_QUEUE_SIZE` is assigned after init. Consider picking up the correct value before `socket.listen()` is called.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,104 @@
...
+    except (OSError, ValueError) as e:
+        logger.warning(
+            "Failed to start metrics server for bind address %r (%s); continuing without /metrics",
+            addr,
+            e,
+        )
+        return "", None
+
+    thread = Thread(target=server.serve_forever, name="jumpstarter-metrics", daemon=True)
+    thread.start()
```
raballew: [HIGH] `ThreadingHTTPServer.__init__` calls `server_bind()` and `server_activate()`, binding the socket into LISTEN state before `Thread.start()` is called. If `Thread.start()` raises `RuntimeError` (e.g., due to OS thread-count limits), the bound socket leaks since the exception propagates without any cleanup. Consider wrapping `thread.start()` in `try/except Exception: server.server_close(); raise` to ensure the socket is released on thread startup failure.

python/packages/jumpstarter/jumpstarter/streams/common.py
raballew: [HIGH] `copy_stream` was a pure byte-forwarding utility and now imports three symbols from `jumpstarter.metrics.registry` and conditionally calls into the metrics registry on every chunk. The `if metrics_direction is not None` guard appears twice (once before the loop to initialize helpers and once inside the hot path per chunk), adding a per-chunk branch even for callers that want no metrics. Consider keeping `copy_stream` free of metrics concerns and moving the accounting into `forward_stream` (which already decides whether metrics are desired), or accepting a callback `on_chunk: Callable[[bytes], None] | None = None` so metrics are injected rather than embedded.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
```diff
@@ -0,0 +1,152 @@
...
+    def add_stream_bytes(
+        self,
+        *,
+        exporter: str,
+        driver_type: str,
+        direction: StreamDirection,
+        nbytes: int,
+        exemplars: dict[str, str] | None = None,
+    ) -> None:
+        # Zero/negative transfers are not observed (no counter change).
```
raballew: [MEDIUM] The comment `# Zero/negative transfers are not observed (no counter change).` just restates what the two-line guard below it already expresses. The guard is self-explanatory; the comment does not add information. Consider removing it.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -0,0 +1,407 @@
...
+    body = reg.generate_latest().decode()
+    # Series may be absent entirely until a positive observation.
+    assert 'direction="tx"' not in body or _sample_value(
+        reg,
+        "jumpstarter_stream_bytes_total",
+        {"exporter": "lab-01", "driver_type": "serial", "direction": "tx"},
+    ) in (None, 0.0)
+
+
+def test_generate_latest_contains_named_series_after_increments():
```
raballew: [MEDIUM] The assertion `assert 'direction="tx"' not in body or _sample_value(...) in (None, 0.0)` is vacuously true with a fresh registry. Since only non-positive byte counts are added, the counter is never created, so the first disjunct (`not in body`) is always True and short-circuits. An implementation that silently writes a zero-increment counter would still pass. Consider replacing with `val = _sample_value(reg, "jumpstarter_stream_bytes_total", {...}); assert val is None`.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,104 @@
...
+                return
+            body = reg.generate_latest()
+            self.send_response(200)
+            self.send_header("Content-Type", CONTENT_TYPE_LATEST)
+            self.send_header("Content-Length", str(len(body)))
+            self.end_headers()
+            self.wfile.write(body)
+
+        def log_message(self, format: str, *args) -> None:  # noqa: A002
+            return
```
raballew: [MEDIUM] The `Handler` class is defined inside `start_metrics_server` and captures `reg` from the enclosing function scope. There is no way to instantiate `Handler` with a mock registry without calling `start_metrics_server`, which starts a real OS thread and binds a real port. The `do_GET` logic (404 guard, response headers, body write) cannot be exercised in isolation. Consider lifting `Handler` to module scope via `_make_handler_class(reg: MetricsRegistry)` returning a `Handler` subclass with `_registry = reg`, enabling direct instantiation in tests without network binding.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,104 @@
...
+    """
+    if addr == "" or addr == "0":
+        return "", None
+
+    reg = registry if registry is not None else get_registry()
+
+    class Handler(BaseHTTPRequestHandler):
+        timeout = _METRICS_REQUEST_TIMEOUT_S
+
+        def do_GET(self):  # noqa: N802
```
raballew: [MEDIUM] `def do_GET(self):  # noqa: N802` has no return type annotation. Since `BaseHTTPRequestHandler` does not declare a `do_GET` method, type checkers cannot infer the return type. Consider adding `-> None` to the signature.

python/packages/jumpstarter/jumpstarter/metrics/server.py
```diff
@@ -0,0 +1,104 @@
...
+            if self.path.split("?", 1)[0] != "/metrics":
+                self.send_error(404)
+                return
+            body = reg.generate_latest()
+            self.send_response(200)
+            self.send_header("Content-Type", CONTENT_TYPE_LATEST)
```
raballew: [MEDIUM] The server imports `CONTENT_TYPE_LATEST` from `prometheus_client.openmetrics.exposition` (value: `application/openmetrics-text; version=1.0.0; charset=utf-8`) and uses `generate_latest_openmetrics`. The test acknowledges the mismatch with `assert "text/plain" in ctype or "openmetrics" in ctype`. If text/plain is the intended format, switch to `from prometheus_client.exposition import generate_latest, CONTENT_TYPE_LATEST`. If OpenMetrics is intentional, update the test to assert only the OpenMetrics content type rather than hedging on both.

python/packages/jumpstarter/jumpstarter/driver/base.py
```diff
@@ -113,11 +129,81 @@ def client(cls) -> str:
...
+                extra={
+                    "operation": operation,
+                    "driver_type": self.driver_type,
+                    "result": result,
+                    "error_type": error_type,
+                },
+                exc_info=True,
+            )
+
+    async def _handle_driver_exception(self, exc: BaseException, op: str, started: float, context) -> None:
```
raballew: Consider adding `context: grpc.aio.ServicerContext` to the signature.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -0,0 +1,407 @@
...
+        reg,
+        "jumpstarter_operation_errors_total",
+        {
+            "exporter": "lab-01",
+            "operation": "flash",
+            "driver_type": "storage",
+            "error_type": "timeout",
+        },
+    )
+    assert errors == 1.0
```
raballew: [MEDIUM] `re.search(r'jumpstarter_operations_total.*# \{.*client="ci-bot".*lease_id="lease-xyz"', body, re.DOTALL)` -- with `re.DOTALL`, `.*` matches newlines, so `client="ci-bot"` and `lease_id="lease-xyz"` can appear in any later metric family's exemplar line. This would pass even if the exemplar is attached to a different metric and `jumpstarter_operations_total` has no exemplar at all. Consider splitting the body into lines, finding the `jumpstarter_operations_total` counter line, and verifying the exemplar is on that line or immediately after. Remove `re.DOTALL`.

python/packages/jumpstarter-cli/jumpstarter_cli/run.py
```diff
@@ -294,8 +326,27 @@ def _serve_with_exc_handling(
...
+    tls_insecure,
+    tls_cert,
+    tls_key,
+    passphrase,
+    exit_on_lease_end,
+    metrics_bind_address,
+):
     """Run an exporter locally."""
     if listener_bind is not None and config is None:
         raise click.UsageError("--exporter-config (or --exporter) is required when using --tls-grpc-listener")
```
raballew: [MEDIUM] The `--metrics-bind-address` Click option added here has no tests. No test verifies the option is registered and parsed, that `"0"` disables the server, or that a valid address causes a bind and is stopped on exit. The `finally: shutdown_metrics()` cleanup path is also untested. Consider adding a Click CLI test using `CliRunner.invoke` with `--metrics-bind-address=0` (disabled path) and a second test with `127.0.0.1:0` that asserts `GET /metrics` responds 200 and verifies shutdown is called on exit.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
raballew: [MEDIUM] `test_generate_latest_contains_named_series_after_increments` drives a success operation, a failure operation with an error type, stream bytes, and an active-session gauge set, then asserts all five series, multiple labels, and several sampled values in one 65-line function. Each behavior is an independent specification point. Consider splitting into `test_record_operation_success_increments_counter`, `test_record_operation_failure_increments_error_counter`, `test_add_stream_bytes_increments_counter`, and `test_set_active_sessions_writes_gauge`, each with a single record/assert pair.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -0,0 +1,407 @@
...
+    structlog.contextvars.clear_contextvars()
+    assert exemplars_from_log_context() is None
+    assert exporter_from_log_context() == "unknown"
+
+    structlog.contextvars.bind_contextvars(client="ci-bot", lease_id="lease-1", exporter="lab-01")
+    try:
+        assert exemplars_from_log_context() == {"client": "ci-bot", "lease_id": "lease-1"}
+        assert exporter_from_log_context() == "lab-01"
+    finally:
+        structlog.contextvars.clear_contextvars()
```
raballew: [MEDIUM] `test_exemplars_and_exporter_from_log_context` asserts the behavior of both `exemplars_from_log_context` and `exporter_from_log_context` in one body. If either function is broken, only one assertion pair fails but the test name does not indicate which contract was violated. Consider splitting into `test_exemplars_from_log_context_reads_bound_keys` and `test_exporter_from_log_context_reads_bound_exporter`, each with its own arrange-act-assert cycle.

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: Added this to [Risks](https://github.com/RoddieKieley/jumpstarter/blob/1bb49a2b0d00d50f0cb5a735c99415ca1be6fb0b/docs/source/contributing/jeps/JEP-0013-observability-telemetry-logs.md#risks) as well as [Unresolved Questions](https://github.com/RoddieKieley/jumpstarter/blob/1bb49a2b0d00d50f0cb5a735c99415ca1be6fb0b/docs/source/contributing/jeps/JEP-0013-observability-telemetry-logs.md#unresolved-questions)

python/packages/jumpstarter/jumpstarter/streams/common.py
RoddieKieley: Added this to [Risks](https://github.com/RoddieKieley/jumpstarter/blob/1bb49a2b0d00d50f0cb5a735c99415ca1be6fb0b/docs/source/contributing/jeps/JEP-0013-observability-telemetry-logs.md#risks) as well as [Unresolved Questions](https://github.com/RoddieKieley/jumpstarter/blob/1bb49a2b0d00d50f0cb5a735c99415ca1be6fb0b/docs/source/contributing/jeps/JEP-0013-observability-telemetry-logs.md#unresolved-questions)

python/packages/jumpstarter/jumpstarter/metrics/registry.py
RoddieKieley: This is indeed an intentional Phase 2 gap and a Phase 3+ concern.

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Kept the logging at WARNING for the moment. If this becomes too noisy will drop it to DEBUG later.

python/packages/jumpstarter-cli/jumpstarter_cli/run.py
RoddieKieley: Moved to the top.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
RoddieKieley: Addressed in [8574c117](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/8574c11786674c8861ed3c67d2d2e84a14395ca3)

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Addressed in [8574c117](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/8574c11786674c8861ed3c67d2d2e84a14395ca3)

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Addressed in [8574c117](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/8574c11786674c8861ed3c67d2d2e84a14395ca3)

python/packages/jumpstarter/jumpstarter/driver/base.py
RoddieKieley: Addressed in [8574c117](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/8574c11786674c8861ed3c67d2d2e84a14395ca3)

python/packages/jumpstarter/jumpstarter/metrics/server.py
RoddieKieley: Addressed via [697d81c4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/697d81c46b325e316ef0aa5477e5b7e780f7f706)

python/packages/jumpstarter-cli/jumpstarter_cli/run.py
RoddieKieley: Addressed via [697d81c4](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/697d81c46b325e316ef0aa5477e5b7e780f7f706)

python/packages/jumpstarter/jumpstarter/streams/common.py
raballew, line 49: All `stream_bytes_total` assertions in `metrics_test.py` call `reg.add_stream_bytes()` directly on the registry, bypassing the actual chunk-iteration wiring in `copy_stream`. No test exercises `forward_stream` or `copy_stream` with metric arguments.

raballew (approved):
Only one nit left. I leave it up to you to fix it, if you think it adds value.

python/packages/jumpstarter/jumpstarter/streams/common.py
RoddieKieley: Addressed via [c7df5c9f](https://github.com/jumpstarter-dev/jumpstarter/pull/934/changes/c7df5c9f38694bdc0a24d2b0417542f90ed9cb59)

RoddieKieley:
@mangelajo I addressed the last nit there from @raballew after his original approval which I think still stands. We need CI go ahead as well as a pass by you I think - anything else outstanding?

mangelajo (approved):

PR #936, opened 2026-08-02
feat: cuttlefish driver

bennyz:
Add jumpstarter-driver-cuttlefish, managing Android Cuttlefish virtual devices (CVDs) through the Host Orchestrator REST API.

Composite driver with three children:
  - power   VirtualPowerInterface; on() creates a CVD from env_config or
            starts an existing one, off() stops it, off(destroy=True)
            deletes it.
  - storage FlasherInterface (stubbed for now)
  - adb     AdbServer, auto-connected and waited on after power on.

Also exposes cuttlefish-specific operations: powerwash, powerbtn, and CVD listing.

### DIFF
--- a/python/packages/jumpstarter-driver-cuttlefish/README.md
+++ b/python/packages/jumpstarter-driver-cuttlefish/README.md
@@ -1,0 +1,313 @@
+# Cuttlefish Driver
+
+`jumpstarter-driver-cuttlefish` manages
+[Android Cuttlefish](https://source.android.com/docs/devices/cuttlefish)
+virtual devices through the
+[Host Orchestrator](https://github.com/google/android-cuttlefish) REST API.
+It provides full CVD (Cuttlefish Virtual Device) lifecycle management through
+standard Jumpstarter interfaces: `VirtualPowerInterface` for on/off/cycle,
+plus cuttlefish-specific operations
+(snapshot, powerwash, restart).
+
+## Installation
+
+```{code-block} console
+:substitutions:
+$ pip3 install --extra-index-url {{index_url}} jumpstarter-driver-cuttlefish
+```
+
+### Prerequisites
+
+- A running Cuttlefish Host Orchestrator (port 2080 by default)
+
+## Host Setup
+
+A `cvd-images` named volume mounted at `/home/vsoc-01/fetch` persists
+fetched AOSP images across container restarts. Instance state (`/var/tmp/cvd`)
+is deliberately kept ephemeral — restarting the container gives you a clean
+slate with no orphaned instance directories.
+
+```bash
+# 1. Pull the orchestration image
+podman pull us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration:stable
+
+# 2. Create a named volume for AOSP images
+podman volume create cvd-images
+
+# 3. Start the container
+#    --network=host: netsim and rootcanal bind to 127.0.0.1 inside the
+#    container, so without host networking they'd be unreachable from
+#    outside. Host networking shares the VM's network namespace directly.
+#
+#    Security note: --privileged + --network=host gives the container full
+#    access to the VM's network stack. HO, netsim, and rootcanal have no
+#    auth — only deploy on dedicated, non-public hosts.
+podman run -d \
+  --name cuttlefish-orchestrator \
+  --restart=always \
+  --privileged \
+  --network=host \
+  -v cvd-images:/home/vsoc-01/fetch:Z \
+  -v /opt/cuttlefish:/opt/cuttlefish:Z \
+  us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration:stable
+
+# 4. Fix permissions
+podman exec cuttlefish-orchestrator chown -R httpcvd:httpcvd /home/vsoc-01/fetch
+
+# 5. Fetch AOSP images (one-time, ~2 minutes)
+podman exec cuttlefish-orchestrator cvd fetch \
+  --default_build=aosp-android-latest-release/aosp_cf_x86_64_auto-userdebug \
+  --target_directory=/home/vsoc-01/fetch
+
+# 6. Verify
+curl -s http://localhost:2080/_debug/statusz  # should return 200
+curl -s http://localhost:2080/cvds             # should return {"cvds":[]}
+```
+
+### Ports
+
+After a CVD boots, the following ports are available on the host.
+All per-instance ports use the same offset: `base + instance_num - 1`.
+
+| Service | Base port | Instance 1 | Instance 2 |
+|---------|-----------|------------|------------|
+| Host Orchestrator | 2080 | 2080 (fixed) | 2080 (fixed) |
+| ADB | 6520 | 6520 | 6521 |
+| Netsim REST | 7681 | 7681 | 7682 |
+| Rootcanal HCI | 7300 | 7300 | 7301 |
+
+When using `instance_num > 1`, update the netsim `port` and bt_peer
+`hci_port` in the exporter config to match.
+
+### SSH tunnel (local development only)
+
+For local development, when running the exporter on your workstation
+instead of as a pod, tunnel ports from the VM. This works because
+`--network=host` places netsim and rootcanal on the VM's loopback -
+the tunnel's `localhost` target reaches them directly.
+
+```bash
+ssh -L 2080:localhost:2080 \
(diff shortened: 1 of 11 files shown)


bennyz:
@kirkbrauer er could you have a preliminary look? the idea is to use cuttlefish's host orchestrator rather than construct CLI commands, but not the full cloud orchestrator which in our case would be jumpstarter's job

python/packages/jumpstarter-driver-cuttlefish/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+```bash
+# Health check
+j cuttlefish status
+
+# List all CVDs
+j cuttlefish list
+
+# Create an AAOS CVD from Android CI (auto-downloads images)
+j cuttlefish create '{"instances":[{"name":"auto1","disk":{"default_build":"@ab/aosp-android-latest-release/aosp_cf_x86_64_auto-userdebug"},"vm":{"cpus":4,"memory_mb":4096}}]}'
```
kirkbrauer, line 84: Might want to provide more user-friendly arguments here with a JSON fallback if desired.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -0,0 +1,166 @@
...
+    host: str = "localhost"
+    port: int = 2080
+
+    @classmethod
+    def client(cls) -> str:
+        return "jumpstarter_driver_cuttlefish.client.CuttlefishClient"
+
+    @property
+    def _base_url(self) -> str:
+        return f"http://{self.host}:{self.port}"
```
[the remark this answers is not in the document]
kirkbrauer: @bennyz This does bring up an interesting point of remote host orchestrators in the case that we would want to do something like that. I know @mangelajo  is looking at similar things for QEMU.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/client.py
```diff
@@ -0,0 +1,214 @@
...
+    def delete_snapshot(self, snapshot_id: str) -> dict | list | str:
+        return _parse(self.call("delete_snapshot", snapshot_id))
+
+    def status(self) -> str:
+        return self.call("status")
+
+    def list_operations(self) -> dict | list | str:
+        return _parse(self.call("list_operations"))
+
+    def get_adb_port(self, group: str, name: str) -> int:
```
kirkbrauer, line 92: We should think about how to handle ADB, maybe a nested ADB driver within a Cuttlefish composite driver or maybe we just make an example of how to handle this since ADB is probably the most important aspect here.

python/packages/jumpstarter-driver-cuttlefish/README.md
```diff
@@ -0,0 +1,216 @@
...
+
+## Configuration
+
+Example exporter configuration:
+
+```yaml
+export:
+  cuttlefish:
+    type: jumpstarter_driver_cuttlefish.driver.Cuttlefish
+    config:
```
kirkbrauer, line 44: We might want to expose more complex configuration here like the image to download or host configuration, so the user can just do `cuttlefish.on()` and get a pre-configured device, but that might also be fine to put into the ExporterClass config. Thoughts @mangelajo?

python/packages/jumpstarter-driver-cuttlefish/README.md
```diff
@@ -0,0 +1,216 @@
...
+j cuttlefish delete <group> <name>
+
+# Get ADB port for a CVD
+j cuttlefish adb-port <group> <name>
+
+# List running operations
+j cuttlefish ops
+
+# Snapshot management
+j cuttlefish snapshot create <group> <name> --id my-snapshot
```
kirkbrauer, line 113: We do need to figure out where snapshots go, maybe we can bundle them as OCI artifacts and either send them back to the client or to some registry?

python/packages/jumpstarter-driver-cuttlefish/README.md
```diff
@@ -0,0 +1,216 @@
...
+delete) are handled asynchronously - the driver polls the `/operations/:wait`
+endpoint until completion or timeout.
+
+### CVD Build Sources
+
+The `create_cvd` config supports two build source formats in the `disk.default_build` field:
+
+- **Android CI**: `@ab/<branch>/<target>` - fetches images from Android Build servers.
+  Example: `@ab/aosp-android-latest-release/aosp_cf_x86_64_auto-userdebug` (AAOS)
+- **Local path**: `/path/to/android/build`  uses a local checkout with built Cuttlefish images.
```
kirkbrauer, line 191: Adding flasher support might also be a nice thing to do later where we can "flash" an image to a local path and then boot from that.

kirkbrauer (commented):
@bennyz I think this is a great step towards Cuttlefish support. We might just want to merge this as-is as a starting point and then expand later.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/client.py
bennyz: in my testing I was using it as a sibling, but a child driver might actually be nicer here, since we have a bit of dance of:

```
j cuttlefish create ...
j adb connect <using output from first command> (added to the adb driver in another PR)
```
i'll check if it's simple to not have to do that

kirkbrauer: The ADB driver already does wrapping of the underlying ADB connection handling, so it might be easy to grab that and then automatically route it via `j cuttlefish adb` or something like that.

python/packages/jumpstarter-driver-cuttlefish/README.md
bennyz: EnvConfig is considered unstable (or they just forgot to remove the comment)
https://github.com/google/android-cuttlefish/blob/762bf3a532d7c7634cfe317d8c71b7cdcb4fdbae/frontend/src/host_orchestrator/api/v1/messages.go#L25

So we need to think about supported version, but perhaps simply adding an env_config field and forwarding would be enough

python/packages/jumpstarter-driver-cuttlefish/README.md
bennyz: i actually thought about removing the snapshots, and i am not sure how portable they are, i need to check, but i do like the OCI idea

bennyz: and it only works on x86:
```
Note: Snapshots are only supported on x86_64 platforms.
```

python/packages/jumpstarter-driver-cuttlefish/README.md
bennyz: removed it for now, env config will be set on the exporter config directly, since this is 1:1 anyway

python/packages/jumpstarter-driver-cuttlefish/README.md
kirkbrauer: Ah, yeah not that useful in this case.

bennyz:
@kirkbrauer i've dropped snapshot for now, we can revisit later i suppose

kirkbrauer (approved):

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -0,0 +1,467 @@
...
+        """Wait for CVD to be ADB-reachable and fully booted."""
+        adb = self.children.get("adb")
+        if not adb:
+            return
+
+        device = self._cvd_device
+
+        deadline = time.monotonic() + timeout
+        adb_path = adb.adb_path
+        adb_env = adb.adb_env()
```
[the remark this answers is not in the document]
kirkbrauer: @coderabbitai Can you take a stab at this?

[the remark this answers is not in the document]
bennyz: this is defined in https://github.com/jumpstarter-dev/jumpstarter/pull/977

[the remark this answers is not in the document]
bennyz: @coderabbitai this is defined in https://github.com/jumpstarter-dev/jumpstarter/pull/977

PR #942, opened 2026-08-03
feat: lease sharing

bennyz:
```
$ jmp shell -n lease-test-exporter --client lease-test-client
[08/16/26 10:49:19] INFO     [jumpstarter.client.lease] Acquiring lease 01a0098b-e942-719f-9cc2-f7398584dabf for
                             selector None for duration 0:30:00
                    INFO     [jumpstarter_cli.shell] Waiting for beforeLease hook to complete...
                    INFO     [jumpstarter.client.status_monitor] Status changed: None -> LEASE_READY (version=2)

jmp share add 01a0098b-e942-719f-9cc2-f7398584dabf lease-test-client2 --client lease-test-client
NAME                                  SELECTOR  EXPIRES AT   REMAINING  CLIENT       EXPORTER     TAGS  SHARED WITH
01a0098b-e942-719f-9cc2-f7398584dabf  <none>    2026-08-16   29m        lease-test…  lease-test…        lease-test…
                                                11:19:19
Shared lease 01a0098b-e942-719f-9cc2-f7398584dabf with: lease-test-client2

$$ j serial start-console

Starting serial port console ... exit with CTRL+B x 3 times

fsdsdsdA hello fhhello dsfsd⏎
```

```
$ jmp shell --lease 01a009c7-8e0e-710f-adfa-cba97d477b73  --client lease-test-client2
[08/16/26 11:55:07] INFO     [jumpstarter_cli.shell] Waiting for beforeLease hook to complete...
                    INFO     [jumpstarter.client.status_monitor] Status changed: None -> LEASE_READY (version=2)
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
$$ j serial start-console

Starting serial port console ... exit with CTRL+B x 3 times

Error: Console in use. Use --observe or release-console.

$$ j serial start-console --observe

Starting serial console in observe mode (read-only) ... exit with CTRL+B x 3 times

fsdsdsdA hello fhhello dsfsd⏎
```


mangelajo:
We will need to fix the serial multi-reader :D and may be other streams too (like the Ble ..) . :)

controller/internal/service/client/v1/client_service.go
```diff
@@ -322,6 +323,18 @@ func (s *ClientService) CreateLease(ctx context.Context, req *cpb.CreateLeaseReq
...
+		for _, name := range jlease.Spec.SharedWith {
+			if name == jclient.Name {
+				return nil, status.Errorf(codes.InvalidArgument, "cannot share lease with the owner")
+			}
+			var sharedClient jumpstarterdevv1alpha1.Client
+			if err := s.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, &sharedClient); err != nil {
+				return nil, status.Errorf(codes.InvalidArgument, "shared client %q not found", name)
+			}
+		}
+	}
```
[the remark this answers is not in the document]
mangelajo: @bennyz  this ^

controller/internal/service/controller_service.go
```diff
@@ -1143,7 +1143,7 @@ func (s *ControllerService) ReleaseLease(
 		return nil, err
 	}
 
-	if lease.Spec.ClientRef.Name != jclient.Name {
+	if !lease.IsAccessibleBy(jclient.Name) {
```
mangelajo: I probably disagree with the agent here, perhaps at some point we can make this configurable?

**Bug: Shared clients can terminate leases they don't own**

`ReleaseLease` sets `Spec.Release = true`, which is a destructive, irreversible operation that ends the lease for **everyone** — the owner and all shared clients. Using `IsAccessibleBy` here means any shared client can unilaterally terminate the lease.

This contradicts the authorization pattern used elsewhere:
- `DeleteLease` (`client_service.go:558`) correctly uses `IsOwnedBy`
- `UpdateLease` gates sharing, time changes, and transfers behind `IsOwnedBy`

Since `ReleaseLease` and `DeleteLease` have the same effect (set `Spec.Release = true`), they should have the same authorization.

```go
if !lease.IsOwnedBy(jclient.Name) {
    return nil, fmt.Errorf("ReleaseLease permission denied: only lease owner can release")
}
```

---

controller/internal/controller/lease_controller.go
```diff
@@ -541,6 +545,52 @@ func (r *LeaseReconciler) attachMatchingPolicies(ctx context.Context, lease *jum
...
+	}
+
+	logger := log.FromContext(ctx)
+	var allowed []string
+	for _, clientName := range lease.Spec.SharedWith {
+		var jclient jumpstarterdevv1alpha1.Client
+		if err := r.Get(ctx, types.NamespacedName{
+			Namespace: lease.Namespace,
+			Name:      clientName,
+		}, &jclient); err != nil {
```
mangelajo, line 595: **Bug: Transient API errors permanently remove shared clients**

This `Get()` treats **all** errors as "not found" — including transient failures (network timeouts, API server 500s, rate limiting). When a transient error occurs, the client is silently dropped from the `SharedWith` list and the filtered result is persisted via the later `r.Update()` call.

Compare with `reconcileStatusExporterRef` (line 255-268), which correctly distinguishes `k8serrors.IsNotFound` from other errors and returns the error to trigger a requeue.

```go
if err := r.Get(ctx, types.NamespacedName{
    Namespace: lease.Namespace,
    Name:      clientName,
}, &jclient); err != nil {
    if k8serrors.IsNotFound(err) {
        logger.Info("removing shared client: not found", "client", clientName)
        continue
    }
    return fmt.Errorf("reconcileSharedWithPolicies: failed to get client %s: %w", clientName, err)
}
```

---

bennyz:
going to do another pass on this, a lot has changed since the last rebase :)

controller/api/v1alpha1/lease_helpers.go
```diff
@@ -396,6 +398,45 @@ func (l *Lease) SetStatusCondition(
...
+
+func ClientAllowedByPolicy(
+	policies []ExporterAccessPolicy,
+	exporter *Exporter,
+	jclient *Client,
+) bool {
+	for _, policy := range policies {
+		exporterSelector, err := metav1.LabelSelectorAsSelector(&policy.Spec.ExporterSelector)
+		if err != nil {
+			continue
```
mangelajo: **Inconsistency: Malformed policy selectors silently skipped**

When `LabelSelectorAsSelector` returns an error for a malformed policy, the function silently skips it — denying access with no log or indication that the policy was misconfigured.

This is inconsistent with `attachMatchingPolicies` (`lease_controller.go:502-505`), which treats the same error as fatal:
```go
exporterSelector, err := metav1.LabelSelectorAsSelector(&policy.Spec.ExporterSelector)
if err != nil {
    return nil, nil, fmt.Errorf("... failed to convert exporter selector: %w", err)
}
```

A misconfigured policy could silently cause shared clients to be denied and then removed by `reconcileSharedWithPolicies`, making the behavior hard to diagnose. Consider at minimum logging a warning, or changing the signature to return `(bool, error)`.

---

controller/internal/service/client/v1/client_service.go
```diff
@@ -361,66 +377,52 @@ func (s *ClientService) UpdateLease(ctx context.Context, req *cpb.UpdateLeaseReq
...
+	hasShareChanges := len(req.AddSharedWith) > 0 || len(req.RemoveSharedWith) > 0
+	hasTransfer := req.Lease.Client != nil && *req.Lease.Client != ""
+
+	if hasTransfer && hasShareChanges {
+		return nil, fmt.Errorf("UpdateLease: cannot transfer and modify sharing in the same request")
+	}
+
+	if hasShareChanges {
+		if !jlease.IsOwnedBy(jclient.Name) {
+			return nil, fmt.Errorf("UpdateLease permission denied: only lease owner can modify sharing")
```
mangelajo: **Nit: Error responses use `fmt.Errorf` instead of gRPC status codes**

Throughout the sharing/transfer logic (lines 384, 389, 392, 400, 403, 417, etc.), errors are returned as `fmt.Errorf(...)`, which produces gRPC `codes.Unknown`. In contrast, `CreateLease` correctly uses `status.Errorf(codes.X, ...)`.

For better client-side error handling, consider using proper gRPC codes:
- Permission denials → `codes.PermissionDenied`
- "lease has already ended" → `codes.FailedPrecondition`
- "client not found" → `codes.NotFound`
- "cannot transfer and modify sharing" → `codes.InvalidArgument`

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,525 @@
...
+                self._pending_drop_marker = True
+        self._buf.append(data)
+        self._current_bytes += len(data)
+        self._event.set()
+
+    async def pull(self) -> bytes:
+        """Async pull. Blocks until data available."""
+        while not self._buf and not self._closed:
+            self._event = anyio.Event()
+            await self._event.wait()
```
mangelajo, line 98: **Concurrency: `pull()` event replacement can miss wakeups from concurrent `push()`**

Between creating the new `Event` on line 90 and awaiting it on line 91, a concurrent `push()` call can set the **old** event (via `self._event.set()` at line 85). The new event will never be set, causing the puller to hang indefinitely.

Sequence:
1. `pull()` checks `while not self._buf` → true, enters loop
2. `push()` runs: appends data to `_buf`, calls `self._event.set()` on old event
3. `pull()` replaces `self._event = anyio.Event()` → new event
4. `pull()` awaits new event → hangs forever (data is in `_buf` but no one sets the new event)

Fix by re-checking the condition after publishing the new event:

```python
async def pull(self) -> bytes:
    while not self._buf and not self._closed:
        event = anyio.Event()
        self._event = event
        # Re-check after publishing the new event
        if self._buf or self._closed:
            break
        await event.wait()
```

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,525 @@
...
+                    break
+                raise
+
+        self._source = None
+        self._reader_running = False
+
+    def _broadcast_data(self, data: bytes) -> None:
+        """Push data to all clients, removing any that error."""
+        disconnected = []
+        for client_id, buf in self._clients.items():
```
mangelajo: **Concurrency: `_broadcast_data` mutates `_clients` dict without holding `_lock`**

`_broadcast_data` is called from `_reader_loop` (line 274) without holding `self._lock`. It iterates `self._clients` (line 306) and may remove entries (lines 311-312). Meanwhile, `attach_exclusive`/`attach_observer` concurrently add to `self._clients` under `self._lock` (lines 362, 387), and `_detach` removes from it under the lock (line 399).

This creates a dict-mutation-during-iteration race. While CPython's GIL provides some protection for single-threaded asyncio, the logical race is real — a client added during broadcast could be missed or cause unexpected iteration behavior.

Consider taking a snapshot before iterating:

```python
def _broadcast_data(self, data: bytes) -> None:
    disconnected = []
    for client_id, buf in list(self._clients.items()):  # snapshot
        try:
            buf.push(data)
        except Exception:
            disconnected.append(client_id)
    for client_id in disconnected:
        self._clients.pop(client_id, None)
        if client_id == self._write_token_holder:
            self._write_token_holder = None
            self._write_token_holder_identity = None
```

---

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/driver.py
```diff
@@ -195,3 +159,30 @@ async def connect(self):
...
         self.logger.info("Disconnected from %s", self.url)
+
+    @export
+    def close(self):
+        transport = self._transport
+        if transport is None:
+            self.logger.debug("close() called but no active connection (no-op)")
+            return
+        self.logger.debug("close() closing transport for %s", self.url)
+        transport.close()
```
mangelajo: **Resource leak: `close()` never calls `super().close()` — fan-out resources not cleaned up**

`PySerial` extends `FanOutStreamMixin`, which provides a `close()` method that shuts down the fan-out reader task, closes all client buffers, and releases the write token. But this override only closes the serial transport without calling `super().close()`.

This means the fan-out's background reader task, scrollback buffer, and client buffers are never cleaned up when the driver is closed.

```python
@export
def close(self):
    transport = self._transport
    if transport is None:
        self.logger.debug("close() called but no active connection (no-op)")
        return
    self.logger.debug("close() closing transport for %s", self.url)
    transport.close()
    super().close()  # <-- add this to clean up fan-out resources
```

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,525 @@
...
+                    if not first_connect:
+                        self._broadcast_system(b"[reconnected]\r\n")
+                    first_connect = False
+                    self._reader_running = True
+                    self._source_ready.set()
+
+                    async for data in source:
+                        if self._shutdown:
+                            break
+                        self._scrollback_append(data)
```
mangelajo, line 280: **Concurrency: `_scrollback_append` called without lock, races with `_scrollback_snapshot`**

`_scrollback_append` (called at line 273) modifies `self._scrollback` and `self._scrollback_bytes` from `_reader_loop`, which does **not** hold `self._lock`. Meanwhile, `_scrollback_snapshot` is called under the lock from `attach_exclusive` (line 361) and `attach_observer` (line 386).

This means an observer attaching concurrently with data arriving could get a partially-consistent scrollback — for example, a chunk could be appended via `_scrollback_append` while `_scrollback_snapshot` is iterating the deque.

Consider either holding the lock during `_scrollback_append` or using a separate lock for scrollback access.

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,525 @@
...
+    @property
+    def extra_attributes(self) -> dict:
+        return {}
+
+    def extra(self, attribute: Any, default: Any = None) -> Any:
+        from anyio import TypedAttributeLookupError
+        attrs = self.extra_attributes
+        if attribute in attrs:
+            return attrs[attribute]()
+        if default is not None:
```
mangelajo: **Minor: `extra(default=None)` raises instead of returning `None`**

If a caller explicitly passes `default=None`, they probably expect `None` back. But the code will raise `TypedAttributeLookupError` because `None` is not truthy. The anyio `extra()` contract uses a sentinel to distinguish "no default provided" from an explicit `None`.

```python
_UNSET = object()

def extra(self, attribute: Any, default: Any = _UNSET) -> Any:
    from anyio import TypedAttributeLookupError
    attrs = self.extra_attributes
    if attribute in attrs:
        return attrs[attribute]()
    if default is not _UNSET:
        return default
    raise TypedAttributeLookupError()
```

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,525 @@
...
+            if not self._shutdown:
+                logger.exception("Reader loop failed unexpectedly")
+
+    async def _ensure_started(self) -> None:
+        """Start the reader loop. Must hold _lock."""
+        if self._started or self._shutdown:
+            return
+        self._started = True
+        self._source_ready = anyio.Event()
+        self._reader_task = asyncio.get_running_loop().create_task(self._run_reader())
```
mangelajo, line 356: **Portability: Uses `asyncio.create_task` directly instead of anyio**

The rest of the file consistently uses anyio primitives (`anyio.Event`, `anyio.Lock`, `anyio.sleep`), but the reader task is launched via `asyncio.get_running_loop().create_task()`. This hard-couples the implementation to asyncio and would break with trio (which anyio supports).

Consider using an `anyio.create_task_group()` to manage the reader task lifecycle, or document that only asyncio is supported.

---

controller/internal/controller/lease_controller.go
```diff
@@ -541,6 +545,52 @@ func (r *LeaseReconciler) attachMatchingPolicies(ctx context.Context, lease *jum
...
+			logger.Info("removing shared client: not found", "client", clientName)
+			continue
+		}
+		if jumpstarterdevv1alpha1.ClientAllowedByPolicy(policies.Items, &exporter, &jclient) {
+			allowed = append(allowed, clientName)
+		} else {
+			logger.Info("removing shared client: denied by policy", "client", clientName)
+		}
+	}
+	lease.Spec.SharedWith = allowed
```
mangelajo: **Design: Reconciler mutates `Spec.SharedWith` — anti-pattern in Kubernetes controllers**

This directly mutates the user's desired state (`Spec`) during reconciliation. In the Kubernetes controller model, `Spec` is the user's declared intent and `Status` reflects the actual state. Controllers typically should not modify `Spec`.

The mutation is persisted because it piggybacks on the later `r.Update(ctx, &lease)` call at line 147 — but this coupling is implicit and fragile. If someone adds an early return between here and line 147, the mutation is silently lost.

Consider either:
1. Tracking denied/removed shared clients in `Status` (e.g., a condition message or `Status.DeniedSharedWith`) instead of mutating `Spec`
2. If mutating `Spec` is intentional (enforcement), making it explicit by persisting within this function and adding a comment explaining the design choice

---

mangelajo: should we add a Status.SharedWith?

mangelajo (commented):
Sharing some comments my local agent generated yesterday which I didn't have time to review until today, I don't agree on one I have noted.

controller/internal/controller/lease_controller.go
bennyz: added

controller/internal/service/controller_service.go
```diff
@@ -832,7 +832,7 @@ func (s *ControllerService) Dial(ctx context.Context, req *pb.DialRequest) (*pb.
 		return nil, err
 	}
 
-	if lease.Spec.ClientRef.Name != client.Name {
+	if !lease.IsAccessibleBy(client.Name) {
```
mangelajo, line 835: **Stale log message and missing gRPC status code after `IsAccessibleBy` change**

The log says `"lease not held by client"` but the check was updated to `IsAccessibleBy` which includes shared clients. Should say something like `"lease not accessible by client"`.

Also, this returns `fmt.Errorf("permission denied")` which maps to gRPC `codes.Unknown`. The Python CLI exception mapper handles specific gRPC codes to show user-friendly messages. Consider `status.Errorf(codes.PermissionDenied, "permission denied")`.

---

controller/internal/service/controller_service.go
```diff
@@ -1143,8 +1143,10 @@ func (s *ControllerService) ReleaseLease(
 		return nil, err
 	}
 
-	if lease.Spec.ClientRef.Name != jclient.Name {
-		return nil, fmt.Errorf("ReleaseLease permission denied")
+	// Release is destructive and ends the lease for everyone, so it is restricted
+	// to the owner. Shared clients keep read/list/dial access via IsAccessibleBy.
+	if !lease.IsOwnedBy(jclient.Name) {
+		return nil, fmt.Errorf("ReleaseLease permission denied: only lease owner can release")
```
mangelajo: **`ReleaseLease` returns `fmt.Errorf` — maps to gRPC `codes.Unknown`**

Same issue as in `DeleteLease` (client_service.go:590) — a bare `fmt.Errorf` produces `codes.Unknown` on the wire. The Python CLI's exception mapper looks for `codes.PERMISSION_DENIED` to display user-friendly messages.

Suggested:
```go
return nil, status.Errorf(codes.PermissionDenied, "ReleaseLease permission denied: only lease owner can release")
```

The same pattern applies to `DeleteLease` at `client_service.go:590`.

---

controller/internal/service/client/v1/client_service.go
```diff
@@ -450,8 +586,8 @@ func (s *ClientService) DeleteLease(ctx context.Context, req *cpb.DeleteLeaseReq
 		return nil, err
 	}
 
-	if jlease.Spec.ClientRef.Name != jclient.Name {
-		return nil, fmt.Errorf("DeleteLease permission denied")
+	if !jlease.IsOwnedBy(jclient.Name) {
+		return nil, fmt.Errorf("DeleteLease permission denied: only lease owner can release")
```
mangelajo: **`DeleteLease` returns `fmt.Errorf` for permission denial — should use `status.Errorf`**

gRPC maps a bare `fmt.Errorf` to `codes.Unknown`. The Python CLI exception mapper handles specific gRPC codes (`PERMISSION_DENIED`, `FAILED_PRECONDITION`, etc.) to show user-friendly messages. With `codes.Unknown` the user sees a raw error dump.

Suggested:
```go
return nil, status.Errorf(codes.PermissionDenied, "DeleteLease permission denied: only lease owner can release")
```

---

controller/internal/controller/lease_controller.go
```diff
@@ -112,6 +112,12 @@ func (r *LeaseReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		return result, err
 	}
 
+	// Recompute the effective shared-with set before persisting status so the
+	// derived Status.SharedWith is saved by the single Status().Update below.
+	if err := r.reconcileSharedWithPolicies(ctx, &lease); err != nil {
+		return result, err
```
mangelajo: **Missing controller-level integration tests for the sharing reconciliation**

The sharing policy reconciliation (`reconcileSharedWithPolicies`) and the `IsAccessibleBy`/`IsOwnedBy` authorization gates touch critical paths (Dial, Release, Delete, Update, GetLease, ListLeases). The unit tests in `lease_helpers_test.go` and `client_service_test.go` cover the helpers well, but there are no controller-level integration tests exercising the full reconcile loop with shared clients.

Consider adding tests in `lease_controller_test.go` that:
1. Create a lease with `Spec.SharedWith`, reconcile, verify `Status.SharedWith` is populated
2. Add a restrictive policy, reconcile, verify a shared client is excluded from `Status.SharedWith`
3. Delete the shared client object, reconcile, verify it is excluded

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,559 @@
...
+                self._dropped_bytes += len(dropped)
+                self._pending_drop_marker = True
+        self._buf.append(data)
+        self._current_bytes += len(data)
+        self._event.set()
+
+    async def pull(self) -> bytes:
+        """Async pull. Blocks until data available."""
+        while not self._buf and not self._closed:
+            self._event = anyio.Event()
```
mangelajo, line 97: **Concurrency: `pull()` event replacement can miss wakeups**

(Reinforcing mangelajo's earlier comment — this is still present.)

Between creating the new `Event` on line 97 and awaiting it on line 98, a concurrent `push()` can `.set()` the **old** event. The new event is never set, so `pull()` hangs until the next `push()` call.

In practice this is safe under CPython's GIL for asyncio (since both sides are in the same event loop and only yield at `await`), but it's fragile and will break if anyio ever runs on trio. A safer pattern:
```python
async def pull(self) -> bytes:
    while not self._buf and not self._closed:
        await self._event.wait()
        if not self._buf and not self._closed:
            self._event = anyio.Event()
    ...
```

Or use `anyio.Condition` which is designed for this producer-consumer pattern.

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,559 @@
...
+                        self._broadcast_system(b"[reconnected]\r\n")
+                    first_connect = False
+                    self._reader_running = True
+                    self._source_ready.set()
+
+                    async for data in source:
+                        if self._shutdown:
+                            break
+                        self._scrollback_append(data)
+                        self._broadcast_data(data)
```
mangelajo, line 281: **Concurrency: `_broadcast_data` and `_scrollback_append` called without lock**

(Reinforcing mangelajo's earlier comments — still present.)

`_broadcast_data` iterates and mutates `self._clients` (pruning closed buffers at line 316) while `attach_exclusive`/`attach_observer` concurrently add to `_clients` under `self._lock`, and `_detach` removes under the lock.

`_scrollback_append` modifies `self._scrollback` while `_scrollback_snapshot` (called under lock from `attach_*`) reads it.

While CPython's GIL prevents memory corruption, the semantic race can cause a newly-attached client to miss its first few bytes (registered after snapshot but before broadcast). Consider at minimum a comment documenting the guarantees, or using a `threading.Lock` for the fast path.

---

python/packages/jumpstarter-driver-pyserial/jumpstarter_driver_pyserial/driver.py
```diff
@@ -195,3 +159,33 @@ async def connect(self):
...
+            self.logger.debug("close() closing transport for %s", self.url)
+            transport.close()
+        # Always chain up so FanOutStreamMixin.close() shuts down the reader
+        # task and client buffers, then the base Driver.close() runs.
+        super().close()
+
+    @export
+    def set_dtr(self, value: bool):
+        """Set the DTR control signal."""
+        s = serial_for_url(self.url, baudrate=self.baudrate)
```
mangelajo: **`set_dtr`/`set_rts` open a second serial connection, potentially conflicting with the fan-out reader**

With the new `FanOutStreamMixin`, the serial port is managed by the fan-out reader loop via `_open_source`. These methods open a **separate** `serial_for_url` connection to the same port URL. Depending on the OS and port type:
- On Linux, opening the same serial device twice can fail with `EBUSY`
- On macOS, concurrent opens may cause DTR/RTS state conflicts
- `loop://` test URLs will open a different virtual port

Consider using `self._transport.serial` when the fan-out has an active connection:
```python
@export
def set_dtr(self, value: bool):
    if self._transport is not None:
        self._transport.serial.dtr = value
    else:
        s = serial_for_url(self.url, baudrate=self.baudrate)
        try:
            s.dtr = value
        finally:
            s.close()
```

---

python/packages/jumpstarter/jumpstarter/streams/fanout.py
```diff
@@ -0,0 +1,559 @@
...
+    @export
+    async def console_status(self) -> dict:
+        """Return current console session status."""
+        return self._get_fanout().status()
+
+    def close(self):
+        if self._fanout is not None:
+            self._fanout.shutdown_sync()
+        if hasattr(super(), "close"):
+            super().close()
```
mangelajo, line 559: **Missing documentation for the new `jmp share` command and lease sharing feature**

This PR adds significant user-facing functionality (`jmp share add/remove/list`, `--observe` flag on serial console, `--lease` for shared lease access in `jmp shell`, `-A` flag on `jmp get leases`), but no documentation updates under `docs/`.

At minimum, the lease sharing workflow shown in the PR description should be documented — it's a great walkthrough. Consider adding:
- A guide or section in the getting-started docs explaining lease sharing
- CLI reference updates for `jmp share`, `jmp shell --lease`, serial `--observe`
- A note in the pyserial driver README about observe mode and `release-console`

---

controller/internal/service/client/v1/client_service.go
```diff
@@ -322,6 +324,30 @@ func (s *ClientService) CreateLease(ctx context.Context, req *cpb.CreateLeaseReq
...
+		var deduped []string
+		for _, name := range jlease.Spec.SharedWith {
+			if name == jclient.Name {
+				return nil, status.Errorf(codes.InvalidArgument, "cannot share lease with the owner")
+			}
+			if slices.Contains(deduped, name) {
+				continue
+			}
+			var sharedClient jumpstarterdevv1alpha1.Client
+			if err := s.Get(ctx, types.NamespacedName{Namespace: namespace, Name: name}, &sharedClient); err != nil {
```
mangelajo: **`CreateLease` skips access policy validation for shared clients**

`applySharedWithChanges` (used in `UpdateLease`) validates each added shared client against the exporter's access policy when the exporter is assigned. `CreateLease` doesn't — it only checks existence and owner-exclusion.

At create time the exporter hasn't been assigned yet, so this is technically correct. However, the `reconcileSharedWithPolicies` reconciler will silently remove policy-violating entries from `Status.SharedWith` on the next reconcile. The user sees a successful `CreateLease` with shared clients, but those clients may never actually get access.

Consider at least documenting this behavior — or adding a note in the `share list` output when a client is in `shared_with` but not `effective_shared_with` (which you already do, nicely).

---

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -273,6 +278,52 @@ async def _monitor_token_expiry(config, lease, cancel_scope, token_state=None) -
...
+
+    Only shared (non-owner) clients are monitored — an owner losing the lease is
+    handled by the natural lease_ended / delete path.
+    """
+    client_name = lease.client_name
+    if not client_name:
+        return
+    while not cancel_scope.cancel_called:
+        try:
+            fresh = await lease.get()
```
mangelajo, line 305: **Consider documenting the shared-access revocation exposure window**

The 15-second poll interval at `_SHARED_ACCESS_POLL_SECONDS` means there's up to a 15-second window where a revoked shared client can still use an active stream. The docstring above nicely explains the soft-guarantee semantics, but the user-facing `jmp share remove` docstring (in share.py) could mention this so admins know removal is not instantaneous.

Also: if the lease is deleted entirely, `lease.get()` would raise a not-found error, which the `except Exception` catches and retries. The lease deletion would eventually be caught by the lease-ended monitor elsewhere, but there could be a compounding delay (poll interval + lease-ended detection).

---

mangelajo: We should document the behavior yes.

mangelajo:
The most important is probably the error codes for gRPC

controller/internal/service/controller_service.go
bennyz: ack

mangelajo (approved):

PR #947, opened 2026-08-03
refactor: extract serve() into composable methods

bennyz:
- serve() → serve / _run_control_plane / _apply_status
 - Transition handlers: _on_lease_acquired, _on_lease_update,
   _on_lease_released, _check_stop_requested
 - Removes the C901 suppression
 - Status transitions are now unit-testable without task-group scaffolding"

Next: #948

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
@@ -17,6 +17,7 @@ from anyio import (
     sleep,
 )
 from anyio.abc import TaskGroup
+from anyio.streams.memory import MemoryObjectReceiveStream, MemoryObjectSendStream
 from google.protobuf import empty_pb2
 from jumpstarter_protocol import (
     jumpstarter_pb2,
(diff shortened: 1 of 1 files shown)


python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -998,111 +998,124 @@ async def process_connections():
...
-        # Buffer status updates to avoid blocking during short processing gaps
         status_tx, status_rx = create_memory_object_stream[jumpstarter_pb2.StatusResponse](max_buffer_size=5)
+        try:
+            await self._run_control_plane(status_tx, status_rx)
+        finally:
+            self._tg = None
+            self._status_drain_active = False
+            clear_log_context()
 
+    async def _run_control_plane(self, status_tx, status_rx):
```
mangelajo: nit: `status_tx` and `status_rx` are missing type annotations, unlike the other new methods which all have proper signatures. Consider:

```python
async def _run_control_plane(
    self,
    status_tx: MemoryObjectSendStream[jumpstarter_pb2.StatusResponse],
    status_rx: MemoryObjectReceiveStream[jumpstarter_pb2.StatusResponse],
) -> None:
```

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -998,111 +998,124 @@ async def process_connections():
...
+                await lease_ctx.after_lease_hook_done.wait()
+            logger.info("afterLease hook completed")
+
+        session_was_created = (
+            self._lease_context is not None and self._lease_context.session is not None
+        )
+        self._lease_context = None
         clear_log_context()
+        if session_was_created:
+            await sleep(0.2)
```
mangelajo, line 1113: The old code had useful context here explaining why this non-obvious delay exists:

```python
# Brief delay to ensure session is fully closed before next lease
# This prevents SSL corruption from overlapping connections
await sleep(0.2)
```

Worth preserving -- without the comment this looks like an arbitrary magic sleep that someone might remove in a future cleanup.

mangelajo (commented):
a couple of nits, I think the sleep(0.2) comment is actually the important one if the types are passing the type checker.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: ack

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: ack

mangelajo (approved):
thank you benny!

PR #948, opened 2026-08-03
refactor: replace _previous_leased with derived LeaseState

bennyz:
Introduce LeaseState enum and _lease_state property derived from
_lease_context, eliminating a class of state-synchronization bugs
where _previous_leased could drift from _lease_context.

Key changes:
- Wrap handle_lease body in try/finally so early returns (stale lease, session setup race) always clean up _lease_context and log context
- Reject overlapping leases in _apply_status instead of silently replacing _lease_context, preventing concurrent handle_lease tasks
- Move before-lease hook spawn into _on_lease_acquired for cleaner ownership of lease startup
- Remove stale _previous_leased from test fixtures

Depends on #947
Next: #949

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
@@ -312,7 +312,9 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     """Stashed status from a lease reassignment, replayed after handle_lease's
     finally clears _lease_context so the new lease can be acquired."""
 
-    _status_replay_tx: MemoryObjectSendStream | None = field(init=False, default=None)
+    _status_replay_tx: MemoryObjectSendStream[jumpstarter_pb2.StatusResponse] | None = field(
+        init=False, default=None
+    )
     """Send side of the status channel, used to replay _pending_lease_status
     back into the status loop after a lease transition."""
     _lease_context: LeaseContext | None = field(init=False, default=None)
@@ -1212,17 +1214,21 @@ class Exporter(AsyncContextManagerMixin, Metadata):
                     clear_log_context()
                     set_log_context(exporter=self.name)
                     logger.debug("Ready for next lease")
-                    pending = self._pending_lease_status
-                    if pending is not None:
-                        self._pending_lease_status = None
-                        if self._status_replay_tx is not None:
-                            try:
-                                await self._status_replay_tx.send(pending)
-                            except (anyio.ClosedResourceError, anyio.EndOfStream):
-                                logger.debug(
-                                    "Status channel closed, skipping replay for %s",
-                                    pending.lease_name,
-                                )
+                # Replay a stashed reassignment status regardless of who cleared
+                # _lease_context. On the reassign-then-leased=False ordering,
+                # _on_lease_released may clear the field before we reach this finally;
+                # gating the replay on ownership above would drop the pending lease.
+                pending = self._pending_lease_status
+                if pending is not None:
+                    self._pending_lease_status = None
+                    if self._status_replay_tx is not None:
+                        try:
+                            await self._status_replay_tx.send(pending)
+                        except (anyio.ClosedResourceError, anyio.EndOfStream):
+                            logger.debug(
+                                "Status channel closed, skipping replay for %s",
+                                pending.lease_name,
+                            )
 
     async def serve(self):
         """Serve the exporter, handling leases until stopped."""
@@ -1322,6 +1328,17 @@ class Exporter(AsyncContextManagerMixin, Metadata):
 
         return self._check_stop_requested() if not current_leased else False
 
+    def _lease_log_context(self, status: jumpstarter_pb2.StatusResponse) -> dict[str, str]:
+        """Build the log context dict for a newly-assigned lease.
+
+        Extracted so tests can verify context propagation without duplicating
+        this logic separately from _on_lease_acquired.
+        """
+        log_ctx: dict[str, str] = {"lease_id": status.lease_name, "exporter": self.name}
+        if status.context:
+            log_ctx.update(status.context)
+        return log_ctx
+
     def _on_lease_acquired(
         self,
         status: jumpstarter_pb2.StatusResponse,
@@ -1335,10 +1352,7 @@ class Exporter(AsyncContextManagerMixin, Metadata):
             before_lease_hook=Event(),
         )
         self._lease_context = lease_scope
-        log_ctx: dict[str, str] = {"lease_id": status.lease_name, "exporter": self.name}
-        if status.context:
-            log_ctx.update(status.context)
-        set_log_context(**log_ctx)
+        set_log_context(**self._lease_log_context(status))
         if self.hook_executor:
             tg.start_soon(
                 self.hook_executor.run_before_lease_hook,
@@ -1360,33 +1374,52 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     async def _on_lease_released(self, previous_state: LeaseState) -> None:
         """Handle not-leased status: signal handle_lease on transition, check exit_on_lease_end.
 
-        Clears _lease_context before waiting for the afterLease hook so that
-        subsequent status ticks see IDLE. handle_lease's outer finally is the
-        fallback if this path never runs (cancellation).
+        Primary cleanup path: signals lease_ended, waits (shielded) for the
+        afterLease hook, then clears _lease_context. _lease_context is kept set
+        for the whole hook so _report_status still reaches the client session
+        (it gates the session update on _lease_context). The status loop is
+        sequential, so no other ticks are processed while we wait. A brief
+        settle delay runs before clearing so the next lease can't grab this
+        slot before handle_lease finishes tearing down the session.
+
+        handle_lease's outer finally is the fallback when this path never
+        runs (e.g. task cancellation, or handle_lease finishing before the
+        controller sends leased=False).
         """
         logger.info("Currently not leased")
 
         if previous_state == LeaseState.LEASED and self._lease_context:
             lease_ctx = self._lease_context
-            self._last_completed_lease = lease_ctx.lease_name
-            self._lease_context = None
             if self.exit_on_lease_end:
                 # Refuse new leases immediately, but keep the runtime up until
                 # afterLease finishes — shutdown SIGTERMs Exec children (QEMU)
                 # that hooks may still be talking to.
                 self._stop_requested = True
-            clear_log_context()
-            set_log_context(exporter=self.name)
 
             logger.info("Lease ended, signaling handle_lease to run afterLease hook")
             lease_ctx.lease_ended.set()
 
+            # Keep _lease_context set while the afterLease hook runs. _report_status
+            # gates its session/client update on _lease_context, so clearing it here
+            # would silently drop status updates the hook emits (they would reach the
+            # controller RPC but never the client). Clear only after the hook is done.
             with CancelScope(shield=True):
                 await lease_ctx.after_lease_hook_done.wait()
             logger.info("afterLease hook completed")
 
-            if self.exit_on_lease_end:
-                await anyio.to_thread.run_sync(shutdown_runtime_sidecar)
+            if lease_ctx.session is not None:
+                # Brief delay to ensure session is fully closed before next lease.
+                # Prevents SSL corruption from overlapping connections.
+                await sleep(0.2)
+
+            self._last_completed_lease = lease_ctx.lease_name
+            self._lease_context = None
+            clear_log_context()
+            set_log_context(exporter=self.name)
+            # exit_on_lease_end shutdown runs once, in serve()'s finally, after
+            # the control-plane loop unwinds. _stop_requested (set above) is
+            # what drives that unwind, so the hook is already done by the time
+            # it fires there — no need to call shutdown_runtime_sidecar here too.
 
     def _check_stop_requested(self) -> bool:
         """Check if stop was requested and initiate shutdown. Returns True to break the status loop."""
--- a/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
@@ -3,6 +3,7 @@ import os
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
+from enum import Enum
 from typing import TYPE_CHECKING, Any, Self
 
 import anyio
@@ -156,6 +157,11 @@ def shutdown_runtime_sidecar(
     return True
 
 
+class LeaseState(Enum):
+    IDLE = "idle"
+    LEASED = "leased"
+
+
 async def _standalone_shutdown_waiter():
     """Wait forever; used so serve_standalone_tcp can be cancelled by stop()."""
     await anyio.sleep_forever()
@@ -284,13 +290,6 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     AFTER_LEASE_HOOK, BEFORE_LEASE_HOOK_FAILED, AFTER_LEASE_HOOK_FAILED.
     """
 
-    _previous_leased: bool = field(init=False, default=False)
-    """Previous lease state used to detect lease state transitions.
-
-    Tracks whether the exporter was leased in the previous status check to
-    determine when to trigger before-lease and after-lease hooks.
-    """
-
     _exit_code: int | None = field(init=False, default=None)
     """Exit code to use when the exporter shuts down.
 
@@ -305,6 +304,17 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     _report_status and __aexit__ skip controller calls when _standalone is True.
     """
 
+    _last_completed_lease: str | None = field(init=False, default=None)
+    """Name of the most recently completed lease, used to filter trailing
+    status ticks after handle_lease's finally has cleaned up."""
+
+    _pending_lease_status: jumpstarter_pb2.StatusResponse | None = field(init=False, default=None)
+    """Stashed status from a lease reassignment, replayed after handle_lease's
+    finally clears _lease_context so the new lease can be acquired."""
+
+    _status_replay_tx: MemoryObjectSendStream | None = field(init=False, default=None)
+    """Send side of the status channel, used to replay _pending_lease_status
+    back into the status loop after a lease transition."""
     _lease_context: LeaseContext | None = field(init=False, default=None)
     """Encapsulates all resources associated with the current lease.
 
@@ -358,6 +368,10 @@ class Exporter(AsyncContextManagerMixin, Metadata):
 
     _status_rpc_event: Event = field(init=False, default_factory=Event)
     """Signals the drain task that a new status update is pending."""
+
+    @property
+    def _lease_state(self) -> LeaseState:
+        return LeaseState.LEASED if self._lease_context is not None else LeaseState.IDLE
 
     def stop(self, wait_for_lease_exit=False, should_unregister=False, exit_code: int | None = None):
         """Signal the exporter to stop.
@@ -1024,7 +1038,7 @@ class Exporter(AsyncContextManagerMixin, Metadata):
         lease_scope.after_lease_hook_done.set()
         return True
 
-    async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseContext) -> None:
+    async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseContext) -> None:  # noqa: C901
         """Handle all incoming client connections for a lease.
 
         This method orchestrates the complete lifecycle of managing connections during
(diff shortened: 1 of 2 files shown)


python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -877,127 +880,133 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
+        finally:
+            if self._lease_context is lease_scope:
+                session_was_created = lease_scope.session is not None
+                self._lease_context = None
+                clear_log_context()
+                if session_was_created:
+                    # Brief delay to ensure session is fully closed before next lease.
+                    # Prevents SSL corruption from overlapping connections.
+                    await sleep(0.2)
+                logger.debug("Ready for next lease")
```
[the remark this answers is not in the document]
bennyz: The status loop is sequential: async for status in status_rx → await _apply_status(status). While _on_lease_released is waiting (shielded) for after_lease_hook_done, no other ticks are processed. By the time
  the next tick arrives, handle_lease already cleared _lease_context and the state is consistent.

  A "trailing leased=true tick for the same lease" would require the controller to send leased=true, name=A after it already sent leased=false. That's a controller bug, not something the exporter should
  silently swallow.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1039,27 +1048,28 @@ async def _apply_status(
...
-            self._on_lease_acquired(status, tg)
-
         if current_leased:
+            if previous_state == LeaseState.IDLE and status.lease_name != "":
+                self._on_lease_acquired(status, tg)
+            elif (
+                previous_state == LeaseState.LEASED
+                and self._lease_context
+                and self._lease_context.lease_name != status.lease_name
+            ):
```
mangelajo, line 1280: Nice addition — prevents concurrent `handle_lease` tasks for different leases.

Worth noting there's a small timing window here: after `_on_lease_released` returns (having waited for `after_lease_hook_done`), `handle_lease`'s outer finally may not have run yet (session teardown still in progress). During that gap `_lease_context` is still set, so a new lease arriving here would be rejected. It gets picked up on the next status tick once the finally clears `_lease_context`.

A brief comment would help future readers understand this is intentional:

```python
# Note: this may briefly reject new leases while handle_lease
# finishes session teardown; the next status tick will acquire them.
```

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1113,6 +1115,315 @@ async def fake_session_for_lease():
...
+
+        status = MagicMock()
+        status.leased = False
+        status.lease_name = ""
+        status.client_name = ""
+        status.context = {}
+
+        async with create_task_group() as tg:
+            await exporter._apply_status(status, tg)
+            tg.cancel_scope.cancel()
```
mangelajo, line 1277: This test runs the LEASED→IDLE transition but has no assertions — it only checks that it doesn't raise. Consider verifying the side effects, e.g.:

```python
assert lease_ctx.lease_ended.is_set()
```

This would confirm the signal actually reached `handle_lease` (the core purpose of the transition).

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1089,11 +1115,11 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
                 set_log_context(client=status.client_name)
         logger.info("Currently leased by %s under %s", status.client_name, status.lease_name)
 
-    async def _on_lease_released(self, previous_leased: bool) -> None:
+    async def _on_lease_released(self, previous_state: LeaseState) -> None:
         """Handle not-leased status: signal handle_lease on transition, clean up context."""
```
mangelajo: nit: The docstring still says "clean up context" but this method no longer clears `_lease_context`, `clear_log_context()`, or does the `sleep(0.2)` — that responsibility moved to `handle_lease`'s finally block. Consider updating to match the reduced scope:

```python
"""Handle not-leased status: signal handle_lease on transition, check exit_on_lease_end."""
```

mangelajo (commented):
A few notes, nothing major.

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
bennyz: ack

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: ack

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -877,127 +881,134 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
+            if self._lease_context is lease_scope:
+                session_was_created = lease_scope.session is not None
+                if session_was_created:
+                    # Brief delay to ensure session is fully closed before next lease.
+                    # Prevents SSL corruption from overlapping connections.
+                    await sleep(0.2)
+                self._last_completed_lease = lease_scope.lease_name
+                self._lease_context = None
+                clear_log_context()
+                logger.debug("Ready for next lease")
```
raballew, line 1197: `_lease_state` derives from `_lease_context`, but this finally block clears `_lease_context = None` concurrently with `_apply_status`. When `handle_lease` completes before the controller's `leased=False` tick arrives, `_apply_status` sees `previous_state == IDLE` and skips the `exit_on_lease_end` check in `_on_lease_released`. The exporter keeps running instead of shutting down.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -877,127 +883,137 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
+                pending = self._pending_lease_status
+                if pending is not None:
+                    self._pending_lease_status = None
+                    if self._status_replay_tx is not None:
+                        await self._status_replay_tx.send(pending)
```
mangelajo, line 1202: If the task group is being cancelled during shutdown, `_status_replay_tx` may already be closed. `send()` on a closed stream would raise `ClosedResourceError`. Replaying a stashed status during shutdown is pointless anyway, so this should be guarded:

```python
pending = self._pending_lease_status
if pending is not None:
    self._pending_lease_status = None
    if self._status_replay_tx is not None:
        try:
            await self._status_replay_tx.send(pending)
        except (ClosedResourceError, EndOfStream):
            logger.debug("Status channel closed, skipping replay for %s", pending.lease_name)
```

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -198,6 +197,9 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     _report_status and __aexit__ skip controller calls when _standalone is True.
     """
 
+    _last_completed_lease: str | None = field(init=False, default=None)
+    _pending_lease_status: jumpstarter_pb2.StatusResponse | None = field(init=False, default=None)
+    _status_replay_tx: MemoryObjectSendStream | None = field(init=False, default=None)
```
mangelajo, line 315: nit: I would document those attributes to make them more evident (also other fields have docstrings), also docstrings need maintenance so it's a double side sword... since a nit, I leave it up to your criteria :D

```python
_last_completed_lease: str | None = field(init=False, default=None)
"""Name of the most recently completed lease, used to filter trailing status ticks
after handle_lease's finally has cleaned up. Cleared on the next leased=False tick."""

_pending_lease_status: jumpstarter_pb2.StatusResponse | None = field(init=False, default=None)
"""Stashed status for a reassigned lease. When the controller reassigns the exporter
to a new lease while the old one is still tearing down, the new status is held here
and replayed by handle_lease's finally block after cleanup completes."""

_status_replay_tx: MemoryObjectSendStream | None = field(init=False, default=None)
"""Reference to the status channel TX end, set by _run_control_plane. Used by
handle_lease's finally to replay stashed reassignment statuses into the status loop."""
```

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -847,7 +853,7 @@ async def _skip_stale_lease(self, lease_name: str, lease_scope: LeaseContext, co
         lease_scope.after_lease_hook_done.set()
         return True
```
mangelajo, line 1040: oh no the noqa is back :'(

mangelajo (changes requested):
This one here: https://github.com/jumpstarter-dev/jumpstarter/pull/948/changes#diff-4566174bc7bf5eb59681380166c9f49879f3e6392e22ef0508f9eae6e0eb8640R1016

mangelajo:
There is a failure in E2E related to lease status: https://github.com/jumpstarter-dev/jumpstarter/actions/runs/30887610848/job/91922876687?pr=948#step:10:1773

I am trying to figure out if it could be related.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: temporarily, needed the lint to pass in this PR 😅

raballew (dismissed):

mangelajo:
This is what claude determined, when I saw it was a status transition I thought it could be related, I am looking at the code now.

## E2E Failure Analysis: `test-exporter-oidc` stuck in `LeaseReady`

**Failed test**: `paginated exporter listing returns all exporters` — timed out after 300s waiting for `test-exporter-oidc` to reach `Available` status. It was stuck in `LeaseReady`.

### What the logs show

The exporter gets a new lease at `07:41:36`, creates a session, reaches `LEASE_READY`, and then stops producing any further log output. No `"Currently not leased"`, no `"Lease ended"`, no `Available` transition.

### Why this is related to the PR

In the old code, `_on_lease_released` was the single owner of `_lease_context` cleanup — it would signal `lease_ended`, wait for `after_lease_hook_done`, clear `_lease_context`, and move on. This meant repeated `leased=False` ticks were harmless: after the first one cleared `_lease_context`, subsequent ones would see `_lease_context is None` and skip the signaling/wait path.

In the new code, `_on_lease_released` no longer clears `_lease_context` — that responsibility moved to `handle_lease`'s outer finally block. This creates a window where multiple `leased=False` ticks arrive while `_lease_context` is still set (pointing at the old lease scope). The logs show exactly this pattern: three consecutive `"Lease ended, signaling handle_lease"` entries for the same old lease right before the new lease starts.

The likely deadlock scenario: a `leased=False` tick arrives and `_on_lease_released` calls `await lease_ctx.after_lease_hook_done.wait()` (shielded) on a `_lease_context` that now belongs to the **new** lease (which hasn't run `_cleanup_after_lease` yet because it just started). The status loop blocks waiting for an event that won't be set until the new lease ends, so the exporter never processes any more status updates and stays stuck in `LeaseReady`.

### Suggested fix

`_on_lease_released` should either:
- Snapshot `_lease_context` atomically and guard against waiting on a context it didn't originate, or
- Continue to clear `_lease_context` itself (as the old code did) so repeated ticks are idempotent, and let `handle_lease`'s finally be the fallback rather than the primary cleanup path.

bkhizgiy (approved):
I don't see any issues, good from my side.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1285,36 +1334,33 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
...
+                self._stop_requested = True
+            clear_log_context()
+            set_log_context(exporter=self.name)
+
             logger.info("Lease ended, signaling handle_lease to run afterLease hook")
             lease_ctx.lease_ended.set()
 
             with CancelScope(shield=True):
                 await lease_ctx.after_lease_hook_done.wait()
             logger.info("afterLease hook completed")
```
raballew, line 1362: Is it intentional that we shutdown the sidecar before the hook completed?

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1113,6 +1116,390 @@ async def fake_session_for_lease():
...
+
+                replayed = await status_rx.receive()
+                await exporter._apply_status(replayed, tg)
+
+            tg.cancel_scope.cancel()
+
+        assert acquired_leases == ["lease-B"]
+        assert exporter._lease_context is not None
+        assert exporter._lease_context.lease_name == "lease-B"
+        assert exporter._last_completed_lease == "lease-A"
```
raballew, line 1385: Imho, this test manually assigns `exporter._last_completed_lease`, `exporter._lease_context = None`, and `exporter._pending_lease_status = None` and replays the pending status. These are exactly the operations that `handle_lease`'s finally block performs. Removing the replay logic from production code would not fail this test.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1024,7 +1038,7 @@ async def _skip_stale_lease(self, lease_name: str, lease_scope: LeaseContext, co
         lease_scope.after_lease_hook_done.set()
         return True
 
-    async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseContext) -> None:
+    async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseContext) -> None:  # noqa: C901
```
raballew, line 1041: Why `# noqa: C901`? I mean you could refactor the function if its too big

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1285,36 +1334,33 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
...
-    async def _on_lease_released(self, previous_leased: bool) -> None:
-        """Handle not-leased status: signal handle_lease on transition, clean up context."""
+    async def _on_lease_released(self, previous_state: LeaseState) -> None:
+        """Handle not-leased status: signal handle_lease on transition, check exit_on_lease_end.
+
+        Clears _lease_context before waiting for the afterLease hook so that
+        subsequent status ticks see IDLE state and new leases can be acquired
+        without waiting for handle_lease's outer finally.  _finalize_lease_context
+        becomes a fallback for abnormal exits (cancellation).
+        """
```
raballew, line 1344: Where is this `_finalize_lease_context`?

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -156,6 +157,11 @@ def shutdown_runtime_sidecar(
     return True
 
 
+class LeaseState(Enum):
+    IDLE = "idle"
+    LEASED = "leased"
```
raballew, line 162: `LeaseState` is defined here but is absent from `jumpstarter/exporter/__init__.py`.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1054,127 +1068,138 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
+                    await listen_tx.aclose()
+                    # Run afterLease hook before closing the session
+                    # This ensures the socket is still available for driver calls within the hook
+                    # Shield from cancellation so the hook can complete even during shutdown
+                    await self._cleanup_after_lease(lease_scope)
+        finally:
+            if self._lease_context is lease_scope:
+                session_was_created = lease_scope.session is not None
+                if session_was_created:
+                    await sleep(0.2)
```
raballew, line 1192: why are we sleeping here?

bennyz: claude keeps removing the comment for some reason... i'll restore it

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: it's removed in the next PR in the stack

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: it is currently internal

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: introduced later, dropped

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
bennyz: removed, will be done in a later PR

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: fixed

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1054,127 +1068,140 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
+                self._last_completed_lease = lease_scope.lease_name
+                self._lease_context = None
+                clear_log_context()
+                set_log_context(exporter=self.name)
+                logger.debug("Ready for next lease")
+                pending = self._pending_lease_status
+                if pending is not None:
+                    self._pending_lease_status = None
+                    if self._status_replay_tx is not None:
+                        await self._status_replay_tx.send(pending)
```
mangelajo: The `_status_replay_tx.send(pending)` call is unguarded against `ClosedResourceError`. During shutdown, the task group hosting `_run_control_plane` may cancel, closing the `status_tx` stream before `handle_lease`'s finally block runs. Sending to a closed stream raises `ClosedResourceError`, which would propagate out of the finally block and mask the original shutdown reason.

@mangelajo already flagged this in an earlier review comment (still unresolved). Consider wrapping in a try/except:

```python
pending = self._pending_lease_status
if pending is not None:
    self._pending_lease_status = None
    if self._status_replay_tx is not None:
        try:
            await self._status_replay_tx.send(pending)
        except (ClosedResourceError, EndOfStream):
            logger.debug("Status channel closed, skipping replay for %s", pending.lease_name)
```

---

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1054,127 +1068,140 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
```
mangelajo, line 1232: Both `_on_lease_released` (line 1350) and `handle_lease`'s finally (line 1195) set `_last_completed_lease` and clear `_lease_context`. In the normal flow, `_on_lease_released` runs first (triggered by the `leased=False` status tick), then `handle_lease`'s finally runs when its task completes. This means the cleanup runs twice for the same lease in the happy path.

The redundancy is intentional (the finally is a fallback for cancellation paths), but it makes the state ownership harder to reason about — two code paths independently setting `_last_completed_lease` and clearing `_lease_context` with slightly different side effects (the finally also does the `sleep(0.2)` and replays `_pending_lease_status`).

Consider adding a brief comment at one of the two sites making the "primary vs fallback" ownership explicit, e.g.:

```python
# Primary cleanup path — _on_lease_released handles the normal case.
# handle_lease's outer finally is the fallback for cancellation/early-exit.
```

This would help future readers understand why there are two cleanup sites.

---

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1285,35 +1336,36 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
...
-            # Prevents SSL corruption from overlapping connections.
-            await sleep(0.2)
-        logger.debug("Ready for next lease")
-
-        if self.exit_on_lease_end and previous_leased:
-            logger.info("Exporter configured to exit after lease, shutting down")
-            await anyio.to_thread.run_sync(shutdown_runtime_sidecar)
-            self._stop_requested = True
+            if self.exit_on_lease_end:
+                await anyio.to_thread.run_sync(shutdown_runtime_sidecar)
```
mangelajo: Race concern on `exit_on_lease_end`: @raballew raised an important point (still unresolved) — when `handle_lease`'s finally block clears `_lease_context` before the controller's `leased=False` tick arrives, `_apply_status` sees `previous_state == IDLE` and skips the `exit_on_lease_end` check in `_on_lease_released`. The exporter keeps running instead of shutting down.

The `handle_lease` finally block does not check `exit_on_lease_end`, so if it's the path that clears `_lease_context`, the `exit_on_lease_end` behavior is lost.

One fix: have the finally block also check `self.exit_on_lease_end` and set `self._stop_requested = True` when applicable.

---

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1054,127 +1068,140 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
+                session.update_status(lease_scope.current_status, lease_scope.status_message)
+                logger.debug("Session sockets: main=%s, hook=%s", main_path, hook_path)
+
+                # Check if lease ended during session creation - serve() often
+                # processes the buffered leased=False while session_for_lease is
+                # setting up sockets and gRPC servers.  Bailing here avoids the
+                # Listen stream, conn_tg, and _cleanup_after_lease overhead.
+                # The session context manager handles teardown on return.
+                if await self._skip_stale_lease(lease_name, lease_scope, "during session setup"):
+                    return
```
mangelajo: `listen_tx` and `listen_rx` are created at line 1083, but the `_skip_stale_lease` check at line 1104 can return out of the `session_for_lease` context manager before the inner `try/finally` that closes `listen_tx`. This leaks both stream ends (anyio will emit resource warnings for unclosed memory object streams).

The CodeRabbit review already flagged this (resolved in a prior commit per thread), but the current code still has this issue. Consider wrapping the streams:

```python
listen_tx, listen_rx = create_memory_object_stream[...](max_buffer_size=10)
async with listen_tx, listen_rx:
    async with self.session_for_lease() as (session, main_path, hook_path):
        ...
```

The explicit `await listen_tx.aclose()` in the inner finally is a no-op on already-closed streams, so it remains safe.

---

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1285,35 +1336,36 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
...
 
-    async def _on_lease_released(self, previous_leased: bool) -> None:
-        """Handle not-leased status: signal handle_lease on transition, clean up context."""
+    async def _on_lease_released(self, previous_state: LeaseState) -> None:
+        """Handle not-leased status: signal handle_lease on transition, check exit_on_lease_end.
+
+        Clears _lease_context before waiting for the afterLease hook so that
+        subsequent status ticks see IDLE. handle_lease's outer finally is the
+        fallback if this path never runs (cancellation).
+        """
```
mangelajo, line 1388: The docstring says "Clears `_lease_context` before waiting for the afterLease hook so that subsequent status ticks see IDLE." — but `_run_control_plane` processes the status stream sequentially (`async for status in status_rx: await _apply_status(status, tg)`). While `_on_lease_released` is awaiting `after_lease_hook_done`, no other status ticks are processed.

The clearing-before-waiting pattern matters only for the reassignment path where `_pending_lease_status` replay happens, not for "subsequent status ticks." The docstring could be more precise about when this early clear actually matters (i.e., only for the `handle_lease` finally-block replay path, not for concurrent status processing).

Nit, but the current wording might mislead readers into thinking status processing is concurrent.

---

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1113,6 +1116,342 @@ async def fake_session_for_lease():
...
+        exporter.hook_executor = hook_executor
+        exporter.labels = {"jumpstarter.dev/name": "test-exporter"}
+        exporter._last_completed_lease = None
+        exporter._pending_lease_status = None
+        exporter._status_replay_tx = None
+        exporter._report_status = AsyncMock()
+        exporter._request_lease_release = AsyncMock()
+        return exporter
+
+    async def test_reassignment_signals_old_lease_ended(self):
```
mangelajo, line 1158: `TestApplyStatus._make_idle_exporter` and the module-level `make_exporter` both construct exporters with manually-set fields, and they diverge on which fields they initialize. For example, `_make_idle_exporter` sets `_pending_lease_status`, `_status_replay_tx`, `_started` — but `make_exporter` doesn't set them, which means tests using `make_exporter` may have uninitialized fields if `handle_lease`'s finally touches them.

Consider unifying these helpers or at least having `make_exporter` call `_make_idle_exporter` and then override the lease context, so that all new fields are always initialized and you don't have to remember to update both helpers when adding new fields.

---

mangelajo (commented):
## Review Summary

Good refactor overall — replacing `_previous_leased` with a derived `_lease_state` property eliminates a class of state-synchronization bugs and makes the state machine easier to reason about. The `LeaseState` enum, trailing-tick guard (`_last_completed_lease`), and lease reassignment handling (`_pending_lease_status` + replay) are well-designed additions. Test coverage is thorough.

### Key Findings

**Potential bugs / correctness issues:**

1. **`ClosedResourceError` on status replay** — `_status_replay_tx.send()` in `handle_lease`'s finally block is unguarded. During shutdown the stream may already be closed. (@mangelajo flagged this in an earlier review, still unresolved.)

2. **`exit_on_lease_end` race** — When `handle_lease`'s finally clears `_lease_context` before the controller's `leased=False` tick arrives, `_apply_status` sees `IDLE` and skips `_on_lease_released`'s `exit_on_lease_end` check. The exporter stays up. (@raballew flagged this, still unresolved.)

3. **Listen stream leak on stale-lease early return** — `listen_tx`/`listen_rx` are created before the `session_for_lease` context, but the stale-lease check inside can `return` before the inner `try/finally` that closes them.

**Code quality / maintainability:**

4. **Dual cleanup ownership** — Both `_on_lease_released` and `handle_lease`'s finally set `_last_completed_lease` and clear `_lease_context`. The relationship (primary path vs. fallback) should be documented.

5. **Missing blank line** between `_status_replay_tx` and `_lease_context` field declarations.

6. **Slightly misleading docstring** on `_on_lease_released` — references "subsequent status ticks" seeing IDLE, but the status loop is sequential.

7. **Test helper divergence** — `_make_idle_exporter` and `make_exporter` initialize different field sets, risking uninitialized fields.

### Files reviewed in depth
- `python/packages/jumpstarter/jumpstarter/exporter/exporter.py` — all changed sections
- `python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py` — all new test classes

---

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
mangelajo: Not sure if wrapping the streams, since that will add more indent... perhaps just close them here?

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1371,6 +1856,96 @@ def test_nonzero_exit_returns_false(self, monkeypatch, tmp_path):
...
+                tg.start_soon(finish_hook)
+                await exporter._apply_status(status, tg)
+                tg.cancel_scope.cancel()
+
+        assert hook_done_at_shutdown == [True]
+        assert exporter._stop_requested is True
+
+
 class TestContextPropagation:
     """Tests for spec.context propagation from StatusResponse to log context."""
```
raballew, line 1976: `TestContextPropagation` re-implements production logic inline instead of calling production methods.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -305,6 +304,17 @@ class Exporter(AsyncContextManagerMixin, Metadata):
...
 
+    _last_completed_lease: str | None = field(init=False, default=None)
+    """Name of the most recently completed lease, used to filter trailing
+    status ticks after handle_lease's finally has cleaned up."""
+
+    _pending_lease_status: jumpstarter_pb2.StatusResponse | None = field(init=False, default=None)
+    """Stashed status from a lease reassignment, replayed after handle_lease's
+    finally clears _lease_context so the new lease can be acquired."""
+
+    _status_replay_tx: MemoryObjectSendStream | None = field(init=False, default=None)
```
raballew: Change the declaration to: `_status_replay_tx: MemoryObjectSendStream[jumpstarter_pb2.StatusResponse] | None = field(init=False, default=None)`

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1113,32 +1149,481 @@ async def fake_session_for_lease():
...
+
+        status = MagicMock()
+        status.leased = True
+        status.lease_name = "hooked-lease"
+        status.client_name = "ci-bot"
+        status.context = {"env": "staging"}
+
+        async with create_task_group() as tg:
+            await exporter._apply_status(status, tg)
+            await anyio.sleep(0.05)
```
raballew: `test_idle_to_leased_spawns_handle_lease` and `test_idle_to_leased_with_hook_executor` rely on fixed 50 ms sleeps for synchronization. Replace the sleep in each test with an `anyio.Event` that the fake coroutine sets on entry; await that event before cancelling the task group.

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1113,32 +1149,481 @@ async def fake_session_for_lease():
...
+        exporter._skip_stale_lease = AsyncMock(return_value=False)
+        exporter._cleanup_after_lease = AsyncMock()
+
+        async with create_task_group() as tg:
+            tg.start_soon(exporter.handle_lease, "conn-lease", tg, lease_ctx)
+            with fail_after(5):
+                await conn_arrived.wait()
+            lease_ctx.lease_ended.set()
+            await anyio.sleep(0.05)
+            tg.cancel_scope.cancel()
```
raballew, line 1434: Mock `_cleanup_after_lease` to set an `anyio.Event` on completion; replace `sleep(0.05)` with `await cleanup_done.wait()` wrapped in `fail_after`.

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
```diff
@@ -1113,32 +1149,481 @@ async def fake_session_for_lease():
...
+            await tx.aclose()
+
+        exporter._retry_stream = fake_retry_stream
+        exporter._listen_stream_factory = MagicMock(return_value=MagicMock())
+
+        async with create_task_group() as tg:
+            tg.start_soon(exporter.handle_lease, "fallback-lease", tg, lease_ctx)
+            await anyio.sleep(0.1)
+            lease_ctx.lease_ended.set()
+            await anyio.sleep(0.1)
```
raballew: Couldnt this also be two events instead?

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1285,35 +1361,47 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
...
-            logger.info("Exporter configured to exit after lease, shutting down")
-            await anyio.to_thread.run_sync(shutdown_runtime_sidecar)
-            self._stop_requested = True
+            self._last_completed_lease = lease_ctx.lease_name
+            self._lease_context = None
+            clear_log_context()
+            set_log_context(exporter=self.name)
+
+            if self.exit_on_lease_end:
+                await anyio.to_thread.run_sync(shutdown_runtime_sidecar)
```
raballew: `shutdown_runtime_sidecar` is called twice in the normal `exit_on_lease_end=True` flow.
Here and `serve()`'s `finally` block calls it again at line 1246 when `exit_on_lease_end=True`. Call `shutdown_runtime_sidecar` in one authoritative location: either remove the call from `_on_lease_released` and rely solely on `serve()`'s `finally`, or add a guard flag to prevent double execution.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1285,35 +1361,47 @@ def _on_lease_update(self, status: jumpstarter_pb2.StatusResponse) -> None:
...
-        logger.debug("Ready for next lease")
-
-        if self.exit_on_lease_end and previous_leased:
-            logger.info("Exporter configured to exit after lease, shutting down")
-            await anyio.to_thread.run_sync(shutdown_runtime_sidecar)
-            self._stop_requested = True
+            self._last_completed_lease = lease_ctx.lease_name
+            self._lease_context = None
+            clear_log_context()
+            set_log_context(exporter=self.name)
```
mangelajo, line 1418: The old `_on_lease_released` had a `sleep(0.2)` settle delay after clearing `_lease_context`, specifically to prevent SSL corruption from overlapping sessions. That delay is gone from this primary cleanup path.

The delay still exists in the fallback path (`handle_lease`'s finally, line 1207), but in the normal flow `_on_lease_released` runs first and clears `_lease_context` — so the fallback's `self._lease_context is lease_scope` check is `False` and the delay is never applied.

At the point `_on_lease_released` clears `_lease_context`, `handle_lease` is still unwinding from `session_for_lease`'s context manager (the old session isn't fully closed yet). A new lease could be acquired on the next status tick, creating overlapping sessions — exactly what the delay was preventing.

Suggest restoring the settle delay here (before clearing `_lease_context`, matching the fallback path ordering), and consider extracting the shared cleanup logic (delay + clear context + log context reset) into a small helper so the intent is documented in one place and both paths stay in sync.

```python
# e.g.
async def _finalize_lease_cleanup(self, lease_ctx: LeaseContext) -> None:
    """Clear lease ownership after session teardown.

    Includes a brief settle delay when a session was created to ensure
    the gRPC/SSL transport is fully closed before the next lease can
    create a new session — prevents SSL corruption from overlapping
    connections.
    """
    if lease_ctx.session is not None:
        await sleep(0.2)
    self._last_completed_lease = lease_ctx.lease_name
    self._lease_context = None
    if self.exit_on_lease_end:
        self._stop_requested = True
    clear_log_context()
    set_log_context(exporter=self.name)
```

Please fix before merge.

---

mangelajo (changes requested):

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
bennyz: done

python/packages/jumpstarter/jumpstarter/exporter/exporter_test.py
bennyz: done

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: restored

mangelajo (approved):

PR #949, opened 2026-08-03
feat: replace count-based stream retry with grace-window

bennyz:
Replace the 5-attempt count-based retry (2.5s total budget) with a
300-second wall-clock grace window and exponential backoff with jitter.
This gives the exporter enough runway to survive controller restarts
that take 30-60s.

Key changes:
 - Add _is_retryable() to classify errors: UNAVAILABLE/INTERNAL/UNKNOWN
   are retried, PERMISSION_DENIED/NOT_FOUND are terminal
 - Extract _stream_once() for single connection attempts with inline
   window/backoff reset when data flows
 - Terminal errors invoke on_terminal callback instead of exhausting
   retries — Listen terminal errors signal lease_ended, Status terminal
   errors cancel the control-plane task group
 - Add _fatal_stream_error field so serve() can log why it stopped

Depends on #948
Next: #950

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
@@ -1,5 +1,8 @@
+import functools
 import logging
 import os
+import random
+import time
 from collections.abc import AsyncGenerator, Awaitable, Callable
 from contextlib import asynccontextmanager
 from dataclasses import dataclass, field
@@ -155,6 +158,76 @@ def shutdown_runtime_sidecar(
 
     logger.info("Runtime sidecar shutdown acknowledged")
     return True
+
+
+# Streams retry the unary transient codes plus INTERNAL/UNKNOWN: the Go
+# controller can surface those during rolling updates (e.g. INTERNAL when
+# the context is cancelled mid-send).
+_RETRYABLE_STREAM_CODES = _TRANSIENT_GRPC_CODES | frozenset({
+    grpc.StatusCode.INTERNAL,
+    grpc.StatusCode.UNKNOWN,
+})
+
+
+class _StreamClosedImmediately(Exception):
+    """Stream connected and returned zero items; treated as retryable degradation."""
+
+
+def _is_retryable(e: Exception) -> bool:
+    """Classify whether a streaming error warrants retry or is terminal."""
+    if isinstance(e, _StreamClosedImmediately):
+        return True
+    if isinstance(e, grpc.aio.AioRpcError):
+        return e.code() in _RETRYABLE_STREAM_CODES
+    if isinstance(e, ConnectionError):
+        return True
+    return False
+
+
+@dataclass
+class _GraceWindow:
+    """Wall-clock degradation window for stream retries."""
+
+    period: float
+    since: float | None = field(default=None, init=False)
+
+    def mark_failure(self) -> float:
+        now = time.monotonic()
+        if self.since is None:
+            self.since = now
+        return now - self.since
+
+    def elapsed(self) -> float:
+        if self.since is None:
+            return 0.0
+        return time.monotonic() - self.since
+
+    def expired(self) -> bool:
+        return self.since is not None and time.monotonic() - self.since > self.period
+
+    def reset(self):
+        self.since = None
+
+
+@dataclass
+class _Backoff:
+    """Exponential backoff with jitter for stream retries."""
+
+    max_delay: float
+    delay: float = field(default=0.5, init=False)
+    _initial: float = field(default=0.5, init=False)
+
+    def __post_init__(self):
+        self._initial = min(0.5, self.max_delay)
+        self.delay = self._initial
+
+    def reset(self):
+        self.delay = self._initial
+
+    async def wait(self):
+        jitter = random.uniform(0, self.delay * 0.3)
+        await sleep(self.delay + jitter)
+        self.delay = min(self.delay * 2, self.max_delay)
 
 
 class LeaseState(Enum):
@@ -369,6 +442,9 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     _status_rpc_event: Event = field(init=False, default_factory=Event)
     """Signals the drain task that a new status update is pending."""
 
+    _fatal_stream_error: tuple[str, Exception] | None = field(init=False, default=None)
+    """Set by _cancel_with_fatal_error when a stream hits a terminal error."""
+
     @property
     def _lease_state(self) -> LeaseState:
         return LeaseState.LEASED if self._lease_context is not None else LeaseState.IDLE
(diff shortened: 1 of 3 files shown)


python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -56,6 +59,77 @@
...
+
+
+def _is_retryable(e: Exception) -> bool:
+    """Classify whether a streaming error warrants retry or is terminal."""
+    if isinstance(e, _StreamClosedImmediately):
+        return True
+    if isinstance(e, grpc.aio.AioRpcError):
+        return e.code() in _RETRYABLE_STREAM_CODES
+    if isinstance(e, (ConnectionError, OSError)):
+        return True
```
raballew: Consider narrowing to `isinstance(e, ConnectionError)` which covers `ConnectionRefusedError`, `ConnectionResetError`, and `ConnectionAbortedError`.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -298,55 +375,146 @@ async def _controller_stub(self) -> AsyncGenerator[jumpstarter_pb2_grpc.Controll
         finally:
             await channel.close()
 
+    def _cancel_with_fatal_error(self, stream_name: str, error: Exception):
+        self._fatal_stream_error = (stream_name, error)
+        if self._tg is not None:
+            self._tg.cancel_scope.cancel()
+
+    def _on_status_exhausted(self, stream_name: str, error: Exception):
+        pass
```
raballew, line 510: An explicit `retry_indefinitely: bool = False` parameter on `_retry_stream` would be clearer unless the callback pattern is kept for future extensibility.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -938,14 +1106,16 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
...
-                            self._listen_stream_factory(lease_name),
-                            listen_tx,
-                        )
+                            stream_name="Listen",
+                            stream_factory=self._listen_stream_factory(lease_name),
+                            send_tx=listen_tx,
+                            on_terminal=lambda name, err: (
+                                logger.info("Listen stream ended (%s: %s), signaling lease end", name, err),
+                                lease_scope.lease_ended.set(),
+                            ),
```
raballew, line 1305: The critical lifecycle signal (`lease_ended.set()`) runs second in the tuple. If `logger.info()` raised, `lease_ended.set()` would never execute.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1022,8 +1192,16 @@ async def serve(self):
...
+            if self._fatal_stream_error:
+                name, err = self._fatal_stream_error
+                logger.warning(
+                    "Control plane down (%s: %s)",
+                    name,
+                    err,
+                )
         finally:
             self._tg = None
+            self._fatal_stream_error = None
```
raballew, line 1401: If the intent is to let callers see the shutdown reason, consider clearing the field at the start of `serve()` instead of the end.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -298,55 +375,146 @@ async def _controller_stub(self) -> AsyncGenerator[jumpstarter_pb2_grpc.Controll
...
+                    logger.warning(
+                        "%s stream degraded, retrying in %.1fs for %.0fs: %s",
                         stream_name,
-                        retry_delay,
-                        retries_left,
-                        e,
+                        backoff.delay,
+                        grace_period,
+                        failure,
                     )
```
raballew, line 634: When the grace window expires and `on_exhausted` is set, the code resets the window, backoff, and `warned = False`, then falls through into the degradation logging block. Doesnt this produce two warnings for the same event?

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -56,6 +59,77 @@
...
 
+# Superset of _TRANSIENT_GRPC_CODES: streams also retry INTERNAL/UNKNOWN because
+# the Go controller can surface these transiently during rolling updates (e.g.,
+# the gRPC server returns INTERNAL when the context is cancelled mid-send).
+_RETRYABLE_STREAM_CODES = frozenset({
+    grpc.StatusCode.UNAVAILABLE,
+    grpc.StatusCode.DEADLINE_EXCEEDED,
+    grpc.StatusCode.INTERNAL,
+    grpc.StatusCode.UNKNOWN,
+})
```
raballew: A comment documents the superset relationship, but the code could express it directly:

```python
_RETRYABLE_STREAM_CODES = _TRANSIENT_GRPC_CODES | frozenset({
    grpc.StatusCode.INTERNAL,
    grpc.StatusCode.UNKNOWN,
})
```

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -298,55 +375,146 @@ async def _controller_stub(self) -> AsyncGenerator[jumpstarter_pb2_grpc.Controll
...
         stream_name: str,
         stream_factory: Callable[[jumpstarter_pb2_grpc.ControllerServiceStub], AsyncGenerator],
         send_tx,
-        retries: int = 5,
-        backoff: float = 1.0,  # Reduced from 3.0 for faster recovery from transient errors
+        grace_period: float = 300.0,
+        max_backoff: float = 10.0,
+        on_terminal: Callable[[str, Exception], None] | None = None,
+        on_exhausted: Callable[[str, Exception], None] | None = None,
     ):
```
raballew: Adding `send_tx: MemoryObjectSendStream[Any]` and `-> None` would improve readability.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
raballew, line 171: `DEADLINE_EXCEEDED` is included in `_RETRYABLE_STREAM_CODES` but only `UNAVAILABLE`, `INTERNAL`, and `UNKNOWN` are called out in the code comments as retryable stream codes. Is this intentional?

python/packages/jumpstarter/jumpstarter/exporter/exporter_retry_test.py
```diff
@@ -21,140 +29,384 @@ async def channel_factory():
...
+class TestGraceWindow:
+    def test_starts_not_expired(self):
+        w = _GraceWindow(period=10.0)
+        assert w.expired() is False
+        assert w.elapsed() == 0.0
+
+    def test_mark_failure_starts_window(self):
+        w = _GraceWindow(period=10.0)
+        elapsed = w.mark_failure()
+        assert elapsed >= 0.0
```
raballew: `assert elapsed >= 0.0` is always true.

raballew (changes requested):

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: found no evidence logger.info can raise here

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: done

python/packages/jumpstarter/jumpstarter/exporter/exporter_retry_test.py
bennyz: done

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: this is for internal logging

PR #950, opened 2026-08-03
refactor: split task group into data-plane and control-plane

bennyz:
Introduce an outer conns_tg (data-plane) that hosts handle_lease
and _handle_client_conn, and an inner tg (control-plane) that hosts
Status/Listen streams and _handle_end_session.

When _cancel_with_fatal_error fires (Status stream terminal error),
only the inner group is cancelled. Active client tunnels on conns_tg
remain alive until serve() explicitly cancels the outer group.

Add TestTaskGroupIsolation to verify a connection task survives
control-plane cancellation.

Depends on #949
Next: #951

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/exporter.py
@@ -382,12 +382,14 @@ class Exporter(AsyncContextManagerMixin, Metadata):
     status ticks after handle_lease's finally has cleaned up."""
 
     _pending_lease_status: jumpstarter_pb2.StatusResponse | None = field(init=False, default=None)
-    """Stashed status from a lease reassignment, replayed after handle_lease's
-    finally clears _lease_context so the new lease can be acquired."""
+    """Stashed status from a lease reassignment, replayed after handle_lease
+    finalizes so the new lease can be acquired. Needed because the controller
+    Status stream suppresses duplicate ticks via proto.Equal."""
 
     _status_replay_tx: MemoryObjectSendStream | None = field(init=False, default=None)
     """Send side of the status channel, used to replay _pending_lease_status
     back into the status loop after a lease transition."""
+
     _lease_context: LeaseContext | None = field(init=False, default=None)
     """Encapsulates all resources associated with the current lease.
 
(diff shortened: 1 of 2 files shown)


python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1250,7 +1255,7 @@ def _on_lease_acquired(
                 self.stop,
                 self._request_lease_release,
             )
-        tg.start_soon(self.handle_lease, status.lease_name, tg, lease_scope)
+        conns_tg.start_soon(self.handle_lease, status.lease_name, conns_tg, lease_scope)
```
raballew, line 1517: if `session_for_lease()` raises during `handle_lease`, execution jumps to the outer `finally` without entering the inner `try/finally` that cleans up. The exporter hence could hang permanently.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1163,21 +1164,23 @@ async def serve(self):
...
+            async with create_task_group() as conns_tg:
+                await self._run_control_plane(status_tx, status_rx, conns_tg)
+                if self._fatal_stream_error:
+                    name, err = self._fatal_stream_error
+                    logger.warning(
+                        "Control plane down (%s: %s), cancelling active connections",
+                        name,
+                        err,
+                    )
+                conns_tg.cancel_scope.cancel()
```
raballew, line 1405: Could we try keeping connections alive during control-plane recovery?

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1192,21 +1195,22 @@ async def _run_control_plane(self, status_tx, status_rx):
...
     ) -> bool:
         """Process a single status update. Returns True to stop the status loop."""
         previous_state = self._lease_state
         current_leased = status.leased
 
         if current_leased:
             if previous_state == LeaseState.IDLE and status.lease_name != "":
-                self._on_lease_acquired(status, tg)
+                self._on_lease_acquired(status, tg, conns_tg)
             elif (
```
raballew, line 1471: Consider clearing `_lease_context` in `_on_lease_released` after the shielded wait completes, and have the `handle_lease` finally guard against double-clearing.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -1075,7 +1075,8 @@ async def handle_lease(self, lease_name: str, tg: TaskGroup, lease_scope: LeaseC
                 # session creation (e.g., BEFORE_LEASE_HOOK when hooks are configured).
 
                 # Start task to handle EndSession requests (runs afterLease hook when client signals done)
-                tg.start_soon(self._handle_end_session, lease_scope)
+                # Runs on control-plane group so it's cancelled with Status/Listen, not data-plane
+                self._tg.start_soon(self._handle_end_session, lease_scope)
```
raballew, line 1314: Consider also moving the Listen stream to `self._tg` so its lifecycle matches the other control-plane streams.

raballew (changes requested):

bkhizgiy (approved):
looks good from my side.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: listen is per lease while status is per process

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
bennyz: they stay up for the grace period (300s)

PR #951, opened 2026-08-03
fix: per-connection Dial retry with dual timeout budget

bennyz:
Replace pre-flight _dial_with_retry with handle_async that retries Dial on every Unix socket connection independently. Use dial_timeout (60s) for FAILED_PRECONDITION ("not ready") and retry_timeout (300s, only when _connected=True) for UNAVAILABLE, so initial lease acquisition stays fast while established sessions survive controller restarts.

Fix permission-denied detection: controller sends fmt.Errorf("permission denied") which maps to UNKNOWN, not PERMISSION_DENIED - match on details text instead of status code.

Depends on #950

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
@@ -12,7 +12,7 @@ import anyio.to_thread
 import click
 import grpc
 import grpc.aio
-from anyio import create_task_group, get_cancelled_exc_class
+from anyio import CancelScope, create_task_group, get_cancelled_exc_class
 from jumpstarter_cli_common.config import opt_config
 from jumpstarter_cli_common.exceptions import find_exception_in_group, handle_exceptions_with_reauthentication
 from jumpstarter_cli_common.oidc import (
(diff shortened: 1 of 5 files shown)


python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -325,31 +326,34 @@ def __contextmanager__(self) -> Generator[Self]:
...
-        deadline = time.monotonic() + self.dial_timeout
+        dial_deadline = started + self.dial_timeout
+        unavail_budget = self.retry_timeout if self._connected else self.dial_timeout
+        unavailable_deadline = started + unavail_budget if unavail_budget > 0 else None
         attempt = 0
+        warned_unavailable = False
         while True:
             try:
-                return await self.controller.Dial(jumpstarter_pb2.DialRequest(lease_name=self.name))
+                response = await self.controller.Dial(jumpstarter_pb2.DialRequest(lease_name=self.name))
```
raballew: No per-RPC timeout on the Dial call. The keepalive bound (200s) can silently exceed `dial_timeout` (60s).

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -504,6 +504,7 @@ async def _shell_with_signal_handling(  # noqa: C901
             try:
                 async with anyio.from_thread.BlockingPortal() as portal:
                     connect_deadline = None
+                    connect_start = None
                     while True:
```
raballew, line 508: Token monitor tasks accumulate without cancellation across outer retry iterations. It would be better tying the monitoring task lifetime to the lease using a per-lease cancel scope, cancelling the monitor in a `finally` block before each `continue`.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -534,11 +535,13 @@ async def _shell_with_signal_handling(  # noqa: C901
                                         "Session is no longer valid."
                                     ) from unreachable
                                 if connect_deadline is None:
-                                    connect_deadline = time.monotonic() + lease.retry_timeout
+                                    connect_start = time.monotonic()
+                                    connect_deadline = connect_start + lease.retry_timeout
```
raballew, line 543: `connect_deadline` in the shell retry loop never resets after successful reconnection.

python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -325,31 +326,34 @@ def __contextmanager__(self) -> Generator[Self]:
...
             yield value
 
-    async def _dial_with_retry(self):
-        """Dial the controller with exponential backoff, waiting for the exporter to be ready.
-
-        Returns DialResponse on success.
-        Raises ExporterUnreachableError on timeout or unrecoverable error.
-        """
-        logger.debug("Dialing controller for lease %s", self.name)
+    async def handle_async(self, stream):  # noqa: C901
```
raballew: function parameters lack type annotations

python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -382,47 +401,31 @@ async def _dial_with_retry(self):
                     await sleep(delay)
                     attempt += 1
                     continue
-                # Exporter went offline or lease ended - raise immediately
                 if "permission denied" in str(e.details()).lower():
```
raballew: Consider adding `or e.code() == grpc.StatusCode.PERMISSION_DENIED` for defense-in-depth.

python/packages/jumpstarter/jumpstarter/client/lease.py
```diff
@@ -382,47 +401,31 @@ async def _dial_with_retry(self):
...
             ):
                 pass
+        except grpc.aio.AioRpcError as e:
+            raise ExporterUnreachableError(
+                f"Router {response.router_endpoint} unreachable: {e.details()}"
+            ) from e
+        except OSError as e:
+            raise ExporterUnreachableError(
+                f"Router {response.router_endpoint} connection failed: {e}"
+            ) from e
```
raballew: `stream` is not closed when Dial fails before the router stream is entered. Consider wrapping the method body with `try/except Exception: await stream.aclose(); raise`.

raballew (changes requested):

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
bennyz: resets in line #556

PR #975, opened 2026-08-05
feat: TLS and env-based endpoint for jumpstarter-telemetry

bkhizgiy:
Follow-up to #930 

jumpstarter-telemetry now always uses TLS. 
Set `EXTERNAL_CERT_PEM` and `EXTERNAL_KEY_PEM` to mount operator-provided certs, 
falls back to a self-signed cert when absent.
The controller now reads `GRPC_TELEMETRY_ENDPOINT` from the
environment instead of auto-deriving `jumpstarter-telemetry.<ns>:9093` 
from the ConfigMap loader.

### DIFF
--- a/controller/cmd/telemetry/main.go
+++ b/controller/cmd/telemetry/main.go
@@ -91,13 +91,15 @@ func main() {
 		Signer:   signer,
 	}
 
+	// Register signal handler before starting the service so no signal
+	// is missed in the window between goroutine start and Notify.
+	sigs := make(chan os.Signal, 1)
+	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
+
 	errCh := make(chan error, 1)
 	go func() {
 		errCh <- svc.Start(ctx)
 	}()
-
-	sigs := make(chan os.Signal, 1)
-	signal.Notify(sigs, syscall.SIGINT, syscall.SIGTERM)
 
 	select {
 	case sig := <-sigs:
--- a/controller/cmd/telemetry/main.go
+++ b/controller/cmd/telemetry/main.go
@@ -18,6 +18,18 @@ limitations under the License.
 // via the PushLogs gRPC RPC and writes them to structured stdout for downstream
 // log shippers (Promtail, Grafana Alloy, Vector) to forward to Loki.
 //
+// TLS: always enabled. Set EXTERNAL_CERT_PEM and EXTERNAL_KEY_PEM to file paths of
+// operator-mounted cert/key (e.g. from a cert-manager Secret); when absent a
+// self-signed certificate is generated. The self-signed cert PEM is logged at
+// startup — copy it into the controller ConfigMap's telemetry.certificate field
+// so exporters can verify the TLS connection.
+//
+// Endpoint: GRPC_TELEMETRY_ENDPOINT must be set on BOTH this pod and the controller
+// pod to the same value (e.g. "jumpstarter-telemetry.jumpstarter.svc:9093").
+// The telemetry service uses it to generate the correct SAN in the self-signed
+// certificate; the controller uses it to advertise the address to exporters via
+// GetServiceEndpoints. A mismatch causes TLS hostname verification failures.
+//
 // Future phases will add direct Loki push and MetricsStream for reverse-scrape
 // of exporter prometheus_client registries.
 package main
--- a/controller/internal/config/config.go
+++ b/controller/internal/config/config.go
@@ -127,12 +127,6 @@ func LoadConfiguration(
 		if err := config.Telemetry.Validate(); err != nil {
 			return nil, err
 		}
-		// Auto-derive the gRPC address when the operator has not overridden it.
-		// The well-known service name follows the same pattern as the controller
-		// and router: <service>.<namespace>.svc (in-cluster DNS).
-		if config.Telemetry.Endpoint == "" {
-			config.Telemetry.Endpoint = "jumpstarter-telemetry." + key.Namespace + ":9093"
-		}
 		telemetry = config.Telemetry
 	}
 
--- a/controller/internal/config/config.go
+++ b/controller/internal/config/config.go
@@ -1,8 +1,11 @@
 package config
 
 import (
+	"cmp"
 	"context"
 	"fmt"
+	"net"
+	"os"
 	"time"
 
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/oidc"
@@ -41,6 +44,34 @@ func LoadRouterConfiguration(
 	}
 
 	return serverOptions, nil
+}
+
+// resolveTelemetryConfig validates and resolves the telemetry endpoint for a
+// Telemetry config block. The GRPC_TELEMETRY_ENDPOINT env var takes priority
+// over the ConfigMap value, allowing operators to override at the pod level.
+// Returns nil when t is nil or disabled.
+func resolveTelemetryConfig(t *Telemetry) (*Telemetry, error) {
+	if t == nil || !t.Enabled {
+		return nil, nil
+	}
+	if err := t.Validate(); err != nil {
+		return nil, err
+	}
+	// Env var takes priority over ConfigMap, allowing operators to override
+	// at the pod level without modifying the ConfigMap. Resolving here ensures
+	// LoadedConfig.Telemetry.Endpoint is always the complete value — callers
+	// don't need to re-check the env var.
+	t.Endpoint = cmp.Or(os.Getenv("GRPC_TELEMETRY_ENDPOINT"), t.Endpoint)
+	if ep := t.Endpoint; ep != "" {
+		host, _, err := net.SplitHostPort(ep)
+		if err != nil {
+			return nil, fmt.Errorf("telemetry endpoint %q is not a valid host:port: %w", ep, err)
+		}
+		if host == "" {
+			return nil, fmt.Errorf("telemetry endpoint %q has no host", ep)
+		}
+	}
+	return t, nil
 }
 
 func LoadConfiguration(
@@ -122,12 +153,9 @@ func LoadConfiguration(
 		return nil, err
 	}
 
-	var telemetry *Telemetry
-	if config.Telemetry != nil && config.Telemetry.Enabled {
-		if err := config.Telemetry.Validate(); err != nil {
-			return nil, err
-		}
-		telemetry = config.Telemetry
+	telemetry, err := resolveTelemetryConfig(config.Telemetry)
+	if err != nil {
+		return nil, err
 	}
 
 	return &LoadedConfig{
--- a/controller/internal/config/types_test.go
+++ b/controller/internal/config/types_test.go
@@ -276,6 +276,97 @@ func TestDeprecatedLabelsOmitEmpty(t *testing.T) {
 	}
 }
 
+func TestTelemetryEndpointResolution(t *testing.T) {
+	tests := []struct {
+		name         string
+		cfg          *Telemetry
+		envValue     string
+		wantNil      bool
+		wantEndpoint string
+		wantErr      bool
+	}{
+		{
+			name:    "nil config returns nil",
+			cfg:     nil,
+			wantNil: true,
+		},
+		{
+			name:    "disabled config returns nil",
+			cfg:     &Telemetry{Enabled: false, Endpoint: "telemetry:9093"},
+			wantNil: true,
+		},
+		{
+			name:         "env var takes precedence over ConfigMap",
+			cfg:          &Telemetry{Enabled: true, Endpoint: "telemetry.ns.svc:9093"},
+			envValue:     "env-telemetry:9093",
+			wantEndpoint: "env-telemetry:9093",
+		},
+		{
+			name:         "ConfigMap fallback when env var is empty",
+			cfg:          &Telemetry{Enabled: true, Endpoint: "telemetry.ns.svc:9093"},
+			wantEndpoint: "telemetry.ns.svc:9093",
+		},
+		{
+			name:         "both empty yields empty endpoint (no error)",
+			cfg:          &Telemetry{Enabled: true},
+			wantEndpoint: "",
+		},
+		{
+			name:    "malformed ConfigMap value is rejected",
+			cfg:     &Telemetry{Enabled: true, Endpoint: "no-port"},
+			wantErr: true,
+		},
+		{
+			name:     "malformed env var is rejected",
+			cfg:      &Telemetry{Enabled: true},
+			envValue: "garbage-no-port",
+			wantErr:  true,
+		},
+		{
+			name:    "port-only ConfigMap value is rejected",
+			cfg:     &Telemetry{Enabled: true, Endpoint: ":9093"},
+			wantErr: true,
+		},
+		{
+			name:     "port-only env var is rejected",
+			cfg:      &Telemetry{Enabled: true},
+			envValue: ":9093",
+			wantErr:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv("GRPC_TELEMETRY_ENDPOINT", tt.envValue)
+
+			resolved, err := resolveTelemetryConfig(tt.cfg)
+
+			if tt.wantErr {
+				if err == nil {
+					t.Fatalf("expected validation error, got nil (resolved=%+v)", resolved)
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if tt.wantNil {
+				if resolved != nil {
+					t.Fatalf("expected nil, got %+v", resolved)
+				}
+				return
+			}
+			var gotEndpoint string
+			if resolved != nil {
+				gotEndpoint = resolved.Endpoint
+			}
+			if gotEndpoint != tt.wantEndpoint {
+				t.Errorf("resolved.Endpoint = %q, want %q", gotEndpoint, tt.wantEndpoint)
+			}
+		})
+	}
+}
+
 func TestParseDuration(t *testing.T) {
 	tests := []struct {
 		input    string
--- a/controller/internal/config/types_test.go
+++ b/controller/internal/config/types_test.go
@@ -276,6 +276,78 @@ func TestDeprecatedLabelsOmitEmpty(t *testing.T) {
 	}
 }
 
+func TestTelemetryEndpointResolution(t *testing.T) {
+	tests := []struct {
+		name         string
+		configValue  string
+		envValue     string
+		wantEndpoint string
+		wantErr      bool
+	}{
+		{
+			name:         "env var takes precedence over ConfigMap",
+			configValue:  "telemetry.ns.svc:9093",
+			envValue:     "env-telemetry:9093",
+			wantEndpoint: "env-telemetry:9093",
+		},
+		{
+			name:         "ConfigMap fallback when env var is empty",
+			configValue:  "telemetry.ns.svc:9093",
+			wantEndpoint: "telemetry.ns.svc:9093",
+		},
+		{
+			name:         "both empty yields empty endpoint (no error)",
+			wantEndpoint: "",
+		},
+		{
+			name:        "malformed ConfigMap value is rejected",
+			configValue: "no-port",
+			wantErr:     true,
+		},
+		{
+			name:     "malformed env var is rejected",
+			envValue: "garbage-no-port",
+			wantErr:  true,
+		},
+		{
+			name:        "port-only ConfigMap value is rejected",
+			configValue: ":9093",
+			wantErr:     true,
+		},
+		{
+			name:     "port-only env var is rejected",
+			envValue: ":9093",
+			wantErr:  true,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			t.Setenv("GRPC_TELEMETRY_ENDPOINT", tt.envValue)
+
+			cfg := &Telemetry{Enabled: true, Endpoint: tt.configValue}
+			resolved, err := resolveTelemetryConfig(cfg)
+
+			if tt.wantErr {
+				if err == nil {
+					t.Fatalf("expected validation error, got nil (resolved=%+v)", resolved)
+				}
+				return
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			var gotEndpoint string
+			if resolved != nil {
+				gotEndpoint = resolved.Endpoint
+			}
+			if gotEndpoint != tt.wantEndpoint {
+				t.Errorf("resolved.Endpoint = %q, want %q", gotEndpoint, tt.wantEndpoint)
+			}
+		})
+	}
+}
+
 func TestParseDuration(t *testing.T) {
 	tests := []struct {
 		input    string
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -327,8 +327,11 @@ func (s *ControllerService) GetServiceEndpoints(
 	resp := &pb.GetServiceEndpointsResponse{}
 
 	if s.TelemetryConfig != nil && s.TelemetryConfig.Enabled {
+		// Prefer the explicit ConfigMap endpoint; fall back to GRPC_TELEMETRY_ENDPOINT
+		// so the operator can pass the address via env var without touching the ConfigMap.
+		ep := cmp.Or(s.TelemetryConfig.Endpoint, telemetryEndpoint())
 		resp.TelemetryEndpoints = append(resp.TelemetryEndpoints, &pb.TelemetryEndpoint{
-			Endpoint:    s.TelemetryConfig.Endpoint,
+			Endpoint:    ep,
 			Certificate: s.TelemetryConfig.Certificate,
 			MinSeverity: cmp.Or(s.TelemetryConfig.Logging.Filter.MinSeverity, "info"),
 		})
--- a/controller/internal/service/controller_service.go
+++ b/controller/internal/service/controller_service.go
@@ -327,11 +327,13 @@ func (s *ControllerService) GetServiceEndpoints(
 	resp := &pb.GetServiceEndpointsResponse{}
 
 	if s.TelemetryConfig != nil && s.TelemetryConfig.Enabled {
-		// Prefer the explicit ConfigMap endpoint; fall back to GRPC_TELEMETRY_ENDPOINT
-		// so the operator can pass the address via env var without touching the ConfigMap.
-		ep := cmp.Or(s.TelemetryConfig.Endpoint, telemetryEndpoint())
+		// Endpoint is resolved at config-load time (ConfigMap value or GRPC_TELEMETRY_ENDPOINT
+		// env var fallback), so TelemetryConfig.Endpoint is always the complete value here.
+		if s.TelemetryConfig.Endpoint == "" {
+			return nil, status.Error(codes.FailedPrecondition, "telemetry is enabled but no endpoint is configured; set telemetry.endpoint in the ConfigMap or GRPC_TELEMETRY_ENDPOINT on the controller pod")
+		}
 		resp.TelemetryEndpoints = append(resp.TelemetryEndpoints, &pb.TelemetryEndpoint{
-			Endpoint:    ep,
+			Endpoint:    s.TelemetryConfig.Endpoint,
 			Certificate: s.TelemetryConfig.Certificate,
 			MinSeverity: cmp.Or(s.TelemetryConfig.Logging.Filter.MinSeverity, "info"),
 		})
@@ -1199,32 +1201,9 @@ func (s *ControllerService) Start(ctx context.Context) error {
 		return err
 	}
 
-	// Load external certificate if provided via environment variables.
-	// Environment variables EXTERNAL_CERT_PEM and EXTERNAL_KEY_PEM should contain the PEM-encoded
-	// certificate and private key respectively. If both are set, they are used; otherwise
-	// a self-signed certificate is generated.
-	var cert *tls.Certificate
-	certPEMPath := os.Getenv("EXTERNAL_CERT_PEM")
-	keyPEMPath := os.Getenv("EXTERNAL_KEY_PEM")
-	if certPEMPath != "" && keyPEMPath != "" {
-		certPEMBytes, err := os.ReadFile(certPEMPath)
-		if err != nil {
-			return fmt.Errorf("failed to read external certificate file: %w", err)
-		}
-		keyPEMBytes, err := os.ReadFile(keyPEMPath)
-		if err != nil {
-			return fmt.Errorf("failed to read external key file: %w", err)
-		}
-		parsedCert, err := tls.X509KeyPair(certPEMBytes, keyPEMBytes)
-		if err != nil {
-			return fmt.Errorf("failed to parse external certificate: %w", err)
-		}
-		cert = &parsedCert
-	} else {
-		cert, err = NewSelfSignedCertificate("jumpstarter controller", dnsnames, ipaddresses)
-		if err != nil {
-			return err
-		}
+	cert, _, err := LoadTLSCertificate("jumpstarter controller", dnsnames, ipaddresses)
+	if err != nil {
+		return err
 	}
 
 	opts := append(s.ServerOptions,
@@ -1267,8 +1246,11 @@ func (s *ControllerService) Start(ctx context.Context) error {
 	// Register gRPC gateway
 	gwmux := gwruntime.NewServeMux()
 
+	// The controller multiplexes gRPC (h2) and REST (http/1.1) on a single port,
+	// so it needs NextProtos — which LoadTLSCredentials doesn't expose.
 	listener, err := tls.Listen("tcp", ":8082", &tls.Config{
 		Certificates: []tls.Certificate{*cert},
+		MinVersion:   tls.VersionTLS12,
 		NextProtos:   []string{"http/1.1", "h2"},
 	})
 	if err != nil {
--- a/controller/internal/service/endpoints.go
+++ b/controller/internal/service/endpoints.go
@@ -21,6 +21,10 @@ func routerEndpoint() string {
 	return ep
 }
 
+func telemetryEndpoint() string {
+	return os.Getenv("GRPC_TELEMETRY_ENDPOINT")
+}
+
 func endpointToSAN(endpoint string) ([]string, []net.IP, error) {
 	host, _, err := net.SplitHostPort(endpoint)
 	if err != nil {
--- a/controller/internal/service/endpoints.go
+++ b/controller/internal/service/endpoints.go
@@ -1,6 +1,7 @@
 package service
 
 import (
+	"fmt"
 	"net"
 	"os"
 )
@@ -21,15 +22,36 @@ func routerEndpoint() string {
 	return ep
 }
 
-func telemetryEndpoint() string {
-	return os.Getenv("GRPC_TELEMETRY_ENDPOINT")
+// telemetryEndpoint returns the GRPC_TELEMETRY_ENDPOINT env var value.
+// Returns ("", nil) if unset, or an error if set but malformed.
+func telemetryEndpoint() (string, error) {
+	ep := os.Getenv("GRPC_TELEMETRY_ENDPOINT")
+	if ep == "" {
+		return "", nil
+	}
+	if err := validateHostPort(ep); err != nil {
+		return "", fmt.Errorf("GRPC_TELEMETRY_ENDPOINT %q is not a valid host:port: %w", ep, err)
+	}
+	return ep, nil
+}
+
+// validateHostPort checks that s is a valid "host:port" with a non-empty host.
+func validateHostPort(s string) error {
+	host, _, err := net.SplitHostPort(s)
+	if err != nil {
+		return err
+	}
+	if host == "" {
+		return fmt.Errorf("endpoint %q has no host", s)
+	}
+	return nil
 }
 
 func endpointToSAN(endpoint string) ([]string, []net.IP, error) {
-	host, _, err := net.SplitHostPort(endpoint)
-	if err != nil {
+	if err := validateHostPort(endpoint); err != nil {
 		return nil, nil, err
 	}
+	host, _, _ := net.SplitHostPort(endpoint)
 	ip := net.ParseIP(host)
 	if ip != nil {
 		return []string{}, []net.IP{ip}, nil
--- a/controller/internal/service/telemetry_service.go
+++ b/controller/internal/service/telemetry_service.go
@@ -18,9 +18,12 @@ package service
 
 import (
 	"context"
+	"crypto/tls"
+	"encoding/pem"
 	"errors"
 	"fmt"
 	"net"
+	"os"
 	"strings"
 	"time"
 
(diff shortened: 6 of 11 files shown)


bkhizgiy:
@coderabbitai review

controller/internal/service/controller_service.go
```diff
@@ -327,8 +327,11 @@ func (s *ControllerService) GetServiceEndpoints(
 	resp := &pb.GetServiceEndpointsResponse{}
 
 	if s.TelemetryConfig != nil && s.TelemetryConfig.Enabled {
+		// Prefer the explicit ConfigMap endpoint; fall back to GRPC_TELEMETRY_ENDPOINT
+		// so the operator can pass the address via env var without touching the ConfigMap.
+		ep := cmp.Or(s.TelemetryConfig.Endpoint, telemetryEndpoint())
 		resp.TelemetryEndpoints = append(resp.TelemetryEndpoints, &pb.TelemetryEndpoint{
-			Endpoint:    s.TelemetryConfig.Endpoint,
+			Endpoint:    ep,
```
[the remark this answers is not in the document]
bkhizgiy: done

controller/internal/service/controller_service.go
```diff
@@ -327,8 +327,11 @@ func (s *ControllerService) GetServiceEndpoints(
...
 			Certificate: s.TelemetryConfig.Certificate,
 			MinSeverity: cmp.Or(s.TelemetryConfig.Logging.Filter.MinSeverity, "info"),
 		})
```
raballew, line 337: When `GRPC_TELEMETRY_ENDPOINT` is unset and `TelemetryConfig.Endpoint` is also empty, `cmp.Or("", "")` returns `""`. The append highlighted here, runs unconditionally, so all exporters receive a `TelemetryEndpoint{Endpoint: ""}` and attempt to dial an empty string. I think you need to guard the append or return error.

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +200,88 @@ func truncate(s string, n int) string {
...
+			"certPEM", selfSignedPEM)
+	}
+
 	lis, err := net.Listen("tcp", s.BindAddr)
 	if err != nil {
 		return fmt.Errorf("telemetry: listen %s: %w", s.BindAddr, err)
 	}
 
-	srv := grpc.NewServer()
+	srv := grpc.NewServer(grpc.Creds(creds))
```
raballew, line 284: The `errCh` branch (`case err := <-errCh: return err`) in the `select` block below returns without stopping the gRPC server.

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +200,88 @@ func truncate(s string, n int) string {
...
+		if genErr != nil {
+			return nil, "", genErr
+		}
+		// Encode the leaf cert as PEM so the caller can log it for the operator.
+		selfSignedPEM = string(pem.EncodeToMemory(&pem.Block{
+			Type:  "CERTIFICATE",
+			Bytes: cert.Certificate[0],
+		}))
+	}
+	return credentials.NewServerTLSFromCert(cert), selfSignedPEM, nil
```
raballew, line 260: TLS credential loading is duplicated between this block and `controller_service.go:1206-1228` and I think they already diverged.

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +200,88 @@ func truncate(s string, n int) string {
...
+
 	lis, err := net.Listen("tcp", s.BindAddr)
 	if err != nil {
 		return fmt.Errorf("telemetry: listen %s: %w", s.BindAddr, err)
 	}
 
-	srv := grpc.NewServer()
+	srv := grpc.NewServer(grpc.Creds(creds))
 	pb.RegisterTelemetryServiceServer(srv, s)
 	reflection.Register(srv)
```
raballew, line 286: The core behavioral change introduced by this PR, `grpc.NewServer(grpc.Creds(creds))` mandating TLS for all connections, has no behavioral test.

controller/internal/service/telemetry_service_test.go
```diff
@@ -91,8 +94,12 @@ func buildTelemetryEndpointsResponse(cfg *config.Telemetry) *pb.GetServiceEndpoi
...
+		ep := cfg.Endpoint
+		if ep == "" {
+			ep = telemetryEndpoint()
+		}
 		resp.TelemetryEndpoints = append(resp.TelemetryEndpoints, &pb.TelemetryEndpoint{
-			Endpoint:    cfg.Endpoint,
+			Endpoint:    ep,
 			Certificate: cfg.Certificate,
 			MinSeverity: minSev,
 		})
```
raballew, line 105: The production line `ep := cmp.Or(s.TelemetryConfig.Endpoint, telemetryEndpoint())` in `controller_service.go` has zero coverage.

controller/internal/service/telemetry_service_test.go
```diff
@@ -153,6 +160,246 @@ func TestGetServiceEndpoints_DefaultsMinSeverityToInfo(t *testing.T) {
...
+	}
+}
+
+func TestTelemetryService_LoadTLSCredentials_SelfSignedFallsBackToLocalhostWhenNoEndpoint(t *testing.T) {
+	// When GRPC_TELEMETRY_ENDPOINT is empty, SAN defaults to "localhost".
+	sans := selfSignedSANs(t, "")
+	if len(sans) != 1 || sans[0] != "localhost" {
+		t.Errorf("expected SAN [localhost], got %v", sans)
+	}
+}
```
raballew, line 310: Neither test actually calls `svc.loadTLSCredentials()`.

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +200,88 @@ func truncate(s string, n int) string {
...
+
+	var cert *tls.Certificate
+	if certPEMPath != "" && keyPEMPath != "" {
+		certPEMBytes, readErr := os.ReadFile(certPEMPath)
+		if readErr != nil {
+			return nil, "", fmt.Errorf("failed to read external certificate file: %w", readErr)
+		}
+		keyPEMBytes, readErr := os.ReadFile(keyPEMPath)
+		if readErr != nil {
+			return nil, "", fmt.Errorf("failed to read external key file: %w", readErr)
```
raballew, line 223: Do we have a test for valid cert but missing/invalid key?

controller/internal/config/config.go
```diff
@@ -127,12 +127,6 @@ func LoadConfiguration(
...
 			return nil, err
 		}
-		// Auto-derive the gRPC address when the operator has not overridden it.
-		// The well-known service name follows the same pattern as the controller
-		// and router: <service>.<namespace>.svc (in-cluster DNS).
-		if config.Telemetry.Endpoint == "" {
-			config.Telemetry.Endpoint = "jumpstarter-telemetry." + key.Namespace + ":9093"
-		}
 		telemetry = config.Telemetry
 	}
```
raballew, line 131: `GRPC_TELEMETRY_ENDPOINT` is read in the service layer (`endpoints.go`) rather than here in `LoadConfiguration`. As a result, `LoadedConfig.Telemetry.Endpoint` is always `""` when the ConfigMap omits the field, even if the env var is set.

controller/internal/service/endpoints.go
```diff
@@ -21,6 +21,10 @@ func routerEndpoint() string {
 	return ep
 }
 
+func telemetryEndpoint() string {
+	return os.Getenv("GRPC_TELEMETRY_ENDPOINT")
+}
```
raballew, line 26: The raw env var value is returned with no `host:port` format validation before it is distributed to all exporters.

controller/internal/service/telemetry_service.go
raballew, line 260: Whats our required TLS min version?

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +200,88 @@ func truncate(s string, n int) string {
...
+		// ConfigMap's telemetry.certificate field. Exporters need this PEM to verify
+		// the TLS connection — a self-signed cert is not trusted by the system CA pool.
+		logger.Info("Using self-signed TLS certificate; copy certPEM into the controller ConfigMap telemetry.certificate so exporters can verify TLS",
+			"certPEM", selfSignedPEM)
+	}
+
 	lis, err := net.Listen("tcp", s.BindAddr)
 	if err != nil {
 		return fmt.Errorf("telemetry: listen %s: %w", s.BindAddr, err)
 	}
```
raballew, line 282: After `net.Listen` succeeds there is no `defer lis.Close()` guard.

controller/cmd/telemetry/main.go
```diff
@@ -18,6 +18,18 @@ limitations under the License.
...
+//
+// Endpoint: GRPC_TELEMETRY_ENDPOINT must be set on BOTH this pod and the controller
+// pod to the same value (e.g. "jumpstarter-telemetry.jumpstarter.svc:9093").
+// The telemetry service uses it to generate the correct SAN in the self-signed
+// certificate; the controller uses it to advertise the address to exporters via
+// GetServiceEndpoints. A mismatch causes TLS hostname verification failures.
+//
 // Future phases will add direct Loki push and MetricsStream for reverse-scrape
 // of exporter prometheus_client registries.
 package main
```
raballew, line 35: In the `main()` function below, the goroutine calling `svc.Start(ctx)` is launched before `signal.Notify` is registered. In the narrow window between goroutine start and signal subscription, SIGINT and SIGTERM use Go's default disposition (immediate process termination). A signal arriving in this window bypasses the graceful shutdown path entirely.

bkhizgiy: Fixed. `signal.Notify` is now registered before the goroutine is launched, so there is no window where a signal can bypass the graceful shutdown path.

controller/internal/service/telemetry_service.go
bkhizgiy: Fixed. Added defer `func() { _ = lis.Close() }()` immediately after `net.Listen succeeds`. The error is intentionally ignored since grpc.Serve already closes the listener on exit , the defer is a safety guard for any future code inserted between Listen and Serve.

controller/internal/service/telemetry_service.go
bkhizgiy: TLS 1.2. Added `MinVersion: tls.VersionTLS12` explicitly in `LoadTLSCredentials` (and on the controller's tls.Listen config).

controller/internal/service/endpoints.go
bkhizgiy: Fixed. `telemetryEndpoint()` in endpoints.go now validates the value with `net.SplitHostPort` and returns "" for any malformed input, so an invalid env var is treated the same as an unset one rather than being propagated to exporters.

controller/internal/config/config.go
bkhizgiy: Fixed. `LoadConfiguration` now resolves the endpoint with `cmp.Or(config.Telemetry.Endpoint, os.Getenv("GRPC_TELEMETRY_ENDPOINT"))` before storing it in `LoadedConfig`.

controller/internal/service/telemetry_service.go
bkhizgiy: Yes, added two tests: `TestTelemetryService_LoadTLSCredentials_MissingKeyFileReturnsError` (valid cert + path that doesn't exist) and `TestTelemetryService_LoadTLSCredentials_ValidCertInvalidKeyReturnsError` (valid cert + empty key file that fails X509KeyPair parsing).

controller/internal/service/telemetry_service_test.go
bkhizgiy: Fixed. Both SAN tests now call `svc.loadTLSCredentials()` directly, decode the returned selfSignedPEM with `x509.ParseCertificate`, and assert on `leaf.DNSNames`. The private `selfSignedSANs` helper was removed.

controller/internal/service/telemetry_service_test.go
bkhizgiy: Fixed as part of the endpoint resolution refactor. The cmp.Or fallback was moved into `LoadConfiguration` in config.go, so `TelemetryConfig.Endpoint` is always the fully resolved value by the time it reaches `GetServiceEndpoints`

controller/internal/service/telemetry_service.go
bkhizgiy: It would require a larger change, and since this feature is still WIP, I think it would be better to address it in a follow-up PR and cover some e2e flows as well. Would that be okay with you?

controller/internal/service/telemetry_service.go
bkhizgiy: Fixed. Extracted a shared `LoadTLSCertificate / LoadTLSCredentials` helper into `internal/service/tls_credentials.go`. All three services (controller, router, telemetry) now call it.

controller/internal/service/telemetry_service.go
bkhizgiy: Fixed. Added `srv.Stop() `before returning in the case `err := <-errCh:` branch, and added `defer func() { _ = lis.Close() }()` as a safety guard on the listener.

controller/internal/service/controller_service.go
bkhizgiy: Fixed. Endpoint resolution was moved to `LoadConfiguration`, `config.Telemetry.Endpoint` is now always the fully resolved value `(cmp.Or(ConfigMap value, GRPC_TELEMETRY_ENDPOINT))`. `GetServiceEndpoints` then guards the append and returns codes.Internal if the resolved endpoint is still empty, so exporters never receive an empty `TelemetryEndpoint`.

controller/internal/service/controller_service.go
```diff
@@ -327,6 +327,11 @@ func (s *ControllerService) GetServiceEndpoints(
 	resp := &pb.GetServiceEndpointsResponse{}
 
 	if s.TelemetryConfig != nil && s.TelemetryConfig.Enabled {
+		// Endpoint is resolved at config-load time (ConfigMap value or GRPC_TELEMETRY_ENDPOINT
+		// env var fallback), so TelemetryConfig.Endpoint is always the complete value here.
+		if s.TelemetryConfig.Endpoint == "" {
+			return nil, status.Error(codes.Internal, "telemetry is enabled but no endpoint is configured; set telemetry.endpoint in the ConfigMap or GRPC_TELEMETRY_ENDPOINT on the controller pod")
+		}
```
raballew, line 334: The empty-endpoint guard at these lines is only reachable after `authenticateExporter` succeeds. All existing tests that call `svc.GetServiceEndpoints` use a failing authenticator.

controller/internal/config/types_test.go
```diff
@@ -276,6 +277,77 @@ func TestDeprecatedLabelsOmitEmpty(t *testing.T) {
...
+			}
+			if err != nil {
+				t.Fatalf("unexpected error: %v", err)
+			}
+			if resolved != tt.wantResult {
+				t.Errorf("resolved = %q, want %q", resolved, tt.wantResult)
+			}
+		})
+	}
+}
```
raballew, line 349: `TestTelemetryEndpointResolution` simulates the `LoadConfiguration` logic instead of calling it.

controller/internal/service/controller_service.go
raballew, line 334: A missing endpoint is a known operator configuration gap, not an unexpected server failure. `codes.Internal` communicates unexpected server-side failure, whereas `codes.FailedPrecondition` better matches this situation per the gRPC status code semantics.

controller/internal/service/endpoints.go
```diff
@@ -21,6 +23,19 @@ func routerEndpoint() string {
...
+		ctrl.Log.WithName("telemetry").Info("GRPC_TELEMETRY_ENDPOINT is not a valid host:port; ignoring",
+			"value", ep, "error", err)
+		return ""
+	}
+	return ep
+}
+
 func endpointToSAN(endpoint string) ([]string, []net.IP, error) {
 	host, _, err := net.SplitHostPort(endpoint)
 	if err != nil {
```
raballew: `endpointToSAN` generates a broken TLS certificate when given a port-only endpoint like `:9093`.

controller/internal/service/endpoints.go
```diff
@@ -21,6 +23,19 @@ func routerEndpoint() string {
...
 }
 
+func telemetryEndpoint() string {
+	ep := os.Getenv("GRPC_TELEMETRY_ENDPOINT")
+	if ep == "" {
+		return ""
+	}
+	if _, _, err := net.SplitHostPort(ep); err != nil {
+		ctrl.Log.WithName("telemetry").Info("GRPC_TELEMETRY_ENDPOINT is not a valid host:port; ignoring",
+			"value", ep, "error", err)
```
raballew: Shouldnt this be error level?

raballew (changes requested):

controller/internal/service/controller_service.go
bkhizgiy: Fixed. Added authSuccessServiceCtx, a test helper that builds a ControllerService with a passing authenticator chain backed by a fake k8s client containing a pre-populated Exporter object.

controller/internal/config/types_test.go
bkhizgiy: Fixed. Extracted `resolveTelemetryConfig` from `LoadConfiguration` and rewrote the test to call it directly. The test no longer re-implements the resolution logic, any change to the real function will now break the test as expected.

controller/internal/service/controller_service.go
bkhizgiy: Fixed. Changed to `codes.FailedPrecondition`, you're right that a missing endpoint is a known operator misconfiguration, not an unexpected server-side failure, updated both the production handler and the test helper to match.

controller/internal/service/endpoints.go
bkhizgiy: Fixed. Added a `validateHostPort` helper that checks for an empty host after `net.SplitHostPor`t, both `endpointToSAN` and `telemetryEndpoint()` now use it, so :9093 is rejected at all validation sites.

controller/internal/service/endpoints.go
bkhizgiy: Fixed. Changed to `ctrl.Log.WithName("telemetry").Error(err, ...)`

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +197,56 @@ func truncate(s string, n int) string {
...
 
-	srv := grpc.NewServer()
+	srv := grpc.NewServer(grpc.Creds(creds))
```
bennyz: we need recovery interceptors

https://github.com/jumpstarter-dev/jumpstarter/blob/965658ad3b9f4213a27929d98f04c1ea04248858/controller/internal/service/controller_service.go#L1235

controller/internal/service/tls_credentials.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+		keyPEMBytes, err := os.ReadFile(keyPEMPath)
+		if err != nil {
+			return nil, "", fmt.Errorf("failed to read external key file: %w", err)
+		}
+		cert, err := tls.X509KeyPair(certPEMBytes, keyPEMBytes)
+		if err != nil {
+			return nil, "", fmt.Errorf("failed to parse external certificate: %w", err)
+		}
+		return &cert, "", nil
```
raballew: [CRITICAL] `EXTERNAL_CERT_PEM` and `EXTERNAL_KEY_PEM` are treated as file paths and read from disk via `os.ReadFile()`, but they are supposed to carry the PEM content directly. Setting them to actual PEM strings fails at runtime because the PEM data gets passed to `os.ReadFile()` as a filename. The fix is to read the env var value as bytes directly:

```go
certPEMBytes := []byte(os.Getenv("EXTERNAL_CERT_PEM"))
keyPEMBytes  := []byte(os.Getenv("EXTERNAL_KEY_PEM"))
```

The `main.go` comment saying "file paths of operator-mounted cert/key" and the tests that call `t.Setenv("EXTERNAL_CERT_PEM", certPath)` both confirm the file-path design was intentional, but it contradicts how the env vars should work. The source comments and tests need to be updated alongside the implementation.

controller/internal/service/tls_credentials.go
```diff
@@ -0,0 +1,88 @@
...
+// should log it so the operator can pin it in the relevant ConfigMap field.
+func LoadTLSCertificate(commonName string, dnsnames []string, ipaddresses []net.IP) (*tls.Certificate, string, error) {
+	certPEMPath := os.Getenv("EXTERNAL_CERT_PEM")
+	keyPEMPath := os.Getenv("EXTERNAL_KEY_PEM")
+
+	// Require both or neither — a partial configuration means a broken Secret
+	// mount that would cause clients trusting the intended certificate to fail TLS.
+	if (certPEMPath == "") != (keyPEMPath == "") {
+		return nil, "", fmt.Errorf("EXTERNAL_CERT_PEM and EXTERNAL_KEY_PEM must be set together; got cert=%q key=%q", certPEMPath, keyPEMPath)
+	}
```
raballew: [HIGH] When only one of `EXTERNAL_CERT_PEM` / `EXTERNAL_KEY_PEM` is set, the function returns an error and the service fails to start. The expected behavior is to log a warning and fall back to a self-signed certificate instead. Tests `TestTelemetryService_LoadTLSCredentials_OnlyCertEnvVarReturnsError` and `TestTelemetryService_LoadTLSCredentials_OnlyKeyEnvVarReturnsError` currently lock in the error behavior, but they should be updated to verify the fallback path:

```go
if (certPEMPath == "") != (keyPEMPath == "") {
    // log warning, then fall through to self-signed
    certPEMPath = ""
    keyPEMPath  = ""
}
```

controller/internal/config/config.go
```diff
@@ -43,6 +46,34 @@ func LoadRouterConfiguration(
...
+		return nil, nil
+	}
+	if err := t.Validate(); err != nil {
+		return nil, err
+	}
+	// Prefer the explicit ConfigMap value; fall back to the env var so the
+	// operator can set the endpoint without touching the ConfigMap. Resolving
+	// here ensures LoadedConfig.Telemetry.Endpoint is always the complete value
+	// — callers don't need to re-check the env var.
+	t.Endpoint = cmp.Or(t.Endpoint, os.Getenv("GRPC_TELEMETRY_ENDPOINT"))
```
raballew: The env var is supposed to take priority when explicitly set, with the ConfigMap acting as a fallback. Swap the arguments:

```go
t.Endpoint = cmp.Or(os.Getenv("GRPC_TELEMETRY_ENDPOINT"), t.Endpoint)
```

controller/internal/service/tls_credentials.go
```diff
@@ -0,0 +1,88 @@
...
+
+	if certPEMPath != "" {
+		certPEMBytes, err := os.ReadFile(certPEMPath)
+		if err != nil {
+			return nil, "", fmt.Errorf("failed to read external certificate file: %w", err)
+		}
+		keyPEMBytes, err := os.ReadFile(keyPEMPath)
+		if err != nil {
+			return nil, "", fmt.Errorf("failed to read external key file: %w", err)
+		}
```
raballew: [LOW] The file paths from `EXTERNAL_CERT_PEM` and `EXTERNAL_KEY_PEM` are passed to `os.ReadFile()` without any restriction on which directory they may point to. An attacker who can inject env vars into the pod (implying control-plane-level access) could point these at arbitrary files. In practice `tls.X509KeyPair` rejects non-PEM content so nothing reaches network clients, but the file is still read. Consider validating the resolved absolute path against the expected Secret mount prefix before opening the file, or document that these vars are trusted operator input.

controller/internal/service/controller_service.go
```diff
@@ -327,6 +327,11 @@ func (s *ControllerService) GetServiceEndpoints(
 	resp := &pb.GetServiceEndpointsResponse{}
 
 	if s.TelemetryConfig != nil && s.TelemetryConfig.Enabled {
+		// Endpoint is resolved at config-load time (ConfigMap value or GRPC_TELEMETRY_ENDPOINT
+		// env var fallback), so TelemetryConfig.Endpoint is always the complete value here.
+		if s.TelemetryConfig.Endpoint == "" {
+			return nil, status.Error(codes.FailedPrecondition, "telemetry is enabled but no endpoint is configured; set telemetry.endpoint in the ConfigMap or GRPC_TELEMETRY_ENDPOINT on the controller pod")
+		}
```
raballew: [LOW] The `FAILED_PRECONDITION` error message returned to authenticated exporter clients names the exact ConfigMap field (`telemetry.endpoint`) and environment variable (`GRPC_TELEMETRY_ENDPOINT`) used to configure the controller. Any caller with a valid bearer token learns the internal configuration schema. Consider a shorter, opaque message like `"telemetry is enabled but endpoint is not configured"` and keep the configuration hints in the controller startup logs.

controller/internal/service/endpoints.go
```diff
@@ -21,11 +24,36 @@ func routerEndpoint() string {
...
+// validateHostPort checks that s is a valid "host:port" with a non-empty host.
+func validateHostPort(s string) error {
+	host, _, err := net.SplitHostPort(s)
 	if err != nil {
+		return err
+	}
+	if host == "" {
+		return fmt.Errorf("endpoint %q has no host", s)
+	}
+	return nil
```
raballew: [MEDIUM] `validateHostPort` here and the equivalent inline validation in `config/config.go` implement the same `net.SplitHostPort` + empty-host check independently. A circular import prevents sharing code directly, but if the validation rule ever changes both sites need to be updated in sync. Extracting this into a shared `controller/internal/endpoint` package, or at minimum adding a cross-reference comment at each site, would make future maintenance safer.

controller/internal/service/endpoints.go
```diff
@@ -21,11 +24,36 @@ func routerEndpoint() string {
...
+	if ep == "" {
+		return ""
+	}
+	if err := validateHostPort(ep); err != nil {
+		ctrl.Log.WithName("telemetry").Error(err, "GRPC_TELEMETRY_ENDPOINT is not a valid host:port; ignoring",
+			"value", ep)
+		return ""
+	}
+	return ep
+}
```
raballew: When `GRPC_TELEMETRY_ENDPOINT` is set but malformed, this function logs and returns an empty string. Consider returning an error here and treating it as a fatal condition in `Start()`.

controller/internal/service/tls_credentials.go
```diff
@@ -0,0 +1,88 @@
...
+	cert, selfSignedPEM, err := LoadTLSCertificate(commonName, dnsnames, ipaddresses)
+	if err != nil {
+		return nil, "", err
+	}
+	creds := credentials.NewTLS(&tls.Config{
+		Certificates: []tls.Certificate{*cert},
+		MinVersion:   tls.VersionTLS12,
+	})
+	return creds, selfSignedPEM, nil
+}
```
raballew: [MEDIUM] Both `LoadTLSCertificate` and `LoadTLSCredentials` return `selfSignedPEM string` as their second value, but two of the three callers (`router_service.go` and `controller_service.go`) discard it with `_`. Only the telemetry path uses it to log the cert for operator pinning. Embedding a telemetry-specific concern in a shared API forces every caller to acknowledge an irrelevant return value. Consider removing `selfSignedPEM` from these signatures and providing a separate helper to extract the PEM from the returned `*tls.Certificate`.

controller/internal/config/config.go
```diff
@@ -43,6 +46,34 @@ func LoadRouterConfiguration(
...
+		host, _, err := net.SplitHostPort(ep)
+		if err != nil {
+			return nil, fmt.Errorf("telemetry endpoint %q is not a valid host:port: %w", ep, err)
+		}
+		if host == "" {
+			return nil, fmt.Errorf("telemetry endpoint %q has no host", ep)
+		}
+	}
+	return t, nil
+}
```
raballew: [MEDIUM] `resolveTelemetryConfig` mutates `t.Endpoint` on the caller's struct in place (line 64), then returns the same pointer. The returned pointer and the input are the same object. If the config struct is ever reused after `LoadConfiguration` returns, the env-resolved endpoint will appear as if it came from the ConfigMap rather than from the env var. Working on a copy inside the function makes this a pure transformation:

```go
resolved := *t
resolved.Endpoint = cmp.Or(os.Getenv("GRPC_TELEMETRY_ENDPOINT"), t.Endpoint)
return &resolved, nil
```

controller/internal/service/telemetry_service.go
```diff
@@ -197,16 +197,56 @@ func truncate(s string, n int) string {
...
+			ctrl.Log.WithName("telemetry").Error(sanErr, "failed to derive SAN from advertised endpoint; falling back to localhost",
+				"endpoint", advertised)
+			dnsnames = []string{"localhost"}
+		}
+	} else {
+		// No advertised endpoint configured — development/local mode.
+		dnsnames = []string{"localhost"}
+	}
+	return LoadTLSCredentials("jumpstarter telemetry", dnsnames, ipaddresses)
+}
```
raballew: [MEDIUM] `loadTLSCredentials` is declared as a `*TelemetryService` method, but the receiver `s` is never referenced inside the function body. It only calls package-level functions (`telemetryEndpoint`, `endpointToSAN`, `LoadTLSCredentials`). Declaring it as a method implies a dependency on receiver state that doesn't exist, which misleads readers and makes the function harder to test without constructing a `TelemetryService`. Converting it to a package-level function like `loadTelemetryTLSCredentials()` would make the intent explicit.

controller/internal/service/telemetry_service_test.go
```diff
@@ -82,11 +87,16 @@ func TestGetServiceEndpoints_RequiresAuthentication(t *testing.T) {
...
+// buildTelemetryEndpointsResponse mirrors the production GetServiceEndpoints
+// response-building logic without going through the auth gate.
+// cfg.Endpoint must already be fully resolved (as LoadConfiguration does);
+// an error is returned when the endpoint is empty, matching production.
+func buildTelemetryEndpointsResponse(cfg *config.Telemetry) (*pb.GetServiceEndpointsResponse, error) {
 	resp := &pb.GetServiceEndpointsResponse{}
 	if cfg != nil && cfg.Enabled {
+		if cfg.Endpoint == "" {
+			return nil, status.Error(codes.FailedPrecondition, "telemetry is enabled but no endpoint is configured")
+		}
```
raballew: [MEDIUM] The empty-endpoint guard added here (lines 97-99) mirrors the same guard in the real `GetServiceEndpoints` handler in `controller_service.go`. When the production error message or status code changes, this helper will silently diverge. `TestGetServiceEndpoints_EnabledNoEndpoint_FailedPrecondition` already exercises the real handler for the empty-endpoint case, making the duplication in this helper redundant. Consider migrating the tests that use `buildTelemetryEndpointsResponse` to go through the real handler path and removing this helper.

controller/internal/service/telemetry_service.go
raballew: [HIGH] The server is now created with mandatory TLS credentials, but there is no test that verifies plaintext clients are actually rejected. Without an end-to-end test that dials with `insecure.NewCredentials()` and confirms the call fails with a TLS handshake error, TLS enforcement is only validated indirectly. A companion test connecting with `credentials.NewTLS(&tls.Config{InsecureSkipVerify: true})` would confirm TLS is active. The project pattern of starting a real server and client for end-to-end validation applies here.

controller/internal/config/config.go
```diff
@@ -43,6 +46,34 @@ func LoadRouterConfiguration(
...
 }
 
+// resolveTelemetryConfig validates and resolves the telemetry endpoint for a
+// Telemetry config block. It prefers the explicit ConfigMap endpoint and falls
+// back to GRPC_TELEMETRY_ENDPOINT, matching the behaviour previously inlined in
+// LoadConfiguration. Returns nil when t is nil or disabled.
+func resolveTelemetryConfig(t *Telemetry) (*Telemetry, error) {
+	if t == nil || !t.Enabled {
+		return nil, nil
+	}
```
raballew: [MEDIUM] The early-return branches at lines 54-55 (`t == nil` and `!t.Enabled`, both returning `nil, nil`) have no test coverage. `TestTelemetryEndpointResolution` only constructs configs with `Enabled: true` and a non-nil struct. The nil case represents a ConfigMap with no `telemetry` section, and `Enabled: false` is a common production state. Both paths should have explicit test cases.

controller/internal/service/tls_credentials.go
```diff
@@ -0,0 +1,88 @@
...
+	cert, err := NewSelfSignedCertificate(commonName, dnsnames, ipaddresses)
+	if err != nil {
+		return nil, "", err
+	}
+	selfSignedPEM := string(pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: cert.Certificate[0],
+	}))
+	return cert, selfSignedPEM, nil
+}
```
raballew: [MEDIUM] `LoadTLSCertificate` is a public function used directly by `controller_service.go`, but no test file calls it by name. All coverage flows through the private `loadTLSCredentials` method, which wraps `LoadTLSCredentials`, which wraps `LoadTLSCertificate`. If the intermediate function is refactored to bypass `LoadTLSCertificate`, the public function's contract becomes silently untested. Adding a `TestLoadTLSCertificate_*` family covering the self-signed path, the operator-cert path, and the partial-env-var path would make the public API's contract explicit and cover the controller's direct call path.

controller/cmd/telemetry/main.go
```diff
@@ -18,6 +18,18 @@ limitations under the License.
...
+// operator-mounted cert/key (e.g. from a cert-manager Secret); when absent a
+// self-signed certificate is generated. The self-signed cert PEM is logged at
+// startup — copy it into the controller ConfigMap's telemetry.certificate field
+// so exporters can verify the TLS connection.
+//
+// Endpoint: GRPC_TELEMETRY_ENDPOINT must be set on BOTH this pod and the controller
+// pod to the same value (e.g. "jumpstarter-telemetry.jumpstarter.svc:9093").
+// The telemetry service uses it to generate the correct SAN in the self-signed
+// certificate; the controller uses it to advertise the address to exporters via
+// GetServiceEndpoints. A mismatch causes TLS hostname verification failures.
```
raballew, line 31: The three env vars introduced by this PR (`EXTERNAL_CERT_PEM`, `EXTERNAL_KEY_PEM`, `GRPC_TELEMETRY_ENDPOINT`) are documented only in these source-level package comments. Operators configuring a deployment will not find them in `controller/README.md` or any configuration reference.

raballew (changes requested):

controller/cmd/telemetry/main.go
bkhizgiy: You're right, that's a good point. I think it would be better to add this at the operator level, where it's more discoverable for users actually configuring deployments. I'll include the changes in the operator related PR #997 to keep this one more focused.

controller/internal/service/telemetry_service.go
bkhizgiy: done.

controller/internal/service/telemetry_service_test.go
```diff
@@ -100,32 +110,41 @@ func buildTelemetryEndpointsResponse(cfg *config.Telemetry) *pb.GetServiceEndpoi
...
 		Logging: config.TelemetryLogging{
 			Filter: config.TelemetryLoggingFilter{MinSeverity: "warning"},
 		},
 	})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
 
 	if len(resp.TelemetryEndpoints) != 1 {
 		t.Fatalf("expected 1 telemetry endpoint, got %d", len(resp.TelemetryEndpoints))
```
raballew, line 133: The success-path tests for `GetServiceEndpoints` (`TestGetServiceEndpoints_NilConfig_ReturnsEmptyList`, `TestGetServiceEndpoints_DisabledConfig_ReturnsEmptyList`, `TestGetServiceEndpoints_WithEndpoint_ReturnsEndpoint`, `TestGetServiceEndpoints_DefaultsMinSeverityToInfo`) all call the test-local `buildTelemetryEndpointsResponse` helper rather than the real `svc.GetServiceEndpoints` handler. Only the error path goes through actual production code.

controller/internal/config/config.go
```diff
@@ -43,6 +46,34 @@ func LoadRouterConfiguration(
...
+		host, _, err := net.SplitHostPort(ep)
+		if err != nil {
+			return nil, fmt.Errorf("telemetry endpoint %q is not a valid host:port: %w", ep, err)
+		}
+		if host == "" {
+			return nil, fmt.Errorf("telemetry endpoint %q has no host", ep)
+		}
+	}
+	return t, nil
+}
```
raballew: [MEDIUM] When telemetry is enabled but neither `GRPC_TELEMETRY_ENDPOINT` nor the ConfigMap `telemetry.endpoint` field is set, `resolveTelemetryConfig` returns successfully with `t.Endpoint == ""` and emits no log. The misconfiguration only surfaces later when an exporter calls `GetServiceEndpoints` and receives `FailedPrecondition`. The config package currently has no logger at all.

Adding a structured warning after endpoint resolution, when `t.Enabled && t.Endpoint == ""`, would surface the misconfiguration at startup rather than requiring a failed gRPC call to discover it. For example:
```go
ctrl.Log.WithName("config").Info("telemetry enabled but no endpoint resolved; GetServiceEndpoints will return FailedPrecondition until configured")
```

controller/internal/config/types_test.go
```diff
@@ -276,6 +276,78 @@ func TestDeprecatedLabelsOmitEmpty(t *testing.T) {
...
+			var gotEndpoint string
+			if resolved != nil {
+				gotEndpoint = resolved.Endpoint
+			}
+			if gotEndpoint != tt.wantEndpoint {
+				t.Errorf("resolved.Endpoint = %q, want %q", gotEndpoint, tt.wantEndpoint)
+			}
+		})
+	}
+}
```
raballew, line 368: `TestTelemetryEndpointResolution` always constructs `&Telemetry{Enabled: true, ...}`. The two short-circuit return paths in `resolveTelemetryConfig` (nil input and `Enabled=false`) are never directly exercised. 

Adding two table entries would cover these branches:
- `(*Telemetry)(nil)` expecting `(nil, nil)`
- `&Telemetry{Enabled: false, Endpoint: "telemetry:9093"}` also expecting `(nil, nil)`

controller/internal/service/telemetry_service_test.go
```diff
@@ -156,6 +178,282 @@ func TestGetServiceEndpoints_DefaultsMinSeverityToInfo(t *testing.T) {
...
+	}
+	if selfSignedPEM == "" {
+		t.Error("expected non-empty selfSignedPEM when no external cert is configured")
+	}
+	// Must be parseable PEM.
+	block, _ := pem.Decode([]byte(selfSignedPEM))
+	if block == nil {
+		t.Errorf("selfSignedPEM is not valid PEM: %s", selfSignedPEM)
+	}
+}
```
raballew: [LOW] `TestTelemetryService_LoadTLSCredentials_SelfSigned` calls `t.Setenv` for `EXTERNAL_CERT_PEM` and `EXTERNAL_KEY_PEM` but not for `GRPC_TELEMETRY_ENDPOINT`. If the test suite runs with `GRPC_TELEMETRY_ENDPOINT` set in the environment (for example in CI), `loadTLSCredentials` will use it as the SAN instead of `"localhost"`. The test only checks that `selfSignedPEM != ""` and is parseable, so it passes silently with the wrong SAN.

`TestTelemetryService_LoadTLSCredentials_SelfSignedFallsBackToLocalhostWhenNoEndpoint` already handles this correctly by isolating the env var. Adding `t.Setenv("GRPC_TELEMETRY_ENDPOINT", "")` at the top of this test would close the isolation gap.

raballew (approved):
no issues in the production code but test can be improved. merge when you think you are ready.

PR #978, opened 2026-08-07
feat: trigger rolling restart on TLS certificate renewal

### LINKED ISSUE #852: Jumpstarter does not reload TLS certificates after they've been renewed
Hi,

Following our discussion on #286, I have deployed Jumpstarter with the operator.
The reverse proxy is configured in passthrough mode, cert-manager creates TLS certificates and Jumpstarter loads them. All run smoothly till certificates get renewed.

cert-manager did the job, secrets contain the new certificates but Jumpstarter does not load them.
For now, I have deleted the pods and new certificates have been loaded.

Not sure what's the best way to handle this... Could Jumpstarter watch for secret changes?

bennyz:
Hash TLS secret data into pod template annotations so that cert-manager renewals automatically trigger rolling restarts, eliminating the need for manual pod restarts.

Extends the existing configmap-sha256 annotation pattern to TLS secrets and expands the secret field indexer to watch TLS cert secrets alongside JWT CA secrets.

Fixes #852

### DIFF
--- a/controller/deploy/operator/internal/controller/jumpstarter/hash_annotations_test.go
+++ b/controller/deploy/operator/internal/controller/jumpstarter/hash_annotations_test.go
@@ -1,0 +1,431 @@
+/*
+Copyright 2025. The Jumpstarter Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package jumpstarter
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	operatorv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/deploy/operator/api/v1alpha1"
+)
+
+var _ = Describe("secretDataHash", func() {
+	It("should produce a deterministic hash", func() {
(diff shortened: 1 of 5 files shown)


bennyz:
@coderabbitai review

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -321,7 +349,12 @@ func (r *JumpstarterReconciler) reconcileRouterDeployment(ctx context.Context, j
 
 	// Create one deployment per replica
 	for i := int32(0); i < jumpstarter.Spec.Routers.Replicas; i++ {
-		desiredDeployment := r.createRouterDeployment(jumpstarter, i)
+		routerTLSHash, err := r.getRouterTLSSecretHash(ctx, jumpstarter, i)
```
bkhizgiy: When `Routers.GRPC.TLS.CertSecret` is set, all replicas use the same secret, so we're fetching and hashing it multiple times. could we handle the shared secret case outside the loop?

controller/deploy/operator/internal/controller/jumpstarter/hash_annotations_test.go
```diff
@@ -0,0 +1,271 @@
...
+							CertSecret: "controller-tls-test",
+						},
+					},
+				},
+			},
+		}
+		hash, err := r.getControllerTLSSecretHash(ctx, js)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(hash).To(HaveLen(64))
+	})
```
bkhizgiy, line 310: We already have an integration test for `getControllerTLSSecretHash`. Could we add a similar one for `getRouterTLSSecretHash`?

bkhizgiy (commented):
looks good overall, just small comments

controller/deploy/operator/internal/controller/jumpstarter/hash_annotations_test.go
bennyz: added

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
bennyz: added

bkhizgiy (approved):

PR #979, opened 2026-08-07
fix: catch Exception in _flush_lines

bennyz:
In case there are unexepected exceptions, to avoid hitting flakes https://github.com/jumpstarter-dev/jumpstarter/actions/runs/31179469197/attempts/1

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/exporter/hooks.py
+++ b/python/packages/jumpstarter/jumpstarter/exporter/hooks.py
@@ -427,6 +427,11 @@ class HookExecutor:
                                 # PTY closed or read error
                                 logger.debug("read_pty_output: OSError in loop: %s", e)
                                 break
+
+                            except Exception as e:
+                                logger.debug("read_pty_output: unexpected error in loop: %s", e)
+                                break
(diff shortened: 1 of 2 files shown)


python/packages/jumpstarter/jumpstarter/exporter/hooks.py
```diff
@@ -427,6 +427,11 @@ async def read_pty_output() -> None:  # noqa: C901
                                 # PTY closed or read error
                                 logger.debug("read_pty_output: OSError in loop: %s", e)
                                 break
+
+                            except Exception as e:
+                                logger.debug("read_pty_output: unexpected error in loop: %s", e)
```
bkhizgiy, line 432: don't we prefer to surface the error anyway in case it glitched? maybe worth changing to `l logger.warning` with `exc_info=True`?

bennyz: not sure, it's about expected edge case failures (like OSError above) so raising level would generate noise

bkhizgiy (approved):

PR #980, opened 2026-08-07
feat: netsim driver

bennyz:
- Adds jumpstarter-driver-netsim — a new driver for controlling Android netsim virtual radios (Bluetooth Classic, BLE, WiFi, UWB) via the netsim REST API
- Exposes device listing, radio toggle, raw patching, device reset, and pcap capture management (start/stop/download)
- Includes full CLI via Click with subcommands: devices, device, radio, patch, reset, capture {list,start,stop,get}, status
- Capture toggle uses netsim CLI binary as workaround for broken REST PATCH in netsim ≤0.3.100

### DIFF
--- a/python/packages/jumpstarter-driver-netsim/jumpstarter_driver_netsim/driver.py
+++ b/python/packages/jumpstarter-driver-netsim/jumpstarter_driver_netsim/driver.py
@@ -1,0 +1,274 @@
+import json
+import subprocess
+from base64 import b64encode
+from collections.abc import AsyncGenerator
+from dataclasses import dataclass
+
+import requests
+
+from jumpstarter.driver import Driver, export
+
+
+class NetsimError(Exception):
+    """Raised when a netsim API call fails."""
+
+
+VALID_STATES = frozenset(("true", "1", "on", "false", "0", "off"))
+ON_STATES = frozenset(("true", "1", "on"))
+RADIO_TYPES = ("bt_classic", "ble", "wifi", "uwb")
+
+
+def _parse_bool(value: str, param_name: str) -> bool:
+    v = value.lower()
+    if v not in VALID_STATES:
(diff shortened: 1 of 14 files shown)


python/packages/jumpstarter-driver-netsim/jumpstarter_driver_netsim/driver.py
```diff
@@ -0,0 +1,268 @@
...
+        except requests.Timeout as e:
+            raise NetsimError(f"{method} {path} timed out after {timeout}s") from e
+        except requests.HTTPError as e:
+            raise NetsimError(f"{method} {path} failed: {e}") from e
+
+    def _request_bytes(self, path: str, timeout: int = 30) -> bytes:
+        try:
+            r = requests.get(f"{self._base_url}{path}", timeout=timeout)
+            r.raise_for_status()
+            return r.content
```
kirkbrauer, line 72: This is a good point, I had similar issues with the MITM proxy where payloads were exceeding limits, so chunked transfer was needed.

kirkbrauer (commented):
I think this looks good, but I do wonder if it would be better to actually tunnel the HTTP API over Jumpstarter? I think there are pros/cons of each approach. The main I pro I see with this current approach is that we can make the client thin and keep all heavy operations on the exporter side which is nice, but it does increase complexity as well.

bennyz:
> I think this looks good, but I do wonder if it would be better to actually tunnel the HTTP API over Jumpstarter? I think there are pros/cons of each approach. The main I pro I see with this current approach is that we can make the client thin and keep all heavy operations on the exporter side which is nice, but it does increase complexity as well.

I looked around, and considering we have some workarounds in place (CLI shell, numeric ids), it would complicate the tests. Another option here is would be to nest the TcpNetwork driver and provide access to not yet exposed endpoints

kirkbrauer (approved):

PR #981, opened 2026-08-07
feat: generic video driver

bennyz:
connecting a camera to a device like ESP32, would not show up as a /dev/ on the exporter, so we need to stream over TCP. Rather than copying a lot of code from ustreamer, create a generic video driver to make code sharing easier

### DIFF
--- a/python/packages/jumpstarter-driver-ustreamer/jumpstarter_driver_ustreamer/client_test.py
+++ b/python/packages/jumpstarter-driver-ustreamer/jumpstarter_driver_ustreamer/client_test.py
@@ -1,21 +1,10 @@
-import base64
-import contextlib
-import io
-from types import SimpleNamespace
-from unittest.mock import AsyncMock, MagicMock, patch
+from unittest.mock import MagicMock
 
-import anyio
-import pytest
 from click.testing import CliRunner
-from PIL import Image
+from jumpstarter_driver_video.client import VideoClient
+from jumpstarter_driver_video.common import VideoState
 
-from jumpstarter_driver_ustreamer.client import (
-    LANDING_PAGE,
-    UStreamerClient,
-    _parse_content_type,
-    _proxy_mjpeg_stream,
-    _run_server,
-)
+from jumpstarter_driver_ustreamer.client import UStreamerClient
 from jumpstarter_driver_ustreamer.common import UStreamerState
 
 
@@ -44,73 +33,6 @@ def _make_state():
     )
 
 
-def _make_jpeg_bytes():
-    buffer = io.BytesIO()
-    Image.new("RGB", (2, 1), color="red").save(buffer, format="JPEG")
-    return buffer.getvalue()
-
-
-def _get_route_handler(app, path):
-    for resource in app.router.resources():
-        if getattr(resource, "canonical", None) == path:
-            return next(iter(resource)).handler
-    raise AssertionError(f"Route {path} was not registered")
-
-
-class _FakeStreamContext:
-    def __init__(self, tunnel):
-        self.tunnel = tunnel
-
-    async def __aenter__(self):
-        return self.tunnel
-
-    async def __aexit__(self, exc_type, exc, tb):
-        return False
-
-
-class _FakeTunnel:
-    def __init__(self, chunks):
-        self._chunks = iter(chunks)
-        self.sent = []
-
-    async def send(self, data):
-        self.sent.append(data)
-
-    async def receive(self):
-        chunk = next(self._chunks)
-        if isinstance(chunk, BaseException):
-            raise chunk
-        return chunk
-
-
-class _FakeStreamResponse:
-    def __init__(self):
-        self.content_type = None
-        self.prepared_request = None
-        self.writes = []
-
-    async def prepare(self, request):
-        self.prepared_request = request
-
-    async def write(self, data):
-        self.writes.append(data)
-
-
-@pytest.fixture
-def anyio_backend():
-    return "asyncio"
-
-
-def test_parse_content_type_returns_header_value():
-    header_bytes = b"HTTP/1.1 200 OK\r\nContent-Type: image/jpeg\r\nX-Test: 1\r\n"
-
-    assert _parse_content_type(header_bytes) == "image/jpeg"
-
-
-def test_parse_content_type_falls_back_to_mjpeg_default():
-    assert _parse_content_type(b"HTTP/1.1 200 OK\r\nX-Test: 1\r\n") == "multipart/x-mixed-replace; boundary=--"
-
-
 def test_state_returns_validated_model():
     client = _make_client()
     client.call = MagicMock(return_value=_make_state().model_dump(mode="python"))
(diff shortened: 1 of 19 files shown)


python/packages/jumpstarter-driver-video/jumpstarter_driver_video/client.py
```diff
@@ -0,0 +1,236 @@
...
+        if buf:
+            yield buf
+        while True:
+            yield await tunnel.receive()
+
+    while True:
+        while b"\r\n" not in buf:
+            buf += await tunnel.receive()
+        size_line, _, buf = buf.partition(b"\r\n")
+        size = int(size_line.split(b";")[0].strip(), 16)
```
mangelajo: **Bug: `ValueError` on malformed chunk-size line**

If the source sends a non-hex chunk-size line (e.g. corrupted stream), `int(..., 16)` raises `ValueError` which propagates as an unhandled 500 to the client. Should catch and convert:

```python
try:
    size = int(size_line.split(b";")[0].strip(), 16)
except ValueError:
    raise web.HTTPBadGateway(reason="invalid chunk size from upstream") from None
```

python/packages/jumpstarter-driver-video/jumpstarter_driver_video/client.py
```diff
@@ -0,0 +1,236 @@
...
+        if size == 0:
+            return
+        while len(buf) < size + 2:
+            buf += await tunnel.receive()
+        yield buf[:size]
+        buf = buf[size + 2 :]  # drop the CRLF terminating the chunk
```
mangelajo, line 81: **Bug: unbounded memory growth in chunked decoder**

A buggy or malicious source can declare an enormous chunk size (e.g. `FFFFFFFF\r\n`), causing the loop at line 73 to buffer ~4GB before yielding anything. The header-reading loop above already has `_MAX_HEADER_SIZE` protection -- the chunked decoder should have a similar cap on chunk size. Something like 16MB would be generous for MJPEG frames.

python/packages/jumpstarter-driver-ustreamer/jumpstarter_driver_ustreamer/client_test.py
```diff
@@ -1,21 +1,9 @@
...
 
-from jumpstarter_driver_ustreamer.client import (
-    LANDING_PAGE,
-    UStreamerClient,
-    _parse_content_type,
-    _proxy_mjpeg_stream,
-    _run_server,
-)
+from jumpstarter_driver_ustreamer.client import UStreamerClient
 from jumpstarter_driver_ustreamer.common import UStreamerState
```
mangelajo, line 10: **Testing gap: deleted ustreamer client tests not replaced by inheritance coverage**

The old test file had tests for `snapshot`, `snapshot_bytes`, the `snapshot` CLI command, and the `stream` CLI command -- all deleted because they moved to `VideoClient`. But there's no test here verifying that `UStreamerClient` actually inherits and correctly delegates these through the inheritance chain.

If someone accidentally overrides or breaks one of these methods on `UStreamerClient`, there's no test to catch it. A simple smoke test would provide a safety net:

```python
def test_ustreamer_inherits_video_client_methods():
    """UStreamerClient must expose the full VideoClient API."""
    assert issubclass(UStreamerClient, VideoClient)
    # snapshot/snapshot_bytes/stream come from VideoClient
    assert UStreamerClient.snapshot is VideoClient.snapshot
    assert UStreamerClient.snapshot_bytes is VideoClient.snapshot_bytes
```

python/packages/jumpstarter-driver-video/jumpstarter_driver_video/driver_test.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+FRAME_A = _jpeg_bytes("red")
+FRAME_B = _jpeg_bytes("blue")
+BOUNDARY = "frameboundary"
+
+
+class FakeCameraHandler(BaseHTTPRequestHandler):
+    def do_GET(self):
+        if self.path == "/snapshot.jpg":
```
mangelajo, line 26: **Testing gap: no integration test for chunked transfer encoding**

The `FakeCameraHandler` uses Python's `BaseHTTPRequestHandler` which sends responses with `Content-Length` (no chunking). There's a unit test in `client_test.py` for the chunked decoder using mocked tunnels, but no integration test connecting `HttpVideo` to a server that actually uses chunked transfer encoding.

Since the ESP-IDF chunked MJPEG use case is explicitly called out in the PR description and the `_iter_body` chunked decoder is new non-trivial code, it would be valuable to add a `FakeChunkedCameraHandler` that sends the MJPEG stream with `Transfer-Encoding: chunked` framing.

python/packages/jumpstarter-driver-video/jumpstarter_driver_video/client.py
```diff
@@ -0,0 +1,236 @@
...
+
+
+_MAX_HEADER_SIZE = 16 * 1024
+
+
+async def proxy_mjpeg_stream(client, request, path):
+    """Proxy the source's native MJPEG stream through the jumpstarter tunnel."""
+    async with client.stream_async("connect") as tunnel:
+        await tunnel.send(f"GET {path} HTTP/1.1\r\nHost: localhost\r\n\r\n".encode("ascii"))
+
```
mangelajo, line 134: **Minor: `Host: localhost` header sent to remote cameras**

The proxy sends `Host: localhost` regardless of the actual camera hostname. For `UStreamer` (which connects over a unix socket) this is fine and matches the old behavior. But for `HttpVideo`, the `connect()` stream is a raw TCP connection to the camera's real host (e.g. `192.168.1.50`), and a virtual-host-aware HTTP server could reject or misroute the request.

Consider having `proxy_mjpeg_stream` accept an optional `host` parameter, or having `stream_path()` return enough context for the client to construct a correct `Host` header.

Most simple cameras / ESP-IDF servers won't care, so this is low priority.

python/packages/jumpstarter-driver-video/jumpstarter_driver_video/client.py
mangelajo: matches coderabbit :)

python/packages/jumpstarter-driver-video/jumpstarter_driver_video/client_test.py
```diff
@@ -0,0 +1,299 @@
...
+
+            request = object()
+            response = await stream_handler(request)
+            assert response is proxied_response
+            mock_proxy.assert_awaited_once_with(client, request, "/video.mjpg")
+
+        anyio.run(exercise_routes)
+
+    client.call_async.assert_awaited_once_with("snapshot")
+
```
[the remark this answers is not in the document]
mangelajo: @bennyz already added a mockup server for testing, right?

mangelajo (changes requested):
A few nits, but overall it's great work, thanks benny :)

bennyz:
@coderabbitai review

mangelajo (approved):

PR #992, opened 2026-08-10
fix(grpc): separate DNS resolution timeout from connect timeout

bennyz:
fail_after(timeout) wrapped both resolution and connection, so a slow resolver consumed the whole budget and reported "Timeout connecting to <host>:<port>" with an empty cause list - blaming the server for a name that was never resolved.

Give resolution its own budget and its own message, and name the resolved IPs in the connect error.

### DIFF
--- a/python/packages/jumpstarter/jumpstarter/common/grpc.py
+++ b/python/packages/jumpstarter/jumpstarter/common/grpc.py
@@ -50,6 +50,9 @@ async def _ssl_channel_credentials_insecure(target: str, timeout: float) -> grpc
 
     Tries to connect to all resolved IPs in parallel and returns credentials
     from the first successful connection.
+
+    ``timeout`` applies independently to DNS resolution and connection, so
+    worst-case wall time is approximately ``2 * timeout``.
     """
     try:
         parsed = urlparse(f"//{target}")
--- a/python/packages/jumpstarter/jumpstarter/common/grpc.py
+++ b/python/packages/jumpstarter/jumpstarter/common/grpc.py
@@ -57,24 +57,31 @@ async def _ssl_channel_credentials_insecure(target: str, timeout: float) -> grpc
     except ValueError as e:
         raise ConfigurationError(f"Failed parsing {target}") from e
 
+    ssl_context = ssl.create_default_context()
+    ssl_context.check_hostname = False
+    ssl_context.verify_mode = ssl.CERT_NONE
+
+    # Resolve all IP addresses for the hostname.
+    #
+    # Resolution gets its own budget, separate from the connect budget below. A
+    # slow or rate-limited resolver would otherwise consume the whole timeout and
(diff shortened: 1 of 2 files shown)


python/packages/jumpstarter/jumpstarter/common/grpc.py
```diff
@@ -57,24 +57,31 @@ async def _ssl_channel_credentials_insecure(target: str, timeout: float) -> grpc
...
+        raise ConnectionError(f"Failed resolving {parsed.hostname}") from e
+    except TimeoutError as e:
+        raise ConnectionError(f"Timeout resolving {parsed.hostname} after {timeout}s") from e
+
+    # Log resolved IPs
+    resolved_ips = [sockaddr[0] for _, _, _, _, sockaddr in addr_info]
+    logger.debug(f"Resolved {parsed.hostname} to {len(resolved_ips)} IP(s): {', '.join(resolved_ips)}")
 
+    try:
+        with fail_after(timeout):
```
bkhizgiy, line 87: Both `fail_after` (here and line 72) calls use the full `timeout`, so `timeout=5` can now take up to 10s total instead of 5s. Could we track the time spent resolving and pass the remaining timeout to the connect phase?

python/packages/jumpstarter/jumpstarter/common/grpc.py
```diff
@@ -57,24 +57,31 @@ async def _ssl_channel_credentials_insecure(target: str, timeout: float) -> grpc
...
-                f"Resolved {parsed.hostname} to {len(resolved_ips)} IP(s): {', '.join(resolved_ips)}"
-            )
+            addr_info = await loop.getaddrinfo(parsed.hostname, port, family=socket.AF_UNSPEC, type=socket.SOCK_STREAM)
+    except socket.gaierror as e:
+        raise ConnectionError(f"Failed resolving {parsed.hostname}") from e
+    except TimeoutError as e:
+        raise ConnectionError(f"Timeout resolving {parsed.hostname} after {timeout}s") from e
+
+    # Log resolved IPs
+    resolved_ips = [sockaddr[0] for _, _, _, _, sockaddr in addr_info]
```
bkhizgiy, line 83: If `getaddrinfo` returns an empty list, this falls through returning `None`. Could we handle it explicitly and raise a `ConnectionError`?

python/packages/jumpstarter/jumpstarter/common/grpc.py
bennyz: this would restore the original bug, i added a comment about worst case scenario instead

python/packages/jumpstarter/jumpstarter/common/grpc.py
bennyz: i am not sure this can happen in practice, getaddrinfo raises gaierror

bkhizgiy (approved):

PR #993, opened 2026-08-10
perf(ci): parallelize e2e operator build prerequisites

bennyz:
* Parallelize independent Make prerequisites (docker-build, build-operator, cluster) with -j so kind cluster creation overlaps with Go compilation
* Add CI-optimized build path that compiles Go on the runner (with setup-go module cache) and packages binaries into minimal runtime-only containers, eliminating the ~1.5GB go-toolset image pull and in-container module downloads

### DIFF
--- a/controller/Containerfile.prebuilt
+++ b/controller/Containerfile.prebuilt
@@ -1,0 +1,13 @@
+# CI-only runtime image for host-compiled Go binaries.
+#
+# BIN selects which binary becomes the default entrypoint.
+# COPY ${BIN} /entrypoint works because COPY supports ARG expansion,
+# while ENTRYPOINT does not (and ubi-micro has no shell for shell-form).
+FROM registry.access.redhat.com/ubi9/ubi-micro:9.8-1784702951@sha256:b1e86b97028b8fcfb6d85f997c39e6b6b67496163ef8d80d243220a4918e8bef
+ARG BIN
+WORKDIR /
+COPY . .
+# ENTRYPOINT can't expand ARG; fixed path lets one file serve all three images
+COPY ${BIN} /entrypoint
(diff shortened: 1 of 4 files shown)


controller/Containerfile.prebuilt
```diff
@@ -0,0 +1,12 @@
+# CI-only runtime image for host-compiled Go binaries.
+#
+# BIN selects which binary becomes the default entrypoint.
+# COPY ${BIN} /entrypoint works because COPY supports ARG expansion,
+# while ENTRYPOINT does not (and ubi-micro has no shell for shell-form).
+FROM registry.access.redhat.com/ubi9/ubi-micro:9.8-1784702951@sha256:b1e86b97028b8fcfb6d85f997c39e6b6b67496163ef8d80d243220a4918e8bef
+ARG BIN
+WORKDIR /
+COPY . .
+COPY ${BIN} /entrypoint
```
bkhizgiy, line 11: `COPY . . ` copies everything from the build context, including the binaries, since `${BIN}` is copied again as `/entrypoint`, we end up shipping duplicate `manager/router` binaries that aren't used.
So it probably can be simplified to just `COPY ${BIN} /entrypoint`

bkhizgiy (commented):
looks goo overall, just a small comment

controller/Containerfile.prebuilt
bennyz: coderabbit commented on this too, 
the controller image has manager and router, and we have a third exporter-set-controller

so in order to get router too we copy it in COPY . . 
at the expense of an extra 40-50 throwaway MBs

I'll add a comment to clarify

bkhizgiy (approved):

PR #995, opened 2026-08-10
test(e2e): improve reliability, parallelism, and CI caching

bennyz:
* Pin nip.io hostnames in /etc/hosts so DNS flakes don't burn the connect budget
* Fix kubectl query helpers to not treat stderr as output (was ending Eventually polls early)
* Add retry, continue-on-failure, and broader hook failure matching to reduce spurious failures
* Cut fixture overhead: single kubectl apply for pagination, reuse hooks exporter when config matches
* Enable ginkgo --procs=2: graceful SIGTERM lifecycle, argv-scoped orphan sweep, --client instead of config client use
* Cache Go modules, controller tools, and load kind images concurrently

### DIFF
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -14,6 +14,10 @@ env:
   # flakes in the ginkgo summary, so this hides nothing; it only stops a single
   # infrastructure hiccup from failing an unrelated PR.
   E2E_FLAKE_ATTEMPTS: "2"
+  # Run the non-Serial containers two at a time. The runner has 4 CPUs and the
+  # suite spends most of its time waiting on the cluster, so a second process
+  # overlaps the two long containers (hooks, core) rather than competing for CPU.
+  E2E_PROCS: "2"
 
 jobs:
   changes:
--- a/.github/workflows/e2e.yaml
+++ b/.github/workflows/e2e.yaml
@@ -76,6 +76,7 @@ jobs:
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
           go-version-file: .go-version
+          cache-dependency-path: "**/go.sum"
 
       - name: Cache controller image
         id: cache
@@ -115,6 +116,7 @@ jobs:
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
           go-version-file: .go-version
+          cache-dependency-path: "**/go.sum"
 
       - name: Cache operator artifacts
         id: cache
@@ -159,6 +161,7 @@ jobs:
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
           go-version-file: .go-version
+          cache-dependency-path: "**/go.sum"
 
       - name: Cache exporterset-controller image
         id: cache
@@ -318,6 +321,7 @@ jobs:
         uses: actions/setup-go@b7ad1dad31e06c5925ef5d2fc7ad053ef454303e # v7
         with:
           go-version-file: .go-version
+          cache-dependency-path: "**/go.sum"
 
       - name: Load e2e artifacts
         uses: ./.github/actions/load-e2e-artifacts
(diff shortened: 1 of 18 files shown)


.github/workflows/e2e.yaml
```diff
@@ -10,6 +10,14 @@ permissions:
 
 env:
   CONTAINER_TOOL: docker
+  # Retry a failed spec once before failing the job. Retries are reported as
+  # flakes in the ginkgo summary, so this hides nothing; it only stops a single
+  # infrastructure hiccup from failing an unrelated PR.
+  E2E_FLAKE_ATTEMPTS: "2"
```
bkhizgiy, line 16: nice addition! :)

.github/workflows/e2e.yaml
```diff
@@ -327,6 +339,16 @@ jobs:
...
           done
 
+      - name: Cache controller tools
+        uses: actions/cache@55cc8345863c7cc4c66a329aec7e433d2d1c52a9 # v6
+        with:
+          path: controller/bin
+          key: controller-tools-${{ matrix.arch }}-kind${{ env.KIND_VERSION }}-kustomize${{ env.KUSTOMIZE_VERSION }}-grpcurl${{ env.GRPCURL_VERSION }}
+        env:
+          KIND_VERSION: v0.27.0
+          KUSTOMIZE_VERSION: v5.4.1
```
bkhizgiy: can we somehow define this once in the file?, we have some duplication of it and it will be better in case the version changes to do it once insted track all places.

e2e/setup-e2e.sh
```diff
@@ -275,6 +275,35 @@ deploy_controller() {
...
+    local basedomain="$1"
+    local ip
+
+    ip=$(echo "${basedomain}" | sed -nE 's/^.*\.([0-9]+\.[0-9]+\.[0-9]+\.[0-9]+)\.nip\.io$/\1/p')
+    if [ -z "${ip}" ]; then
+        log_info "baseDomain ${basedomain} is not a nip.io name, leaving DNS resolution alone"
+        return 0
+    fi
+
+    if grep -q "grpc.${basedomain}" /etc/hosts 2>/dev/null; then
```
bkhizgiy: Since we're matching a literal hostname here, mayve we should we use `grep -Fq` instead? -F avoids treating characters like `.`  as regex wildcards and makes the match more precise.

e2e/lib/common.sh
```diff
@@ -79,7 +79,23 @@ run_ginkgo() {
         timeout="60m"
     fi
 
-    local flags=(-v --show-node-events --trace --timeout "${timeout}")
+    # Retry a failed spec instead of failing the whole suite. The e2e suite talks
+    # to a real cluster over the network, so a spec can fail for reasons that have
+    # nothing to do with the code under test (a slow DNS answer, a pod scheduled
+    # late, a router connection dropped). A retried spec is still reported as
+    # flaky in the summary, so genuine instability stays visible.
+    local flake_attempts="${E2E_FLAKE_ATTEMPTS:-1}"
```
bkhizgiy, line 87: I think we should set it to 2, since ginkgo treats it as the total number of runs rather than the number of retries, otherwise we'll end up with the same behavior as before.

e2e/test/utils.go
```diff
@@ -642,36 +708,139 @@ func AnyPIDAlive(pids []int) bool {
...
+	// Reap in the background. A zombie still answers signal 0, so a child that
+	// nobody waited on would look alive for the whole grace period. Processes
+	// started through StartExporter* already have a Wait goroutine; the extra
+	// waiter just loses the race and gets an error, which is harmless.
 	for _, pid := range pids {
-		proc, err := os.FindProcess(pid)
-		if err != nil {
-			continue
+		if proc, err := os.FindProcess(pid); err == nil {
+			go func() { _, _ = proc.Wait() }()
```
bkhizgiy, line 762: Do we need another `Wait()` here? It looks like `StartExporter*` already waits for these processes, so we might end up with two goroutines trying to reap the same process. Could we rely on the existing one instead?

e2e/lib/common.sh
bennyz: we have it set to 2 in the workflow yaml, it will default to 1 only for local testing and such

e2e/test/utils.go
bennyz: it's addressed in the comment on L#756

e2e/setup-e2e.sh
bennyz: added

.github/workflows/e2e.yaml
bennyz: added Makefile extraction

bkhizgiy (approved):

PR #997, opened 2026-08-12
feat(operator): deploy and manage jumpstarter-telemetry in the operator (JEP-0013)

bkhizgiy:
## Summary

Integrates the `jumpstarter-telemetry` service introduced in #930 into the Jumpstarter operator.

Telemetry can now be configured through the `Jumpstarter` CR and is automatically deployed and managed by the operator, following the same patterns as the controller and router.

## What changed

* Added telemetry Deployment and ClusterIP Service reconciliation, including cleanup when telemetry is disabled.
* Added a new optional `spec.telemetry` configuration with:

  * `enabled`
  * `image` / `imagePullPolicy`
  * `replicas`
  * `logging.filter.minSeverity`
  * `resources`
* Added telemetry reconciliation to the main controller loop.
* Configured the controller to advertise the telemetry endpoint to exporters through `GetServiceEndpoints`.
* Added telemetry endpoint, certificate, and log filter configuration to the controller ConfigMap.
* Added TLS certificate reconciliation through cert-manager, supporting both self-signed and external issuers.
* Added a `TelemetryDeploymentReady` status condition.
* Added integration and unit tests covering the telemetry lifecycle, replicas, TLS, configuration, probes, and status handling.

## Usage

Telemetry can be enabled through the `Jumpstarter` CR:

```yaml
spec:
  telemetry:
    enabled: true
    image: quay.io/jumpstarter-dev/jumpstarter-telemetry:latest
    imagePullPolicy: IfNotPresent
    replicas: 1
    logging:
      filter:
        minSeverity: info
```

When enabled, the operator creates the telemetry Deployment and Service, configures TLS when cert-manager is enabled, and configures the controller to advertise the telemetry endpoint to exporters.

To disable telemetry and clean up its resources:

```yaml
spec:
  telemetry:
    enabled: false
```


bkhizgiy:
@coderabbitai review

bkhizgiy:
@coderabbitai review

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -266,6 +276,55 @@ type DeprecatedLabelsConfig struct {
...
+	Filter TelemetryLoggingFilterConfig `json:"filter,omitempty"`
+}
+
+// TelemetryLoggingFilterConfig controls which log entries are forwarded to the telemetry service.
+type TelemetryLoggingFilterConfig struct {
+	// Minimum log severity to forward.
+	// Accepted values: debug, info, warning, error, critical. Defaults to "info".
+	// +kubebuilder:default="info"
+	// +kubebuilder:validation:Enum=debug;info;warning;error;critical
+	MinSeverity string `json:"minSeverity,omitempty"`
```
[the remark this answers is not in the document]
bkhizgiy: was fixed

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
```diff
@@ -369,6 +376,28 @@ func (r *JumpstarterReconciler) reconcileRouterCertificate(ctx context.Context,
...
+	if includeInternalNames {
+		dnsNames = append(dnsNames,
+			telemetryServiceName,
+			fmt.Sprintf("%s.%s", telemetryServiceName, js.Namespace),
+			fmt.Sprintf("%s.%s.svc", telemetryServiceName, js.Namespace),
+			fmt.Sprintf("%s.%s.svc.cluster.local", telemetryServiceName, js.Namespace),
+		)
+	}
+	return dnsNames
+}
```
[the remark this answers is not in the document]
bkhizgiy: was fixed

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+	svc.Name = telemetryServiceName
+	svc.Namespace = jumpstarter.Namespace
+	if err := r.Delete(ctx, svc); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete telemetry service: %w", err)
+	} else if err == nil {
+		log.Info("Deleted telemetry service", "name", telemetryServiceName)
+	}
+
+	return nil
+}
```
raballew, line 420: Do we want to clean up certificates once telemetry is disabled (`jumpstarter.Spec.CertManager.Enabled` flips from true to false)? GC wont work here.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+						},
+					},
+					Volumes: volumes,
+					SecurityContext: &corev1.PodSecurityContext{
+						RunAsNonRoot: boolPtr(true),
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
+					ServiceAccountName: fmt.Sprintf("%s-controller-manager", jumpstarter.Name),
```
raballew: Create a dedicated ServiceAccount (e.g., `{name}-telemetry`) with no RBAC bindings and assign it to the telemetry pod instead. This is too broad right now.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+			Name: "CONTROLLER_KEY",
+			ValueFrom: &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: "jumpstarter-controller-secret",
+					},
+					Key: "key",
+				},
+			},
+		},
```
raballew: The telemetry service is a gRPC log ingestion endpoint with no documented need for this credential. Why are we passing `CONTROLLER_KEY`  into it then?

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1292,6 +1312,30 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
...
+		// When an external issuer is used, use its CA bundle if provided.
+		if jumpstarter.Spec.CertManager.Enabled {
+			caCert, err := r.resolveTelemetryCA(ctx, jumpstarter)
+			if err != nil {
+				logf.FromContext(ctx).V(1).Info("Telemetry CA not yet available, will retry", "error", err)
+			} else if caCert != "" {
+				telemetryCfg.Certificate = caCert
+			}
+		}
+		cfg.Telemetry = telemetryCfg
```
raballew: Remove `telemetryCfg.Certificate = caCert` from `buildConfig` and remove the `EXTERNAL_CERT_PEM`/`EXTERNAL_KEY_PEM` and secret volume mount from `createTelemetryDeployment` until the telemetry binary is updated to serve TLS. It will fail otherwise (non-TLS path).

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
```diff
@@ -369,6 +376,28 @@ func (r *JumpstarterReconciler) reconcileRouterCertificate(ctx context.Context,
...
 
+// reconcileTelemetryCertificate creates the TLS certificate for the telemetry service.
+func (r *JumpstarterReconciler) reconcileTelemetryCertificate(ctx context.Context, js *operatorv1alpha1.Jumpstarter, issuerRef cmmeta.ObjectReference) error {
+	certName := GetTelemetryCertSecretName(js)
+	includeInternalNames := !isExternalIssuer(js)
+	dnsNames := r.collectTelemetryDNSNames(js, includeInternalNames)
+	return r.reconcileServerCertificate(ctx, js, issuerRef, certName, "telemetry", dnsNames, nil)
+}
+
+// collectTelemetryDNSNames collects all DNS names for the telemetry certificate.
```
raballew, line 387: Add an integration test inside `Describe("Telemetry Lifecycle")` that sets both `CertManager.Enabled: true` and `Telemetry.Enabled: true`, calls `doReconcile()`, and asserts a `certmanagerv1.Certificate` named by `GetTelemetryCertSecretName(js)` is created with the expected DNS names.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1292,6 +1312,30 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
...
+		// Resolve the CA certificate for TLS verification by exporters.
+		// When cert-manager is enabled in self-signed mode, read it from the CA secret.
+		// When an external issuer is used, use its CA bundle if provided.
+		if jumpstarter.Spec.CertManager.Enabled {
+			caCert, err := r.resolveTelemetryCA(ctx, jumpstarter)
+			if err != nil {
+				logf.FromContext(ctx).V(1).Info("Telemetry CA not yet available, will retry", "error", err)
+			} else if caCert != "" {
+				telemetryCfg.Certificate = caCert
+			}
```
raballew: Add a ConfigMap propagation test variant that enables cert-manager with a pre-created CA secret and asserts the serialised config contains the expected CA PEM.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,754 @@
...
+			Namespace: crNamespace,
+		}, deployment)
+		Expect(errors.IsNotFound(err)).To(BeTrue(), "telemetry deployment should be deleted")
+
+		err = k8sClient.Get(ctx, types.NamespacedName{
+			Name:      telemetryServiceName,
+			Namespace: crNamespace,
+		}, svc)
+		Expect(errors.IsNotFound(err)).To(BeTrue(), "telemetry service should be deleted")
+	})
```
raballew, line 316: Following up on a previous comment (the first one in this review). I think you could add an `It` block that enables telemetry with `CertManager.Enabled: true`, reconciles to verify the Certificate is created, then disables telemetry, reconciles again, and asserts the Certificate is absent if (big if) we want to delete the cert when to bool flag toggles to false.

controller/deploy/operator/internal/controller/jumpstarter/status.go
```diff
@@ -111,6 +111,21 @@ func (r *JumpstarterReconciler) updateStatus(ctx context.Context, js *operatorv1
...
+			telReady,
+			conditionReason(telReady, "DeploymentAvailable", "DeploymentNotAvailable"),
+			telMsg)
+		if !telReady {
+			allReady = false
+			messages = append(messages, telMsg)
+		}
+	} else {
+		meta.RemoveStatusCondition(&js.Status.Conditions, operatorv1alpha1.ConditionTypeTelemetryDeploymentReady)
+	}
```
raballew: [MEDIUM] `RemoveStatusCondition` is called here when telemetry is disabled, but the cleanup test only checks for Deployment and Service deletion. It does not verify the condition was actually removed from the CR status. A regression that silently stopped this call would pass the existing test.

Extend the cleanup test to read the CR status after the second `doReconcile()` and assert `meta.FindStatusCondition(js.Status.Conditions, operatorv1alpha1.ConditionTypeTelemetryDeploymentReady)` returns nil.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+// resolveTelemetryCA reads the CA certificate that exporters need to verify the
+// telemetry TLS connection. For self-signed CA mode, the cert is in the CA secret;
+// for external issuers, the user-provided caBundle is used.
+func (r *JumpstarterReconciler) resolveTelemetryCA(ctx context.Context, jumpstarter *operatorv1alpha1.Jumpstarter) (string, error) {
+	if jumpstarter.Spec.CertManager.Server != nil && jumpstarter.Spec.CertManager.Server.IssuerRef != nil {
+		if len(jumpstarter.Spec.CertManager.Server.IssuerRef.CABundle) > 0 {
+			return string(jumpstarter.Spec.CertManager.Server.IssuerRef.CABundle), nil
+		}
+		return "", nil
+	}
```
raballew, line 436: Add a test case in `Describe("resolveTelemetryCA")` with an `IssuerRef` that has a nil or empty `CABundle` and assert the return is `("", nil)`.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+						SeccompProfile: &corev1.SeccompProfile{
+							Type: corev1.SeccompProfileTypeRuntimeDefault,
+						},
+					},
+					ServiceAccountName: fmt.Sprintf("%s-controller-manager", jumpstarter.Name),
+				},
+			},
+		},
+	}
+}
```
raballew: The signature is `func (r *JumpstarterReconciler) createTelemetryDeployment(...)` but there are zero `r.` accesses in the body.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+	} else if err == nil {
+		log.Info("Deleted telemetry service", "name", telemetryServiceName)
+	}
+
+	return nil
+}
+
+// GetTelemetryCertSecretName returns the name of the telemetry TLS secret.
+func GetTelemetryCertSecretName(js *operatorv1alpha1.Jumpstarter) string {
+	return js.Name + telemetryCertSuffix
```
raballew: Rename to `getTelemetryCertSecretName` as its only consumed internally.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+
+	operatorv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/deploy/operator/api/v1alpha1"
+)
+
+const (
+	telemetryPort         = 9093
+	telemetryCertSuffix   = "-telemetry-tls"
+	telemetryServiceName  = "jumpstarter-telemetry"
+	telemetryComponentApp = "jumpstarter-telemetry"
+)
```
raballew: [MEDIUM] `telemetryComponentApp = "jumpstarter-telemetry"` is identical in value to `telemetryServiceName = "jumpstarter-telemetry"` and has only one use in `telemetryLabels`. Two constants with the same value create a maintenance hazard where renaming one silently leaves the other stale.

Remove `telemetryComponentApp` and replace its sole use in `telemetryLabels` with `telemetryServiceName`.

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
```diff
@@ -114,6 +114,13 @@ func (r *JumpstarterReconciler) reconcileCertificates(ctx context.Context, js *o
 		}
 	}
 
+	// Create telemetry certificate if telemetry is enabled
```
raballew: [LOW] Three inline comments directly paraphrase the condition guard and function call they precede, adding no information: "Create telemetry certificate if telemetry is enabled" (certificates.go:117), "Add telemetry endpoint env var when telemetry is enabled" (jumpstarter_controller.go:866), and "Telemetry configuration" (jumpstarter_controller.go:1315).

Delete all three; the code is self-explanatory without them.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+	}
+
+	if err := r.reconcileTelemetryService(ctx, jumpstarter); err != nil {
+		return fmt.Errorf("failed to reconcile telemetry service: %w", err)
+	}
+
+	return nil
+}
+
+// reconcileTelemetryDeployment creates or updates the telemetry Deployment.
```
raballew: [LOW] Several functions have single-line doc comments that are direct paraphrases of their identifiers with nothing else to say: `reconcileTelemetryDeployment`, `reconcileTelemetryService`, `createTelemetryDeployment`, `telemetryEndpointFor`, `telemetryLabels`, `checkTelemetryDeploymentReady`, and `defaultTelemetryResources`.

Remove those doc comments. Comments that explain non-obvious intent (such as the multi-paragraph explanation in `cleanupTelemetry`) are worth keeping.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,754 @@
...
+		spec := makeJumpstarterSpec()
+		Expect(k8sClient.Create(ctx, &operatorv1alpha1.Jumpstarter{
+			ObjectMeta: metav1.ObjectMeta{Name: crName, Namespace: crNamespace},
+			Spec:       spec,
+		})).To(Succeed())
+
+		doReconcile()
+
+		configData := getConfigData()
+		Expect(configData).NotTo(ContainSubstring("telemetry"))
```
raballew: [LOW] `Expect(configData).NotTo(ContainSubstring("telemetry"))` would produce a false negative if any future config field contains the word "telemetry". The enabled-case checks have the same problem with blunt substring matching.

Unmarshal `configData` into a `config.Config` struct and assert `cfg.Telemetry == nil` for the disabled case. For the enabled case, assert individual struct fields.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,754 @@
...
+		found := false
+		expectedEndpoint := telemetryEndpointFor(crNamespace)
+		for _, env := range container.Env {
+			if env.Name == "GRPC_TELEMETRY_ENDPOINT" {
+				found = true
+				Expect(env.Value).To(Equal(expectedEndpoint))
+				break
+			}
+		}
+		Expect(found).To(BeTrue(), "GRPC_TELEMETRY_ENDPOINT should be set on controller container")
```
raballew: [LOW] The `for` loop over `container.Env` with a `break` on match cannot detect a duplicate `GRPC_TELEMETRY_ENDPOINT` entry and is more verbose than necessary.

Replace with `Expect(container.Env).To(ContainElement(corev1.EnvVar{Name: "GRPC_TELEMETRY_ENDPOINT", Value: telemetryEndpointFor(crNamespace)}))`.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,432 @@
...
+							},
+							Resources:                defaultTelemetryResources(t.Resources),
+							TerminationMessagePath:   "/dev/termination-log",
+							TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: boolPtr(false),
+								Capabilities: &corev1.Capabilities{
+									Drop: []corev1.Capability{"ALL"},
+								},
+							},
```
raballew: [LOW] The SecurityContext sets `AllowPrivilegeEscalation: false` and drops all capabilities but does not set `ReadOnlyRootFilesystem: true`. A compromised process can write to the container filesystem, weakening post-exploitation detection.

Add `ReadOnlyRootFilesystem: ptr.To(true)` to the SecurityContext. Mount explicit `emptyDir` volumes for any paths the binary requires write access to.

raballew (changes requested):
sorry, i only saw your request today. a few nits but also a substantial number of biggies that need to be resolved.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,759 @@
...
+
+		expectedSA := crName + telemetrySASuffix
+		Expect(telemetryDep.Spec.Template.Spec.ServiceAccountName).To(Equal(expectedSA))
+
+		By("verifying the dedicated ServiceAccount exists with no RBAC bindings")
+		sa := &corev1.ServiceAccount{}
+		Expect(k8sClient.Get(ctx, types.NamespacedName{
+			Name:      expectedSA,
+			Namespace: crNamespace,
+		}, sa)).To(Succeed())
```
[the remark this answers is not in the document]
bkhizgiy: not valid

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,442 @@
...
+	telemetryComponentApp = "jumpstarter-telemetry"
+	telemetrySASuffix     = "-telemetry"
+)
+
+// reconcileTelemetryDeploymentStage reconciles only the telemetry Deployment (and cleanup).
+// It is called from the Deployment stage of the reconcile loop, before Services.
+// Certificate reconciliation is handled in certificates.go alongside other certs.
+func (r *JumpstarterReconciler) reconcileTelemetryDeploymentStage(ctx context.Context, jumpstarter *operatorv1alpha1.Jumpstarter) error {
+	if jumpstarter.Spec.Telemetry == nil || !jumpstarter.Spec.Telemetry.Enabled {
+		return r.cleanupTelemetry(ctx, jumpstarter)
```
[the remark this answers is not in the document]
bkhizgiy: not valid.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,443 @@
...
+			existingService.Spec.Selector = desiredService.Spec.Selector
+			existingService.Spec.Ports = desiredService.Spec.Ports
+			return controllerutil.SetControllerReference(jumpstarter, existingService, r.Scheme)
+		}
+
+		existingService.Labels = desiredService.Labels
+		existingService.Spec.Type = desiredService.Spec.Type
+		existingService.Spec.Selector = desiredService.Spec.Selector
+		existingService.Spec.Ports = desiredService.Spec.Ports
+		return controllerutil.SetControllerReference(jumpstarter, existingService, r.Scheme)
```
mangelajo, line 237: Nit: Both branches of the `CreationTimestamp.IsZero()` guard are identical — same labels, same spec fields, same `SetControllerReference`. The guard can be removed and the body collapsed into a single path.

```go
op, err := controllerutil.CreateOrUpdate(ctx, r.Client, existingService, func() error {
	existingService.Labels = desiredService.Labels
	existingService.Spec.Type = desiredService.Spec.Type
	existingService.Spec.Selector = desiredService.Spec.Selector
	existingService.Spec.Ports = desiredService.Spec.Ports
	return controllerutil.SetControllerReference(jumpstarter, existingService, r.Scheme)
})
```

AI Generated, but reviewed/edited by me.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,443 @@
...
+
+	log.V(1).Info("Telemetry ServiceAccount reconciled", "name", saName, "operation", op)
+	return nil
+}
+
+// reconcileTelemetryServiceStage reconciles only the telemetry ClusterIP Service.
+// It is called from the Services/networking stage of the reconcile loop.
+func (r *JumpstarterReconciler) reconcileTelemetryServiceStage(ctx context.Context, jumpstarter *operatorv1alpha1.Jumpstarter) error {
+	if jumpstarter.Spec.Telemetry == nil || !jumpstarter.Spec.Telemetry.Enabled {
+		return nil
```
mangelajo: This early-return when telemetry is disabled means Service cleanup depends entirely on `cleanupTelemetry` being called first from `reconcileTelemetryDeploymentStage`. If the call order in the main reconcile loop ever changes (e.g. services are reconciled before deployments), the Service would be orphaned.

Consider adding a symmetric cleanup call here:
  ```go
  if jumpstarter.Spec.Telemetry == nil || !jumpstarter.Spec.Telemetry.Enabled {
      return r.cleanupTelemetryService(ctx, jumpstarter)
  }
  ```

AI Generated, but reviewed/edited by me.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -851,6 +863,14 @@ func (r *JumpstarterReconciler) createControllerDeployment(jumpstarter *operator
 		},
 	}
 
+	// Add telemetry endpoint env var when telemetry is enabled
+	if jumpstarter.Spec.Telemetry != nil && jumpstarter.Spec.Telemetry.Enabled {
+		envVars = append(envVars, corev1.EnvVar{
+			Name:  "GRPC_TELEMETRY_ENDPOINT",
+			Value: telemetryEndpointFor(jumpstarter.Namespace),
+		})
+	}
```
mangelajo: This env var is set on the controller deployment but nothing in the controller binary reads `GRPC_TELEMETRY_ENDPOINT` — the controller loads its telemetry config from the ConfigMap via `cfg.Telemetry` (wired at `cmd/main.go:288`), which is already correctly populated by `buildConfig` below (lines 1315-1327).

Is this intended for a future consumer, or can it be removed? If kept, it should at least have a comment explaining why it exists alongside the ConfigMap path.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,443 @@
...
+	// Self-signed CA mode — read from the CA secret created by cert-manager
+	caSecretName := jumpstarter.Name + caCertificateSuffix
+	caSecret := &corev1.Secret{}
+	if err := r.Get(ctx, client.ObjectKey{Name: caSecretName, Namespace: jumpstarter.Namespace}, caSecret); err != nil {
+		return "", fmt.Errorf("CA secret %s not found: %w", caSecretName, err)
+	}
+	if cert, ok := caSecret.Data["tls.crt"]; ok {
+		return string(cert), nil
+	}
+	return "", fmt.Errorf("CA secret %s missing tls.crt", caSecretName)
```
mangelajo, line 447: Missing test coverage: the `resolveTelemetryCA` tests cover external issuer with CA bundle, self-signed with valid secret, missing secret, nil CABundle, and empty CABundle — but there's no test for the case where the CA secret exists but is **missing the `tls.crt` key** (line 410). This error path should have a test to prevent silent regressions.

Same as @raballew comment.

AI Generated, but reviewed/edited by me.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1292,6 +1312,20 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
...
+	// Telemetry configuration.
+	// Certificate is intentionally omitted until the telemetry binary supports TLS serving.
+	if jumpstarter.Spec.Telemetry != nil && jumpstarter.Spec.Telemetry.Enabled {
+		t := jumpstarter.Spec.Telemetry
+		telemetryCfg := &config.Telemetry{
+			Enabled:  true,
+			Endpoint: telemetryEndpointFor(jumpstarter.Namespace),
+		}
+		if t.Logging.Filter.MinSeverity != "" {
+			telemetryCfg.Logging.Filter.MinSeverity = t.Logging.Filter.MinSeverity
```
mangelajo, line 1316: Minor naming inconsistency: the CRD field here uses `minSeverity` (camelCase, standard for K8s CRDs), but the config struct at `controller/internal/config/types.go:57` serializes as `min_severity` (snake_case in JSON/YAML tags).

Would recommend updating the config struct's JSON tag to `minSeverity` as well for uniformity — it keeps the ConfigMap YAML consistent with the CR spec and avoids confusion for operators inspecting both resources. Since the ConfigMap is machine-generated (not hand-edited), the rename should be safe.

AI Generated, but reviewed by me.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,443 @@
...
+								PeriodSeconds:       10,
+								TimeoutSeconds:      1,
+								SuccessThreshold:    1,
+								FailureThreshold:    3,
+							},
+							Resources:                defaultTelemetryResources(t.Resources),
+							TerminationMessagePath:   "/dev/termination-log",
+							TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+							SecurityContext: &corev1.SecurityContext{
+								AllowPrivilegeEscalation: boolPtr(false),
```
mangelajo: Nit: `boolPtr(false)` / `boolPtr(true)` are used here, but `ptr.To` from `k8s.io/utils/ptr` is already imported and used in this same file (e.g. `ptr.To(int32(600))` at line 248). Consider using `ptr.To(false)` / `ptr.To(true)` for consistency instead of the local `boolPtr` helper.


AI Generated, but reviewed/edited by me.

mangelajo (changes requested):
Very solid incremental update. Just a few comments.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: fixed, collapsed into a single path without the `CreationTimestamp.IsZero()` guard since the logic was identical.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: Fixed. Added `cleanupTelemetryService` call when telemetry is disabled

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: Added, new test case: "returns an error when the CA secret exists but is missing tls.crt".

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
bkhizgiy: Fixed. Updated controller/internal/config/types.go to use `minSeverity` in JSON/YAML tags for consistency with the CRD field.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: Fixed. Replaced `boolPtr(false)/boolPtr(true)` with `ptr.To(false)/ptr.To(true)` for consistency with the rest of the file.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,784 @@
...
+		Expect(errors.IsNotFound(err)).To(BeTrue(), "telemetry deployment should not exist")
+
+		By("verifying no telemetry Service exists")
+		svc := &corev1.Service{}
+		err = k8sClient.Get(ctx, types.NamespacedName{
+			Name:      telemetryServiceName,
+			Namespace: crNamespace,
+		}, svc)
+		Expect(errors.IsNotFound(err)).To(BeTrue(), "telemetry service should not exist")
+	})
```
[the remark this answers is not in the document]
bkhizgiy: Added. The "does not create telemetry resources when telemetry is disabled" test now also asserts that the telemetry ServiceAccount returns IsNotFound.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
bkhizgiy: This was related to changes on the [tls](https://github.com/jumpstarter-dev/jumpstarter/pull/975) branch that haven't been merged yet, but after reviewing it again specifically for the operator, I realized the ConfigMap alone should be sufficient since the operator always populates Telemetry.Endpoint there. The env var was intended as a fallback for manual deployments, but operator-managed deployments don't need it, dropped.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: Done, was added :
```
"returns ('', nil) when an external IssuerRef has a nil CABundle"
"returns ('', nil) when an external IssuerRef has an empty CABundle"
```

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: done, the `cleanupTelemetry` function now explicitly deletes the certificate.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
bkhizgiy: Added. New test "cleans up Certificate when telemetry is disabled" enables CertManager + Telemetry, reconciles to verify the Certificate is created, then disables telemetry, reconciles again, and asserts the Certificate is deleted.

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
bkhizgiy: Added. New test "creates a Certificate when CertManager and Telemetry are both enabled" sets both flags to true, reconciles, and asserts the Certificate is created with the expected DNS names.

mangelajo:
looks good but some of the controller tests aren't passing.

bkhizgiy:
> looks good but some of the controller tests aren't passing.

yes. its a test I've added in the last commit, looking into it.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
bkhizgiy: I ended up removing these and the next one for now, since they would require adding the cert manager CRDs to envtest otherwise, the tests fail. I think that's a bit out of scope for this PR, so I'll add proper integration tests for telemetry, including certificate creation and cleanup in a follow-up PR.

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
bkhizgiy: same her, was dropped.

bkhizgiy:
@mangelajo @raballew Fixed the testing issue, for now I dropped the latest integration tests introduced during the PR review, since they require adding some cert manager CRDs to envtest in order to run. I think that's a bit out of scope for this PR and would be better handled in a follow-up. We'll need to create a separate PR for e2e and testing anyway, so I'll include everything there if that's okay with you.

mangelajo:
@bkhizgiy I see there are some regressions in the last change, the GRPC telemetry endpoint details aren't passed anymore to the controller for announcement:

https://github.com/jumpstarter-dev/jumpstarter/pull/997/changes/03358124d8ad18dade294c8f732c66dfeee07f73#diff-55b7b5cfeb5b2fcd9f1d83431b263d743cfa0b0ea3c778db2395e81984e6ee60L866

bkhizgiy:
@mangelajo I did another verification, looks like everything presents, when looking a the last commit I see only the 2 tests being removed... 
maybe you talking about the `GRPC_TELEMETRY_ENDPOINT` env var that was intentionally dropped? other then that looks like everything else is present.
https://github.com/jumpstarter-dev/jumpstarter/compare/9666e49584479366e64dd252d86747724daeb087..03358124d8ad18dade294c8f732c66dfeee07f73

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
```diff
@@ -114,6 +114,13 @@ func (r *JumpstarterReconciler) reconcileCertificates(ctx context.Context, js *o
...
+	if js.Spec.Telemetry != nil && js.Spec.Telemetry.Enabled {
+		if err := r.reconcileTelemetryCertificate(ctx, js, issuerRef); err != nil {
+			return fmt.Errorf("failed to reconcile telemetry certificate: %w", err)
+		}
+	}
```
raballew, line 122: `reconcileTelemetryCertificate` has no coverage through the reconcile loop. Every `doReconcile()` call site in the test suite uses `makeJumpstarterSpec()` which hardcodes `CertManager.Enabled: false`.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
...
+							Image:           t.Image,
+							ImagePullPolicy: t.ImagePullPolicy,
+							Command:         []string{"/telemetry"},
+							Args: []string{
+								fmt.Sprintf("--grpc-bind=:%d", telemetryPort),
+							},
+							Ports: []corev1.ContainerPort{
+								{
+									ContainerPort: int32(telemetryPort),
+									Name:          "grpc",
```
raballew: define `const grpcPortName = "grpc"` alongside the other telemetry constants in telemetry.go and replace all four occurrences in both files.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -266,6 +276,55 @@ type DeprecatedLabelsConfig struct {
...
+	// +kubebuilder:default="IfNotPresent"
+	// +kubebuilder:validation:Enum=Always;IfNotPresent;Never
+	ImagePullPolicy corev1.PullPolicy `json:"imagePullPolicy,omitempty"`
+
+	// Number of telemetry replicas to run.
+	// Multiple replicas provide HA; each exporter connects to exactly one replica
+	// via a persistent MetricsStream, so Prometheus sum-by queries across replicas
+	// yield exact totals without double-counting (see JEP-0013 DD-8).
+	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=1
```
raballew, line 303: The intent is that setting `replicas: 0` should scale the Deployment to zero without deleting it, but the API server rejects `replicas: 0` at admission, making that behavior unreachable.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
...
+	caSecretName := jumpstarter.Name + caCertificateSuffix
+	caSecret := &corev1.Secret{}
+	if err := r.Get(ctx, client.ObjectKey{Name: caSecretName, Namespace: jumpstarter.Namespace}, caSecret); err != nil {
+		return "", fmt.Errorf("CA secret %s not found: %w", caSecretName, err)
+	}
+	if cert, ok := caSecret.Data["tls.crt"]; ok {
+		return string(cert), nil
+	}
+	return "", fmt.Errorf("CA secret %s missing tls.crt", caSecretName)
+}
```
raballew: [MEDIUM] `resolveTelemetryCA` (lines 421-442) has no production callers. Running `grep -rn "resolveTelemetryCA" controller/ --include="*.go" | grep -v "_test.go"` returns only the function definition. `buildConfig` in jumpstarter_controller.go explicitly leaves `config.Telemetry.Certificate` empty with a comment that TLS serving is not yet supported. The six dedicated tests for `resolveTelemetryCA` give false confidence that CA resolution is active during reconciliation when it is not. Either wire `resolveTelemetryCA` into `buildConfig` to populate the certificate field when cert-manager is enabled, or remove the function and its six tests and leave a `// TODO(phase2): wire resolveTelemetryCA into buildConfig` comment at the Certificate field site.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,744 @@
...
+			Name:      crName + telemetrySASuffix,
+			Namespace: crNamespace,
+		}, sa)
+		Expect(errors.IsNotFound(err)).To(BeTrue(), "telemetry service account should be deleted")
+
+		By("verifying telemetry is absent from the controller ConfigMap")
+		configData := getConfigData()
+		Expect(configData).NotTo(ContainSubstring("telemetry"),
+			"telemetry should not appear in ConfigMap after disabling")
+	})
```
raballew: [MEDIUM] The test 'cleans up telemetry resources when telemetry is disabled after being enabled' (lines 256-316) verifies that Deployment, Service, and ServiceAccount are deleted after the second reconcile, but never asserts that the `TelemetryDeploymentReady` condition is removed. `status.go:126` calls `meta.RemoveStatusCondition` on the disable path, but this branch is never checked in the cleanup test. After the second `doReconcile()`, add: `Expect(k8sClient.Get(ctx, types.NamespacedName{Name: crName, Namespace: crNamespace}, js)).To(Succeed()); cond := meta.FindStatusCondition(js.Status.Conditions, operatorv1alpha1.ConditionTypeTelemetryDeploymentReady); Expect(cond).To(BeNil(), "TelemetryDeploymentReady condition should be removed when telemetry is disabled")`.

controller/internal/config/types.go
```diff
@@ -54,7 +54,7 @@ type TelemetryLogging struct {
 type TelemetryLoggingFilter struct {
 	// MinSeverity is the minimum log severity to forward.
 	// Accepted values: debug, info, warning, error, critical. Defaults to "info" when empty.
-	MinSeverity string `json:"min_severity,omitempty" yaml:"min_severity,omitempty"`
+	MinSeverity string `json:"minSeverity,omitempty" yaml:"minSeverity,omitempty"`
 }
 
 var validSeverities = map[string]struct{}{
```
raballew: [MEDIUM] `Telemetry.Validate()` has zero test coverage. It checks `MinSeverity` against the five accepted values in `validSeverities` and is called at controller startup, making it the sole guard against invalid ConfigMap values. `grep -c "Validate" controller/internal/config/types_test.go` returns 0. Add a table-driven `TestTelemetryValidate` covering: empty MinSeverity (valid), each of the five accepted values (valid), `"WARN"` (invalid), `"WARNING"` (invalid), and `"verbose"` (invalid). The `TestParseDuration` function in the same file is a good model for the test structure.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1292,6 +1304,20 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
 		Keys: jumpstarter.Spec.DeprecatedLabels.Keys,
 	}
 
+	// Telemetry configuration.
+	// Certificate is intentionally omitted until the telemetry binary supports TLS serving.
```
raballew: [LOW] Line 1307 has a redundant comment `// Telemetry configuration.` immediately above the more informative `// Certificate is intentionally omitted until the telemetry binary supports TLS serving.`. The first line just restates what the surrounding `if jumpstarter.Spec.Telemetry != nil && jumpstarter.Spec.Telemetry.Enabled` block already makes obvious. Remove line 1307 and keep only the intentionality comment.

controller/deploy/operator/internal/controller/jumpstarter/status.go
```diff
@@ -465,6 +480,30 @@ func (r *JumpstarterReconciler) checkExporterSetControllersReady(ctx context.Con
...
+// checkTelemetryDeploymentReady checks if the telemetry deployment is available.
+func (r *JumpstarterReconciler) checkTelemetryDeploymentReady(ctx context.Context, js *operatorv1alpha1.Jumpstarter) (bool, string) {
+	deploymentName := fmt.Sprintf("%s-telemetry", js.Name)
+	deployment := &appsv1.Deployment{}
+	err := r.Get(ctx, types.NamespacedName{Name: deploymentName, Namespace: js.Namespace}, deployment)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return false, fmt.Sprintf("Telemetry deployment %s not found", deploymentName)
+		}
+		return false, fmt.Sprintf("Error getting telemetry deployment: %v", err)
```
raballew: Log the full error to the operator log and surface a generic message in the condition, e.g. `log.Error(err, "failed to get telemetry deployment", "deployment", deploymentName)` then return `"error querying telemetry deployment; check operator logs for details"`

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
+/*
+Copyright 2026 by the Jumpstarter Authors
```
raballew: [LOW] The copyright line reads `Copyright 2026 by the Jumpstarter Authors` but the established format in this package is `Copyright YYYY. The Jumpstarter Authors.` (with a period after the year, "The" before the author, and a trailing period). The companion `telemetry_test.go` correctly uses `Copyright 2026. The Jumpstarter Authors.` Change line 2 to match: `Copyright 2026. The Jumpstarter Authors.`

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
...
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+
+	operatorv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/deploy/operator/api/v1alpha1"
+)
+
+const (
+	telemetryPort         = 9093
+	telemetryCertSuffix   = "-telemetry-tls"
+	telemetryServiceName  = "jumpstarter-telemetry"
+	telemetryComponentApp = "jumpstarter-telemetry"
```
raballew: [LOW] `telemetryServiceName` and `telemetryComponentApp` are both defined as `"jumpstarter-telemetry"` (lines 42-43). The Service selector in `reconcileTelemetryService` uses `telemetryLabels()` which sets `"app": telemetryComponentApp`. If either constant is changed independently, the Service selector will silently stop matching Deployment pod labels with no compile-time or test-time signal. Either remove `telemetryComponentApp` and use `telemetryServiceName` everywhere, or add a comment explicitly stating the two values must remain in sync and explaining why they are kept separate.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
...
+	if err := r.reconcileTelemetryServiceAccount(ctx, jumpstarter); err != nil {
+		return fmt.Errorf("failed to reconcile telemetry service account: %w", err)
+	}
+
+	if err := r.reconcileTelemetryDeployment(ctx, jumpstarter); err != nil {
+		return fmt.Errorf("failed to reconcile telemetry deployment: %w", err)
+	}
+
+	return nil
+}
```
raballew: [LOW] The godoc at line 47 says `reconcileTelemetryDeploymentStage` manages only 'the telemetry Deployment (and cleanup)', but the function body also calls `reconcileTelemetryServiceAccount`. A ServiceAccount is an identity resource, not a deployment artifact, so the name and doc are misleading. Consider renaming to `reconcileTelemetryWorkloadStage` and updating the godoc to accurately list all managed resources: ServiceAccount and Deployment, plus full cleanup of all four telemetry resources when telemetry is disabled.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
...
+	}
+
+	svc := &corev1.Service{}
+	svc.Name = telemetryServiceName
+	svc.Namespace = jumpstarter.Namespace
+	if err := r.Delete(ctx, svc); err != nil && !errors.IsNotFound(err) {
+		return fmt.Errorf("failed to delete telemetry service: %w", err)
+	} else if err == nil {
+		log.Info("Deleted telemetry service", "name", telemetryServiceName)
+	}
```
raballew: The Service is deleted twice on every reconcile cycle when telemetry is disabled. `cleanupTelemetry` (called from `reconcileTelemetryDeploymentStage`) already deletes the Service at lines 384-391. Then `reconcileTelemetryServiceStage` calls `cleanupTelemetryService`, which issues a second Delete on the same service name at lines 114-127. The second call always returns 404 (silently ignored via `errors.IsNotFound`), but it is a superfluous Kubernetes API call on every disabled-telemetry reconcile.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -0,0 +1,474 @@
...
+			return controllerutil.SetControllerReference(jumpstarter, existingDeployment, r.Scheme)
+		}
+
+		desiredDeployment.Spec.Template.Spec.DeprecatedServiceAccount = existingDeployment.Spec.Template.Spec.DeprecatedServiceAccount
+		desiredDeployment.Spec.Template.Spec.SchedulerName = existingDeployment.Spec.Template.Spec.SchedulerName
+
+		if !deploymentNeedsUpdate(existingDeployment, desiredDeployment) {
+			log.V(1).Info("Telemetry deployment is up to date", "name", existingDeployment.Name)
+			return nil
+		}
```
raballew, line 154: Inside the `controllerutil.CreateOrUpdate` mutate closure, when the Deployment already exists and `deploymentNeedsUpdate` returns false, the function returns nil at line 151 without calling `SetControllerReference`. If the OwnerReference is stripped externally (e.g. by `kubectl edit` or a mutating webhook), the reconciler never re-attaches it, and on CR deletion the Deployment is not garbage-collected.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -0,0 +1,744 @@
...
+		}
+
+		result := defaultTelemetryResources(custom)
+
+		Expect(result.Claims).To(HaveLen(1))
+		Expect(result.Claims[0].Name).To(Equal("gpu"))
+		Expect(result.Requests).To(BeNil())
+		Expect(result.Limits).To(BeNil())
+	})
+})
```
raballew: [LOW] The four `It` blocks in `Describe("defaultTelemetryResources")` (lines 529-576) use a `"should"` prefix while all other tests in the file use factual present-tense statements. Consider renaming to match the established style: `"returns defaults when no resource constraints are specified"`, `"passes through caller-provided CPU requests unchanged"`, `"passes through caller-provided memory limits unchanged"`, `"passes through claims-only input without applying CPU or memory defaults"`.

raballew (dismissed):

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
bkhizgiy: yep, we discussed this in the comments above, will be done in a follow up PR.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
bkhizgiy: Minimum=1 is intentional and consistent with controller, router, and provisioner replicas. To disable telemetry, users should set enabled: false, which cleanly removes all telemetry resources, scaling to 0 replicas would leave orphaned resources without running pods.

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
raballew: okay! thanks for clarifying. must have missed it in this incredible painful github issue review UI ;)

bkhizgiy: haha no worries, thanks for the review!!

mangelajo (approved):

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
mangelajo: This could be useful in some cases (where deleting all the resources is not wanted but you just want to scale it down temporarily). Since this is consistent with the rest of the controllers, I'd leave it as-is, we can always add this for all as an new feature at some point

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
mangelajo: External stripping of the reference is something almost no controller handles.  If an admin does that it's probably for a reason and not even supported

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
raballew: It seems I made a mistake - please double check if PushLogs requires a bearer token. It seems so to me, which mean push logs uses CONTROLLER_KEY. If we now use `sha256.Sum256([]byte(""))` it produces a deterministic ECDSA key.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
bkhizgiy: yep agreed, I kept this one consistent with the others but we can add this, I"ll open an issue for us to track this.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: Yes, you’re right, my bad for dropping it. I got confused with the other env vars we removed here, but we actually need this one to verify the correct identity otherwise it will fail.

I’m posting a fix now, thanks for catching this!

bkhizgiy: Pushed back the missing `CONTROLLER_KEY` env var, also ran a full test on a local environment to verify everything works as expected.

PR #998, opened 2026-08-13
perf(mitmproxy): share proxy process across tests in a class

bennyz:
Class-scoped fixtures keep mitmdump running across tests within each class. Mocks are cleared and reconfigured between tests via hot-reload instead of full process restart (~3s savings per test).


bennyz:
seems to reduce 1-3 minutes from the python tests

kirkbrauer (approved):

PR #999, opened 2026-08-13
perf(ci): use host-compiled Go for e2e image builds on cache miss

bennyz:
Switch the e2e workflow's controller, operator, and exporterset-controller build jobs from multi-stage docker-build (pulls ~1.5GB go-toolset image) to the CI-optimized path that compiles on the runner with cached Go modules and packages into minimal ubi-micro containers.

Add docker-build-exporter-set-controller-ci as a standalone target so the e2e workflow can build the ESC image independently (docker-build-ci builds controller + ESC together, but e2e needs them in separate parallel jobs). Update cache keys to include Containerfile.prebuilt.


mangelajo:
hmm controller image goes down to 1.5min from 4min, nice!

mangelajo (approved):

PR #1000, opened 2026-08-13
fix: eliminate socket permission race in jumpstarter-exec serve

bennyz:
bind() creates the listen socket as 0o777 (with umask=0), then set_permissions() tightens it to 0o666. Any observer that stat()s the socket between those two syscalls sees the wrong mode — this caused flaky CI failures in the e2e_large_output integration test.

Set umask(0o111) before bind() so the socket is born 0o666 atomically, then restore umask(0) for child processes that need 0o777 (QEMU QMP/serial/VNC sockets).


mangelajo (approved):
Thank you!!

PR #1002, opened 2026-08-14
fix(e2e): wait for controller leader election before running tests

bennyz:
The controller pod can be "Available" (passing readiness probes) before it has acquired the leader lease, which means it cannot reconcile CRDs yet. Legacy exporter creation depends on reconciliation to generate a Secret, so it times out when leader election is slow (~57s observed in CI). Combined with ContinueOnFailure, this caused every downstream spec to wait 5 min on kubectl wait before failing and retrying — effectively hanging the job for hours.

Two fixes:
- Poll the controller's Lease object in wait_for_jumpstarter_resources() until holderIdentity is set, proving reconciliation is possible.
- Move client/exporter creation from It blocks into BeforeAll so that a setup failure aborts the Ordered container immediately instead of letting dependent specs time out one by one.

### DIFF
--- a/controller/hack/utils
+++ b/controller/hack/utils
@@ -236,6 +236,35 @@ wait_for_jumpstarter_resources() {
     --for=condition=ready pod \
     --selector=app=jumpstarter-router-0 \
     --timeout=180s
+
+  # Wait for the controller to acquire its leader lease — without this the
+  # controller pod is "Available" but cannot reconcile, so any operation that
+  # depends on reconciliation (e.g. legacy exporter secret creation) will time
+  # out.
+  #
+  # We check renewTime freshness (within the last 30s) rather than just
+  # holderIdentity presence, because a stale lease from a dead pod retains
+  # its holderIdentity until a new leader overwrites it.
+  echo -e "${GREEN} * Waiting for controller leader election ...${NC}"
+  local le_elapsed=0
+  while [ $le_elapsed -lt 120 ]; do
+    renew_time=$(kubectl get lease a38b78e7.jumpstarter.dev -n "${namespace}" \
+      -o jsonpath='{.spec.renewTime}' 2>/dev/null)
+    if [ -n "$renew_time" ]; then
+      renew_epoch=$(date -d "$renew_time" +%s 2>/dev/null || date -j -f "%Y-%m-%dT%H:%M:%S" "${renew_time%%.*}" +%s 2>/dev/null)
+      now_epoch=$(date +%s)
+      if [ -n "$renew_epoch" ] && [ $((now_epoch - renew_epoch)) -lt 30 ]; then
+        echo -e "${GREEN} * Controller acquired leader lease (renewTime: ${renew_time})${NC}"
+        break
+      fi
+    fi
+    sleep 2
+    le_elapsed=$((le_elapsed + 2))
+  done
+  if [ $le_elapsed -ge 120 ]; then
+    echo -e "${RED}ERROR: Controller did not acquire leader lease within 120s${NC}"
+    exit 1
+  fi
 }
 
 # Wait for gRPC endpoint to be ready
--- a/e2e/test/e2e_test.go
+++ b/e2e/test/e2e_test.go
@@ -119,13 +119,13 @@ var _ = Describe("Core E2E Tests", Label("core"), Ordered, ContinueOnFailure, fu
 			MergeExporterConfig(exporterConfigPath, overlayPath)
 		})
 
-		It("clients were created", func() {
+		It("can create clients with admin cli", func() {
 			out, err := Jmp("config", "client", "list", "-o", "yaml")
 			Expect(err).NotTo(HaveOccurred(), out)
 			Expect(out).To(ContainSubstring("test-client-legacy"))
 		})
 
-		It("exporters were created", func() {
+		It("can create exporters with admin cli", func() {
 			out, err := Jmp("config", "exporter", "list", "-o", "yaml")
 			Expect(err).NotTo(HaveOccurred(), out)
 			Expect(out).To(ContainSubstring("test-exporter-legacy"))
--- a/e2e/test/e2e_test.go
+++ b/e2e/test/e2e_test.go
@@ -70,7 +70,7 @@ var _ = Describe("Core E2E Tests", Label("core"), Ordered, ContinueOnFailure, fu
 	// Client and Exporter creation
 	// -------------------------------------------------------------------
 	Context("Admin CLI resource creation", func() {
-		It("can create clients with admin cli", func() {
+		BeforeAll(func() {
 			ns := Namespace()
 
 			// Idempotent under flake-retries: a prior failed attempt may have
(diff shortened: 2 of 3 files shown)


controller/hack/utils
```diff
@@ -236,6 +236,35 @@ wait_for_jumpstarter_resources() {
     --for=condition=ready pod \
     --selector=app=jumpstarter-router-0 \
     --timeout=180s
+
+  # Wait for the controller to acquire its leader lease — without this the
+  # controller pod is "Available" but cannot reconcile, so any operation that
```
mangelajo, line 241: Should we also/instead fix the controllers to make sure they don't declare healthy status until leader acquired? wdyt?

controller/hack/utils
```diff
@@ -236,6 +236,35 @@ wait_for_jumpstarter_resources() {
...
+  # out.
+  #
+  # We check renewTime freshness (within the last 30s) rather than just
+  # holderIdentity presence, because a stale lease from a dead pod retains
+  # its holderIdentity until a new leader overwrites it.
+  echo -e "${GREEN} * Waiting for controller leader election ...${NC}"
+  local le_elapsed=0
+  while [ $le_elapsed -lt 120 ]; do
+    renew_time=$(kubectl get lease a38b78e7.jumpstarter.dev -n "${namespace}" \
+      -o jsonpath='{.spec.renewTime}' 2>/dev/null)
```
mangelajo, line 252: The script runs under `set -eo pipefail` (line 4 of this file) and `set -exo pipefail` in `deploy_with_operator.sh` which sources it. If the Lease object doesn't exist yet, `kubectl get lease` returns non-zero. In a simple variable assignment (`var=$(cmd)`), `set -e` **will** abort the script — `2>/dev/null` only suppresses stderr, not the exit code.

Consider adding `|| true`:
```bash
renew_time=$(kubectl get lease a38b78e7.jumpstarter.dev -n "${namespace}" \
      -o jsonpath='{.spec.renewTime}' 2>/dev/null || true)
```

---

controller/hack/utils
mangelajo, line 241: Re @mangelajo's question — the idea of tying readiness to leader election is the right instinct, but it's more nuanced than it seems.

**The current situation is already broken for HA (2+ replicas):**

- The controller defaults to **2 replicas** (`jumpstarter_types.go:335`).
- The gRPC `ControllerService` (port 8082) only runs on the leader — it doesn't implement `NeedLeaderElection()`, so controller-runtime defaults to `true` and `Start()` never executes on standby replicas. Port 8082 is **not listening** on non-leader pods.
- But `healthz.Ping` always returns OK, so Kubernetes includes **both** pods in the gRPC Service endpoints.
- Result: with 2 replicas, ~50% of gRPC connections hit the non-leader pod and get **connection refused**.


Doing this would make all gRPC traffic go to one controller when in HA (login and gRPC client/exporter traffic) but seems like the right thing to do as well, it was always designed with active/passive in mind.


---

e2e/test/e2e_test.go
```diff
@@ -113,8 +105,16 @@ var _ = Describe("Core E2E Tests", Label("core"), Ordered, ContinueOnFailure, fu
...
+			out, err := Jmp("config", "client", "list", "-o", "yaml")
+			Expect(err).NotTo(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("test-client-legacy"))
+		})
+
+		It("exporters were created", func() {
+			out, err := Jmp("config", "exporter", "list", "-o", "yaml")
 			Expect(err).NotTo(HaveOccurred(), out)
 			Expect(out).To(ContainSubstring("test-exporter-legacy"))
 		})
```
mangelajo: CodeRabbit suggested moving these list-verification assertions back into the `BeforeAll`. Their reasoning is that under `ContinueOnFailure`, a failed `It` won't stop later specs, so if the verification fails the downstream specs would still run and time out.

However, I think the current approach is fine: the purpose of the `BeforeAll` is to *create* resources, and these `It` blocks verify creation succeeded. If creation itself fails in `BeforeAll`, Ginkgo v2 already skips all specs in the container (even with `ContinueOnFailure`). These `It` blocks provide better diagnostics by separating "creation command succeeded" (checked in `BeforeAll`) from "resource is visible in config" (checked here). If the list check fails, it's a less severe issue that shouldn't necessarily abort the entire suite.

---

mangelajo (commented):
## Review Summary

Good fix for a real CI problem. The diagnosis is solid — the controller pod being "Available" before leader election completes causes downstream reconciliation-dependent operations to time out.

### Findings

**Shell script (`controller/hack/utils`):**
- **Bug:** The `kubectl get lease` lookup will abort the script under `set -e` if the Lease object doesn't exist yet. Needs `|| true` to let the retry loop continue. (See inline comment.)
- The leader election wait logic is well thought out — checking `renewTime` freshness rather than just `holderIdentity` presence is the right approach for avoiding stale leases.

**Test refactor (`e2e/test/e2e_test.go`):**
- Moving resource creation into `BeforeAll` is a good improvement. Under `ContinueOnFailure`, a failed `BeforeAll` skips all specs in the container, which is exactly the fail-fast behavior needed here.
- The separate `It` blocks for verification provide good diagnostics.

**Architectural note (re: @mangelajo's comment):**
- The underlying issue is that the controller's readiness probe (`healthz.Ping`) is not leader-election-aware. However, naively adding `LeaderElectionCheck` would break the `LoginService` which intentionally runs on all replicas and shares the same pod selector. This is an existing HA design issue that should be tracked as a follow-up, separate from this PR's scope. See inline comment for details.

### Actionable items
1. Add `|| true` to the `kubectl get lease` command to prevent `set -e` from aborting the script
2. Consider opening a follow-up issue for the HA readiness probe architecture

---

controller/hack/utils
mangelajo: If the LoginService being leader-only is acceptable (it's low throughput), then this becomes straightforward. The fix is a one-liner in each controller binary — replace `healthz.Ping` with `healthz.LeaderElectionCheck`:

```go
// controller/cmd/main.go (line 324):
mgr.AddReadyzCheck("readyz", healthz.LeaderElectionCheck(mgr))
```

Same change in `controller/cmd/exporter-set-controller/main.go` and `controller/deploy/operator/cmd/main.go`.

This would:
- Make standby replicas report not-ready → removed from all Service endpoints
- Fix this PR's root cause (pod "Available" before leader election) at the source
- Also fix the existing HA bug where ~50% of gRPC traffic hits non-leader pods that aren't listening on port 8082

The shell-side wait in this PR is still good to keep as belt-and-suspenders for e2e, and the `BeforeAll` refactor is a standalone improvement. But the `LeaderElectionCheck` change could either be added to this PR or tracked as a follow-up.

---

mangelajo:
@bennyz I am preparing an alternate PR proposal

mangelajo:
@bennyz https://github.com/jumpstarter-dev/jumpstarter/pull/1012 WDYT?

mangelajo (approved):
It's not that easy, so I think this is an ok workaround for our CI for now.

mangelajo (approved):

PR #1009, opened 2026-08-18
fix: flush stderr after listening log for socket readiness visibility

bennyz:
seems like https://github.com/jumpstarter-dev/jumpstarter/pull/1008 did not fix the issue properly

### DIFF
--- a/rust/jumpstarter-exec/src/server.rs
+++ b/rust/jumpstarter-exec/src/server.rs
@@ -137,6 +137,9 @@ pub fn serve_with(socket_path: &str, opts: ServeOptions) -> std::io::Result<()> 
         "listening",
         &[("socket", json!(socket_path)), ("debug", json!(opts.debug))],
     );
+    // Ensure listening log is flushed so tests that poll the socket get
+    // visible evidence that the server is ready.
+    let _ = std::io::stderr().flush();
 
     while !state.is_shutdown() {
         match listener.accept() {
@@ -366,9 +369,8 @@ fn handle_exec(
                 _ => {}
             }
         }
-        if !reaped_ref.load(Ordering::Acquire) {
-            unsafe { kill(pid as i32, 15) }; // SIGTERM on client disconnect
-        }
+        // Don't kill child on client disconnect; let it finish naturally.
+        // The main thread will reap it with wait() after forwarding threads exit.
     });
 
     let status = child.wait()?;
--- a/rust/jumpstarter-exec/src/server.rs
+++ b/rust/jumpstarter-exec/src/server.rs
@@ -103,21 +103,56 @@ pub fn serve(socket_path: &str) -> std::io::Result<()> {
     serve_with(socket_path, ServeOptions::default())
 }
 
+/// Bind a Unix listener that is world-accessible (mode 0o666) the instant it
+/// appears at `socket_path`, with no window where it exists at another mode.
+/// Binds on a private temp path in the same directory, chmods it there, then
+/// publishes it at `socket_path` via `hard_link` — see the comment in
+/// `serve_with` for why this avoids `umask()`.
+///
+/// `hard_link` (not `rename`) is deliberate: `rename` atomically *replaces*
+/// an existing destination, so two `serve` processes racing to start on the
+/// same path could both publish successfully, with the loser's rename
+/// silently orphaning the winner's listener (still running, but no longer
+/// reachable through socket_path). `hard_link` fails with `AlreadyExists`
+/// instead of clobbering, so the loser gets a clean "already listening"
+/// error and the winner's listener is never displaced.
+fn bind_listen_socket(socket_path: &str) -> std::io::Result<UnixListener> {
+    let tmp_path = format!("{socket_path}.tmp-{}", std::process::id());
+    let result = (|| {
+        let listener = UnixListener::bind(&tmp_path)?;
+        #[cfg(unix)]
+        {
+            use std::os::unix::fs::PermissionsExt;
+            std::fs::set_permissions(&tmp_path, std::fs::Permissions::from_mode(0o666))?;
+        }
+        listener.set_nonblocking(true)?;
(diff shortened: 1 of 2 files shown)


rust/jumpstarter-exec/src/server.rs
```diff
@@ -366,9 +369,8 @@ fn handle_exec(
                 _ => {}
             }
         }
-        if !reaped_ref.load(Ordering::Acquire) {
-            unsafe { kill(pid as i32, 15) }; // SIGTERM on client disconnect
-        }
+        // Don't kill child on client disconnect; let it finish naturally.
+        // The main thread will reap it with wait() after forwarding threads exit.
```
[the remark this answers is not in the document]
mangelajo: @bennyz wdyt about coderabbitai comment?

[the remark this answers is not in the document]
bennyz: checking it, tbh i am not sure about the PR yet, I am still trying to understand what happened with that test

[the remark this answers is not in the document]
mangelajo: @bennyz do you have a reference to the issue? I'll have a look at it as well.

rust/jumpstarter-exec/tests/integration.rs
```diff
@@ -23,9 +23,27 @@ fn binary_path() -> PathBuf {
...
 /// Wait (up to 10s) for the server's listening socket to appear.
-fn wait_for_socket(sock: &std::path::Path) {
+///
+/// The socket file is created synchronously by `UnixListener::bind`, so its
+/// presence is the readiness signal. If the server *process* dies before
+/// binding (spawn failure, panic, bind error), polling the file alone would
+/// block for the full deadline and then report a bare "never appeared" with no
+/// cause. So we also poll the child: on early exit, drain its stderr and fail
+/// immediately with the exit status and logs.
+fn wait_for_socket(sock: &std::path::Path, child: &mut Child) {
```
mangelajo, line 33: Nit: the PR title says "flush stderr after listening log for socket readiness visibility" and the commit message says "flush stderr after listening log and don't kill child on disconnect" — but neither of those things happen in this diff. The actual change is improving test infrastructure to detect early server failures during socket-wait polling. Consider updating the PR title/description to match what the code actually does, e.g. "fix: detect early server exit in test socket-wait loops".

---

rust/jumpstarter-exec/tests/integration.rs
```diff
@@ -23,9 +23,27 @@ fn binary_path() -> PathBuf {
...
+            let mut stderr = String::new();
+            if let Some(mut err) = child.stderr.take() {
+                use std::io::Read;
+                err.read_to_string(&mut stderr).ok();
+            }
+            panic!(
+                "server process exited early ({status}) before creating socket {}; stderr:\n{stderr}",
+                sock.display()
+            );
+        }
```
mangelajo, line 46: Good improvement — polling the child for early exit gives much better diagnostics than waiting for the full 10s deadline.

One thing to consider: `child.stderr.take()` will only have content if stderr was piped at spawn time. Looking at `start_server_process()`, it sets `.stderr(Stdio::piped())`, so this is fine there. But the two e2e tests (`e2e_debug_json_logs_commands_and_io` and `e2e_log_fields_appear_on_every_line`) that also call `wait_for_socket` also pipe stderr — so this works correctly everywhere.

Minor: if the child exits *successfully* (status 0) before creating the socket (e.g. some misconfiguration), this would panic with `"server process exited early (exit status: 0)"` which is correct but might be slightly confusing. Not actionable — just noting.

---

rust/jumpstarter-exec/tests/integration.rs
```diff
@@ -108,10 +126,41 @@ fn start_server() -> (TempDir, String) {
...
+            // Still starting up.
+            Err(std::sync::mpsc::TryRecvError::Empty) => {}
+        }
+        assert!(
+            std::time::Instant::now() < deadline,
+            "in-process server socket never appeared at {}",
+            sock.display()
+        );
+        thread::sleep(Duration::from_millis(10));
+    }
```
mangelajo, line 163: This is a solid improvement over the previous `thread::sleep(Duration::from_millis(100))` — the fixed sleep was inherently racy under CPU pressure.

The channel-based approach here is thorough: it correctly handles all four cases:
- `Ok(Err(e))`: `serve()` returned an error (bind failure, etc.)
- `Ok(Ok(()))`: `serve()` returned successfully before binding (shouldn't happen, but defended)
- `Disconnected`: thread panicked before sending
- `Empty`: still starting up (continue polling)

One edge case worth considering: if `serve()` successfully binds but the socket file creation and the `sock.exists()` check race such that `serve()` returns `Ok(())` (e.g., from a shutdown on another connection) *after* binding but *before* this loop sees the socket file, the `Ok(Ok(()))` arm would fire a false-positive panic. In practice this shouldn't happen in tests since nothing sends Shutdown before the test starts, so this is fine.

---

rust/jumpstarter-exec/tests/integration.rs
```diff
@@ -108,10 +126,41 @@ fn start_server() -> (TempDir, String) {
...
+            ),
+            // Still starting up.
+            Err(std::sync::mpsc::TryRecvError::Empty) => {}
+        }
+        assert!(
+            std::time::Instant::now() < deadline,
+            "in-process server socket never appeared at {}",
+            sock.display()
+        );
+        thread::sleep(Duration::from_millis(10));
```
mangelajo, line 162: Minor inconsistency: `wait_for_socket` (the subprocess version) sleeps for 20ms between polls, while this in-process version sleeps for 10ms. Not a big deal — both are fine — but aligning them would be slightly cleaner.

---

mangelajo (approved):

bennyz:
seems like this fix helped catch the bug, pushing another commit

PR #1011, opened 2026-08-20
docs(e2e): add gen-spec-index to keep README synced

maboras-rh:
## Summary

This PR adds a lightweight, repeatable workflow to keep `e2e/README.md` aligned with the actual Go/Ginkgo E2E suite.

- Adds `e2e/scripts/gen-spec-index.sh` to generate a lane-grouped spec index (group, test name, labels, source) from `ginkgo --dry-run`.
- Adds `make e2e-spec-index` in the top-level `Makefile`.
- Updates `e2e/README.md` to document how to use the generated spec index when checking for drift.
- Updates `.cursor/rules/e2e-doc-sync.mdc` with an explicit “recommended workflow” that uses `make e2e-spec-index`.

## Why

The E2E README is hand-maintained for prerequisites/steps/pass checks, which is useful but can drift on lane/test/source metadata over time.

This change gives contributors a quick, no-cluster-needed way to verify metadata before editing the README, reducing drift and review churn.

## Validation

- Ran `make e2e-spec-index > /tmp/e2e-spec-index.md`.
- Cross-checked generated lane/label/test/source output against `e2e/README.md`.
- Confirmed generated scratch output is not committed.

## Impact

- Easier ongoing maintenance of E2E documentation.
- Faster and more reliable README sync checks during E2E test changes.

Reference: https://github.com/centos-automotive-suite/automotive-dev-operator/pull/556#issuecomment-5328601468


maboras-rh:
@coderabbit review

e2e/scripts/gen-spec-index.sh
```diff
@@ -0,0 +1,179 @@
...
+hint: verify dependencies and suite compilation with:
+  cd e2e/test && go build ./...
+EOF
+  exit 1
+fi
+
+jq -e -r --arg label_filter "$LABEL_FILTER" '
+  def known_lanes:
+    ["core", "hooks", "direct-listener", "dut-network", "exit-on-lease-end",
+     "exporterset-qemu", "auth-logging", "old-controller", "old-client", "compat"];
```
mangelajo: The `known_lanes` list is hardcoded in the jq filter. If a new lane label is added to the test suite (e.g., a new `Describe` with a new `Label`), this list must be manually updated or the new lane will silently fall back to alphabetic-first-label classification — which can produce confusing groupings.

Consider adding a comment near the top of the script (or in the jq block) noting that this list must stay in sync with the Ginkgo labels used in `e2e/test/`. Alternatively, this could be extracted to a variable or external config to make it more discoverable.

Some agent instructions to keep this up to date can also be beneficial in `.cursor/rules/e2e-doc-sync.mdc` 

---

e2e/scripts/gen-spec-index.sh
```diff
@@ -0,0 +1,179 @@
...
+  def rel_file:
+    (.LeafNodeLocation.FileName | sub("^.*/e2e/test/"; ""));
+
+  def source:
+    rel_file + ":" + (.LeafNodeLocation.LineNumber | tostring);
+
+  def esc_md:
+    tostring
+    | gsub("\\|"; "\\\\|")
+    | gsub("\n"; " ");
```
mangelajo, line 144: `esc_md` escapes `|` (for table cell boundaries) and newlines, but doesn't handle other Markdown inline formatting characters like backticks, `*`, or `_`. If a test name or group text contains these (e.g., a test named `` can use `j power on` ``), the resulting Markdown table could render incorrectly.

This is unlikely to be a problem in practice given the current test naming conventions, but worth noting for robustness. If you want to be thorough:

```jq
  def esc_md:
    tostring
    | gsub("\\\\"; "\\\\\\\\")
    | gsub("\\|"; "\\\\|")
    | gsub("`"; "\\\\`")
    | gsub("\\*"; "\\\\*")
    | gsub("_"; "\\\\_")
    | gsub("\n"; " ");
```

---

mangelajo (changes requested):
a couple of tiny comments. mostly nits to be honest..

e2e/scripts/gen-spec-index.sh
maboras-rh: good catch, addressed in d080af3
`known_lanes` is now a top-level bash variable `KNOWN_LANES_JSON` at the top of
the script, with a prominent "Keep in sync with e2e/test/*.go lane Label(...) values"
comment both in the file header and inline. It's passed to jq via
`--argjson known_lanes "$KNOWN_LANES_JSON"`, so the list is no longer buried
inside the jq filter body and is easy to spot/update.
Also updated agent rule `e2e-doc-sync` to update the known lanes when adding/renaming/removing a Ginkgo lane `Label(...)`, in the same PR and keep it ordered as the desired Markdown section order.

e2e/scripts/gen-spec-index.sh
maboras-rh: applied exactly the suggestion `esc_md` now also escapes backslash first (to
avoid double-escaping later passes), then backtick, `*`, and `_`, in addition
to the existing `|` and newline handling.

mangelajo (approved):

PR #1012, opened 2026-08-20
fix: tie readiness probe to leader election in all controllers

mangelajo:
## Summary

Alternative approach to #1002 — instead of polling the Lease object from shell scripts, fix the root cause: controllers report ready before they've won leader election.

Replaces `healthz.Ping` with a leader-election-aware readiness checker for the `/readyz` endpoint in all three controller binaries. The checker uses `mgr.Elected()` (a channel closed when the manager wins leader election, or immediately if leader election is disabled) to gate readiness.

## What this fixes

**1. E2E race condition (the #1002 root cause)**

Pods were reporting ready (passing `/readyz`) before leader election completed (~57s observed in CI). The E2E `wait_for_jumpstarter_resources` uses `kubectl wait --for=condition=available`, which was satisfied immediately — but the controller couldn't reconcile yet. With this change, the existing readiness wait naturally blocks until the leader is elected.

**2. HA traffic routing bug**

With 2+ replicas (the default is 2), the gRPC `ControllerService` (port 8082) only runs on the leader pod — `controller-runtime` defaults `NeedLeaderElection()` to `true`, so `Start()` never executes on standby replicas. But `healthz.Ping` always returned OK, so Kubernetes included **both** pods in Service endpoints. Result: ~50% of gRPC connections hit the non-leader pod and got **connection refused**. Now non-leader pods report not-ready and are removed from endpoints.

## Changes

| File | Change |
|------|--------|
| `controller/cmd/main.go` | Readyz check: `healthz.Ping` → `leaderElectionCheck(mgr)` |
| `controller/cmd/exporter-set-controller/main.go` | Same |
| `controller/deploy/operator/cmd/main.go` | Same |
| `controller/deploy/operator/.../jumpstarter_controller.go` | Controller Deployment `maxUnavailable`: `25%` → `max(replicas-1, 1)` |
| `controller/hack/deploy_with_operator.sh` | E2E controller replicas: 1 → 2 |

The `leaderElectionCheck` helper is a simple non-blocking select on `mgr.Elected()`:

```go
func leaderElectionCheck(mgr manager.Manager) healthz.Checker {
    return func(_ *http.Request) error {
        select {
        case <-mgr.Elected():
            return nil
        default:
            return fmt.Errorf("not yet leader")
        }
    }
}
```

Since the controller is active/passive by design (only the leader runs gRPC and reconciliation), at most 1 replica will ever pass readiness. The operator now sets `maxUnavailable` to `replicas - 1` (minimum 1) so the Deployment reaches `Available` with just the leader pod ready.

## E2E impact

- Controller replicas bumped from 1 to 2 so the HA readiness behavior is exercised in CI
- No changes needed to `wait_for_jumpstarter_resources` — the existing `kubectl wait --for=condition=available` now correctly blocks until leader election completes, since pods won't pass readiness until then
- The dynamic `maxUnavailable` ensures the deployment becomes `Available` once the leader is ready, regardless of how many standby replicas exist

Relates to #1002

### DIFF
--- a/controller/cmd/exporter-set-controller/main.go
+++ b/controller/cmd/exporter-set-controller/main.go
@@ -19,7 +19,6 @@ package main
 import (
 	"flag"
 	"fmt"
-	"net/http"
 	"os"
 	"time"
 
@@ -32,13 +31,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	jumpstarterdevv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/api/v1alpha1"
 	virtualtargetv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/api/virtualtarget/v1alpha1"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/exporterset"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/exporterset/provisioners/qemu"
+	jmphealthz "github.com/jumpstarter-dev/jumpstarter/controller/internal/healthz"
 )
 
 var (
@@ -145,7 +144,7 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", leaderElectionCheck(mgr)); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", jmphealthz.LeaderElectionCheck(mgr)); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
@@ -165,18 +164,5 @@ func selectProvisioner(name string) (exporterset.Provisioner, error) {
 		return qemu.New(version), nil
 	default:
 		return nil, fmt.Errorf("unknown provisioner %q; supported: %s", name, qemu.ProvisionerName)
-	}
-}
-
-// leaderElectionCheck returns a healthz.Checker that reports not-ready until
-// the manager has been elected leader (or leader election is disabled).
-func leaderElectionCheck(mgr manager.Manager) healthz.Checker {
-	return func(_ *http.Request) error {
-		select {
-		case <-mgr.Elected():
-			return nil
-		default:
-			return fmt.Errorf("not yet leader")
-		}
 	}
 }
--- a/controller/cmd/exporter-set-controller/main.go
+++ b/controller/cmd/exporter-set-controller/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"net/http"
 	"os"
 	"time"
 
@@ -31,6 +32,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 
 	jumpstarterdevv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/api/v1alpha1"
@@ -143,7 +145,7 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", leaderElectionCheck(mgr)); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
@@ -163,5 +165,18 @@ func selectProvisioner(name string) (exporterset.Provisioner, error) {
 		return qemu.New(version), nil
 	default:
 		return nil, fmt.Errorf("unknown provisioner %q; supported: %s", name, qemu.ProvisionerName)
+	}
+}
+
+// leaderElectionCheck returns a healthz.Checker that reports not-ready until
+// the manager has been elected leader (or leader election is disabled).
+func leaderElectionCheck(mgr manager.Manager) healthz.Checker {
+	return func(_ *http.Request) error {
+		select {
+		case <-mgr.Elected():
+			return nil
+		default:
+			return fmt.Errorf("not yet leader")
+		}
 	}
 }
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -21,7 +21,9 @@ import (
 	"crypto/tls"
 	"encoding/pem"
 	"flag"
+	"fmt"
 	"net"
+	"net/http"
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -40,6 +42,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
@@ -324,7 +327,7 @@ func main() {
 		setupLog.Error(err, "unable to set up health check")
 		os.Exit(1)
 	}
-	if err := mgr.AddReadyzCheck("readyz", healthz.Ping); err != nil {
+	if err := mgr.AddReadyzCheck("readyz", leaderElectionCheck(mgr)); err != nil {
 		setupLog.Error(err, "unable to set up ready check")
 		os.Exit(1)
 	}
@@ -383,4 +386,20 @@ func jwtAuthenticatorsToOIDCConfigs(authenticators []apiserverv1beta1.JWTAuthent
 		})
 	}
 	return configs
+}
+
+// leaderElectionCheck returns a healthz.Checker that reports not-ready until
+// the manager has been elected leader (or leader election is disabled).
+// This ensures that non-leader replicas are removed from Service endpoints,
+// preventing traffic from reaching pods that are not actively reconciling
+// or serving gRPC.
+func leaderElectionCheck(mgr manager.Manager) healthz.Checker {
+	return func(_ *http.Request) error {
+		select {
+		case <-mgr.Elected():
+			return nil
+		default:
+			return fmt.Errorf("not yet leader")
+		}
+	}
 }
--- a/controller/cmd/main.go
+++ b/controller/cmd/main.go
@@ -21,9 +21,7 @@ import (
 	"crypto/tls"
 	"encoding/pem"
 	"flag"
-	"fmt"
 	"net"
-	"net/http"
 	"os"
 
 	// Import all Kubernetes client auth plugins (e.g. Azure, GCP, OIDC, etc.)
@@ -42,7 +40,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
-	"sigs.k8s.io/controller-runtime/pkg/manager"
 	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
@@ -51,6 +48,7 @@ import (
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/authorization"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/config"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/controller"
+	jmphealthz "github.com/jumpstarter-dev/jumpstarter/controller/internal/healthz"
 	jmpmetrics "github.com/jumpstarter-dev/jumpstarter/controller/internal/metrics"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/oidc"
 	"github.com/jumpstarter-dev/jumpstarter/controller/internal/service"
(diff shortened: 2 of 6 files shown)


controller/cmd/main.go
```diff
@@ -384,3 +387,19 @@ func jwtAuthenticatorsToOIDCConfigs(authenticators []apiserverv1beta1.JWTAuthent
...
+
+// leaderElectionCheck returns a healthz.Checker that reports not-ready until
+// the manager has been elected leader (or leader election is disabled).
+// This ensures that non-leader replicas are removed from Service endpoints,
+// preventing traffic from reaching pods that are not actively reconciling
+// or serving gRPC.
+func leaderElectionCheck(mgr manager.Manager) healthz.Checker {
+	return func(_ *http.Request) error {
+		select {
+		case <-mgr.Elected():
```
mangelajo: The `leaderElectionCheck` implementation is correct — `mgr.Elected()` returns a channel that's closed when the manager wins the election (or immediately if LE is disabled), and the non-blocking `select` is the right pattern for a health checker.

One thing worth noting: if leader election is disabled (the default for `--leader-elect` is `false`), the `Elected()` channel is closed immediately, so this is backward compatible and won't break single-replica / non-LE deployments.

controller/cmd/main.go
```diff
@@ -384,3 +387,19 @@ func jwtAuthenticatorsToOIDCConfigs(authenticators []apiserverv1beta1.JWTAuthent
...
+			return nil
+		default:
+			return fmt.Errorf("not yet leader")
+		}
```
mangelajo: Nit: the `leaderElectionCheck` function is copy-pasted identically across all three binaries (`cmd/main.go`, `cmd/exporter-set-controller/main.go`, `deploy/operator/cmd/main.go`). It's small enough that I don't think it's a blocker, but it could be extracted to a shared internal package (e.g., `internal/healthz`) to avoid the duplication. Up to you whether that's worth it for a ~10 line function.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -898,8 +898,12 @@ func (r *JumpstarterReconciler) createControllerDeployment(jumpstarter *operator
...
 				RollingUpdate: &appsv1.RollingUpdateDeployment{
-					MaxSurge:       &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
-					MaxUnavailable: &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
+					MaxSurge: &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
+					// The controller is active/passive: only the leader pod passes
+					// readiness (see leaderElectionCheck), so at most 1 replica will
+					// ever be ready. Allow all standby replicas to be unavailable so
+					// the Deployment can reach the Available condition.
+					MaxUnavailable: &intstr.IntOrString{Type: intstr.Int, IntVal: max(jumpstarter.Spec.Controller.Replicas-1, 1)},
 				},
```
mangelajo, line 907: `Replicas` is `int32`, so `Replicas - 1` is `int32` arithmetic. When `Replicas == 1` (the minimum enforced by the kubebuilder validation `+kubebuilder:validation:Minimum=1`), this evaluates to `max(0, 1) = 1`, which is correct.

For any `Replicas >= 2`, it evaluates to `Replicas - 1`, which is also correct — it allows all standby pods to be unavailable since only the leader will ever be ready.

The Go built-in `max()` (available since Go 1.21, project uses 1.24) handles `int32` correctly. Looks good.

controller/hack/deploy_with_operator.sh
```diff
@@ -213,7 +213,7 @@ ${AUTH_CONFIG}
   controller:
     image: ${IMAGE_REPO}
     imagePullPolicy: IfNotPresent
-    replicas: 1
+    replicas: 2
```
mangelajo, line 216: Good call bumping this to 2 replicas in CI — this exercises the HA readiness behavior end-to-end. Since the `maxUnavailable` change ensures the Deployment reaches `Available` with just 1 ready pod, the existing `kubectl wait --for=condition=available` should work without needing a longer timeout.

Worth watching the first few CI runs to confirm there's no additional delay from the second replica's readiness probe failures before it stabilizes (3 failures * 10s period = 30s of probe attempts on the standby pod, but this shouldn't affect the deployment's `Available` condition).

mangelajo (commented):
self review from another agent, this is what I found, and handled the duplication.

This is a clean, well-reasoned fix that addresses the root cause rather than papering over it with shell-script polling (as in #1002).

**What I like:**

1. **Correct approach** — tying readiness to `mgr.Elected()` is the idiomatic controller-runtime solution. The non-blocking select on the channel is exactly right for a health checker.

2. **Backward compatible** — when leader election is disabled (the `--leader-elect` default is `false`), the `Elected()` channel is closed immediately, so this doesn't break single-replica deployments.

3. **Fixes two bugs at once** — the E2E race (pods reporting ready before they can reconcile) and the HA traffic routing issue (~50% of gRPC connections hitting the non-leader pod).

4. **`maxUnavailable` math is correct** — `max(Replicas-1, 1)` handles the `Replicas=1` edge case properly, and the `int32` arithmetic is safe given the kubebuilder `Minimum=1` validation.

5. **Good E2E coverage** — bumping CI replicas to 2 exercises the HA path that was previously untested.

**Minor nit (non-blocking):** `leaderElectionCheck` is copy-pasted across all 3 binaries. Could be extracted to a shared internal package but it's small enough that it's not a real concern.

CI is still running — worth confirming the e2e-test-operator passes with 2 replicas before merging.

LGTM overall.

controller/hack/deploy_with_operator.sh
mangelajo, line 216: just to make sure we test this config in CI. it will mean a 100MB idle process, but ok..

controller/cmd/exporter-set-controller/main.go
```diff
@@ -165,3 +167,16 @@ func selectProvisioner(name string) (exporterset.Provisioner, error) {
 		return nil, fmt.Errorf("unknown provisioner %q; supported: %s", name, qemu.ProvisionerName)
 	}
 }
+
+// leaderElectionCheck returns a healthz.Checker that reports not-ready until
```
mangelajo: ok, this function is replicated in 3 places, let's see if it's viable to keep it in a single one...

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -898,8 +898,12 @@ func (r *JumpstarterReconciler) createControllerDeployment(jumpstarter *operator
...
 				Type: appsv1.RollingUpdateDeploymentStrategyType,
 				RollingUpdate: &appsv1.RollingUpdateDeployment{
-					MaxSurge:       &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
-					MaxUnavailable: &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
+					MaxSurge: &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
+					// The controller is active/passive: only the leader pod passes
+					// readiness (see leaderElectionCheck), so at most 1 replica will
+					// ever be ready. Allow all standby replicas to be unavailable so
+					// the Deployment can reach the Available condition.
+					MaxUnavailable: &intstr.IntOrString{Type: intstr.Int, IntVal: max(jumpstarter.Spec.Controller.Replicas-1, 1)},
```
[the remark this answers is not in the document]
mangelajo: Looking at this... when do the MaxUnavailable mechanisms trigger in contrast to leader election handover.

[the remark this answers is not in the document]
mangelajo: The single-replica rollout concern is valid in theory but not a real problem for this use case. Here's why:

**What actually happens with `Replicas=1, MaxUnavailable=1, MaxSurge=25%` (rounds to 1):**

1. Rollout starts → surge pod created (2 pods total)
2. Surge pod can't become ready (old pod holds the lease)
3. `MaxUnavailable=1` means K8s requires `desired - maxUnavailable = 0` minimum ready pods, so it **can** terminate the old leader immediately
4. Old pod receives SIGTERM. Since `LeaderElectionReleaseOnCancel` is `false` (commented out in `cmd/main.go:182`), the lease is **not** released on shutdown — it expires naturally
5. New pod acquires the lease after `LeaseDuration` (15s default) expires, then passes `/readyz`

**Total gap with no ready controller: ~15-20s.** This is identical to what happens with *every* controller-runtime operator during rollouts (including `kube-controller-manager` itself). It's expected and safe because:
- Reconciliation events are queued and replayed when the new leader starts
- gRPC clients reconnect automatically
- `ProgressDeadlineSeconds=600s` gives plenty of headroom

**Why `MaxUnavailable=0` would be wrong (deadlock):**
- Old leader stays ready, new pod can't become ready (not leader), K8s can't terminate the old pod → rollout stalls for 600s until `ProgressDeadlineSeconds` fails it. This is strictly worse.

**Why "coordinated leader handoff" is unnecessary:**
- This controller is not a latency-sensitive user-facing service — it's a reconciliation controller. A 15s gap during upgrades is a non-event.
- If reducing the gap were desired, the simplest fix is uncommenting `LeaderElectionReleaseOnCancel: true` (already in the code at `cmd/main.go:182`), which would drop the gap from ~15s to ~2s. But that's a separate improvement, not a blocker for this PR.

**Summary:** The current `max(Replicas-1, 1)` is the correct formula. It avoids the `MaxUnavailable=0` deadlock, and the brief unavailability during single-replica rollouts is standard Kubernetes controller behavior — not a bug.

bkhizgiy (commented):
I looked at both this PR and #1002. I do think this one is the better direction long term, since it addresses the actual production issue rather than just the E2E race. If we want a quick fix, I’d go with merging #1002 first to stabilize the CI, and then spend a bit more time on this one, since it’s a more complex solution and proboly needs a bit more work before merging.

I also got the following feedback after running an AI review on this PR, it looks like we may have a potential deadlock with the current implementation:

```
Consider the steady state with replicas: 2:

Old leader pod: Ready (holds the lease)
Old standby pod: Not Ready (can't win election)
Available = 1, Unavailable = 1
Now a rolling update is triggered (image change, config change, etc.):

maxSurge = ceil(2 * 0.25) = 1 → max total pods = 3
maxUnavailable = max(2-1, 1) = 1 → minAvailable = replicas - maxUnavailable = 2 - 1 = 1
The Deployment controller's logic:

scaleDownCount = currentAvailable - minAvailable = 1 - 1 = 0
It can never scale down the old ReplicaSet because the leader is the only available pod and terminating it would violate minAvailable.

Meanwhile, new pods cannot become Ready because:

Leader election requires the current holder to stop renewing the lease
The old leader is alive and healthy, continuously renewing
New pods will never win election → never become Ready
Result: circular dependency.

New pods need old leader terminated → to win election → to become Ready
Old leader can't be terminated → because available(1) - 1 < minAvailable(1)
The ProgressDeadlineSeconds: 600 will eventually mark the Deployment as Progressing=False, but the update will never actually complete.

This affects all replica counts
With replicas = N:

maxUnavailable = max(N-1, 1) = N-1
minAvailable = N - (N-1) = 1
Available is always exactly 1 (only the leader)
scaleDownCount = 1 - 1 = 0 — always blocked
```

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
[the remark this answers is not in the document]
mangelajo: Correction to my earlier analysis — after reading the [Kubernetes Deployment controller source](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/rolling.go), **Bella is right** about the deadlock for `Replicas >= 2`. My previous analysis was incorrect for multi-replica scenarios.

Here's the actual Kubernetes scale-down logic in `scaleDownOldReplicaSetsForRollingUpdate`:

```go
minAvailable := *(deployment.Spec.Replicas) - maxUnavailable
availablePodCount := deploymentutil.GetAvailableReplicaCountForReplicaSets(allRSs)
if availablePodCount <= minAvailable {
    return 0, nil  // Cannot scale down
}
```

**Replicas=2, MaxUnavailable=1 trace:**

| Step | State | Result |
|------|-------|--------|
| Steady state | Old leader (available), old standby (not available) | available=1 |
| Rollout starts | Surge pod created (new RS, not available) | allPods=3 |
| `cleanupUnhealthyReplicas` | Removes old standby (unhealthy) | allPods=2 |
| Next cycle: `scaleDownOldReplicaSetsForRollingUpdate` | `availablePodCount=1, minAvailable=1` → `1 <= 1` | **Cannot scale down** |

The old leader can't be terminated because it's the only available pod. The new pod can't become available because the old leader holds the lease. **Deadlock.** The rollout stalls until `ProgressDeadlineSeconds` (600s) marks it failed.

**Replicas=1, MaxUnavailable=1 still works:**
- `minAvailable = 1 - 1 = 0`
- `availablePodCount=1 > 0` → can scale down the old leader

So the formula `max(Replicas-1, 1)` is correct for `Replicas=1` but causes a deadlock for `Replicas >= 2`. This needs to be fixed.

mangelajo:
Verified on my side, and also makes sense to me:

@bkhizgiy Good catch — I verified your deadlock analysis against the [Kubernetes Deployment controller source](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/rolling.go) and it's correct.

The key is in `scaleDownOldReplicaSetsForRollingUpdate`:

```go
availablePodCount := deploymentutil.GetAvailableReplicaCountForReplicaSets(allRSs)
if availablePodCount <= minAvailable {
    return 0, nil  // Cannot scale down
}
```

With `Replicas=2, MaxUnavailable=1`: `minAvailable = 2 - 1 = 1`. Only the leader is ever available (=1), so `1 <= 1` → K8s refuses to scale down the old leader. New pods can't win election while the old leader is alive. Deadlock.

**`Replicas=1` is fine** — `minAvailable = 0`, so the leader can be terminated and the new pod eventually acquires the lease after the 15s `LeaseDuration` expires.

**To fix `Replicas >= 2`, the options are:**

1. **`MaxUnavailable = Replicas`** — allows K8s to terminate all old pods including the leader. Simple, but means a brief period with no ready pods during rollouts (same as the `Replicas=1` case). This is acceptable for a reconciliation controller.

2. **Enable `LeaderElectionReleaseOnCancel: true`** — makes the leader release the lease on SIGTERM, so the surge pod can acquire it before the old leader fully terminates. This helps but doesn't fix the `scaleDownCount` calculation — K8s still won't terminate the old leader if `availablePodCount <= minAvailable`.

3. **Use `Recreate` strategy instead of `RollingUpdate`** — terminates all old pods first, then creates new ones. No deadlock possible. Downtime is bounded by lease acquisition (~15s or ~2s with `ReleaseOnCancel`).

Option 1 is the simplest fix: change `max(Replicas-1, 1)` to just `Replicas`. For a controller that tolerates brief unavailability during rollouts, this is the right trade-off.

mangelajo:
Until we identify how to properly handle this I have opened https://github.com/jumpstarter-dev/jumpstarter/pull/1014/changes which we can probably merge along with #1002

PR #1022, opened 2026-08-25
fix: gate lease slot release on session teardown

### LINKED ISSUE #1024: Lease context clearing introduced a race condition causing a lease to get stuck in IDLE
#948 introduced a regressions:

What 948 changed. Before 948, the handle_lease fallback was a check-then-clear with no await in between:
```
  if self._lease_context is lease_scope:
      self._lease_context = None
```

948 wrapped that in try/finally, and put a 0.2s settle sleep inside the if, then still cleared unconditionally:
``
   python/packages/jumpstarter/jumpstarter/exporter/exporter.py lines 1204-1209

                  if self._lease_context is lease_scope:
                      session_was_created = lease_scope.session is not None
                      if session_was_created:
                          await sleep(0.2)
                      self._last_completed_lease = lease_scope.lease_name
                      self._lease_context = None
```
The identity check is not repeated after the sleep. That is the TOCTOU.

How it matches the logs. The pagination spec created 10 leases that all selected test-exporter-oidc, then delete --all. The exporter did this:

  1. pagination-lease-1 ended (Currently not leased → AVAILABLE)
  2. ~100ms later it acquired pagination-lease-8 and reported LEASE_READY
  3. Then another Currently not leased without Lease ended, signaling handle_lease…
  4. Status stuck at LeaseReady| (empty leaseRef) for the rest of the suite

That third step is _on_lease_released seeing _lease_state == IDLE because _lease_context was already None. The only writer that can do that while lease-8 is live is lease-1’s finally, waking from the 0.2s sleep and wiping the field.

bennyz:
#948 moved `await sleep(0.2)`
1. _on_lease_released finishes its own sleep, clears, returns
2. status loop acquires lease-B
3. A’s finally wakes and sets _lease_context = None
4. B’s handle_lease is now running with IDLE state: LeaseReady, empty leaseRef. 

This PR switches to a single control-plane writer for the lease's slot:
image

fixes https://github.com/jumpstarter-dev/jumpstarter/issues/1024

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,10 @@ help:
 	@echo ""
 	@echo "End-to-end testing:"
 	@echo "  make e2e-setup             - Setup e2e test environment (one-time)"
-	@echo "  make e2e-run               - Run full e2e suite (includes ExporterSet QEMU)"
+	@echo "  make e2e-run               - Run e2e suite (excludes lease-churn; includes ExporterSet QEMU)"
 	@echo "  make e2e                   - Same as e2e-run"
 	@echo "  make e2e-exporterset-qemu  - Run ExporterSet QEMU e2e only"
+	@echo "  make e2e-lease-churn       - Run core e2e including lease-churn cycles"
 	@echo "  make e2e-full              - Full setup + run (for CI or first time)"
 	@echo "  make e2e-clean             - Clean up e2e test environment (delete cluster, certs, etc.)"
 	@echo ""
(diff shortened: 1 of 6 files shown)


mangelajo:
@bennyz , do you have a link to the bug? what's going on?

bennyz:
> @bennyz , do you have a link to the bug? what's going on?

i've seen e2e failures https://github.com/jumpstarter-dev/jumpstarter/actions/runs/32812149379/job/97694251760

mangelajo (approved):

PR #1023, opened 2026-08-25
feat: Add TLS support for jumpstarter-telemetry with the operator

bkhizgiy:
Add TLS support to the telemetry deployment when cert-manager is enabled.
Include the telemetry CA certificate in the controller ConfigMap so exporters can verify TLS connections.
Add test coverage for the this functionality


controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -273,6 +273,49 @@ func createTelemetryDeployment(jumpstarter *operatorv1alpha1.Jumpstarter) *appsv
...
+		volumes = append(volumes, corev1.Volume{
+			Name: "tls-certs",
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName:  tlsSecretName,
+					DefaultMode: &defaultMode,
+				},
+			},
+		})
+	}
```
mangelajo: **Missing TLS hash annotation — cert renewals won't trigger pod restarts**

The controller and router deployments both compute a TLS secret hash and inject it as a pod annotation (see `getControllerTLSSecretHash` at line 262, `getRouterTLSSecretHash` at line 282 of `jumpstarter_controller.go`). This ensures that when cert-manager renews a certificate, the hash changes and triggers a rolling restart so the pod picks up the new cert.

The telemetry deployment has no equivalent. When a telemetry TLS certificate is renewed by cert-manager, the telemetry pod will **not** automatically restart. The only mechanism to pick up the new cert would be the 30-minute requeue timer, and even that just reconciles the deployment spec — it doesn't force a pod restart because the pod template hasn't changed.

Consider adding a `getTelemetryTLSSecretHash` function (following the pattern of `getControllerTLSSecretHash`) and passing the hash into `createTelemetryDeployment` as a pod annotation.

---

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -273,6 +273,49 @@ func createTelemetryDeployment(jumpstarter *operatorv1alpha1.Jumpstarter) *appsv
...
+			},
+		},
+	}
+
+	var volumeMounts []corev1.VolumeMount
+	var volumes []corev1.Volume
+
+	// Add TLS certificate mount when cert-manager is enabled
+	if jumpstarter.Spec.CertManager.Enabled {
+		tlsSecretName := getTelemetryCertSecretName(jumpstarter)
```
mangelajo: **Missing secret index — changes to the telemetry TLS secret won't trigger reconciliation**

In `SetupWithManager` (`jumpstarter_controller.go:1726-1740`), the controller and router TLS secrets are indexed via `indexReferencedSecret`, so when those secrets change (e.g., cert-manager issues a new cert), the watch fires and triggers a reconcile immediately.

The telemetry TLS secret (`{name}-telemetry-tls`) is **not** indexed here, so changes to it won't trigger a reconciliation via the Secret watch. Only the 30-minute requeue would eventually detect it. This is a notable gap compared to how the other components are wired up.

---

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -273,6 +273,49 @@ func createTelemetryDeployment(jumpstarter *operatorv1alpha1.Jumpstarter) *appsv
...
+				},
+			},
+		},
+	}
+
+	var volumeMounts []corev1.VolumeMount
+	var volumes []corev1.Volume
+
+	// Add TLS certificate mount when cert-manager is enabled
+	if jumpstarter.Spec.CertManager.Enabled {
```
mangelajo: **No manual TLS path when cert-manager is disabled — inconsistent with controller/router**

The controller and router deployments both support a dual TLS path:
1. **cert-manager enabled** → auto-managed secret (e.g., `{name}-controller-tls`)
2. **cert-manager disabled + manual `CertSecret`** → user-provided secret via `Spec.Controller.GRPC.TLS.CertSecret` / `Spec.Routers.GRPC.TLS.CertSecret`

See the controller pattern at `jumpstarter_controller.go:870-875`:
```go
if jumpstarter.Spec.CertManager.Enabled {
    tlsSecretName = GetControllerCertSecretName(jumpstarter)
} else if jumpstarter.Spec.Controller.GRPC.TLS.CertSecret != "" {
    tlsSecretName = jumpstarter.Spec.Controller.GRPC.TLS.CertSecret
}
```

The telemetry implementation only supports cert-manager with no fallback for manual certs. The `TelemetryConfig` CRD struct has no `TLS`/`CertSecret` field.

This may be intentional for now, but it means users who deploy without cert-manager (e.g., providing their own certificates) can have TLS on the controller and routers but **not** on telemetry. Worth documenting this limitation, or considering adding a `TLS` field to `TelemetryConfig` for parity.

---

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -750,3 +803,92 @@ var _ = Describe("resolveTelemetryCA", func() {
...
+		r := &JumpstarterReconciler{Client: k8sClient, Scheme: k8sClient.Scheme()}
+		cfg, err := r.buildConfig(ctx, js)
+		// Should not fail - logs warning but continues
+		Expect(err).NotTo(HaveOccurred())
+
+		Expect(cfg.Telemetry).NotTo(BeNil())
+		Expect(cfg.Telemetry.Enabled).To(BeTrue())
+		// Certificate is empty because CA secret doesn't exist
+		Expect(cfg.Telemetry.Certificate).To(BeEmpty())
+	})
```
mangelajo, line 978: **Good test coverage — minor suggestion**

The `buildConfig` tests are thorough and cover the key scenarios (cert-manager enabled + CA exists, cert-manager disabled, CA secret missing). The deployment unit tests also properly verify both the TLS-enabled and TLS-disabled paths.

One gap: there's no test for the case where cert-manager is enabled with an **external issuer** (i.e., `Spec.CertManager.Server.IssuerRef` set with a `CABundle`). The `resolveTelemetryCA` function has a branch for this at `telemetry.go:464-468`. Existing tests in the `resolveTelemetryCA` describe block may cover this, but a `buildConfig`-level integration test for this path would add confidence that the external issuer CA bundle flows through correctly to the config.

---

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
mangelajo: We should probably add the same GRPC field to configure this in an uniform way with the other controllers. This wasn't covered on the JEP, let's update it too :)

mangelajo (changes requested):
Looking good, but we need to fix some gaps compared to router/controller certs.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: fixed now following the controller/router pattern.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
bkhizgiy: good point! now aligned with other components pattern and support both paths.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -308,6 +308,21 @@ type TelemetryConfig struct {
...
+
+// TelemetryGRPCConfig defines gRPC configuration for the telemetry service.
+// This is a simplified version of GRPCConfig since telemetry is internal-only
+// (ClusterIP) and doesn't need external endpoints or keepalive settings.
+type TelemetryGRPCConfig struct {
+	// TLS configuration for secure gRPC communication with the telemetry service.
+	// When spec.certManager.enabled is true, this is ignored and certificates are
+	// automatically managed by cert-manager.
+	// When spec.certManager.enabled is false, you can provide your own TLS secret here.
+	TLS TLSConfig `json:"tls,omitempty"`
```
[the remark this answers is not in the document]
bkhizgiy: I see the controller/route also don't support this, keeping everything consistent so wont fix.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1315,6 +1314,18 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
...
+		if jumpstarter.Spec.CertManager.Enabled {
+			caCert, err := r.resolveTelemetryCA(ctx, jumpstarter)
+			if err != nil {
+				// Log but don't fail - telemetry will still work with self-signed cert
+				// and exporters will use insecure mode if certificate is missing
+				logf.FromContext(ctx).V(1).Info("Could not resolve telemetry CA certificate",
+					"error", err)
+			} else if caCert != "" {
+				telemetryCfg.Certificate = caCert
+			}
```
mangelajo, line 1333: **Soft-fail on CA resolution may leave exporters unable to verify TLS**

When `resolveTelemetryCA` fails (e.g., CA secret not yet created by cert-manager), the error is logged at V(1) and reconciliation continues without a certificate. This means the ConfigMap will be written with an empty `Certificate` field, and exporters will either fail to verify the telemetry TLS connection or fall back to insecure mode.

While this is documented in the comment, the V(1) log level makes this easy to miss operationally. Since cert-manager secrets are created asynchronously, this will commonly happen during initial setup before the CA secret is ready. The 30-minute requeue will eventually pick it up, but the gap could be confusing.

Two suggestions to consider:
1. Use `log.Info` (V(0)) instead of `V(1)` so the message is visible by default.
2. Alternatively, return a shorter `RequeueAfter` (e.g., 30 seconds) when the CA is missing, to ensure quicker convergence.

This matches the importance level of similar missing-resource scenarios in the codebase.

---

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -308,6 +308,21 @@ type TelemetryConfig struct {
...
 }
 
 // TelemetryLoggingConfig configures the log push path to the telemetry service.
```
mangelajo, line 328: Nice, now we would be missing the Endpoints field as for other services I believe, but that's out of the scope of this PR.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -127,10 +127,38 @@ func (r *JumpstarterReconciler) cleanupTelemetryService(ctx context.Context, jum
...
+	return ""
+}
+
+// getTelemetryTLSSecretHash resolves the telemetry TLS secret name and returns its data hash.
+// This ensures that when cert-manager renews a certificate (or a manual secret changes),
+// the hash changes and triggers a rolling restart so the pod picks up the new cert.
+func (r *JumpstarterReconciler) getTelemetryTLSSecretHash(ctx context.Context, jumpstarter *operatorv1alpha1.Jumpstarter) (string, error) {
+	tlsSecretName := telemetryTLSSecretName(jumpstarter)
+	return r.getTLSSecretHash(ctx, jumpstarter.Namespace, tlsSecretName)
+}
```
mangelajo, line 151: **Good: TLS hash annotation and cert renewal restart mechanism**

This addresses the previous review's feedback about missing TLS hash annotations. The implementation correctly follows the controller/router pattern:
- `getTelemetryTLSSecretHash` delegates to the shared `getTLSSecretHash`
- The hash is injected as a pod annotation via `createTelemetryDeployment`
- When cert-manager renews certificates, the hash changes → pod template changes → rolling restart

The extraction of `telemetryTLSSecretName` into its own function is actually a nice improvement over the inline approach used by `getControllerTLSSecretHash`. Could be worth refactoring the controller one to match this pattern in a follow-up, but not needed in this PR.

---

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -418,19 +434,110 @@ var _ = Describe("Telemetry Lifecycle", func() {
 			},
```
mangelajo, line 434: **Consider E2E test coverage for telemetry TLS**

The unit and envtest coverage here is thorough. For completeness, consider whether an E2E test that validates the full telemetry TLS flow (operator deploys telemetry with cert-manager → exporter connects via TLS → metrics flow) would be valuable. This would be in `controller/deploy/operator/test/e2e/`.

This is not a blocker — the envtest coverage is solid — but worth tracking as a follow-up.

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
mangelajo: I Like suggestions 1 and 2 here.

mangelajo (changes requested):
@bkhizgiy have an eye at the comment in https://github.com/jumpstarter-dev/jumpstarter/pull/1023/changes#diff-55b7b5cfeb5b2fcd9f1d83431b263d743cfa0b0ea3c778db2395e81984e6ee60R1327 (that's the important one)

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
bkhizgiy: currently working on the e2e PR, will include this as well there

bkhizgiy:
@mangelajo can you please take another look when you have a moment:)

mangelajo (approved):

PR #1026, opened 2026-08-25
feat: allow replica=0 on jumpstarter components

bkhizgiy:
## Summary
This PR adds support for setting replicas: 0 on Jumpstarter controller components. This allows workloads to be temporarily suspended without deleting their Deployments, Services, certificates, or other configuration, making it easy to resume them later.

Components that can now be scaled to zero:

`spec.controller.replicas` — keeps the controller Deployment and scales it to 0 pods.
`spec.routers.replicas` — scales router Deployments to 0 while keeping their Services and TLS certificates.
`spec.telemetry.replicas` — scales the telemetry Deployment to 0 and keeps its Service. The telemetry endpoint is removed from the controller ConfigMap while suspended.
`spec.exporterSets.provisioners[].replicas` — scales individual provisioner Deployments to 0 while keeping their Deployment and RBAC configuration.

Suspended components report Ready=True with the Suspended reason, since being scaled to zero is intentional and shouldn't be treated as a failure.

ExporterSetControllersReady is also updated when all provisioners are suspended or disabled, so a stale False condition doesn't block the overall resource readiness.

The CRD minimum replica validation for these fields is changed from 1 to 0. Existing configurations with one or more replicas are unaffected.

## Examples
Suspend the controller

```
kubectl patch jumpstarter jumpstarter-cr \
  -n jumpstarter-system \
  --type=merge \
  -p '{"spec":{"controller":{"replicas":0}}}'
```

Suspend everything
```
spec:
  controller:
    replicas: 0
  routers:
    replicas: 0
  telemetry:
    enabled: true
    replicas: 0
  exporterSets:
    provisioners:
      - name: qemu.jumpstarter.dev
        replicas: 0
```

### DIFF
--- a/controller/deploy/operator/internal/controller/jumpstarter/certificates.go
+++ b/controller/deploy/operator/internal/controller/jumpstarter/certificates.go
@@ -108,7 +108,11 @@ func (r *JumpstarterReconciler) reconcileCertificates(ctx context.Context, js *o
 	}
 
 	// Create router certificates (one per replica)
-	for i := int32(0); i < js.Spec.Routers.Replicas; i++ {
+	certReplicas := int32(0)
+	if js.Spec.Routers.Replicas != nil {
+		certReplicas = *js.Spec.Routers.Replicas
+	}
+	for i := int32(0); i < certReplicas; i++ {
 		if err := r.reconcileRouterCertificate(ctx, js, issuerRef, i); err != nil {
 			return fmt.Errorf("failed to reconcile router %d certificate: %w", i, err)
 		}
--- a/controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
+++ b/controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
@@ -164,14 +164,14 @@ func (r *JumpstarterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 	// gRPC stream coordination (Dial/Listen pairing), so only one replica can
 	// serve traffic correctly. Multiple replicas would cause connection failures
 	// when Dial and Listen land on different pods.
-	if jumpstarter.Spec.Controller.Replicas > 1 {
+	if jumpstarter.Spec.Controller.Replicas != nil && *jumpstarter.Spec.Controller.Replicas > 1 {
 		log.Info("WARNING: controller.replicas > 1 is not yet supported — the controller "+
 			"uses in-memory state for gRPC stream coordination. Clamping to 1.",
-			"requested", jumpstarter.Spec.Controller.Replicas)
+			"requested", *jumpstarter.Spec.Controller.Replicas)
 		r.emitEventf(&jumpstarter, corev1.EventTypeWarning, "ReplicasClamped",
 			"controller.replicas=%d is not yet supported (in-memory gRPC state requires a single replica), clamping to 1",
-			jumpstarter.Spec.Controller.Replicas)
-		jumpstarter.Spec.Controller.Replicas = 1
+			*jumpstarter.Spec.Controller.Replicas)
+		jumpstarter.Spec.Controller.Replicas = ptr.To(int32(1))
 	}
 
 	// Reconcile RBAC resources first
@@ -378,11 +378,22 @@ func (r *JumpstarterReconciler) reconcileControllerDeployment(ctx context.Contex
 func (r *JumpstarterReconciler) reconcileRouterDeployment(ctx context.Context, jumpstarter *operatorv1alpha1.Jumpstarter) error {
 	log := logf.FromContext(ctx)
 
+	// When replicas is 0, suspend all existing router Deployments in-place
+	// (scale to 0 pods) without deleting them or their associated resources.
+	routerReplicas := int32(0)
+	if jumpstarter.Spec.Routers.Replicas != nil {
+		routerReplicas = *jumpstarter.Spec.Routers.Replicas
+	}
+
+	if routerReplicas == 0 {
+		return r.suspendAllRouterDeployments(ctx, jumpstarter)
+	}
+
 	// Cache hashes by secret name so a shared CertSecret is fetched once across replicas.
 	tlsHashBySecret := make(map[string]string)
 
 	// Create one deployment per replica
-	for i := int32(0); i < jumpstarter.Spec.Routers.Replicas; i++ {
+	for i := int32(0); i < routerReplicas; i++ {
 		secretName := routerTLSSecretName(jumpstarter, i)
 		routerTLSHash, ok := tlsHashBySecret[secretName]
 		if !ok {
@@ -499,8 +510,14 @@ func (r *JumpstarterReconciler) reconcileServices(ctx context.Context, jumpstart
 		}
 	}
 
-	// Reconcile router services - one per replica, all endpoints per replica
-	for i := int32(0); i < jumpstarter.Spec.Routers.Replicas; i++ {
+	// Reconcile router services - one per replica, all endpoints per replica.
+	// When replicas == 0 the router is suspended; skip service reconciliation and
+	// cleanup so existing Services are preserved for quick resume.
+	svcRouterReplicas := int32(0)
+	if jumpstarter.Spec.Routers.Replicas != nil {
+		svcRouterReplicas = *jumpstarter.Spec.Routers.Replicas
+	}
+	for i := int32(0); i < svcRouterReplicas; i++ {
 		if len(jumpstarter.Spec.Routers.GRPC.Endpoints) > 0 {
 			// Each replica gets ALL configured endpoints with replica substitution
 			for endpointIdx, baseEndpoint := range jumpstarter.Spec.Routers.GRPC.Endpoints {
@@ -549,10 +566,14 @@ func (r *JumpstarterReconciler) reconcileServices(ctx context.Context, jumpstart
 		}
 	}
 
-	// Clean up services for scaled-down replicas
-	if err := r.cleanupExcessRouterServices(ctx, jumpstarter); err != nil {
-		log.Error(err, "Failed to cleanup excess router services")
-		return err
+	// Clean up services for scaled-down replicas.
+	// Skip when replicas == 0 (suspended): keep existing Services so the router
+	// configuration can be restored without reconfiguration.
+	if svcRouterReplicas > 0 {
+		if err := r.cleanupExcessRouterServices(ctx, jumpstarter); err != nil {
+			log.Error(err, "Failed to cleanup excess router services")
+			return err
+		}
 	}
 
 	// Reconcile login endpoints (if configured)
@@ -918,7 +939,7 @@ func (r *JumpstarterReconciler) createControllerDeployment(jumpstarter *operator
 			Labels:    labels,
 		},
 		Spec: appsv1.DeploymentSpec{
-			Replicas:                &jumpstarter.Spec.Controller.Replicas,
+			Replicas:                jumpstarter.Spec.Controller.Replicas,
 			ProgressDeadlineSeconds: ptr.To(int32(600)),
 			RevisionHistoryLimit:    ptr.To(int32(10)),
 			Strategy: appsv1.DeploymentStrategy{
@@ -1323,16 +1344,20 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
 
 	// Telemetry configuration.
 	// Certificate is intentionally omitted until the telemetry binary supports TLS serving.
+	// When replicas==0 the telemetry Deployment is suspended (no ready endpoints), so omit
+	// the telemetry block from the config to avoid directing exporters to a dead endpoint.
 	if jumpstarter.Spec.Telemetry != nil && jumpstarter.Spec.Telemetry.Enabled {
 		t := jumpstarter.Spec.Telemetry
-		telemetryCfg := &config.Telemetry{
-			Enabled:  true,
-			Endpoint: telemetryEndpointFor(jumpstarter.Namespace),
+		if t.Replicas == nil || *t.Replicas > 0 {
+			telemetryCfg := &config.Telemetry{
+				Enabled:  true,
+				Endpoint: telemetryEndpointFor(jumpstarter.Namespace),
+			}
+			if t.Logging.Filter.MinSeverity != "" {
+				telemetryCfg.Logging.Filter.MinSeverity = t.Logging.Filter.MinSeverity
+			}
+			cfg.Telemetry = telemetryCfg
 		}
-		if t.Logging.Filter.MinSeverity != "" {
-			telemetryCfg.Logging.Filter.MinSeverity = t.Logging.Filter.MinSeverity
-		}
-		cfg.Telemetry = telemetryCfg
 	}
 
 	// gRPC keepalive configuration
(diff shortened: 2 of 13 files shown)


controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -378,11 +378,22 @@ func (r *JumpstarterReconciler) reconcileControllerDeployment(ctx context.Contex
...
+	// When replicas is 0, suspend all existing router Deployments in-place
+	// (scale to 0 pods) without deleting them or their associated resources.
+	routerReplicas := int32(0)
+	if jumpstarter.Spec.Routers.Replicas != nil {
+		routerReplicas = *jumpstarter.Spec.Routers.Replicas
+	}
+
+	if routerReplicas == 0 {
+		return r.suspendAllRouterDeployments(ctx, jumpstarter)
+	}
```
mangelajo, line 390: The nil-to-zero fallback pattern is repeated in ~8 places throughout this file (and in `certificates.go` and `status.go`). When `Replicas` is `nil`, the code treats it as `0` — i.e. suspended — but the CRD default for routers is `3`. While objects that pass through the API server will always have the default applied, a defensive helper would make the intent explicit and protect against in-process callers or pre-existing stored objects.

Consider introducing a small helper:

```go
func routerReplicas(spec *operatorv1alpha1.RoutersConfig) int32 {
    if spec.Replicas != nil {
        return *spec.Replicas
    }
    return 3 // CRD default
}
```

This would replace the repeated `if spec.Routers.Replicas != nil { ... } else { 0 }` blocks and eliminate the risk of nil being silently treated as suspension. The same pattern applies to `controllerReplicas` (default 1).

This echoes the unresolved coderabbit comment, but I want to emphasize: the fallback to `0` rather than the documented default is the specific concern — a nil pointer should mean "no opinion, use default", not "suspend everything".

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller_test.go
```diff
@@ -81,10 +83,10 @@ var _ = Describe("Jumpstarter Controller", func() {
...
 							},
 						},
-						Routers: operatorv1alpha1.RoutersConfig{
-							Image:           "quay.io/jumpstarter/jumpstarter:latest",
-							ImagePullPolicy: "IfNotPresent",
-							Replicas:        1,
+					Routers: operatorv1alpha1.RoutersConfig{
+						Image:           "quay.io/jumpstarter/jumpstarter:latest",
+						ImagePullPolicy: "IfNotPresent",
+						Replicas:        ptr.To(int32(1)),
```
mangelajo, line 89: The indentation here is broken — `Controller:` and `Routers:` field declarations were de-indented by one tab (they're now at the `Spec:` level rather than inside the `JumpstarterSpec` struct literal), while their nested fields like `Resources` and `GRPC` remain at the old deeper indentation. This compiles because Go struct literals allow mixed indentation, but it makes the code very hard to read and inconsistent with the rest of the test file.

Looks like this was an artifact of the `int32` → `*int32` replacement — the lines with `Replicas:` were touched but the surrounding context wasn't re-indented properly.

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1548,7 +1610,11 @@ func (r *JumpstarterReconciler) cleanupExcessRouterDeployments(ctx context.Conte
 
 		// Check if this deployment's name indicates it's beyond the current replica count
 		// We need to check all indices from current replicas onwards
-		for idx := jumpstarter.Spec.Routers.Replicas; idx < 100; idx++ { // reasonable upper bound
+		cleanupRouterReplicas := int32(0)
+		if jumpstarter.Spec.Routers.Replicas != nil {
+			cleanupRouterReplicas = *jumpstarter.Spec.Routers.Replicas
+		}
```
mangelajo, line 1616: The `cleanupRouterReplicas` variable is recalculated inside the inner `for i := range deploymentList.Items` loop, meaning it's re-evaluated (with the same result) for every deployment in the list. This should be hoisted outside the loop — both for clarity and to avoid the repeated nil-check.

```go
cleanupRouterReplicas := int32(0)
if jumpstarter.Spec.Routers.Replicas != nil {
    cleanupRouterReplicas = *jumpstarter.Spec.Routers.Replicas
}
for i := range deploymentList.Items {
    deployment := &deploymentList.Items[i]
    for idx := cleanupRouterReplicas; idx < 100; idx++ {
        ...
    }
}
```

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -378,11 +378,22 @@ func (r *JumpstarterReconciler) reconcileControllerDeployment(ctx context.Contex
...
+	routerReplicas := int32(0)
+	if jumpstarter.Spec.Routers.Replicas != nil {
+		routerReplicas = *jumpstarter.Spec.Routers.Replicas
+	}
+
+	if routerReplicas == 0 {
+		return r.suspendAllRouterDeployments(ctx, jumpstarter)
+	}
+
 	// Cache hashes by secret name so a shared CertSecret is fetched once across replicas.
```
mangelajo, line 392: When `routerReplicas == 0`, `suspendAllRouterDeployments` returns early, which skips both the deployment creation loop and the `cleanupExcessRouterDeployments` call at line 482. This means suspended router deployments (from a previous higher replica count, e.g., 3 → 0) are preserved in-place — which is the stated intent.

However, when resuming from 0 to a *smaller* count (e.g., 3 → 0 → 1), the excess suspended deployments (router-1, router-2) are cleaned up by `cleanupExcessRouterDeployments` — which works correctly because the reconcile path now enters the loop for replicas=1, creates/updates router-0, then calls cleanup which deletes router-1 and router-2.

The test at line ~1134 ("resumes to 1 replica after 3->0 suspension") covers this path well. Just flagging that this interaction between the suspend path and the cleanup path is a bit subtle, and a code comment noting "cleanup of stale suspended deployments happens on the resume path, not the suspend path" would help future readers.

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1323,16 +1344,20 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
...
+			}
+			if t.Logging.Filter.MinSeverity != "" {
+				telemetryCfg.Logging.Filter.MinSeverity = t.Logging.Filter.MinSeverity
+			}
+			cfg.Telemetry = telemetryCfg
 		}
-		cfg.Telemetry = telemetryCfg
 	}
 
 	// gRPC keepalive configuration
```
mangelajo, line 1363: When telemetry is enabled and `replicas == 0`, the telemetry block is omitted from the ConfigMap entirely. This is a good choice since there are no ready endpoints to connect to, but it's worth considering: when the controller is resumed (replicas restored > 0), does the ConfigMap get reconciled again to re-add the telemetry block?

Looking at the reconcile loop, `buildConfig` is called as part of `reconcileConfigMaps` on every reconciliation, so the telemetry block will be automatically restored on the next reconcile after replicas are bumped back up. This looks correct.

One minor thought: exporters that are already connected when telemetry is suspended will lose their `MetricsStream` targets. That's inherent to the feature, but it might be worth documenting in the PR description or in the CRD field comment that suspending telemetry will disrupt existing exporter metric streams until it's resumed.

---

controller/deploy/operator/internal/controller/jumpstarter/certificates.go
```diff
@@ -108,7 +108,11 @@ func (r *JumpstarterReconciler) reconcileCertificates(ctx context.Context, js *o
 	}
 
 	// Create router certificates (one per replica)
-	for i := int32(0); i < js.Spec.Routers.Replicas; i++ {
+	certReplicas := int32(0)
+	if js.Spec.Routers.Replicas != nil {
+		certReplicas = *js.Spec.Routers.Replicas
+	}
```
mangelajo, line 114: Same nil-to-zero fallback issue here. When `Replicas` is `nil`, `certReplicas` becomes `0` and no router certificates are reconciled. If an object somehow has a nil `Replicas` field (pre-existing stored object, in-process caller), this would silently skip certificate creation for all routers.

With the helper suggested in the main controller comment, this would become:
```go
certReplicas := routerReplicas(js.Spec.Routers)
```

---

controller/deploy/operator/internal/controller/jumpstarter/status.go
```diff
@@ -352,13 +378,24 @@ func (r *JumpstarterReconciler) checkControllerDeploymentReady(ctx context.Conte
...
 
-// checkRouterDeploymentsReady checks if all router deployments are available
+// checkRouterDeploymentsReady checks if all router deployments are available.
+// When spec.routers.replicas == 0 all routers are intentionally suspended and
+// are reported as ready (true) with a "Suspended" reason.
 func (r *JumpstarterReconciler) checkRouterDeploymentsReady(ctx context.Context, js *operatorv1alpha1.Jumpstarter) (bool, string) {
+	routerReplicas := int32(0)
+	if js.Spec.Routers.Replicas != nil {
+		routerReplicas = *js.Spec.Routers.Replicas
+	}
```
mangelajo, line 388: Same nil-to-zero fallback. If `Replicas` is nil here, `routerReplicas` becomes `0` and the function returns "Router deployments suspended (replicas: 0)" — reporting suspension when no suspension was actually requested.

---

controller/deploy/operator/test/e2e/e2e_test.go
```diff
@@ -2049,20 +2050,20 @@ dSignatureRotatedSignatureRotatedSignatureRotatedSignatureRotatedSig==
...
+						Endpoints: []operatorv1alpha1.Endpoint{
+							{Address: fmt.Sprintf("grpc.%s:8082", jwtCATestNamespace)},
 						},
 					},
-					Routers: operatorv1alpha1.RoutersConfig{
-						Image:           image,
-						ImagePullPolicy: corev1.PullIfNotPresent,
-						Replicas:        1,
+				},
+				Routers: operatorv1alpha1.RoutersConfig{
```
mangelajo, line 2063: Same broken indentation issue as in the unit test file — `Controller:` and `Routers:` were de-indented while their nested fields (`GRPC:`, etc.) remain at the old deeper indentation level.

---

controller/deploy/operator/test/e2e/e2e_test.go
mangelajo, line 1: The E2E test changes here update existing tests for the `int32` → `*int32` type change, which is necessary. However, the scale-to-zero feature itself (the core of this PR) doesn't have dedicated E2E tests. The existing E2E tests cover replica scale-up/down (3 → 1), but not the 0-replica suspension behavior.

Consider adding E2E tests for:
- Setting `controller.replicas=0` and verifying the Deployment exists with 0 pods
- Setting `routers.replicas=0` and verifying Deployments are suspended, Services preserved
- Restoring from `replicas=0` back to `replicas=1` and verifying pods come back up
- Verifying that status conditions show `Suspended` reason when at 0 replicas

These are core operator behavioral changes that would benefit from E2E coverage.

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1525,6 +1554,39 @@ func (r *JumpstarterReconciler) buildEndpointForReplica(jumpstarter *operatorv1a
...
+		return fmt.Errorf("failed to list router deployments for suspension: %w", err)
+	}
+
+	zero := int32(0)
+	for i := range deploymentList.Items {
+		dep := &deploymentList.Items[i]
+		if dep.Spec.Replicas != nil && *dep.Spec.Replicas == 0 {
+			continue
+		}
+		dep.Spec.Replicas = &zero
```
mangelajo, line 1578: Nit: all deployments in the list end up pointing to the same `&zero` variable. This is safe in Go (the Kubernetes API serializes the value, and the Deployment spec stores its own copy on the server side), but it's slightly cleaner to use `ptr.To(int32(0))` per iteration to match the style used elsewhere in this PR (e.g., the `ptr.To(int32(1))` calls in test code). Not a correctness issue.

---

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1434,7 +1459,11 @@ func (r *JumpstarterReconciler) buildRouter(jumpstarter *operatorv1alpha1.Jumpst
 	router := make(config.Router)
 
 	// Create router entry for each replica
-	for i := int32(0); i < jumpstarter.Spec.Routers.Replicas; i++ {
+	routerReplicaCount := int32(0)
+	if jumpstarter.Spec.Routers.Replicas != nil {
+		routerReplicaCount = *jumpstarter.Spec.Routers.Replicas
+	}
```
mangelajo, line 1465: When routers are suspended (`replicas == 0`), `buildRouter` returns an empty map. This means the ConfigMap's `router.yaml` will contain `{}`. If the controller is running (not itself suspended) and reads this ConfigMap, it will see zero router entries.

This is probably fine — if only the routers are suspended but the controller is running, clients attempting to Dial would fail to find a router anyway. But it's worth noting the coupling: suspending routers effectively disables Dial/Listen routing at the controller level too, since the ConfigMap no longer has router endpoint information. This might be worth documenting — suspending routers doesn't just stop router pods, it also removes router entries from the controller's config.

---

mangelajo (commented):
I havent' fully reviewed the comments, but at least the first part is good.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1324,29 +1303,36 @@ func (r *JumpstarterReconciler) buildConfig(ctx context.Context, jumpstarter *op
...
+					// Log at default verbosity so operators notice during initial cert-manager setup.
+					// Reconciliation continues without a certificate; telemetryCANeedsRequeue
+					// triggers a short requeue until the CA secret is ready.
+					logf.FromContext(ctx).Info("Could not resolve telemetry CA certificate; exporters cannot verify telemetry TLS until the CA is available",
+						"error", err)
+				} else if caCert != "" {
+					telemetryCfg.Certificate = caCert
+				}
 			}
+			cfg.Telemetry = telemetryCfg
```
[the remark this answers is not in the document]
mangelajo: Right, that is the behavior of the other CRDs, no cert manager... you can still bring your own certsecret

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
```diff
@@ -1539,6 +1530,154 @@ func (r *JumpstarterReconciler) buildEndpointForReplica(jumpstarter *operatorv1a
...
+					"Router deployment suspended: name=%s namespace=%s", dep.Name, dep.Namespace)
+			}
+			continue
+		}
+
+		idx, err := strconv.ParseInt(idxStr, 10, 32)
+		if err != nil {
+			log.Info("Skipping router deployment with invalid router-index during suspension",
+				"name", dep.Name, "router-index", idxStr, "error", err)
+			continue
```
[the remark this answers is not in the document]
mangelajo: I am not sure what you mean @coderabbitai, can you elaborate?

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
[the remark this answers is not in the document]
bkhizgiy: done.

controller/deploy/operator/internal/controller/jumpstarter/jumpstarter_controller.go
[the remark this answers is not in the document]
bkhizgiy: fixed.

PR #1033, opened 2026-08-27
feat(adb): connect exporter devices to a client-owned ADB server

kirkbrauer:
## What

`j <device>.adb connect` puts a device attached to a remote exporter into **the ADB server your machine already runs**, so it shows up in Android Studio, `adb`, `logcat`, tradefed and gradle with no configuration.

```console
$ jmp shell --tls-grpc <exporter-host>:8083 -- j phone connect
connected as 127.0.0.1:41000
```

## Why

The existing path pointed your tooling at *the exporter's* ADB server, which only works if you own your local one — so it fails the most common case, an IDE already running. Android Studio holds port 5037 and respawns its server **1.18 s** after `adb kill-server` (measured), so the port cannot reliably be taken over.

`adb connect` is additive instead: the exporter forwards a device's adbd, Jumpstarter tunnels it, and `adb connect` adds it to whatever server you already have. Devices from several exporters coexist beside your own emulators.

## Design

**Devices are declared, not discovered.** One `AdbDevice` per device, identified by its **bench USB port** (`usb_port: "1-4.2"`) rather than its serial, so a DUT can be swapped or power-cycled without a config change. The serial is re-resolved from the port on every stream, which is what makes re-enumeration self-healing.

**One ADB server per port, shared.** A server claims the USB devices it finds and only one can hold a device, so two servers on one port leave the second blind while `adb start-server` reports success. Drivers share a refcounted registry keyed on the port, adopt a server already running there, and never kill one they did not start.

**The driver undoes only what it did.** A forward or `adb connect` already present in the shared server is reused but left in place at teardown — `adb forward --list` cannot say who created it. A forward the driver did create is re-checked before removal, since `adb forward --remove` matches the local port alone and ignores `-s`.

## Breaking changes

- **The `j adb <command>` passthrough is removed.** Use `j <device>.adb connect`, then your own `adb -s <address> ...`.
- New commands are `connect` and `serve` (not `attach`/`endpoint`); adb already owns `attach`/`detach` with a conflicting meaning. The server-level `tunnel` is now `serve`.

Exporter configs need no changes. See *Upgrading from an earlier `AdbServer`* in the README.

## `AdbServer`

Deprecated in favour of `AdbDevice`, but retained: it covers pointing tooling at the exporter's whole server and devices that cannot expose adbd over TCP, and the Cuttlefish and Android emulator drivers embed it. `AdbDevice` gains a small driver-side API (`adb_env`, `ensure_reachable`, `disconnect`) so Cuttlefish can migrate in a follow-up.

## Testing

138 unit tests against a stateful fake adb that models devices, device state and forwards; the Cuttlefish (86) and Android emulator (8) suites pass unchanged.

Verified end to end on hardware — a Raspberry Pi exporter with an Android tablet, a macOS client, direct mode, and Android Studio. Adoption, `connect`, 17 re-enumeration cycles and rootless-container USB isolation are summarised in [this comment](https://github.com/jumpstarter-dev/jumpstarter/pull/1033#issuecomment-5719300215). Not yet measured: whether the macOS `usb_port` form stays stable across an unplug.


python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -64,21 +103,251 @@ def __post_init__(self):
...
+                [self.adb_path, "forward", "--list"],
+                check=True,
+                capture_output=True,
+                text=True,
+                timeout=self.connect_timeout,
+                env=self.adb_env(),
+            )
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired, OSError) as e:
+            self.logger.warning("could not list adb forwards (%s); assuming none", e)
+            return {}
```
bennyz: this might cause used slots to be cleaned up if we fail listing?

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -64,21 +103,251 @@ def __post_init__(self):
...
+            )
+        except (subprocess.TimeoutExpired, OSError):
+            self.logger.warning(
+                "something is listening on %s:%d but does not answer as an ADB server; "
+                "not adopting it. Free the port, or set a different 'port' in the exporter config.",
+                self.host,
+                self.port,
+            )
+            return False
+        return result.returncode == 0
```
mangelajo, line 148: `adb version` is a client-local command — it prints the local binary's version and exits without contacting `ANDROID_ADB_SERVER_PORT`. So the adoption probe here only confirms that the `adb` binary works, not that an ADB server is actually serving on our port. A non-ADB listener that accepted the TCP connect on line 142 would still be "adopted", and every subsequent `adb` call directed at it (devices, forward, etc.) would hang.

Consider replacing this with a command that actually talks to the server, e.g. `adb devices` (bounded by the same timeout). That way the second check genuinely confirms the peer speaks the ADB protocol.

Alternatively, reading the raw ADB protocol greeting from the socket (the `host:version` service) would avoid spawning a subprocess entirely, but `adb devices` is simpler and good enough here.

---

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
```diff
@@ -114,8 +543,22 @@ def cli(self):
...
+            help="attach: keep tracking devices connected or removed while running "
+            "(off by default; most exporters have a fixed set of devices)",
+        )
+        @click.option(
+            "--poll-interval",
+            type=float,
+            default=2.0,
+            show_default=True,
+            help="attach: seconds between device checks, with --hotplug",
+        )
```
mangelajo: With `--poll-interval 0` (or negative), the hotplug loop calls `devices()` on the exporter without any pause, effectively busy-looping. `anyio.sleep(0)` returns immediately, so this saturates both the gRPC link and the exporter's ADB server.

Consider constraining to a positive minimum, e.g.:

```python
@click.option(
    "--poll-interval",
    type=click.FloatRange(min=0.1),
    default=2.0,
    show_default=True,
    help="attach: seconds between device checks, with --hotplug",
)
```

---

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
```diff
@@ -155,6 +608,16 @@ def adb(host: str, port: int, adb: str, args: tuple[str, ...]):
 
             _validate_adb_args(args)
 
+            if args[0] == "attach":
+                serials = [a for a in args[1:] if not a.startswith("-")]
```
mangelajo: Parsing serials from `args[1:]` by filtering `not a.startswith("-")` means that if someone passes e.g. `j adb attach --hotplug --poll-interval 5 myserial`, the value `"5"` gets treated as a serial because it doesn't start with `-`. It happens to not match any real device so it's harmless in practice, but it's still confusing.

Since `attach` is a Jumpstarter-specific subcommand being parsed manually out of the generic `args`, you might consider making it a proper Click subcommand or group. That said, this is a minor nit — the current approach works and the options are flags or explicit `--key value` pairs that Click already consumed, so `args` probably won't contain the option values by the time we get here. Just flagging in case the arg parsing behaves unexpectedly with custom orderings.

---

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
```diff
@@ -24,21 +36,191 @@ def _validate_adb_args(args: tuple[str, ...]) -> None:
...
+    have to own, configure, or even have an ADB server.
+
+    Returns:
+        adb's own message, for logging.
+
+    Raises:
+        RuntimeError: adb reported a failure, or timed out.
+    """
+    try:
+        result = subprocess.run([adb, "connect", target], check=False, capture_output=True, text=True, timeout=60)
```
mangelajo: The 60-second timeout here is hardcoded while the driver side uses `connect_timeout` (default 30s) for its ADB calls. Consider making this configurable or at least documenting that `adb connect` uses a different timeout from the rest of the driver.

In practice, `adb connect` to a tunneled local port should resolve quickly, so 60s is likely fine as a ceiling. But if the exporter is far away or the tunnel is slow, the mismatch could be confusing.

Minor nit, not a blocker.

---

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -64,21 +103,251 @@ def __post_init__(self):
...
+            raise RuntimeError(
+                f"could not attach {device}: {stderr or e}. The device may be offline, "
+                f"or adbd may not be listening on tcp:{adbd_port} (try `adb tcpip {adbd_port}`)."
+            ) from e
+        except subprocess.TimeoutExpired as e:
+            raise RuntimeError(f"attaching {device} timed out after {self.connect_timeout}s") from e
+
+        self._slots[slot_port] = device
+        self.logger.info("attached %s on slot tcp:%d (device tcp:%d)", device, slot_port, adbd_port)
+        return self._slot_name(slot_port)
```
mangelajo: The `subprocess.run` that creates the forward does not catch `OSError`, which can happen if e.g. the `adb_path` binary disappears or becomes non-executable at runtime (unlikely but possible after a package update). Both `CalledProcessError` and `TimeoutExpired` are caught, but an `OSError` would propagate unhandled.

Consider adding `OSError` to the except chain, or wrapping it in a RuntimeError with context:

```python
except (subprocess.CalledProcessError, subprocess.TimeoutExpired, OSError) as e:
```

The `_remove_forward` method already handles all three, so this would be consistent.

---

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
```diff
@@ -50,6 +232,113 @@ def _remove_tunnel_state() -> None:
...
+        for device in list(self.attached):
+            if device not in here:
+                click.echo(f"{device} disconnected")
+                self.attached.pop(device).close()
+
+        # Forget failures for devices that are no longer here, so re-plugging one is a
+        # genuine retry. Done for every remembered failure, not just attached devices:
+        # a device that *failed* and then vanished never entered `attached`, so
+        # clearing only those left it permanently blacklisted.
+        self._failed -= {device for device in self._failed if device not in here}
```
mangelajo: The failed-set cleanup logic is correct but the set comprehension reads slightly backwards. Consider:

```python
self._failed &= set(here)
```

This is equivalent (keep only failures that are still present) and is a bit more readable than building a set of things to remove and subtracting it. Not a functional issue at all — just a readability suggestion.

---

python/packages/jumpstarter-driver-adb/README.md
```diff
@@ -35,6 +70,35 @@ export:
...
+come up seeing no devices at all while looking healthy.
+
+By default the driver therefore **adopts** a server already on its port, and leaves
+it running at teardown rather than killing a server other processes are using. You
+will see:
+
+```
+adopting the ADB server already listening on 127.0.0.1:15037; it owns the
+connected devices, and this driver will leave it running
+```
```
mangelajo: This fenced block has no language identifier. Since these are Sphinx docs rendered with `myst`, the linter will flag this (MD040). Add `text` or `console`:

```markdown
```text
DUT ──USB──▶ EXPORTER ──Jumpstarter tunnel──▶ YOU
...
```

---

mangelajo (changes requested):

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
kirkbrauer: Humm, so I wonder which ADB version we should even report here, the client's or the exporter's? Maybe it's best to just keep it client-local and show the local ADB server instead for consistency.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
kirkbrauer: This was originally done because we were basically intercepting the ADB commands, but I think maybe this is a better approach.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
kirkbrauer: Yeah, I need to look into this. I'm not convinced this slot mechanism is the best way to handle this anyways, but good suggestion.

kirkbrauer:
## Substantial redesign following review — please re-review

@bennyz @mangelajo the slot mechanism you both pushed back on is gone. Rather than patching it, this replaces discovery-based attach with **declared, per-device drivers**. Summary of what changed and why, then the disposition of every review comment.

### Why the model changed

The pool existed because a per-device *child* cannot express hotplug — children are resolved at lease establishment, and one added later is invisible to the client (I verified this with a prototype). But declaring the device sidesteps that entirely: the child exists from the start, and only the serial behind it changes. Four reasons declared beats discovered here:

- **USB permissions.** An exporter runs its drivers in a container, so devices must be passed in deliberately — which means they have to be described.
- **Re-enumeration.** DUTs are power-cycled with relays; every cycle re-enumerates USB and can change the ADB serial.
- **Bench swapping.** Hardware moves between benches, so the stable identity is the bench **USB port**, not the device.
- **Fit.** Declared hardware is how every other driver here works (`dutlink`/`sdwire`/`yepkit` `serial`, `pyserial` `url`), and it lets a DUT be a composite of its power relay *plus* its ADB, so leasing the DUT leases the right device.

```yaml
export:
  dut1:
    type: jumpstarter_driver_composite.driver.Composite
    children:
      power:
        type: jumpstarter_driver_yepkit.driver.Ykush
        config: { serial: "YK112233", port: "1" }
      adb:
        type: jumpstarter_driver_adb.driver.AdbDevice
        config:
          usb_port: "1-4.2"   # bench port: swap the DUT, config unchanged
```

### What that removes

`_Slot`, the reserve/pending dance, slot exhaustion, `attach_slots`, `attach_base_port`, and the concurrent-attach race — with one device per instance there is nothing to allocate. A single `@exportstream connect()` resolves its endpoint per stream (bench port → current serial → `adb forward tcp:0`), which makes re-enumeration self-healing: forwards vanish with the device, so a stale one is always detected.

**The ADB server is now implicit** — a module-level registry keyed by `(adb_path, port)`, refcounted, acquired lazily on first stream. An ADB server *claims* the USB devices it finds and only one can hold a given device, so sharing is a correctness requirement: two servers on one port would leave the second blind while `start-server` reported success. An adopted server is never killed. Declaring `AdbServer` is now optional, and a declared one is what devices adopt — so **cuttlefish and androidemulator work unchanged** (94 tests, untouched, plus a signature-level guard test on the `AdbServer` surface they call).

**Jumpstarter no longer wraps the `adb` CLI.** The `j adb <cmd>` passthrough is gone, and with it the persistent tunnel state file and its ownership/symlink hardening — that file existed only so passthrough commands could share a tunnel, so the local-writable-endpoint surface disappears rather than being defended. `j adb shell` becomes `adb -s <addr> shell`, which is what you'd type anyway. `attach` keeps exactly one `adb connect`; `endpoint` runs no adb at all.

### Two decisions worth flagging

**Device selection uses the documented `-s SERIAL`, not `-s usb:<path>`.** `-s usb:1-4.2` *does* work — `atransport::MatchesTarget` falls through to `return (target == devpath) || ...` — but it's undocumented, and it turns out unnecessary: resolving the port to a serial via `devices -l` gives the documented contract *and* the serial needed for `forward --list` reconciliation. A test asserts no `usb:` ever reaches an adb argv.

**There is deliberately no `transport: serial`.** adb has no UART transport: `adb.h` defines only `kTransportUsb` and `kTransportLocal` ("local" = TCP), `connect_device()` coerces every address to `tcp:` (`adb connect serial:/dev/ttyUSB0` → `bad port number '/dev/ttyUSB0'`), and `dev:`/`dev-raw:` are forward targets executed *inside adbd on the device*. Device-side there's no adbd-over-UART property either. Serial-only DUTs are reached by getting them onto TCP; the README documents both routes and the raw-UART caveats (no retransmission/checksum, ~11.5 KB/s at 115200). Unsupported transport values name the route rather than implying a typo. `vsock` is real but deferred.

### Review comments

| Comment | Disposition |
|---|---|
| **@bennyz** — failing `forward --list` could clean up used slots | Structurally impossible; no slots to release |
| **@mangelajo** — adoption probe must be server-backed, not `adb version` | Kept as `adb devices`, with the test that pins it |
| **@mangelajo** — `--poll-interval 0` busy-loops | Moot; hotplug polling deleted |
| **@mangelajo** — `args[1:]` sniffing treats `"5"` as a serial | Moot; `attach`/`endpoint`/`info` are real click subcommands |
| **@mangelajo** — hardcoded 60s vs `connect_timeout` | **Was still open — fixed.** Now documented `ADB_CONNECT_TIMEOUT`/`ADB_DISCONNECT_TIMEOUT` constants, overridable via `attach(timeout=...)`, explicitly separate from the exporter's `connect_timeout` because they bound a command on the *client* against a local port-forward. Test pins defaults and override. |
| **@mangelajo** — `OSError` uncaught on forward creation | Caught |
| **@mangelajo** — `_failed &= set(here)` | Moot; `_AttachSet` deleted |
| **@mangelajo** — unlanguaged fence (MD040) | Verified zero unlanguaged fences in the rewritten README |

### Verification

97 tests in this package (from 96), 191 including both consumers. ruff, format and `ty` clean; 100% docstrings on production code. The concurrency test was checked to **fail** with the lock removed. Verified end-to-end against a stateful fake adb that models `devices -l` devpaths and per-serial forwards, and the README's exporter YAML is instantiated through the real config path in a test.

**Not yet verified on hardware** — these are the parts no unit test can cover, and I'd value a second pair of eyes on whether the list is complete:

1. relay power-cycle changing the serial while `usb_port` stays valid;
2. physically swapping two DUTs between ports;
3. exactly one ADB server serving both devices, not one per device;
4. `attach` with no local ADB server running at all.

### Open question for reviewers

Should **cuttlefish** adopt `AdbDevice`? It's already `transport: tcp` in all but name (`_cvd_device` is `host:port`, and it calls `connect_device`/`disconnect_device`), and it would gain `attach` — and therefore Android Studio support — for free. I left it alone here because `_wait_boot` shells out using `adb.adb_path`/`adb.adb_env()`, so either `AdbDevice` grows those or cuttlefish keeps both objects. That's a design call, not a mechanical port, and its failure mode is a 300s timeout on real hardware. Happy to do it in a follow-up.

mangelajo:
@bennyz have an eye on this one :)

bennyz:
I did not do live testing yet, but it seems like it would fit well with the cuttlefish pod exporters

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -179,17 +403,398 @@ def disconnect_device(self, device: str) -> str:
...
+        # `adb connect` exits 0 even when it fails, reporting the reason on stdout, so
+        # match adb's own success strings instead of the exit status.
+        if result.returncode != 0 or not message.startswith(("connected to", "already connected to")):
+            raise RuntimeError(f"could not connect to {target}: {message or 'no output'}")
+        self._connected = target
+
+        host, _, port = target.rpartition(":")
+        return host, int(port)
+
+    def _ensure_forward(self) -> int:
```
bennyz, line 898: we might deleted an adopted forward?

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -1,122 +1,355 @@
...
+    adopt_existing_server: bool,
+    logger,
+) -> _SharedServer:
+    """Take a reference to the ADB server on *port*, starting or adopting it once.
+
+    The probe and start happen while holding the lock. That serialises concurrent
+    first-acquirers, which is the point: without it two callers could both decide no
+    server was running and both start one.
+    """
+    key = (adb_path, port)
```
bennyz: so in theory we could have different paths use the same port?
I suppose we won't have that, so perhaps we want just port?

kirkbrauer: It's keyed on `port` now (7d87d16a).

`AdbServer.port` and `AdbDevice.server_port` both default to `15037` while `adb_path` is per-instance, so two drivers naming different binaries produced two registry entries for one real server, each with its own refcount which will never reach zero with `adb kill-server` while the other still held a reference. With `adopt_existing_server: false` on one of them, both claimed ownership.

Now the first acquirer's `adb_path` is the one the server runs as, and a later caller naming a different binary shares it and gets a warning: an adb client whose version differs from the running server kills and restarts it, which would drop every device claim on that port. `_resolve_adb_path` also normalizes through `realpath`, so `adb`, `/usr/bin/adb` and a symlink aren't three "different" binaries.

Tests: `test_two_adb_paths_on_one_port_share_one_server`, `test_a_differently_named_adb_does_not_get_its_own_refcount`.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
kirkbrauer: Confirmed, and fixed in 7d87d16a — thanks, this one had a twin I'd have missed.

Reuse and creation both just wrote `_forward_port`, and `close()` removed it unconditionally, so a forward already present in the shared server — another lease's, another exporter's, or a person's at the bench — was deleted at teardown, silently. It's now tracked with `_owns_forward`: reuse still happens, since re-forwarding per stream would churn the server, but only a forward this driver created is removed. The TCP path had the same bug, recording `already connected to` as ours and disconnecting it on close; same fix.

Two things I got wrong on the first attempt, both settled against AOSP:

- **Removal can't be scoped to a device.** `killforward` hands its transport to `remove_listener`, which matches the local spec alone and ignores it (`adb_listeners.cpp`). Worse, `-s` actively breaks teardown: `handle_forward_request` acquires the transport first and `acquire_one_transport` rejects any non-`device` state, so an offline DUT — where a relay leaves it mid power-cycle — refuses the removal while its listener is still there. Removal is unscoped, but re-checked against `forward --list` first.
- **The port can stop being ours with nothing freed.** `install_listener` matches on the local spec and *repurposes the listener in place*, so one `adb -s OTHER forward tcp:<our port> tcp:5555` moves it to another device, exit 0, silently. `--no-rebind` can't protect us: a `tcp:0` listener is renamed to its resolved port, so nothing ever matches `tcp:0` again.

Both reproduced on hardware: a hand-made forward is reused and left in place, a driver-created one is removed, and a port rebound to another device is left alone.

Tests: `test_a_forward_we_did_not_create_is_reused_but_never_removed`, `test_close_leaves_a_forward_adb_rebound_to_another_device`, `test_close_removes_the_forward_of_an_offline_device`, `test_a_tcp_device_someone_else_connected_is_left_connected`.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
kirkbrauer: The slot pool is gone, but you'd found a real bug and it survived the redesign in a different form — worth re-reading this one, so thanks.

`_live_forwards()` returned an empty set both when the server said "no forwards" and when it couldn't be asked. So a transient `adb forward --list` failure looked like "everything is gone": the driver dropped ownership of the forward it had created and made a second one, orphaning the first — invisible to teardown for the life of the ADB server.

It now returns `None` for "couldn't ask", and each caller decides what unknown means: `_ensure_forward` keeps the forward it already has rather than risking a duplicate, and teardown declines to remove anything it can't confirm.

Test: `test_a_failed_listing_does_not_orphan_our_forward`, which fails against the old empty-set behaviour.

kirkbrauer:
## Hardware verification

Now verified end to end, which the description previously listed as outstanding. Setup: a Raspberry Pi exporter with an Android tablet on USB, a macOS client, and Jumpstarter's direct mode (`jmp run --tls-grpc-listener` on the exporter, `jmp shell --tls-grpc` on the client) — no controller involved.

**Adoption against a real Android Studio server.** The driver adopted Studio's ADB server, created its own forward, and `close()` left both Studio's server and a forward it had merely reused untouched. I also measured the claim in the description: Studio respawns its server **1.18 s** after `adb kill-server`, so the stated ~3 s is conservative.

**`connect` end to end.** `adb -s <local address> shell getprop ro.serialno` from the Mac returned the tablet attached to the Pi, and both ADB servers held the device at once — the additive property this PR exists for.

**Re-enumeration self-heals.** Across physical unplug/replug and sysfs deauthorize cycles, 17 forward creations, each on a new port, with the client recovering every time and no operator action. During the gap the driver raises the relay-first error rather than handing out a dead endpoint.

**Two faults found only on hardware,** both fixed here:

- `info()` ran adb without taking a server reference, so the adb client silently started a server the driver didn't know about; the next acquire adopted it, and adopted servers are never killed — so it outlived `close()`.
- A failed `adb connect` still registers the address as an `offline` entry. `attach`/`connect` ran its `adb disconnect` only on the success path, so a device that wasn't ready littered the developer's ADB server. Cleanup now covers the failure path too.

**Container isolation** (documented in 8216ad58). A pinned `--device /dev/bus/usb/001/NNN` cannot survive re-enumeration: the node number moved across a dozen values while the port stayed fixed, and since neither Podman nor Docker can add a device to a running container, the next run fails outright with `stat: no such file or directory`. Mounting `/dev/bus/usb` and pinning *authority* by port with a udev rule gave a rootless container exactly one device — every other USB node `Permission denied` — and a long-running container survived a detach/reattach with no restart. Two portability notes are in the README: the rule file must sort before `73-seat-late.rules`, or systemd's `uaccess` ACL silently overrides `MODE="0600"`; and on SELinux hosts the bind mount keeps its label, so `container_use_devices` is needed.

**Also fixed on the way:** `adopt_existing_server: false` used to claim a server it hadn't started — `adb start-server` is silent and exits 0 either way — so `close()` would have killed Android Studio's. It now leaves ownership with whoever started it (`test_adopt_false_does_not_claim_a_server_it_could_not_start`).

One thing still unverified: the macOS `usb_port` form. It is IOKit's location ID, which encodes the device's position in the USB tree, so it should be stable per port — but I did not get to measure an unplug/replug on macOS, and the README says only that it encodes position.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
kirkbrauer: Switched to `adb devices` — thanks, that was a real hole. `test_the_adoption_probe_asks_the_server_not_the_client` pins it so it can't regress to `adb version`.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
kirkbrauer: Moot now: hotplug polling was removed in the declared-device redesign, so there is no `--poll-interval` left to validate.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
kirkbrauer: Moot now: `connect`, `serve` and `info` are real click subcommands, so nothing sniffs `args` for serials.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
kirkbrauer: Now named `ADB_CONNECT_TIMEOUT`/`ADB_DISCONNECT_TIMEOUT` and overridable via `connect(timeout=...)`. Kept deliberately separate from the exporter's `connect_timeout`, since these bound a command on *your* machine against a local port-forward — the docstring says so explicitly.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
kirkbrauer: Caught — `_create_forward` now handles `CalledProcessError`, `TimeoutExpired` and `OSError`, matching `_remove_forward`.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/client.py
kirkbrauer: Moot now: `_AttachSet` and its `_failed` set went away with discovery-based attach.

python/packages/jumpstarter-driver-adb/README.md
kirkbrauer: Fixed — the README has no unlanguaged fences left. The docs build (warnings as errors) later caught a related one from me, a `udev` tag with no Pygments lexer, now `text`.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -1,122 +1,355 @@
...
+        adb_path = resolved
+
+    try:
+        subprocess.run(
+            [adb_path, "version"],
+            check=True,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True,
+        )
```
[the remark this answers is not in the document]
kirkbrauer: Fixed in 3d1cf1c8 — you were right that this was the one unbounded adb call, and it runs from `__post_init__` of both drivers, so a wedged binary hung exporter startup.

It is now bounded by `connect_timeout` (both call sites validate it first), and the except arm gains `TimeoutExpired` and widens `FileNotFoundError` to `OSError` — a superset, so a missing binary still reports as before, and a present-but-unreadable one is now covered too.

Test: `test_an_adb_that_cannot_answer_version_is_a_config_error`, parametrized over a hang and an OSError, asserting the timeout is actually passed.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -1,122 +1,427 @@
...
+        server adopted from another process.
+        """
+        with _SERVERS_LOCK:
+            entry = _SERVERS.get(self.port) or self._server
+            if entry is None:  # pragma: no cover - __post_init__ always sets one
+                return self.port
+            if entry._is_listening(self.connect_timeout, self.logger):
+                self.logger.info("ADB server already listening on port %d", self.port)
+            else:
+                entry.start(self.connect_timeout, self.logger)
```
bennyz: might want to propagate the errors from this and check them here?

kirkbrauer: Yes — fixed in 546d7115, and it was worse than it looked here: `start()` logged the failure and returned normally, so both callers went on to set `entry.owns = True`. A server that never came up was registered as ours and running, and every later adb call failed far from the cause.

`start()` now raises `RuntimeError` on a non-zero exit, a timeout or an `OSError`, carrying adb's stderr. The raise happens before the entry is marked owned or registered, so a failed start leaves nothing behind and a retry starts cleanly. `test_start_server_survives_a_hung_port` had pinned the old swallow and is replaced by `test_a_server_that_fails_to_start_is_an_error_not_a_running_server`, `test_a_failed_start_can_be_retried` and `test_start_server_reports_a_restart_failure`.

python/packages/jumpstarter-driver-adb/jumpstarter_driver_adb/driver.py
```diff
@@ -1,122 +1,434 @@
...
+            if entry is None:  # pragma: no cover - __post_init__ always sets one
+                return self.port
+            if entry.refs > 1:
+                self.logger.warning(
+                    "killing the ADB server on port %d, which %d drivers share; their "
+                    "forwards and device claims go with it",
+                    self.port,
+                    entry.refs,
+                )
+            entry.kill(self.connect_timeout, self.logger)
```
bennyz: here we also want to reraise right?

kirkbrauer: Yes — same bug as `start()`, fixed in 26021d80. `kill()` logged and returned, so `kill_server` cleared ownership and reported success while the server kept running.

`kill()` now raises on a non-zero exit, a timeout or an `OSError`, and also checks the port is no longer served, since `adb kill-server` exits 0 even when there was no server to kill. `kill_server` raises before clearing ownership. The one place it doesn't propagate is teardown (`_release_server` from `close()`), where a lease has to finish releasing even if the server won't die, so the failure is logged there.

Tests: `test_kill_server_reports_a_failed_kill`, `test_kill_server_notices_a_server_that_survived`, `test_teardown_finishes_even_if_the_server_will_not_die`.

bennyz (approved):

PR #1039, opened 2026-08-30
fix(cli): stop reporting hook waits and status changes that never happened

kirkbrauer:
`jmp shell` announced "waiting for beforeLease hook" before the status monitor had polled even once, so attaching to an already-ready lease reported a wait that was not happening. The monitor then logged its first observation as a transition, so a lease that had been in the same state all along looked like it had just changed.

Waits for the first observation before reporting, and logs that first one at debug level rather than as a change.

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
@@ -54,6 +54,11 @@ logger = logging.getLogger(__name__)
 # Refresh token when less than this many seconds remain
 _TOKEN_REFRESH_THRESHOLD_SECONDS = 120
 
+# Total time to wait for the beforeLease hook, and the initial slice of it spent
+# settling the exporter's current status before announcing that we are waiting.
+HOOK_TIMEOUT = 300.0
+HOOK_PROBE_TIMEOUT = 2.0
+
 
 def _run_shell_only(lease, config, command, path: str, motd: str | None = None) -> int:
     """Run just the shell command without log streaming."""
@@ -336,12 +341,18 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
                             # Wait for beforeLease hook to complete while logs are streaming
                             # This allows hook output to be displayed in real-time
                             # Uses non-blocking polling instead of streaming for robustness
-                            logger.info("Waiting for beforeLease hook to complete...")
+                            targets = [ExporterStatus.LEASE_READY, ExporterStatus.BEFORE_LEASE_HOOK_FAILED]
 
-                            # Wait for LEASE_READY or hook failure using background monitor
-                            result = await monitor.wait_for_any_of(
-                                [ExporterStatus.LEASE_READY, ExporterStatus.BEFORE_LEASE_HOOK_FAILED], timeout=300.0
-                            )
+                            # The monitor reports no status until its first poll, so settle
+                            # that first: attaching to a lease that is already LEASE_READY
+                            # must not claim to be waiting on a hook that already ran.
+                            result = await monitor.wait_for_any_of(targets, timeout=HOOK_PROBE_TIMEOUT)
+
+                            if result is None and not monitor.connection_lost:
+                                logger.info("Waiting for beforeLease hook to complete...")
+                                result = await monitor.wait_for_any_of(
+                                    targets, timeout=HOOK_TIMEOUT - HOOK_PROBE_TIMEOUT
+                                )
 
                             if result == ExporterStatus.BEFORE_LEASE_HOOK_FAILED:
                                 reason = monitor.status_message or "beforeLease hook failed"
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
@@ -54,10 +54,8 @@ logger = logging.getLogger(__name__)
 # Refresh token when less than this many seconds remain
 _TOKEN_REFRESH_THRESHOLD_SECONDS = 120
 
-# Total time to wait for the beforeLease hook, and the initial slice of it spent
-# settling the exporter's current status before announcing that we are waiting.
-HOOK_TIMEOUT = 300.0
-HOOK_PROBE_TIMEOUT = 2.0
+# Total time to wait for the beforeLease hook.
+_HOOK_TIMEOUT: float = 300.0
 
 
 def _run_shell_only(lease, config, command, path: str, motd: str | None = None) -> int:
@@ -343,15 +341,21 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
                             # Uses non-blocking polling instead of streaming for robustness
                             targets = [ExporterStatus.LEASE_READY, ExporterStatus.BEFORE_LEASE_HOOK_FAILED]
 
-                            # The monitor reports no status until its first poll, so settle
-                            # that first: attaching to a lease that is already LEASE_READY
-                            # must not claim to be waiting on a hook that already ran.
-                            result = await monitor.wait_for_any_of(targets, timeout=HOOK_PROBE_TIMEOUT)
+                            # The monitor reports no status until its first poll, so
+                            # wait for that observation before saying anything:
+                            # attaching to a lease that is already LEASE_READY must
+                            # not claim to be waiting on a hook that already ran.
+                            # Waiting on the observation rather than a fixed settle
+                            # time keeps that true on a slow or distant link, where a
+                            # wall-clock probe would expire before the first answer.
+                            deadline = anyio.current_time() + _HOOK_TIMEOUT
+                            await monitor.wait_for_first_observation(timeout=_HOOK_TIMEOUT)
+                            result = monitor.current_status if monitor.current_status in targets else None
 
                             if result is None and not monitor.connection_lost:
                                 logger.info("Waiting for beforeLease hook to complete...")
                                 result = await monitor.wait_for_any_of(
-                                    targets, timeout=HOOK_TIMEOUT - HOOK_PROBE_TIMEOUT
+                                    targets, timeout=max(0.0, deadline - anyio.current_time())
                                 )
 
                             if result == ExporterStatus.BEFORE_LEASE_HOOK_FAILED:
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/shell_test.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/shell_test.py
@@ -839,6 +839,9 @@ class _FakeStatusMonitor:
     def connection_lost(self):
         return self._connection_lost
 
+    async def wait_for_first_observation(self, timeout=None):
+        return self.current_status is not None
+
     async def wait_for_any_of(self, targets, timeout=None):
         for s in self._statuses:
             if s in targets:
--- a/python/packages/jumpstarter/jumpstarter/client/status_monitor.py
+++ b/python/packages/jumpstarter/jumpstarter/client/status_monitor.py
@@ -88,11 +88,18 @@ class StatusMonitor:
         # Track if connection was lost (UNAVAILABLE)
         self._connection_lost: bool = False
 
+        # Set once the first GetStatus answer has been processed, or once the
+        # poll loop stops without ever getting one. Until then current_status is
+        # None, which a caller cannot tell apart from "observed, but not the
+        # status you asked for".
+        self._first_observation: Event = Event()
+
     def _signal_unsupported(self):
         """Mark GetStatus as unsupported and signal waiters with LEASE_READY."""
         self._get_status_unsupported = True
         self._current_status = ExporterStatus.LEASE_READY
         self._running = False
+        self._first_observation.set()
         self._any_change_event.set()
         self._any_change_event = Event()
 
@@ -216,6 +223,27 @@ class StatusMonitor:
             return False
         else:
             return await wait_loop()
+
+    async def wait_for_first_observation(self, timeout: float | None = None) -> bool:
+        """Wait until the first GetStatus answer has been processed.
+
+        Until that happens current_status is None, which reads the same as "not
+        the status you asked for". A caller that has to tell those apart — so it
+        does not report waiting on something that already finished — waits for
+        this rather than guessing a settle time, which a slow or distant link
+        would outlast.
+
+        Returns True once a status has been observed, False if the wait timed
+        out or the monitor stopped without ever getting an answer. Returns
+        immediately when GetStatus is unsupported, where LEASE_READY is assumed
+        without polling.
+        """
+        if timeout is None:
+            await self._first_observation.wait()
+        else:
+            with anyio.move_on_after(timeout):
+                await self._first_observation.wait()
+        return self._current_status is not None
 
     async def wait_for_any_of(  # noqa: C901
         self, targets: list[ExporterStatus], timeout: float | None = None
@@ -358,6 +386,7 @@ class StatusMonitor:
                 self._status_message = response.message or ""
                 self._status_version = new_version
                 self._previous_status = previous
+                self._first_observation.set()
 
                 # Fire events if status changed
                 if old_status != new_status:
@@ -462,6 +491,9 @@ class StatusMonitor:
                 break
 
         logger.debug("Status monitor poll loop exited (running=%s)", self._running)
+        # Nothing else will observe a status now, so release anyone waiting on
+        # the first one rather than leaving them to sit out their timeout.
+        self._first_observation.set()
 
     async def start(self, task_group=None):
         """Start the background polling task.
--- a/python/packages/jumpstarter/jumpstarter/client/status_monitor.py
+++ b/python/packages/jumpstarter/jumpstarter/client/status_monitor.py
@@ -361,7 +361,12 @@ class StatusMonitor:
 
                 # Fire events if status changed
                 if old_status != new_status:
-                    logger.info(f"Status changed: {old_status} -> {new_status} (version={new_version})")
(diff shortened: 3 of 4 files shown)


python/packages/jumpstarter/jumpstarter/client/status_monitor.py
```diff
@@ -361,7 +361,12 @@ async def _poll_loop(self):  # noqa: C901
 
                 # Fire events if status changed
                 if old_status != new_status:
-                    logger.info(f"Status changed: {old_status} -> {new_status} (version={new_version})")
+                    # The first poll is an observation, not a transition: reporting
+                    # it as one is noise when attaching to an already-ready lease.
+                    if old_status is None:
+                        logger.debug(f"Exporter status: {new_status} (version={new_version})")
+                    else:
+                        logger.info(f"Status changed: {old_status} -> {new_status} (version={new_version})")
```
raballew: [HIGH] The code correctly emits DEBUG on the first poll and INFO on subsequent transitions, but there is no test verifying this distinction. `status_monitor_test.py` has no `caplog` usage at all, so a future refactor could accidentally promote the first-observation log to INFO without any test catching it.

Suggested fix: add `test_first_observation_logged_at_debug_not_info` to `TestStatusMonitorPolling` using `caplog.at_level(logging.DEBUG)`. Assert zero INFO records containing "Status changed" and at least one DEBUG record containing "Exporter status" after one poll cycle. Pair it with a test confirming a genuine second status change does produce an INFO "Status changed" record.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -336,12 +341,18 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
...
+                            # The monitor reports no status until its first poll, so settle
+                            # that first: attaching to a lease that is already LEASE_READY
+                            # must not claim to be waiting on a hook that already ran.
+                            result = await monitor.wait_for_any_of(targets, timeout=HOOK_PROBE_TIMEOUT)
+
+                            if result is None and not monitor.connection_lost:
+                                logger.info("Waiting for beforeLease hook to complete...")
+                                result = await monitor.wait_for_any_of(
+                                    targets, timeout=HOOK_TIMEOUT - HOOK_PROBE_TIMEOUT
+                                )
```
raballew, line 359: Hmmm, slow networks might not work with 2s delays. I am thinking about max distance and higher latency of some of the newer use-cases (lets say you are on a plane with starlink and need to dial to the other side of the planet) 

I wonder if you add a `_first_observation_done: asyncio.Event` to `StatusMonitor.__init__`, set it after the first successful `GetStatus` response is processed, and expose `async def wait_for_first_observation(self, timeout)` if that might be a better solution. In `shell.py`, await that instead of relying on the wall-clock probe.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -54,6 +54,11 @@
 # Refresh token when less than this many seconds remain
 _TOKEN_REFRESH_THRESHOLD_SECONDS = 120
 
+# Total time to wait for the beforeLease hook, and the initial slice of it spent
+# settling the exporter's current status before announcing that we are waiting.
+HOOK_TIMEOUT = 300.0
+HOOK_PROBE_TIMEOUT = 2.0
```
raballew: Rename to private `_HOOK_TIMEOUT: float = 300.0` and `_HOOK_PROBE_TIMEOUT: float = 2.0`?

python/packages/jumpstarter/jumpstarter/client/status_monitor.py
```diff
@@ -361,7 +361,12 @@ async def _poll_loop(self):  # noqa: C901
 
                 # Fire events if status changed
                 if old_status != new_status:
-                    logger.info(f"Status changed: {old_status} -> {new_status} (version={new_version})")
+                    # The first poll is an observation, not a transition: reporting
+                    # it as one is noise when attaching to an already-ready lease.
+                    if old_status is None:
+                        logger.debug(f"Exporter status: {new_status} (version={new_version})")
+                    else:
```
raballew: [LOW] `test_on_status_change_callback` records `(new_status, old_status)` tuples but only asserts `len(callback_invocations) >= 1`. After this PR, the first poll fires the callback with `old_status=None` (handled by the new `if old_status is None` branch here). Nothing checks that, so a future change that suppresses callbacks on first observation would pass silently.

Suggested fix: assert explicitly in the callback test that the first invocation has `old_status is None`, and add a separate test confirming that genuine subsequent transitions produce an invocation with `old_status is not None`.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -54,6 +54,9 @@
 # Refresh token when less than this many seconds remain
 _TOKEN_REFRESH_THRESHOLD_SECONDS = 120
 
+# Total time to wait for the beforeLease hook.
+_HOOK_TIMEOUT: float = 300.0
```
raballew: [Critical] All four changed files fail `ruff format --check`. CI's ruff action runs `ruff format --check` alongside lint, so this will block the merge. Fix: run `uv run ruff format` on `status_monitor.py`, `status_monitor_test.py`, `shell.py`, and `shell_test.py`, then commit.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -336,12 +339,24 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
...
+                            # wall-clock probe would expire before the first answer.
+                            deadline = anyio.current_time() + _HOOK_TIMEOUT
+                            await monitor.wait_for_first_observation(timeout=_HOOK_TIMEOUT)
+                            result = monitor.current_status if monitor.current_status in targets else None
+
+                            if result is None and not monitor.connection_lost:
+                                logger.info("Waiting for beforeLease hook to complete...")
+                                result = await monitor.wait_for_any_of(
+                                    targets, timeout=max(0.0, deadline - anyio.current_time())
+                                )
```
raballew, line 359: The return value of `await monitor.wait_for_first_observation(timeout=_HOOK_TIMEOUT)` is not captured. What if `GetStatus` stays `UNAVAILABLE` for the entire timeout budget?

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -54,6 +54,9 @@
 # Refresh token when less than this many seconds remain
 _TOKEN_REFRESH_THRESHOLD_SECONDS = 120
```
raballew: [Low] `_TOKEN_REFRESH_THRESHOLD_SECONDS = 120` has no type annotation, while the adjacent `_HOOK_TIMEOUT: float = 300.0` at line 58 is correctly typed. Add `: int` for consistency: `_TOKEN_REFRESH_THRESHOLD_SECONDS: int = 120`.

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
```diff
@@ -336,12 +339,24 @@ async def _run_shell_with_lease_async(lease, exporter_logs, config, command, can
...
+                            targets = [ExporterStatus.LEASE_READY, ExporterStatus.BEFORE_LEASE_HOOK_FAILED]
+
+                            # The monitor reports no status until its first poll, so
+                            # wait for that observation before saying anything:
+                            # attaching to a lease that is already LEASE_READY must
+                            # not claim to be waiting on a hook that already ran.
+                            # Waiting on the observation rather than a fixed settle
+                            # time keeps that true on a slow or distant link, where a
+                            # wall-clock probe would expire before the first answer.
+                            deadline = anyio.current_time() + _HOOK_TIMEOUT
```
raballew: [Medium] The beforeLease path here correctly uses `timeout=_HOOK_TIMEOUT`, but the afterLease hook wait at line 456 uses `timeout=300.0` as a bare literal. A maintainer who changes `_HOOK_TIMEOUT` would miss that call site. Replace `timeout=300.0` at line 456 with `timeout=_HOOK_TIMEOUT`.

python/packages/jumpstarter-cli/jumpstarter_cli/shell_test.py
```diff
@@ -839,6 +839,9 @@ def status_message(self):
     def connection_lost(self):
         return self._connection_lost
 
+    async def wait_for_first_observation(self, timeout=None):
```
raballew: [High] The existing `TestRunShellWithLeaseAsync` tests use `_FakeStatusMonitor(statuses=[ExporterStatus.LEASE_READY])` but none captures log output. This means you can delete the guard at the `if result is None and not monitor.connection_lost:` branch in `shell.py` (which prevents spurious hook-wait logging) and all tests still pass. Consider adding a `caplog.at_level(logging.INFO)` assertion confirming 'Waiting for beforeLease hook' does not appear when the monitor starts at `LEASE_READY`.

python/packages/jumpstarter-cli/jumpstarter_cli/shell_test.py
```diff
@@ -839,6 +839,9 @@ def status_message(self):
...
+        return self.current_status is not None
```
raballew: [High] No test exercises the path where `monitor.current_status` begins as `BEFORE_LEASE_HOOK`. The 'Waiting for beforeLease hook to complete...' message in `shell.py` is untested; deleting that log call leaves all tests green. Add a test with `_FakeStatusMonitor(statuses=[ExporterStatus.BEFORE_LEASE_HOOK, ExporterStatus.LEASE_READY])` using `caplog` at INFO level, asserting the message appears exactly once.

raballew: [High] `wait_for_first_observation` is missing type annotations. The production signature is `async def wait_for_first_observation(self, timeout: float | None = None) -> bool:` but the fake has no parameter type and no return type. Type-checker drift between the fake and real class goes undetected. Fix: `async def wait_for_first_observation(self, timeout: float | None = None) -> bool:`

raballew: [Medium] The fake returns immediately without any `await`, so no shell test can exercise: (a) the case where `wait_for_first_observation` genuinely suspends, (b) the deadline arithmetic `max(0.0, deadline - anyio.current_time())`, or (c) a timeout returning `False`. Consider adding a `first_observation_delay` parameter to `_FakeStatusMonitor` with `await anyio.sleep(delay)` before returning.

python/packages/jumpstarter-cli/jumpstarter_cli/shell_test.py
```diff
@@ -839,6 +839,9 @@ def status_message(self):
...
+
     async def wait_for_any_of(self, targets, timeout=None):
         for s in self._statuses:
             if s in targets:
```
raballew: [Medium] `_FakeStatusMonitor` is a hand-rolled duck-type with no declared `Protocol` or ABC. Any API change in `StatusMonitor` not reflected here produces silent test gaps rather than a type error. Consider defining a `StatusMonitorProtocol(Protocol)` in `status_monitor.py` covering `current_status`, `status_message`, `connection_lost`, `wait_for_first_observation`, and `wait_for_any_of`, and annotating the `monitor` parameter in `_run_shell_with_lease_async` with it.

python/packages/jumpstarter-cli/jumpstarter_cli/shell_test.py
raballew: [Low] Lines 750 and 956 suppress a `ty` checker false-positive on `nonlocal call_count` with `# ty: ignore[unresolved-reference]`. The `nonlocal` declaration is correct Python; the suppression is a workaround for a checker deficiency. Worth tracking the upstream ty issue so these can be removed when the false-positive is fixed.

python/packages/jumpstarter/jumpstarter/client/status_monitor.py
```diff
@@ -88,11 +88,18 @@ def __init__(self, stub, poll_interval: float = 0.3, get_status_unsupported: boo
...
+        # None, which a caller cannot tell apart from "observed, but not the
+        # status you asked for".
+        self._first_observation: Event = Event()
+
     def _signal_unsupported(self):
         """Mark GetStatus as unsupported and signal waiters with LEASE_READY."""
         self._get_status_unsupported = True
         self._current_status = ExporterStatus.LEASE_READY
         self._running = False
+        self._first_observation.set()
```
raballew: [Medium] In `wait_for_status`, the inner `wait_loop()` exits the `while self._running` loop without a post-loop status check. By contrast, `wait_for_any_of` has an explicit post-loop check. When `_signal_unsupported()` fires (as in this hunk), `_running` is set False before `_any_change_event.set()`, so a waiter on `LEASE_READY` can wake, see `_running=False`, and return `False` even though `_current_status == LEASE_READY`. The test comment 'will never be reached' in `test_wait_for_status_unimplemented_returns_promptly` is factually incorrect. Add a post-loop check to `wait_loop()`: `if self._current_status == target: return True; return False`.

python/packages/jumpstarter/jumpstarter/client/status_monitor.py
```diff
@@ -217,6 +224,27 @@ async def wait_loop():
...
+        out or the monitor stopped without ever getting an answer. Returns
+        immediately when GetStatus is unsupported, where LEASE_READY is assumed
+        without polling.
+        """
+        if timeout is None:
+            await self._first_observation.wait()
+        else:
+            with anyio.move_on_after(timeout):
+                await self._first_observation.wait()
+        return self._current_status is not None
```
raballew: [High] The `if timeout is None` branch (direct `await self._first_observation.wait()` without a `move_on_after` guard) is never exercised by any test. All four `TestWaitForFirstObservation` tests pass an explicit float timeout. Removing this branch entirely leaves all tests passing. Add a test calling `await monitor.wait_for_first_observation()` with no argument, guarded by `anyio.fail_after(2.0)`, asserting it returns `True` once the poll loop fires.

raballew: [Low] The `wait_for_first_observation` docstring does not document the intended two-phase usage pattern: call `wait_for_first_observation` first, read `current_status`, then conditionally call `wait_for_any_of` under a shared deadline. A future developer has to read `shell.py` to discover this pattern. Adding a Usage section to the docstring would make it self-explanatory.

python/packages/jumpstarter/jumpstarter/client/status_monitor_test.py
```diff
@@ -995,3 +995,62 @@ async def test_long_after_hook_survives_deadline_exceeded(self) -> None:
 
         assert result == ExporterStatus.AVAILABLE
         assert monitor.connection_lost is False
```
raballew: [High] `status_monitor_test.py` has no `caplog` usage anywhere. This means changing `logger.debug(...)` back to `logger.info(...)` for the first-observation path restores the old spurious 'Status changed: None ->' INFO message without breaking any test. Add a `caplog` test asserting no INFO record contains 'Status changed: None ->' on first observation, and that a subsequent real transition does produce an INFO record.

python/packages/jumpstarter/jumpstarter/client/status_monitor_test.py
```diff
@@ -995,3 +995,62 @@ async def test_long_after_hook_survives_deadline_exceeded(self) -> None:
...
+        async with anyio.create_task_group() as tg:
+            await monitor.start(tg)
+            assert await monitor.wait_for_first_observation(timeout=2.0) is True
+            assert monitor.current_status == ExporterStatus.LEASE_READY
+            await monitor.stop()
+
+    async def test_releases_waiters_when_the_monitor_stops(self) -> None:
+        """A stopped monitor will never observe anything, so waiters must not
+        sit out their whole timeout."""
+        stub = MockExporterStub([AioRpcError(StatusCode.UNAVAILABLE, None, None)])
```
raballew, line 1049: Consider using `create_mock_rpc_error(StatusCode.UNAVAILABLE)` for consistency.

mangelajo (approved):

python/packages/jumpstarter-cli/jumpstarter_cli/shell.py
mangelajo: Good catch. Tracing through the code, here's the concrete scenario:

1. `wait_for_first_observation()` returns `True` (a status was observed), but the return value is discarded.
2. The observed status is `UNAVAILABLE` — not in `targets`, so `result = None`.
3. `UNAVAILABLE` sets `monitor.connection_lost = True`, so the `if result is None and not monitor.connection_lost` guard is `False` — the `wait_for_any_of` call is skipped entirely.
4. Execution falls through to the existing `elif result is None` block (line 349), which checks `monitor.connection_lost` (True) and **silently returns 0**, as if the lease expired gracefully.

So if `GetStatus` stays `UNAVAILABLE` for the full timeout budget, the shell exits successfully instead of raising an error — the user gets no indication that the exporter never became ready.

mangelajo: IDK, since this is mostly aesthetic. I'd propose that we leave this alone, and eventually tackle it when we have a FSM in rust , python or whatever :D

WDYT?

raballew: Is the FSM work scheduled already or somewhere deeply nested in the backlog? I think if we implement the FSM soonish, its okay to leave it as is.

mangelajo:
Based on our later community conversations, we will better leave this alone until we go full FSM

PR #1042, opened 2026-08-30
refactor(client): move driver introspection into jumpstarter.client for CLI/MCP parity

kirkbrauer:
Driver introspection lived in the MCP server, so it was reachable over MCP and nowhere else. This moves it into `jumpstarter.client`, with the MCP server as its first caller, so the CLI and anything written in Python can ask the same questions.

Two fixes come with the move: introspection is now driven from outside the event loop — it previously opened a blocking portal inside one and drove sync driver clients from that thread, which raised at runtime, and no test caught it because the fake client never dispatched — and each command's parameters are now described well enough for a caller to prompt for them: kind, type, required, choices, flags.

### DIFF
--- a/python/packages/jumpstarter-mcp/jumpstarter_mcp/introspect.py
+++ b/python/packages/jumpstarter-mcp/jumpstarter_mcp/introspect.py
@@ -1,205 +1,25 @@
-"""Introspection utilities for Click CLI trees and driver object trees."""
+"""Backward-compatible re-exports of the driver introspection utilities.
 
-from __future__ import annotations
+The implementation lives in jumpstarter.client.introspect so it can be shared
+between the CLI and the MCP server.
+"""
 
-import inspect
-import logging
-from typing import Any
+from jumpstarter.client.introspect import (
+    _BASE_METHODS,
+    _INTERNAL_METHODS,
+    _get_public_method_names,
+    _inspect_method,
+    get_driver_methods,
+    list_drivers,
+    walk_click_tree,
+)
 
-import click
-import click.core
-
-logger = logging.getLogger(__name__)
-
-
-def walk_click_tree(cmd: click.core.BaseCommand, path: list[str] | None = None) -> dict[str, Any]:  # ty: ignore[unresolved-attribute]
-    """Recursively walk a Click command tree and return structured JSON.
-
-    Returns command names, help text, parameters (with types and defaults),
-    and nested subcommands.
-    """
-    path = path or []
-    result: dict[str, Any] = {
-        "name": cmd.name,
-        "help": cmd.help,
-        "params": [
-            {
-                "name": p.name,
-                "type": str(p.type),
-                "help": getattr(p, "help", None),
-                "required": getattr(p, "required", False),
-                "default": p.default if p.default is not None else None,
-            }
-            for p in cmd.params
-            if not getattr(p, "hidden", False) and p.name != "help"
-        ],
-    }
-    if isinstance(cmd, click.Group):
-        result["subcommands"] = {
-            name: walk_click_tree(sub, path + [name])
-            for name, sub in cmd.commands.items()
-        }
-    return result
-
-
-# Methods on the base DriverClient that should be excluded from introspection
-_BASE_METHODS = {
-    "call",
-    "streamingcall",
-    "stream",
-    "stream_async",
-    "log_stream",
-    "log_stream_async",
-    "open_stream",
-    "close",
-    "reset",
-    "cli",
-    "call_async",
-    "streamingcall_async",
-    "report",
-    "report_async",
-    "get_status_async",
-    "end_session_async",
-    "status_monitor_async",
-}
-
-# Methods inherited from base classes that add noise to driver listings
-_INTERNAL_METHODS = {
-    "check_exporter_status",
-    "resource_async",
-    "wait_for_hook_complete_monitored",
-    "wait_for_hook_status",
-    "wait_for_lease_ready",
-    "wait_for_lease_ready_monitored",
-}
-
-
-def _get_public_method_names(obj: Any) -> list[str]:
-    """Return public method names for a driver instance, excluding base internals.
-
-    Uses getmembers_static to avoid triggering property descriptors, which can
-    make gRPC calls that fail when invoked from the event loop thread.
-    """
-    names = []
-    try:
-        members = inspect.getmembers_static(obj)
-    except Exception:
-        logger.debug("inspect.getmembers_static failed for %s", type(obj).__name__, exc_info=True)
-        return names
-    for name, value in members:
-        if name.startswith("_") or name in _BASE_METHODS or name in _INTERNAL_METHODS:
-            continue
-        if isinstance(value, property):
-            continue
-        if not callable(value):
-            continue
-        try:
-            inspect.signature(value)
-        except (ValueError, TypeError, AttributeError):
-            continue
-        names.append(name)
-    return names
-
-
-def list_drivers(client: Any, _keys: list[str] | None = None) -> list[dict[str, Any]]:
-    """Flatten the driver client tree into a list with dot-separated paths.
-
-    Returns path (Python access path like ``client.power``), driver_path
-    (children-key list like ``["power"]`` for use with get_driver_methods),
-    class name, description, and method names for each driver.
-    """
-    keys = _keys or []
-    cls = type(client)
-    results = [
-        {
-            "path": f"client.{'.'.join(keys)}" if keys else "client",
-            "driver_path": keys,
-            "class": f"{cls.__module__}.{cls.__qualname__}",
-            "description": getattr(client, "description", None),
-            "methods": _get_public_method_names(client),
-        }
-    ]
-    children = getattr(client, "children", {})
-    for name, child in children.items():
-        results.extend(list_drivers(child, keys + [name]))
-    return results
-
-
-def _inspect_method(name: str, method: Any, driver_path: list[str]) -> dict[str, Any] | None:
-    """Build a method descriptor dict, or return None if the method can't be inspected."""
-    try:
-        sig = inspect.signature(method)
-    except (ValueError, TypeError, AttributeError):
-        return None
-
-    is_streaming = False
-    try:
-        source = inspect.getsource(method)
-        is_streaming = "streamingcall" in source
-    except (OSError, TypeError):
-        pass
-
-    params = [
-        {
-            "name": pname,
-            "annotation": str(p.annotation) if p.annotation != inspect.Parameter.empty else None,
-            "default": str(p.default) if p.default != inspect.Parameter.empty else None,
-        }
-        for pname, p in sig.parameters.items()
-        if pname != "self"
-    ]
-
-    attr_path = ".".join(driver_path)
-    call_args = ", ".join(f"{p['name']}=..." for p in params)
-    method_call = f"client.{attr_path}.{name}({call_args})"
-
-    return {
-        "name": name,
-        "signature": str(sig),
-        "docstring": inspect.getdoc(method),
-        "parameters": params,
-        "return_type": str(sig.return_annotation) if sig.return_annotation != inspect.Signature.empty else None,
-        "is_streaming": is_streaming,
-        "call_example": (
-            "from jumpstarter.utils.env import env\n\n"
-            "with env() as client:\n"
-            f"    {method_call}"
-        ),
-    }
-
-
-def get_driver_methods(client: Any, driver_path: list[str]) -> dict[str, Any]:
-    """Inspect a specific driver client at the given path in the children tree.
-
-    Uses Python inspect to return detailed method information for all public
-    methods defined on the concrete driver class.
-    """
-    target = client
-    for key in driver_path:
-        children = getattr(target, "children", {})
-        if key not in children:
-            raise KeyError(f"Driver path component '{key}' not found. Available: {list(children.keys())}")
-        target = children[key]
-
-    cls = type(target)
-    try:
-        members = inspect.getmembers_static(target)
-    except Exception:
-        logger.debug("inspect.getmembers_static failed for %s", cls.__name__, exc_info=True)
-        members = []
-
-    methods = []
-    for name, value in members:
-        if name.startswith("_") or name in _BASE_METHODS or name in _INTERNAL_METHODS:
-            continue
-        if isinstance(value, property) or not callable(value):
-            continue
-        info = _inspect_method(name, value, driver_path)
-        if info is not None:
-            methods.append(info)
-
-    return {
-        "class": f"{cls.__module__}.{cls.__qualname__}",
-        "driver_path": driver_path,
-        "methods": methods,
-    }
+__all__ = [
+    "_BASE_METHODS",
+    "_INTERNAL_METHODS",
+    "_get_public_method_names",
+    "_inspect_method",
+    "get_driver_methods",
+    "list_drivers",
+    "walk_click_tree",
+]
--- a/python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
+++ b/python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
@@ -10,16 +10,9 @@ from dataclasses import dataclass
 from datetime import datetime
 from unittest.mock import AsyncMock, MagicMock, patch
 
-import click
 import pytest
 
 from jumpstarter_mcp.connections import Connection, ConnectionManager
-from jumpstarter_mcp.introspect import (
-    _get_public_method_names,
-    get_driver_methods,
-    list_drivers,
-    walk_click_tree,
-)
 from jumpstarter_mcp.server import (
     TOKEN_REFRESH_THRESHOLD_SECONDS,
     _ensure_fresh_token,
@@ -112,135 +105,20 @@ class FakeLease:
 
 
 # ---------------------------------------------------------------------------
-# walk_click_tree
+# Introspection re-exports
 # ---------------------------------------------------------------------------
 
 
-class TestWalkClickTree:
-    def test_simple_command(self):
-        @click.command("hello")
-        @click.option("--name", help="Your name")
-        def hello(name):
-            """Say hello."""
+class TestIntrospectReexports:
+    def test_reexports_core_implementation(self):
+        import jumpstarter_mcp.introspect as shim
 
-        result = walk_click_tree(hello)
-        assert result["name"] == "hello"
-        assert result["help"] == "Say hello."
-        assert len(result["params"]) == 1
-        assert result["params"][0]["name"] == "name"
-        assert result["params"][0]["help"] == "Your name"
-        assert "subcommands" not in result
+        import jumpstarter.client.introspect as core
 
-    def test_group_with_subcommands(self):
-        @click.group("root")
-        def root():
-            """Root group."""
-
-        @root.command("sub1")
-        def sub1():
-            """First sub."""
-
-        @root.command("sub2")
-        @click.option("--count", type=int, default=5)
-        def sub2(count):
-            """Second sub."""
-
-        result = walk_click_tree(root)
-        assert result["name"] == "root"
-        assert "subcommands" in result
-        assert "sub1" in result["subcommands"]
-        assert "sub2" in result["subcommands"]
-        assert result["subcommands"]["sub2"]["params"][0]["name"] == "count"
-        assert result["subcommands"]["sub2"]["params"][0]["default"] == 5
-
-    def test_hidden_params_excluded(self):
-        @click.command("cmd")
-        @click.option("--visible", help="shown")
-        @click.option("--secret", hidden=True)
-        def cmd(visible, secret):
-            pass
-
-        result = walk_click_tree(cmd)
-        names = [p["name"] for p in result["params"]]
-        assert "visible" in names
-        assert "secret" not in names
-
-
-# Introspection tests
-# ---------------------------------------------------------------------------
-
-
-class TestGetPublicMethodNames:
-    def test_filters_private_and_base_methods(self):
-        names = _get_public_method_names(FakePowerClient())
-        assert "on" in names
-        assert "off" in names
-        assert "cycle" in names
-        assert "_private" not in names
-        assert "call" not in names
-        assert "check_exporter_status" not in names
-
-
-class TestListDrivers:
-    def test_flat_tree(self):
-        client = FakeCompositeClient()
-        result = list_drivers(client)
-        paths = [d["path"] for d in result]
-        assert "client" in paths
-        assert "client.power" in paths
-        assert "client.serial" in paths
-
-    def test_driver_path_field(self):
-        client = FakeCompositeClient()
-        result = list_drivers(client)
-        root = next(d for d in result if d["path"] == "client")
-        assert root["driver_path"] == []
-        power = next(d for d in result if d["path"] == "client.power")
-        assert power["driver_path"] == ["power"]
-
-    def test_methods_populated(self):
-        client = FakeCompositeClient()
-        result = list_drivers(client)
-        power = next(d for d in result if d["path"] == "client.power")
-        assert "on" in power["methods"]
-        assert "off" in power["methods"]
-
-
-class TestGetDriverMethods:
-    def test_returns_method_details(self):
-        client = FakeCompositeClient()
-        result = get_driver_methods(client, ["power"])
-        assert result["driver_path"] == ["power"]
-        method_names = [m["name"] for m in result["methods"]]
-        assert "on" in method_names
-        assert "off" in method_names
-        assert "cycle" in method_names
-
-    def test_call_example_uses_dot_notation(self):
-        client = FakeCompositeClient()
-        result = get_driver_methods(client, ["power"])
-        cycle = next(m for m in result["methods"] if m["name"] == "cycle")
-        assert "client.power.cycle(" in cycle["call_example"]
-        assert 'children["power"]' not in cycle["call_example"]
-
-    def test_invalid_path_raises(self):
-        client = FakeCompositeClient()
-        with pytest.raises(KeyError, match="nonexistent"):
-            get_driver_methods(client, ["nonexistent"])
-
-    def test_docstrings_captured(self):
-        client = FakeCompositeClient()
-        result = get_driver_methods(client, ["power"])
-        on_method = next(m for m in result["methods"] if m["name"] == "on")
-        assert on_method["docstring"] == "Power on the device."
-
-    def test_parameters_captured(self):
-        client = FakeCompositeClient()
-        result = get_driver_methods(client, ["power"])
-        cycle = next(m for m in result["methods"] if m["name"] == "cycle")
-        assert len(cycle["parameters"]) == 1
-        assert cycle["parameters"][0]["name"] == "wait"
-        assert cycle["parameters"][0]["default"] == "2"
+        assert shim.walk_click_tree is core.walk_click_tree
+        assert shim.list_drivers is core.list_drivers
+        assert shim.get_driver_methods is core.get_driver_methods
+        assert shim._get_public_method_names is core._get_public_method_names
 
 
 # ---------------------------------------------------------------------------
--- a/python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
+++ b/python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
@@ -12,7 +12,8 @@ import anyio.to_thread
 import click
 
 from jumpstarter_mcp.connections import ConnectionManager
-from jumpstarter_mcp.introspect import get_driver_methods, list_drivers, walk_click_tree
+
+from jumpstarter.client.introspect import get_driver_methods, list_drivers, walk_click_tree
 
 logger = logging.getLogger(__name__)
 
--- a/python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
+++ b/python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
@@ -120,7 +120,10 @@ async def drivers(
 ) -> list[dict]:
     """List all drivers in the client tree."""
     conn = manager.get_connection(connection_id)
-    return list_drivers(conn.client)
+    # Off the loop thread, as explore() does. These walk a sync client facade
+    # over a blocking portal; they read attributes statically today and so do
+    # not dispatch, but a driver property that ever did would deadlock the loop.
+    return await anyio.to_thread.run_sync(list_drivers, conn.client)
 
 
 async def driver_methods(
@@ -130,4 +133,4 @@ async def driver_methods(
 ) -> dict:
     """Inspect methods on a specific driver."""
     conn = manager.get_connection(connection_id)
-    return get_driver_methods(conn.client, driver_path)
+    return await anyio.to_thread.run_sync(get_driver_methods, conn.client, driver_path)
--- a/python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/connections.py
+++ b/python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/connections.py
@@ -8,8 +8,8 @@ from datetime import timedelta
 import anyio.to_thread  # noqa: F401
 
 from jumpstarter_mcp.connections import ConnectionManager
-from jumpstarter_mcp.introspect import list_drivers, walk_click_tree
 
+from jumpstarter.client.introspect import list_drivers, walk_click_tree
 from jumpstarter.config.client import ClientConfigV1Alpha1
 
 logger = logging.getLogger(__name__)
--- a/python/packages/jumpstarter/jumpstarter/client/introspect.py
+++ b/python/packages/jumpstarter/jumpstarter/client/introspect.py
@@ -252,8 +252,16 @@ def describe_client(client: Any) -> dict[str, Any]:
     (including when the root client itself is a stub).
     """
     cli_tree = None
+    # Inherited counts: QemuFlasherClient, for one, defines no cli of its own and
+    # gets a real one from FlasherClientInterface, so looking only at
+    # type(client).__dict__ would drop the CLI of every such driver.
     if not isinstance(client, StubDriverClient) and getattr(type(client), "cli", None) is not None:
-        cli_tree = walk_click_tree(client.cli())
+        try:
+            cli_tree = walk_click_tree(client.cli())
+        except Exception:
+            # A driver whose cli() is broken should cost us its CLI tree, not the
+            # whole description — the driver listing below is still worth having.
+            logger.warning("could not build the CLI tree for %s", type(client).__name__, exc_info=True)
     return {
         "drivers": list_drivers(client),
         "cli_tree": cli_tree,
@@ -271,7 +279,11 @@ async def _connect_lease(config: ClientConfigV1Alpha1, lease_name: str, portal: 
         selector=None,
         exporter_name=None,
         lease_name=lease_name,
-        duration=timedelta(minutes=30),
+        # Attaching by name never reaches Lease._create, and with selector None
+        # the "selector changed, make a new one" branch cannot fire either, so
+        # no duration is ever sent to the controller. Naming 30 minutes here
+        # only suggested this call could extend a lease that it cannot.
+        duration=timedelta(0),
         portal=portal,
     ) as lease:
         async with lease.serve_unix_async() as path:
--- a/python/packages/jumpstarter/jumpstarter/client/introspect.py
+++ b/python/packages/jumpstarter/jumpstarter/client/introspect.py
@@ -1,0 +1,278 @@
+"""Introspection utilities for Click CLI trees and driver object trees."""
+
+from __future__ import annotations
+
+import inspect
+import logging
+from contextlib import ExitStack
+from datetime import timedelta
+from typing import TYPE_CHECKING, Any
+
+import click
+import click.core
+from anyio.from_thread import BlockingPortal, start_blocking_portal
+
+from .base import StubDriverClient
+from .client import client_from_path
+
+if TYPE_CHECKING:
+    from jumpstarter.config.client import ClientConfigV1Alpha1
+
+logger = logging.getLogger(__name__)
+
+
+def _json_safe(value: Any) -> Any:
+    """Coerce a value to a JSON-serializable equivalent, stringifying as a last resort."""
+    if value is None or isinstance(value, (bool, int, float, str)):
+        return value
+    if isinstance(value, (list, tuple)):
+        return [_json_safe(v) for v in value]
+    return str(value)
+
+
+def walk_click_tree(cmd: click.core.BaseCommand, path: list[str] | None = None) -> dict[str, Any]:  # ty: ignore[unresolved-attribute]
+    """Recursively walk a Click command tree and return structured JSON.
+
+    Returns command names, help text, parameters (with types and defaults),
+    and nested subcommands.
+    """
+    path = path or []
+    result: dict[str, Any] = {
+        "name": cmd.name,
+        "help": cmd.help,
+        "params": [
+            {
+                "name": p.name,
+                "type": str(p.type),
+                "help": getattr(p, "help", None),
+                "required": getattr(p, "required", False),
+                "default": _json_safe(p.default) if p.default is not None else None,
+            }
+            for p in cmd.params
+            if not getattr(p, "hidden", False) and p.name != "help"
+        ],
+    }
+    if isinstance(cmd, click.Group):
+        result["subcommands"] = {
+            name: walk_click_tree(sub, path + [name])
+            for name, sub in cmd.commands.items()
+        }
+    return result
+
+
+# Methods on the base DriverClient that should be excluded from introspection
+_BASE_METHODS = {
+    "call",
+    "streamingcall",
+    "stream",
+    "stream_async",
+    "log_stream",
+    "log_stream_async",
+    "open_stream",
+    "close",
+    "reset",
+    "cli",
+    "call_async",
+    "streamingcall_async",
+    "report",
+    "report_async",
+    "get_status_async",
+    "end_session_async",
+    "status_monitor_async",
+}
+
+# Methods inherited from base classes that add noise to driver listings
+_INTERNAL_METHODS = {
+    "check_exporter_status",
+    "resource_async",
+    "wait_for_hook_complete_monitored",
+    "wait_for_hook_status",
+    "wait_for_lease_ready",
+    "wait_for_lease_ready_monitored",
+}
+
+
+def _get_public_method_names(obj: Any) -> list[str]:
+    """Return public method names for a driver instance, excluding base internals.
+
+    Uses getmembers_static to avoid triggering property descriptors, which can
+    make gRPC calls that fail when invoked from the event loop thread.
+    """
+    names = []
+    try:
+        members = inspect.getmembers_static(obj)
+    except Exception:
+        logger.debug("inspect.getmembers_static failed for %s", type(obj).__name__, exc_info=True)
+        return names
+    for name, value in members:
+        if name.startswith("_") or name in _BASE_METHODS or name in _INTERNAL_METHODS:
+            continue
+        if isinstance(value, property):
+            continue
+        if not callable(value):
+            continue
+        try:
+            inspect.signature(value)
+        except (ValueError, TypeError, AttributeError):
+            continue
+        names.append(name)
+    return names
+
+
+def list_drivers(client: Any, _keys: list[str] | None = None) -> list[dict[str, Any]]:
+    """Flatten the driver client tree into a list with dot-separated paths.
+
+    Returns path (Python access path like ``client.power``), driver_path
+    (children-key list like ``["power"]`` for use with get_driver_methods),
+    class name, description, and method names for each driver.
+    """
+    keys = _keys or []
+    cls = type(client)
+    results = [
+        {
+            "path": f"client.{'.'.join(keys)}" if keys else "client",
+            "driver_path": keys,
+            "class": f"{cls.__module__}.{cls.__qualname__}",
+            "description": getattr(client, "description", None),
+            "methods": _get_public_method_names(client),
+        }
+    ]
+    children = getattr(client, "children", {})
+    for name, child in children.items():
+        results.extend(list_drivers(child, keys + [name]))
+    return results
+
+
+def _inspect_method(name: str, method: Any, driver_path: list[str]) -> dict[str, Any] | None:
+    """Build a method descriptor dict, or return None if the method can't be inspected."""
+    try:
+        sig = inspect.signature(method)
+    except (ValueError, TypeError, AttributeError):
+        return None
+
+    is_streaming = False
+    try:
+        source = inspect.getsource(method)
+        is_streaming = "streamingcall" in source
+    except (OSError, TypeError):
+        pass
+
+    params = [
+        {
+            "name": pname,
+            "annotation": str(p.annotation) if p.annotation != inspect.Parameter.empty else None,
+            "default": str(p.default) if p.default != inspect.Parameter.empty else None,
+        }
+        for pname, p in sig.parameters.items()
+        if pname != "self"
+    ]
+
+    attr_path = ".".join(driver_path)
+    call_args = ", ".join(f"{p['name']}=..." for p in params)
+    method_call = f"client.{attr_path}.{name}({call_args})"
+
+    return {
+        "name": name,
+        "signature": str(sig),
+        "docstring": inspect.getdoc(method),
+        "parameters": params,
+        "return_type": str(sig.return_annotation) if sig.return_annotation != inspect.Signature.empty else None,
+        "is_streaming": is_streaming,
+        "call_example": (
+            "from jumpstarter.utils.env import env\n\n"
+            "with env() as client:\n"
+            f"    {method_call}"
+        ),
+    }
+
+
+def get_driver_methods(client: Any, driver_path: list[str]) -> dict[str, Any]:
+    """Inspect a specific driver client at the given path in the children tree.
+
+    Uses Python inspect to return detailed method information for all public
+    methods defined on the concrete driver class.
+    """
+    target = client
+    for key in driver_path:
+        children = getattr(target, "children", {})
+        if key not in children:
+            raise KeyError(f"Driver path component '{key}' not found. Available: {list(children.keys())}")
+        target = children[key]
+
+    cls = type(target)
+    try:
+        members = inspect.getmembers_static(target)
+    except Exception:
+        logger.debug("inspect.getmembers_static failed for %s", cls.__name__, exc_info=True)
+        members = []
+
+    methods = []
+    for name, value in members:
+        if name.startswith("_") or name in _BASE_METHODS or name in _INTERNAL_METHODS:
+            continue
+        if isinstance(value, property) or not callable(value):
+            continue
+        info = _inspect_method(name, value, driver_path)
+        if info is not None:
+            methods.append(info)
+
+    return {
+        "class": f"{cls.__module__}.{cls.__qualname__}",
+        "driver_path": driver_path,
+        "methods": methods,
+    }
+
+
+def describe_client(client: Any) -> dict[str, Any]:
+    """Build a plain-serializable description of a connected driver client tree.
+
+    Returns the flattened driver listing and the Click CLI tree. Drivers whose
+    client packages are not installed appear as StubDriverClient entries in the
+    listing, and cli_tree is None when the root client does not expose a CLI
+    (including when the root client itself is a stub).
+    """
+    cli_tree = None
+    if not isinstance(client, StubDriverClient) and getattr(type(client), "cli", None) is not None:
+        cli_tree = walk_click_tree(client.cli())
+    return {
+        "drivers": list_drivers(client),
+        "cli_tree": cli_tree,
+    }
+
+
+async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+    """Attach to an existing lease and describe its driver clients and CLI tree.
+
+    Attaches to the lease named lease_name without creating a new lease and
+    without releasing it on exit, builds the root driver client, and returns
+    a plain-serializable dict with drivers and cli_tree keys.
+
+    :raises ValueError: if lease_name is empty
+    :raises jumpstarter.client.exceptions.LeaseError: if the lease has ended
+        or cannot be acquired
+    :raises jumpstarter.common.exceptions.ConnectionError: if the lease does
+        not exist or the controller is unreachable
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
+
+    async with BlockingPortal() as portal:
+        async with config.lease_async(
+            selector=None,
+            exporter_name=None,
+            lease_name=lease_name,
+            duration=timedelta(minutes=30),
+            portal=portal,
+        ) as lease:
+            async with lease.serve_unix_async() as path:
+                with ExitStack() as stack:
+                    async with client_from_path(
+                        path, portal, stack, allow=lease.allow, unsafe=lease.unsafe
+                    ) as client:
+                        return describe_client(client)
+
+
+def describe_devices(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+    """Blocking convenience wrapper around describe_devices_async."""
+    with start_blocking_portal() as portal:
+        return portal.call(describe_devices_async, config, lease_name)
--- a/python/packages/jumpstarter/jumpstarter/client/introspect.py
+++ b/python/packages/jumpstarter/jumpstarter/client/introspect.py
@@ -30,7 +30,13 @@ def _json_safe(value: Any) -> Any:
     if isinstance(value, (list, tuple)):
         return [_json_safe(v) for v in value]
     if isinstance(value, Mapping):
-        return {str(key): _json_safe(item) for key, item in value.items()}
+        result: dict[str, Any] = {}
+        for key, item in value.items():
+            json_key = str(key)
+            if json_key in result:
+                raise ValueError(f"Mapping keys collide after stringification: {json_key!r}")
+            result[json_key] = _json_safe(item)
+        return result
     return str(value)
 
 
--- a/python/packages/jumpstarter/jumpstarter/client/introspect.py
+++ b/python/packages/jumpstarter/jumpstarter/client/introspect.py
@@ -4,12 +4,13 @@ from __future__ import annotations
 
 import inspect
 import logging
-from contextlib import ExitStack
+from contextlib import ExitStack, asynccontextmanager
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any
 
 import click
 import click.core
+from anyio import to_thread
 from anyio.from_thread import BlockingPortal, start_blocking_portal
 
 from .base import StubDriverClient
@@ -240,12 +241,36 @@ def describe_client(client: Any) -> dict[str, Any]:
     }
 
 
-async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+@asynccontextmanager
+async def _connect_lease(config: ClientConfigV1Alpha1, lease_name: str, portal: BlockingPortal):
+    """Attach to an existing lease and yield its root driver client.
+
+    Passing lease_name into lease_async attaches to that lease rather than
+    creating one, and leaves it unreleased on exit.
+    """
+    async with config.lease_async(
+        selector=None,
+        exporter_name=None,
+        lease_name=lease_name,
+        duration=timedelta(minutes=30),
+        portal=portal,
+    ) as lease:
+        async with lease.serve_unix_async() as path:
+            with ExitStack() as stack:
+                async with client_from_path(
+                    path, portal, stack, allow=lease.allow, unsafe=lease.unsafe
+                ) as client:
+                    yield client
+
+
+def describe_devices(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
     """Attach to an existing lease and describe its driver clients and CLI tree.
 
-    Attaches to the lease named lease_name without creating a new lease and
-    without releasing it on exit, builds the root driver client, and returns
-    a plain-serializable dict with drivers and cli_tree keys.
+    Returns a plain-serializable dict with drivers and cli_tree keys.
+
+    Driver clients are synchronous facades over a blocking portal, so the
+    portal has to run its event loop in its own thread and be driven from
+    this one — introspection therefore happens outside the loop.
 
     :raises ValueError: if lease_name is empty
     :raises jumpstarter.client.exceptions.LeaseError: if the lease has ended
@@ -256,23 +281,18 @@ async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) 
     if not lease_name:
         raise ValueError("lease_name must be a non-empty existing lease name")
 
-    async with BlockingPortal() as portal:
-        async with config.lease_async(
-            selector=None,
-            exporter_name=None,
-            lease_name=lease_name,
-            duration=timedelta(minutes=30),
-            portal=portal,
-        ) as lease:
-            async with lease.serve_unix_async() as path:
-                with ExitStack() as stack:
-                    async with client_from_path(
-                        path, portal, stack, allow=lease.allow, unsafe=lease.unsafe
-                    ) as client:
-                        return describe_client(client)
-
-
-def describe_devices(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
-    """Blocking convenience wrapper around describe_devices_async."""
     with start_blocking_portal() as portal:
-        return portal.call(describe_devices_async, config, lease_name)
+        with portal.wrap_async_context_manager(_connect_lease(config, lease_name, portal)) as client:
+            return describe_client(client)
+
+
+async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+    """Async wrapper around describe_devices, run in a worker thread.
+
+    The driver clients cannot be driven from an event loop thread, so the
+    whole attach-and-introspect flow runs off-loop.
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
+
+    return await to_thread.run_sync(describe_devices, config, lease_name)
--- a/python/packages/jumpstarter/jumpstarter/client/introspect.py
+++ b/python/packages/jumpstarter/jumpstarter/client/introspect.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import inspect
 import logging
+from collections.abc import Mapping
 from contextlib import ExitStack, asynccontextmanager
 from datetime import timedelta
 from typing import TYPE_CHECKING, Any
@@ -28,6 +29,8 @@ def _json_safe(value: Any) -> Any:
         return value
     if isinstance(value, (list, tuple)):
         return [_json_safe(v) for v in value]
+    if isinstance(value, Mapping):
+        return {str(key): _json_safe(item) for key, item in value.items()}
     return str(value)
 
 
--- a/python/packages/jumpstarter/jumpstarter/client/introspect.py
+++ b/python/packages/jumpstarter/jumpstarter/client/introspect.py
@@ -69,7 +69,7 @@ def walk_click_tree(cmd: click.core.BaseCommand, path: list[str] | None = None) 
         "name": cmd.name,
         "help": cmd.help,
         "params": [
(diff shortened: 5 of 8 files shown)


mangelajo (approved):
nice, btw, I believe our mcp dependency in jumpstarter-mcp is wrong and should be fastmcp, not sure what happened.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+    """Async wrapper around describe_devices, run in a worker thread.
+
+    The driver clients cannot be driven from an event loop thread, so the
+    whole attach-and-introspect flow runs off-loop.
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
+
+    return await to_thread.run_sync(describe_devices, config, lease_name)
```
mangelajo: `describe_devices` and `describe_devices_async` are new public API added to `jumpstarter.client` but they have no consumers yet — the MCP server's `_register_discovery_tools` still calls `cmd_tools.explore/drivers/driver_methods` which use the lower-level functions directly.

Are you planning to wire the MCP server (or the CLI) to call these in a follow-up? If so, consider mentioning that in the PR description. If this is purely forward-looking library code, it might be cleaner to defer adding it until there's an actual consumer, to avoid shipping dead code that constrains the signature before real usage validates it.

---

python/packages/jumpstarter-mcp/jumpstarter_mcp/introspect.py
```diff
@@ -1,205 +1,25 @@
...
+
+from jumpstarter.client.introspect import (
+    _BASE_METHODS,
+    _INTERNAL_METHODS,
+    _get_public_method_names,
+    _inspect_method,
+    get_driver_methods,
+    list_drivers,
+    walk_click_tree,
+)
```
mangelajo: The backward-compat shim re-exports private names (`_BASE_METHODS`, `_INTERNAL_METHODS`, `_get_public_method_names`, `_inspect_method`). Since `commands.py` now imports directly from `jumpstarter.client.introspect`, and the old `server_test.py` tests are removed, does anything still import from this shim?

If the answer is "nothing outside this package", consider either:
1. Dropping the shim entirely (it was an internal module, not a public API), or
2. At minimum, re-exporting only the public names (`walk_click_tree`, `list_drivers`, `get_driver_methods`) — leading-underscore names were implementation details and re-exporting them freezes them as part of this module's surface.

---

python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
```diff
@@ -12,7 +12,8 @@
 import click
 
 from jumpstarter_mcp.connections import ConnectionManager
-from jumpstarter_mcp.introspect import get_driver_methods, list_drivers, walk_click_tree
+
+from jumpstarter.client.introspect import get_driver_methods, list_drivers, walk_click_tree
```
mangelajo, line 16: The PR description says the original bug was that introspection was "driven from inside the event loop — it opened a blocking portal inside one and drove sync driver clients from that thread, which raised at runtime."

The new `describe_devices` function fixes this correctly by calling `start_blocking_portal()` from outside the loop. However, the MCP tool functions `drivers()` and `driver_methods()` in this file (lines 117-133) are `async` functions that still call `list_drivers` / `get_driver_methods` synchronously, directly on the event loop thread. These pure-introspection functions use `inspect.getmembers_static` specifically to avoid triggering gRPC, so they're safe today, but the risk is subtle — if a driver property or `__getattr__` inadvertently fires a gRPC call, it would deadlock.

Consider wrapping these in `anyio.to_thread.run_sync()` for consistency with how `explore()` already handles `client.cli()` on line 105.

---

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+    listing, and cli_tree is None when the root client does not expose a CLI
+    (including when the root client itself is a stub).
+    """
+    cli_tree = None
+    if not isinstance(client, StubDriverClient) and getattr(type(client), "cli", None) is not None:
+        cli_tree = walk_click_tree(client.cli())
+    return {
+        "drivers": list_drivers(client),
+        "cli_tree": cli_tree,
+    }
```
mangelajo, line 278: The `cli` attribute check uses `getattr(type(client), "cli", None) is not None`, which returns True even for a base `DriverClient` that inherits `cli` as a method. When calling `client.cli()` on a leaf driver (e.g. a `PowerClient`) that doesn't override `cli`, this may return an unexpected Click command or raise.

Consider strengthening the guard — e.g. check whether `cli` is overridden on the concrete class rather than just inherited:

```python
if not isinstance(client, StubDriverClient) and "cli" in type(client).__dict__:
```

Or keep the current approach but catch potential exceptions from `client.cli()`.

---

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+        lease_name=lease_name,
+        duration=timedelta(minutes=30),
+        portal=portal,
+    ) as lease:
+        async with lease.serve_unix_async() as path:
+            with ExitStack() as stack:
+                async with client_from_path(
+                    path, portal, stack, allow=lease.allow, unsafe=lease.unsafe
+                ) as client:
+                    yield client
```
mangelajo, line 304: The `duration=timedelta(minutes=30)` is hardcoded here and serves as a "lease keepalive" timeout, but since `_connect_lease` only attaches to an existing lease (doesn't create one), this value is passed but likely ignored by `lease_async` when `lease_name` is provided. If that's the case, the hardcoded value is misleading — a comment clarifying this would help readers (or consider omitting `duration` if the API allows it when attaching to an existing lease).

---

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
+"""Introspection utilities for Click CLI trees and driver object trees."""
```
mangelajo, line 1: The new public functions (`describe_client`, `describe_devices`, `describe_devices_async`) are not exported from `jumpstarter/client/__init__.py`. If these are intended as public API for the `jumpstarter.client` package (which the PR title suggests — "move driver introspection into jumpstarter.client"), consider adding at least the public names to `__init__.py`'s `__all__` and imports. If they're intentionally kept as submodule-only imports, that's fine too but worth noting explicitly.

---

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,366 @@
...
+        paths = [d["path"] for d in result["drivers"]]
+        assert "client" in paths
+        assert "client.power" in paths
+        assert "client.nested.power" in paths
+        power = next(d for d in result["drivers"] if d["path"] == "client.power")
+        assert power["driver_path"] == ["power"]
+        assert "on" in power["methods"]
+        assert "power" in result["cli_tree"]["subcommands"]
+        assert "nested" in result["cli_tree"]["subcommands"]
+        json.dumps(result)
```
mangelajo: Great to see an E2E test with real `Composite` and `MockPower` drivers via the `serve()` helper — this validates the full introspection path end-to-end. The `json.dumps(result)` call at the end is a nice touch to ensure the output is serializable.

---

python/packages/jumpstarter/jumpstarter/client/introspect.py
mangelajo: yep, we can drop lease time there

mangelajo (changes requested):
some nits, but looks great.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
+
+    with start_blocking_portal() as portal:
+        with portal.wrap_async_context_manager(_connect_lease(config, lease_name, portal)) as client:
+            return describe_client(client)
+
+
+async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
```
raballew: [CRITICAL] Both `describe_devices` and `describe_devices_async` currently take `(config, lease_name)` and internally call `_connect_lease` to acquire a new connection. This means callers who already hold a connected client have no way to pass it in and are forced to bypass the public API by calling `list_drivers` or `get_driver_methods` directly. Change the signatures to `describe_devices(client: Any) -> dict[str, Any]` and `describe_devices_async(client: Any) -> dict[str, Any]` so the functions accept an existing client. Move the `_connect_lease` helper and its associated imports (`ExitStack`, `asynccontextmanager`, `timedelta`, `BlockingPortal`, `start_blocking_portal`, `client_from_path`) into the CLI and MCP callers that own config and lease management.

python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
```diff
@@ -112,135 +105,20 @@ class FakeLease:
...
-        assert len(cycle["parameters"]) == 1
-        assert cycle["parameters"][0]["name"] == "wait"
-        assert cycle["parameters"][0]["default"] == "2"
+        assert shim.walk_click_tree is core.walk_click_tree
+        assert shim.list_drivers is core.list_drivers
+        assert shim.get_driver_methods is core.get_driver_methods
+        assert shim._get_public_method_names is core._get_public_method_names
 
 
 # ---------------------------------------------------------------------------
```
raballew: [HIGH] There are no behavioral tests for the `explore`, `drivers`, or `driver_methods` MCP tools. `TestIntrospectReexports` only verifies symbol identity and never calls these tools end-to-end. A regression in the wiring between any tool handler and the core library would go undetected. Add a `TestMCPIntrospectionTools` class that invokes these tools using the existing `FakeCompositeClient` helpers and asserts on the returned structure (e.g., `driver_path`, method names, CLI tree name).

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+from datetime import timedelta
+from typing import TYPE_CHECKING, Any
+
+import click
+import click.core
+from anyio import to_thread
+from anyio.from_thread import BlockingPortal, start_blocking_portal
+
+from .base import StubDriverClient
+from .client import client_from_path
```
raballew: [HIGH] The introspect module imports `ExitStack`, `asynccontextmanager`, `timedelta`, `BlockingPortal`, `start_blocking_portal`, and `client_from_path` (lines 7-17) solely to implement `_connect_lease` (lines 264-282). This puts connection lifecycle management inside a module whose stated purpose is driver and CLI tree introspection, giving it two distinct responsibilities. Callers that already hold a connected client cannot use `describe_devices` at all. Moving `_connect_lease` and its imports into the CLI and MCP callers (or a dedicated `connect.py` helper) would clean up this layering; fixing the signature issue in `describe_devices` to accept an existing client resolves both at once.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+
+
+@asynccontextmanager
+async def _connect_lease(config: ClientConfigV1Alpha1, lease_name: str, portal: BlockingPortal):
+    """Attach to an existing lease and yield its root driver client.
+
+    Passing lease_name into lease_async attaches to that lease rather than
+    creating one, and leaves it unreleased on exit.
+    """
+    async with config.lease_async(
```
raballew: [MEDIUM] `ruff format --check` reports this file would be reformatted. At least four locations need adjustment: the list comprehension around lines 71-73 and dict comprehension around lines 76-79 should each collapse to a single line, the implicit string concatenation around lines 193-201 should be merged, and the `client_from_path(...)` call around lines 267-270 should fit on one line. Run: `cd python && uv run --no-sync ruff format packages/jumpstarter/jumpstarter/client/introspect.py`

python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
raballew: [MEDIUM] `ruff format --check` reports this file would be reformatted. Two locations not in the diff: the list comprehension around lines 457-461 should collapse to a single line, and there is a triple space before an inline comment at line 534 that should be normalized. Run: `cd python && uv run --no-sync ruff format packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py`

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+        "required": getattr(param, "required", False),
+        "default": _json_safe(param.default) if param.default is not None else None,
+        "opts": list(param.opts),
+        "is_flag": bool(getattr(param, "is_flag", False)),
+        "multiple": bool(getattr(param, "multiple", False)),
+        "nargs": param.nargs,
+    }
+    if isinstance(param.type, click.Choice):
+        described["choices"] = [str(choice) for choice in param.type.choices]
+    return described
```
raballew: [MEDIUM] The `choices` key is only set inside the `isinstance(param.type, click.Choice)` branch, so for any other parameter type the key is absent from the returned dict. A caller doing `param["choices"]` on a non-Choice parameter gets a `KeyError` instead of `None`. Set it unconditionally: `described["choices"] = [str(c) for c in param.type.choices] if isinstance(param.type, click.Choice) else None`. The existing test asserts the Choice case but does not assert `choices is None` for a non-Choice parameter; that assertion should be added too.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+            "annotation": str(p.annotation) if p.annotation != inspect.Parameter.empty else None,
+            "default": str(p.default) if p.default != inspect.Parameter.empty else None,
+        }
+        for pname, p in sig.parameters.items()
+        if pname != "self"
+    ]
+
+    attr_path = ".".join(driver_path)
+    call_args = ", ".join(f"{p['name']}=..." for p in params)
+    method_call = f"client.{attr_path}.{name}({call_args})"
```
raballew: [MEDIUM] When `driver_path` is an empty list, `attr_path` becomes an empty string and the f-string produces `"client..method_name()"`, which is syntactically invalid Python. `list_drivers` handles this correctly at line 153 but `get_driver_methods` does not. Fix: `attr_path = ".".join(driver_path); prefix = f"client.{attr_path}" if attr_path else "client"; method_call = f"{prefix}.{name}({call_args})"`.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+        results.extend(list_drivers(child, keys + [name]))
+    return results
+
+
+def _inspect_method(name: str, method: Any, driver_path: list[str]) -> dict[str, Any] | None:
+    """Build a method descriptor dict, or return None if the method can't be inspected."""
+    try:
+        sig = inspect.signature(method)
+    except (ValueError, TypeError, AttributeError):
+        return None
```
raballew: [MEDIUM] `_inspect_method` returns `None` when `inspect.signature` raises `ValueError`, `TypeError`, or `AttributeError`, and `get_driver_methods` silently skips those entries. This fallback path has no test coverage at all. Add a test that patches `inspect.signature` to raise `ValueError` and verifies `_inspect_method` returns `None`. Also add a test that `get_driver_methods` omits the affected method from its results in that case.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+
+    Uses getmembers_static to avoid triggering property descriptors, which can
+    make gRPC calls that fail when invoked from the event loop thread.
+    """
+    names = []
+    try:
+        members = inspect.getmembers_static(obj)
+    except Exception:
+        logger.debug("inspect.getmembers_static failed for %s", type(obj).__name__, exc_info=True)
+        return names
```
raballew: [MEDIUM] The broad `except Exception` here catches any failure from `inspect.getmembers_static` and returns an empty list with a debug log. Only the happy path is tested; no test forces `getmembers_static` to raise. Add a test using `unittest.mock.patch("inspect.getmembers_static", side_effect=RuntimeError("boom"))` and assert the function returns `[]`.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _json_safe(value: Any) -> Any:
+    """Coerce a value to a JSON-serializable equivalent, stringifying as a last resort."""
+    if value is None or isinstance(value, (bool, int, float, str)):
+        return value
+    if isinstance(value, (list, tuple)):
+        return [_json_safe(v) for v in value]
+    return str(value)
```
raballew: [MEDIUM] `_json_safe` has three branches (primitives, list/tuple recursion, str() fallback) but only the primitive branch is exercised indirectly. There are no direct tests for the list/tuple or fallback paths. Add a `TestJsonSafe` class with explicit tests for each branch: `_json_safe([1, "a", None])`, `_json_safe((1, 2))`, and `_json_safe(SomeOpaqueObject())`.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,366 @@
...
+        paths = [d["path"] for d in result["drivers"]]
+        assert "client.power" in paths
+        assert result["cli_tree"]["name"] == "j"
+        json.dumps(result)
+
+    @pytest.mark.anyio
+    async def test_empty_lease_name_raises(self, mock_config):
+        with pytest.raises(ValueError, match="non-empty"):
+            await describe_devices_async(mock_config, "")
+        mock_config.lease_async.assert_not_called()
```
raballew: [MEDIUM] `test_empty_lease_name_raises` only tests the async guard. The sync guard in `describe_devices` (lines 300-301 of introspect.py) is never exercised. Add a sync variant: `def test_sync_empty_lease_name_raises(self, mock_config): with pytest.raises(ValueError, match="non-empty"): describe_devices(mock_config, ""); mock_config.lease_async.assert_not_called()`.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,366 @@
...
+            stub = StubDriverClient(
+                labels={"jumpstarter.dev/client": "missing_pkg.client.MissingClient"},
+                stub=None,
+                portal=portal,
+                stack=ExitStack(),
+            )
+            result = describe_client(stub)
+
+        assert result["cli_tree"] is None
+        assert result["drivers"][0]["class"].endswith("StubDriverClient")
```
raballew: [MEDIUM] `json.dumps(result)` is called in `test_composite_client_e2e` and `test_attaches_to_named_lease` but not in `test_returns_drivers_and_cli_tree`, `test_cli_tree_none_without_cli`, or `test_stub_client_represented`. A non-serializable value introduced in any of those code paths would slip through undetected. Add `json.dumps(result)` as the last assertion in all three missing tests to match the pattern established by `test_composite_client_e2e`.

python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
raballew: [LOW] `import os` appears inside the body of `run_command` at line 43. `os` is a stdlib module with no import-time side effects and ruff flags this as PLC0415 ("import should be at top-level of a file"). Move `import os` to the top-level import block alongside `asyncio`, `logging`, and `shutil`.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+    """Inspect a specific driver client at the given path in the children tree.
+
+    Uses Python inspect to return detailed method information for all public
+    methods defined on the concrete driver class.
+    """
+    target = client
+    for key in driver_path:
+        children = getattr(target, "children", {})
+        if key not in children:
+            raise KeyError(f"Driver path component '{key}' not found. Available: {list(children.keys())}")
```
raballew: [LOW] The `KeyError` message here includes the full list of available child driver names: `f"Driver path component '{key}' not found. Available: {list(children.keys())}"}`. In a multi-tenant context an unauthorized caller who guesses an invalid path receives a complete enumeration of the driver tree at that level. Drop the available names from the exception message and log them at DEBUG instead: `logger.debug("get_driver_methods: %r not in %r", key, list(children.keys()))` then `raise KeyError(f"Driver path component {key!r} not found")`.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+    """
+    described: dict[str, Any] = {
+        "name": param.name,
+        "kind": param.param_type_name,
+        # Click's own type name ("integer", "path", "choice"); str() on some
+        # types renders an object repr, which is no use in a prompt.
+        "type": getattr(param.type, "name", None) or str(param.type),
+        "help": getattr(param, "help", None),
+        "required": getattr(param, "required", False),
+        "default": _json_safe(param.default) if param.default is not None else None,
```
raballew: [LOW] CLI parameter default values are serialized verbatim into the introspection output. If a driver's Click command has a default that contains a sensitive value (e.g., a default API endpoint with embedded credentials), that value appears in every `walk_click_tree` or `describe_devices` response. Driver authors must know: any Click parameter whose default is sensitive should be marked `hidden=True` to exclude it from introspection output.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+        "opts": list(param.opts),
+        "is_flag": bool(getattr(param, "is_flag", False)),
+        "multiple": bool(getattr(param, "multiple", False)),
+        "nargs": param.nargs,
```
raballew: [LOW] The `nargs` field is included in every `ParameterDescriptor` but is not listed in the parameter schema documentation, and the `multiple` field already communicates variadic arity. There is no test asserting `nargs` presence or value either. Either remove `nargs` (since `multiple` covers the documented use cases) or add it explicitly to the parameter spec and add a test asserting its value for a `nargs=-1` parameter.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,317 @@
...
+        if pname != "self"
+    ]
+
+    attr_path = ".".join(driver_path)
+    call_args = ", ".join(f"{p['name']}=..." for p in params)
+    method_call = f"client.{attr_path}.{name}({call_args})"
+
+    return {
+        "name": name,
+        "signature": str(sig),
```
raballew: [LOW] `_inspect_method` exposes method default values verbatim in both the `parameters[n]["default"]` field and the `signature` string. If a driver method has a sensitive hardcoded default (e.g., `def authenticate(self, token="hardcoded-api-key"):`), that value is returned through the MCP `driver_methods` tool. Add a note to the `_inspect_method` docstring so driver authors know that method defaults must not contain secrets.

python/packages/jumpstarter/jumpstarter/client/introspect.py
raballew: [LOW] `_json_safe` has no `dict` branch. A Click parameter with a dict default (e.g., `default={"host": "localhost"}`) falls through to `str(value)`, producing the Python repr string `"{'host': 'localhost'}"` rather than a structured JSON object. The `json.dumps` assertion still passes because a string is valid JSON, but the caller receives a mangled value. Add: `if isinstance(value, dict): return {k: _json_safe(v) for k, v in value.items()}`.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,366 @@
...
+        with pytest.raises(ValueError, match="non-empty"):
+            await describe_devices_async(mock_config, "")
+        mock_config.lease_async.assert_not_called()
+
+    def test_blocking_wrapper(self, mock_config, mock_client_from_path):
+        result = describe_devices(mock_config, "existing-lease")
+
+        kwargs = mock_config.lease_async.call_args.kwargs
+        assert kwargs["lease_name"] == "existing-lease"
+        assert result["cli_tree"]["name"] == "j"
```
raballew: [LOW] `test_attaches_to_named_lease` (async) calls `json.dumps(result)` but `test_blocking_wrapper` (sync) does not. If a non-serializable value were introduced specifically in the sync wrapper path, it would go undetected. Add `json.dumps(result)` at the end of `test_blocking_wrapper` to mirror the async test.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,366 @@
...
+            pass
+
+        @root.command()
+        @click.argument("path", type=click.Path())
+        def send(path):
+            pass
+
+        params = walk_click_tree(root)["subcommands"]["send"]["params"]
+        # str(click.Path()) is an object repr, which is useless in a prompt.
+        assert params[0]["type"] == "path"
```
raballew: [LOW] There is no test for a parameterless Click command. The expected behavior is `params == []` (an empty list), but no test asserts this. Add a test with a Click command that defines no parameters and assert `result["params"] == []`.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,366 @@
...
+        result = await describe_devices_async(mock_config, "existing-lease")
+
+        kwargs = mock_config.lease_async.call_args.kwargs
+        assert kwargs["lease_name"] == "existing-lease"
+        assert kwargs["selector"] is None
+        assert kwargs["exporter_name"] is None
+        paths = [d["path"] for d in result["drivers"]]
+        assert "client.power" in paths
+        assert result["cli_tree"]["name"] == "j"
+        json.dumps(result)
```
raballew: [LOW] The `mock_client_from_path` fixture unconditionally yields `FakeCompositeClient()`, ignoring all arguments. `test_attaches_to_named_lease` asserts that `lease_async` was called with `selector=None` and `exporter_name=None` but does not verify what `allow` and `unsafe` values were passed to `client_from_path`. A regression that hardcoded or swapped those arguments would pass all current tests. Capture call arguments in the fixture and add assertions for `allow` and `unsafe` in the test.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+    whether the parameter is positional or an option (and which flag spells
+    it), whether it is a boolean flag that takes no value, whether it repeats,
+    and the values it accepts when it is a choice.
+    """
+    described: dict[str, Any] = {
+        "name": param.name,
+        "kind": param.param_type_name,
+        # Click's own type name ("integer", "path", "choice"); str() on some
+        # types renders an object repr, which is no use in a prompt.
+        "type": getattr(param.type, "name", None) or str(param.type),
```
raballew: [HIGH] The parameter descriptor key at this line uses `"type"` but downstream callers expect `"type_name"`. Any caller accessing `descriptor["type_name"]` will get a `KeyError`. The tests at `introspect_test.py:372` and `392` also assert against `["type"]` and will need updating. Fix: rename the key from `"type"` to `"type_name"` in `_describe_param` and update the corresponding test assertions.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+            "annotation": str(p.annotation) if p.annotation != inspect.Parameter.empty else None,
+            "default": str(p.default) if p.default != inspect.Parameter.empty else None,
+        }
+        for pname, p in sig.parameters.items()
+        if pname != "self"
+    ]
+
+    attr_path = ".".join(driver_path)
+    call_args = ", ".join(f"{p['name']}=..." for p in params)
+    method_call = f"client.{attr_path}.{name}({call_args})"
```
raballew: [HIGH] When `driver_path` is an empty list (root client), `'.'.join([])` returns `""`, so the format string produces `"client..method_name(...)"` with a double dot. This is syntactically invalid Python and the output is silently wrong with no exception raised. Fix: guard the format string on whether `attr_path` is non-empty. If `attr_path` is empty, use `f"client.{name}({call_args})"` directly instead.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+    """Async wrapper around describe_devices, run in a worker thread.
+
+    The driver clients cannot be driven from an event loop thread, so the
+    whole attach-and-introspect flow runs off-loop.
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
+
+    return await to_thread.run_sync(describe_devices, config, lease_name)
```
raballew: [HIGH] This file mixes two distinct concerns. Lines 1-244 are purely stateless helpers with no I/O. Lines 272-329 introduce `_connect_lease`, `describe_devices`, and `describe_devices_async`, which manage a full lease connection lifecycle including blocking portal setup, Unix socket serving, and config-driven lease creation. This forces every importer of the stateless utilities to pull in heavy I/O dependencies (`start_blocking_portal`, `BlockingPortal`, `client_from_path`). Consider extracting the lifecycle functions into a separate module (e.g. `jumpstarter.client.describe`), leaving `introspect.py` as a pure, I/O-free module.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+    portal has to run its event loop in its own thread and be driven from
+    this one — introspection therefore happens outside the loop.
+
+    :raises ValueError: if lease_name is empty
+    :raises jumpstarter.client.exceptions.LeaseError: if the lease has ended
+        or cannot be acquired
+    :raises jumpstarter.common.exceptions.ConnectionError: if the lease does
+        not exist or the controller is unreachable
+    """
+    if not lease_name:
```
raballew: [HIGH] The sync `describe_devices` function has an empty-string guard at this line, but it has no test coverage. The only test for the empty-name error path (`test_empty_lease_name_raises`) calls `describe_devices_async`, not the sync variant. Removing this guard would not break any existing test. Fix: add a dedicated synchronous test that calls `describe_devices(config, "")` and asserts `ValueError` is raised, verifying `mock_config.lease_async` is never called.

python/packages/jumpstarter/jumpstarter/client/introspect.py
raballew: [MEDIUM] Both `describe_devices` and `describe_devices_async` take `config` as their first positional argument, but these are exported as public API. A caller writing `describe_devices("my-lease")` gets `TypeError: missing 1 required positional argument: 'lease_name'`. Consider either reading config implicitly via `ClientConfigV1Alpha1.load()` inside the function, or making `config` an optional keyword argument with a `None` default that triggers auto-loading.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,392 @@
...
+        with pytest.raises(ValueError, match="non-empty"):
+            await describe_devices_async(mock_config, "")
+        mock_config.lease_async.assert_not_called()
+
+    def test_blocking_wrapper(self, mock_config, mock_client_from_path):
+        result = describe_devices(mock_config, "existing-lease")
+
+        kwargs = mock_config.lease_async.call_args.kwargs
+        assert kwargs["lease_name"] == "existing-lease"
+        assert result["cli_tree"]["name"] == "j"
```
raballew: [MEDIUM] The tests that exercise `_connect_lease` (`test_attaches_to_named_lease` and `test_blocking_wrapper`) assert `lease_name`, `selector`, and `exporter_name` but never verify the `duration` value passed to `lease_async`. If `duration=timedelta(0)` were changed to any other value, no test would catch it. Add an assertion like `assert kwargs.get("duration") == timedelta(0)` to both tests to lock in this invariant.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+    """
+    async with config.lease_async(
+        selector=None,
+        exporter_name=None,
+        lease_name=lease_name,
+        # Attaching by name never reaches Lease._create, and with selector None
+        # the "selector changed, make a new one" branch cannot fire either, so
+        # no duration is ever sent to the controller. Naming 30 minutes here
+        # only suggested this call could extend a lease that it cannot.
+        duration=timedelta(0),
```
raballew: [MEDIUM] `_connect_lease` passes `duration=timedelta(0)` to `lease_async`. The stated intent is "no duration" but passing an explicit zero may still be treated differently from omitting the argument entirely. Check whether `lease_async` accepts `duration=None` or has a default that omits the duration. If passing zero is genuinely equivalent to omitting it, document this with an inline comment. If not, remove the argument.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+            # whole description — the driver listing below is still worth having.
+            logger.warning("could not build the CLI tree for %s", type(client).__name__, exc_info=True)
+    return {
+        "drivers": list_drivers(client),
+        "cli_tree": cli_tree,
+    }
+
+
+@asynccontextmanager
+async def _connect_lease(config: ClientConfigV1Alpha1, lease_name: str, portal: BlockingPortal):
```
raballew: [MEDIUM] `_connect_lease` is the only private function in this file without a return type annotation. All other private helpers carry full annotations. Add `-> AsyncGenerator[Any, None]` (importing `AsyncGenerator` from `collections.abc`) to the function signature for consistency.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+
+    Uses getmembers_static to avoid triggering property descriptors, which can
+    make gRPC calls that fail when invoked from the event loop thread.
+    """
+    names = []
+    try:
+        members = inspect.getmembers_static(obj)
+    except Exception:
+        logger.debug("inspect.getmembers_static failed for %s", type(obj).__name__, exc_info=True)
+        return names
```
raballew: [MEDIUM] The exception handler in `_get_public_method_names` silently returns an empty list when `inspect.getmembers_static` raises. This fallback has no test coverage. `TestGetPublicMethodNames` only covers the happy path. Add a test that patches `inspect.getmembers_static` with `side_effect=RuntimeError` and asserts the function returns `[]`.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+        "docstring": inspect.getdoc(method),
+        "parameters": params,
+        "return_type": str(sig.return_annotation) if sig.return_annotation != inspect.Signature.empty else None,
+        "is_streaming": is_streaming,
+        "call_example": (
+            "from jumpstarter.utils.env import env\n\n"
+            "with env() as client:\n"
+            f"    {method_call}"
+        ),
+    }
```
raballew: [MEDIUM] `_inspect_method` returns `None` when `inspect.signature` raises `(ValueError, TypeError, AttributeError)`, causing `get_driver_methods` to silently skip that method. No test exercises this code path. Add a test to `TestGetDriverMethods` with a fake driver whose callable raises `ValueError` from `inspect.signature`, asserting the method is absent from `result["methods"]` rather than causing an error.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+    try:
+        sig = inspect.signature(method)
+    except (ValueError, TypeError, AttributeError):
+        return None
+
+    is_streaming = False
+    try:
+        source = inspect.getsource(method)
+        is_streaming = "streamingcall" in source
+    except (OSError, TypeError):
```
raballew: [MEDIUM] Streaming detection scans the raw source text for the literal string `"streamingcall"`. This matches comments, docstrings, and variable names, not just actual usage. Additionally, no test ever asserts `is_streaming == True`, so the classification logic is entirely untested. Consider detecting streaming via a type annotation check (e.g. return type is `Iterator`/`AsyncIterator`), and add at minimum one test with a method containing `"streamingcall"` in its source, plus a test covering the `getsource` failure path.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,392 @@
...
+        # so a caller building a command line has to tell them apart.
+        assert params["port"]["kind"] == "argument"
+        assert params["port"]["required"] is True
+        assert params["port"]["type"] == "integer"
+        assert params["address"]["kind"] == "option"
+        assert params["address"]["opts"] == ["--address"]
+        assert params["address"]["help"] == "Local address to bind"
+        assert params["verbose"]["is_flag"] is True
+        assert params["mode"]["choices"] == ["fast", "slow"]
+        assert params["tag"]["multiple"] is True
```
raballew: [MEDIUM] The test defines a `--tag` option with `multiple=True`. Click sets its default to `()`, which `_json_safe` converts to `[]`. The test asserts `params["tag"]["multiple"] is True` but never checks `params["tag"]["default"] == []`, so the `tuple -> list` conversion is executed but not verified. Also, no test covers the `str()` fallback path in `_json_safe` for non-serializable objects. Add a `default` assertion to this test and a dedicated `TestJsonSafe` class covering all type branches including the fallback.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+        lease_name=lease_name,
+        # Attaching by name never reaches Lease._create, and with selector None
+        # the "selector changed, make a new one" branch cannot fire either, so
+        # no duration is ever sent to the controller. Naming 30 minutes here
+        # only suggested this call could extend a lease that it cannot.
+        duration=timedelta(0),
+        portal=portal,
+    ) as lease:
+        async with lease.serve_unix_async() as path:
+            with ExitStack() as stack:
```
raballew: [LOW] `ExitStack` is created inside the async generator body here. The pattern used elsewhere in the codebase places `ExitStack` at the sync level and passes it into async functions. The current usage is safe for read-only introspection since `open_stream()` is never called, but if this code path ever calls `open_stream()`, `ExitStack.__exit__` would deadlock by calling `portal.call()` from within the same event-loop thread. Consider hoisting `ExitStack` creation to the sync level in `describe_devices` and passing it as a parameter to `_connect_lease`, matching the established pattern.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+    """Inspect a specific driver client at the given path in the children tree.
+
+    Uses Python inspect to return detailed method information for all public
+    methods defined on the concrete driver class.
+    """
+    target = client
+    for key in driver_path:
+        children = getattr(target, "children", {})
+        if key not in children:
+            raise KeyError(f"Driver path component '{key}' not found. Available: {list(children.keys())}")
```
raballew: [LOW] When a path component is not found, the error message enumerates all sibling driver names at that level via `f"Driver path component '{key}' not found. Available: {list(children.keys())}"`
. This allows a caller to enumerate the full child-driver namespace by probing with intentionally wrong path components, without needing `list_drivers`. Remove the `Available: ...` clause so the error reads only `f"Driver path component '{key}' not found"`. Callers needing available names should call `list_drivers` explicitly.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+logger = logging.getLogger(__name__)
+
+
+def _json_safe(value: Any) -> Any:
+    """Coerce a value to a JSON-serializable equivalent, stringifying as a last resort."""
+    if value is None or isinstance(value, (bool, int, float, str)):
+        return value
+    if isinstance(value, (list, tuple)):
+        return [_json_safe(v) for v in value]
+    return str(value)
```
raballew: [LOW] The `str()` fallback in `_json_safe` converts non-serializable parameter defaults to strings. Click parameters can carry connection strings, credential helpers, or factory callables whose `__str__` or `__repr__` may include sensitive data. This stringified value ends up in tool responses. Replace the `str()` fallback with a fixed placeholder like `"<non-serializable>"` to avoid unintended data leakage.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+    this one — introspection therefore happens outside the loop.
+
+    :raises ValueError: if lease_name is empty
+    :raises jumpstarter.client.exceptions.LeaseError: if the lease has ended
+        or cannot be acquired
+    :raises jumpstarter.common.exceptions.ConnectionError: if the lease does
+        not exist or the controller is unreachable
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
```
raballew: [LOW] `lease_name` is only checked for non-emptiness before being passed unchanged to the Kubernetes API. Characters illegal in Kubernetes resource names are rejected server-side, but the resulting error from the controller may expose internal details. Add client-side validation against the Kubernetes name pattern (`^[a-z0-9][a-z0-9\-]{0,251}[a-z0-9]$`) before making any network call.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+
+# Methods inherited from base classes that add noise to driver listings
+_INTERNAL_METHODS = {
+    "check_exporter_status",
+    "resource_async",
+    "wait_for_hook_complete_monitored",
+    "wait_for_hook_status",
+    "wait_for_lease_ready",
+    "wait_for_lease_ready_monitored",
+}
```
raballew: [LOW] Method filtering relies on static deny-lists (`_BASE_METHODS`, `_INTERNAL_METHODS`) of known internal method names. If a sensitive method is added to `DriverClient` or a base class in the future without updating these sets, it will immediately appear in introspection responses. An allow-list approach, or a dedicated decorator/marker for driver-public methods, would make the set of introspectable methods opt-in rather than opt-out.

python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
```diff
@@ -112,135 +105,20 @@ class FakeLease:
...
-        client = FakeCompositeClient()
-        result = get_driver_methods(client, ["power"])
-        cycle = next(m for m in result["methods"] if m["name"] == "cycle")
-        assert len(cycle["parameters"]) == 1
-        assert cycle["parameters"][0]["name"] == "wait"
-        assert cycle["parameters"][0]["default"] == "2"
+        assert shim.walk_click_tree is core.walk_click_tree
+        assert shim.list_drivers is core.list_drivers
+        assert shim.get_driver_methods is core.get_driver_methods
+        assert shim._get_public_method_names is core._get_public_method_names
```
raballew: [LOW] The shim re-export test verifies identity for `walk_click_tree`, `list_drivers`, `get_driver_methods`, and `_get_public_method_names`, but skips `_inspect_method`, `_BASE_METHODS`, and `_INTERNAL_METHODS`, which are also exported from `jumpstarter_mcp.introspect.__all__`. If any of these were accidentally omitted or re-defined locally, the test would not catch it. Add identity assertions: `assert shim._inspect_method is core._inspect_method`, `assert shim._BASE_METHODS is core._BASE_METHODS`, and `assert shim._INTERNAL_METHODS is core._INTERNAL_METHODS`.

python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
kirkbrauer: Addressed in 5d9e7920: both drivers() and driver_methods() run their synchronous introspection through anyio.to_thread.run_sync(). Added regression coverage in 1dd009a5 using a live BlockingPortal: driver attributes and CLI construction dispatch through that portal, so running them on the event-loop thread would fail. The drivers, driver_methods, and explore cases all pass.

python/packages/jumpstarter/jumpstarter/client/introspect.py
kirkbrauer: Addressed in 5d9e7920 by catching CLI-construction failures and retaining the driver listing with cli_tree=None. Kept inherited CLIs supported: QemuFlasherClient inherits its working CLI from FlasherClientInterface, so a concrete-class __dict__ check would incorrectly hide it. Tests cover both an inherited CLI and a broken CLI; the current core suite passes.

python/packages/jumpstarter/jumpstarter/client/introspect.py
kirkbrauer: Removed the misleading 30-minute value in 5d9e7920. lease_async currently requires the duration argument, so this now passes timedelta(0) with a comment explaining that attach-by-name never sends a duration to the controller and cannot extend the lease. Omitting the argument would require a separate signature change. Added an assertion in 1dd009a5 covering the neutral duration alongside the named-lease/no-selector arguments.

python/packages/jumpstarter/jumpstarter/client/introspect.py
kirkbrauer: Addressed in 5d9e7920: describe_client, describe_devices, and describe_devices_async are imported and included in jumpstarter.client.__all__. Added regression coverage in 1dd009a5 checking that each package-level export is the same public helper from the introspection submodule. Follow-on #1070 adds the driver-named helpers while retaining these names as compatibility aliases.

kirkbrauer:
Verified the fixes already present in 5d9e7920 and replied to all four inline review threads. Pushed 1dd009a5 with regression coverage for off-loop MCP discovery using a live BlockingPortal, public package exports, and the neutral duration argument when attaching to a named lease.

Validation:
- `make pkg-test-jumpstarter-mcp`: 36 passed
- `make pkg-test-jumpstarter`: 829 passed (20 warnings)
- `make lint-fix`: passed
- `make pkg-ty-jumpstarter-mcp`: passed

On the dependency question: the current server imports `FastMCP` from `mcp.server.fastmcp` and also uses `mcp.server.stdio`. Those come from the declared `mcp` SDK dependency. The standalone `fastmcp` distribution is a separate API/migration, so this PR leaves the dependency unchanged rather than replacing a package the server still imports.

Also updated follow-on #1070 to include the latest version of this branch and preserve these fixes under the new `--drivers` / `driver_tree` terminology. Review threads are left open for confirmation; re-review requested.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,329 @@
...
+
+    Returns command names, help text, parameters (with types and defaults),
+    and nested subcommands.
+    """
+    path = path or []
+    result: dict[str, Any] = {
+        "name": cmd.name,
+        "help": cmd.help,
+        "params": [
+            _describe_param(p) for p in cmd.params if not getattr(p, "hidden", False) and p.name != "help"
```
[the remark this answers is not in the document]
kirkbrauer: Fixed in 080009eb. The filter now excludes only hidden parameters. Regression tests preserve both a required positional argument and an explicit option named help, and verify that Click’s automatically generated help option is absent from the described parameters. Core tests: 833 passed; MCP: 36 passed; lint passed. Also carried into #1070.

python/packages/jumpstarter/jumpstarter/client/introspect.py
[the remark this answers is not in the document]
kirkbrauer: Fixed in 080009eb by building the receiver as client for an empty driver path. Added a root-driver regression asserting client.cycle(wait=...) and parsing the generated example with ast.parse; the nested-driver example is also parsed to retain coverage of client.power.cycle(...). Core/MCP suites and lint pass. Also carried into #1070.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,333 @@
...
+
+
+def _json_safe(value: Any) -> Any:
+    """Coerce a value to a JSON-serializable equivalent, stringifying as a last resort."""
+    if value is None or isinstance(value, (bool, int, float, str)):
+        return value
+    if isinstance(value, (list, tuple)):
+        return [_json_safe(v) for v in value]
+    if isinstance(value, Mapping):
+        return {str(key): _json_safe(item) for key, item in value.items()}
```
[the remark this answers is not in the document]
kirkbrauer: Fixed in cf391f2f and carried into #1070. _json_safe now raises a clear ValueError on colliding stringified keys instead of silently overwriting a value. Regression tests cover both insertion orders for 1 / "1" and a collision inside a nested read-only mapping. Normal mappings retain their existing structured representation. Validation: core 838 / MCP 36 passed on this branch, core 841 / CLI 265 passed on #1070; lint passed.

AI generated, human reviewed

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+        "help": cmd.help,
+        "params": [
+            _describe_param(p) for p in cmd.params if not getattr(p, "hidden", False)
+        ],
+    }
+    if isinstance(cmd, click.Group):
+        result["subcommands"] = {
+            name: walk_click_tree(sub, path + [name])
+            for name, sub in cmd.commands.items()
+        }
```
raballew: [CRITICAL] `ruff format --check` exits 1 for this file. Four hunks need reformatting: the multi-line list/dict comprehensions at lines 80-88 that ruff wants inlined, the parenthesised string concat at 211-214 that ruff wants as a single f-string, and the multi-line `client_from_path(...)` call at 289-291. `introspect_test.py` also fails format-check (the `@pytest.mark.parametrize` decorators at 131-134 and 412-416, plus the `mapping_type({...})` calls at 400-403). Run `ruff format python/packages/jumpstarter/jumpstarter/client/introspect.py python/packages/jumpstarter/jumpstarter/client/introspect_test.py` and commit the result. CI `lint-python` runs `ruff format --check` and will fail until this is done.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+    described: dict[str, Any] = {
+        "name": param.name,
+        "kind": param.param_type_name,
+        # Click's own type name ("integer", "path", "choice"); str() on some
+        # types renders an object repr, which is no use in a prompt.
+        "type": getattr(param.type, "name", None) or str(param.type),
+        "help": getattr(param, "help", None),
+        "required": getattr(param, "required", False),
+        "default": _json_safe(param.default) if param.default is not None else None,
+        "opts": list(param.opts),
```
raballew: [HIGH] The parameter descriptor dict uses wrong field names: `'type'` at line 56 should be `'type_name'`, and `'opts'` at line 60 should be `'option_decls'`. Any caller using the documented names gets a `KeyError` on both keys. Additionally, for `click.Argument` the value should be an empty list rather than `param.opts`. Rename `'type'` to `'type_name'` at line 56, rename `'opts'` to `'option_decls'` at line 60, and use `list(param.opts) if isinstance(param, click.Option) else []` as the value. Update all test assertions accordingly.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+        sig = inspect.signature(method)
+    except (ValueError, TypeError, AttributeError):
+        return None
+
+    is_streaming = False
+    try:
+        source = inspect.getsource(method)
+        is_streaming = "streamingcall" in source
+    except (OSError, TypeError):
+        pass
```
raballew: [HIGH] The `is_streaming=True` branch at lines 182-187 is permanently dead in the test suite. No test method has `streamingcall` in its source, so `is_streaming` is always `False` and the true branch is never exercised. The `except (OSError, TypeError)` source-unavailable fallback is also uncovered. Add a method to `FakePowerClient` whose source contains `self.streamingcall(...)` and assert `result['is_streaming'] is True` for it. Add a second test patching `inspect.getsource` to raise `OSError` and verify the fallback returns `is_streaming=False` with the method still present in the result.

python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/connections.py
```diff
@@ -8,8 +8,8 @@
 import anyio.to_thread  # noqa: F401
 
 from jumpstarter_mcp.connections import ConnectionManager
-from jumpstarter_mcp.introspect import list_drivers, walk_click_tree
 
+from jumpstarter.client.introspect import list_drivers, walk_click_tree
```
raballew: [HIGH] `list_drivers(conn.client)` at line 55 is called directly on the event-loop thread. The `anyio.to_thread` import at line 8 is dead (the `# noqa: F401` comment confirms it is never wired up). If a driver property ever dispatches through a blocking portal, calling from the event-loop thread will deadlock. `commands.py` already handles this correctly with `await anyio.to_thread.run_sync(list_drivers, conn.client)`. Apply the same pattern to the `list_drivers` call in `connect()` and drop the `# noqa`.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+        "help": getattr(param, "help", None),
+        "required": getattr(param, "required", False),
+        "default": _json_safe(param.default) if param.default is not None else None,
+        "opts": list(param.opts),
+        "is_flag": bool(getattr(param, "is_flag", False)),
+        "multiple": bool(getattr(param, "multiple", False)),
+        "nargs": param.nargs,
+    }
+    if isinstance(param.type, click.Choice):
+        described["choices"] = [str(choice) for choice in param.type.choices]
```
raballew: [MEDIUM] The `choices` key is only added for `click.Choice` parameters (lines 64-66). For every other type the key is absent, so `param['choices']` raises `KeyError` at runtime. The field should always be present. Initialise `'choices': None` in the base descriptor dict and overwrite it inside the `isinstance(param.type, click.Choice)` branch. Add an assertion in `test_describes_arguments_and_options` that `choices` is `None` for a non-Choice option.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+        return [_json_safe(v) for v in value]
+    if isinstance(value, Mapping):
+        result: dict[str, Any] = {}
+        for key, item in value.items():
+            json_key = str(key)
+            if json_key in result:
+                raise ValueError(f"Mapping keys collide after stringification: {json_key!r}")
+            result[json_key] = _json_safe(item)
+        return result
+    return str(value)
```
raballew: [MEDIUM] `_json_safe` recurses without a depth limit or cycle guard (lines 26-40). A self-referential Mapping default (e.g. `d = {}; d['self'] = d`) causes `RecursionError` that propagates unhandled through the MCP `explore` handler. Add a `_depth: int = 0` parameter and raise `ValueError` when it exceeds a safe threshold (64 is a reasonable choice), threading `_depth + 1` through recursive calls. Alternatively, track `id(value)` to detect cycles before recursing.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+            continue
+        if isinstance(value, property):
+            continue
+        if not callable(value):
+            continue
+        try:
+            inspect.signature(value)
+        except (ValueError, TypeError, AttributeError):
+            continue
+        names.append(name)
```
raballew: [MEDIUM] Two error branches in `_get_public_method_names` are untested (lines 131-147): the `except Exception` fallback that returns an empty list when `inspect.getmembers_static` fails, and the per-member `except (ValueError, TypeError, AttributeError)` that silently skips uninspectable callables. The `getmembers_static` failure is logged at DEBUG only, so it is invisible at the default log level. Raise it to WARNING. Add `test_returns_empty_list_when_static_inspection_fails` (patch `inspect.getmembers_static` to raise, assert `[]` returned) and `test_skips_members_whose_signature_cannot_be_resolved` (add an uninspectable callable, assert it is absent while valid methods are present).

python/packages/jumpstarter/jumpstarter/client/introspect.py
raballew: [MEDIUM] Neither the `is_streaming=False` outcome nor the `except (OSError, TypeError)` source-unavailable fallback at lines 182-187 is asserted by any test. All test methods have accessible source and none contains `streamingcall`, so `is_streaming` silently defaults to `False` in every case. Add `test_non_streaming_method_is_not_flagged` asserting `is_streaming=False` for a plain method. Add a second test patching `inspect.getsource` to raise `OSError` and asserting `is_streaming=False` while the method still appears in the output.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+        results.extend(list_drivers(child, keys + [name]))
+    return results
+
+
+def _inspect_method(name: str, method: Any, driver_path: list[str]) -> dict[str, Any] | None:
+    """Build a method descriptor dict, or return None if the method can't be inspected."""
+    try:
+        sig = inspect.signature(method)
+    except (ValueError, TypeError, AttributeError):
+        return None
```
raballew: [MEDIUM] `_inspect_method` returns `None` when `inspect.signature` raises (lines 177-180), and the `if info is not None` guard at line 244 filters those out silently. No test places an uninspectable callable on a fake driver and verifies it is absent from the result while other methods remain. The guard could be removed without any test catching it. Add `test_omits_methods_whose_signature_cannot_be_resolved`: attach a callable for which `inspect.signature` raises to a fake driver, call `get_driver_methods`, and assert the uninspectable method is absent while inspectable methods are still returned.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+        if key not in children:
+            raise KeyError(f"Driver path component '{key}' not found. Available: {list(children.keys())}")
+        target = children[key]
+
+    cls = type(target)
+    try:
+        members = inspect.getmembers_static(target)
+    except Exception:
+        logger.debug("inspect.getmembers_static failed for %s", cls.__name__, exc_info=True)
+        members = []
```
raballew: [MEDIUM] When `inspect.getmembers_static` raises inside `get_driver_methods` (lines 233-237), the exception is caught and logged at DEBUG only, returning `{'methods': []}`. This is indistinguishable from a driver that genuinely has no methods, and the failure is completely invisible at the default production log level. Raise it to WARNING. Add `test_returns_empty_methods_when_member_inspection_raises` that patches `inspect.getmembers_static` to raise and asserts `result['methods'] == []` while any other returned fields are still present.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+    """Inspect a specific driver client at the given path in the children tree.
+
+    Uses Python inspect to return detailed method information for all public
+    methods defined on the concrete driver class.
+    """
+    target = client
+    for key in driver_path:
+        children = getattr(target, "children", {})
+        if key not in children:
+            raise KeyError(f"Driver path component '{key}' not found. Available: {list(children.keys())}")
```
raballew: [LOW] The error message at line 229 includes `list(children.keys())`, which lets an authenticated caller enumerate the full driver component tree by submitting invalid path strings without calling any listing endpoint first. Remove the available-keys list from the message. `raise KeyError(f"Driver path component {key!r} not found")` is sufficient to communicate the problem without exposing sibling names.

python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
raballew: [LOW] The test asserts identity of `_get_public_method_names` between the shim and core modules (lines 112-121). This is a private helper not required by the public re-export contract. Renaming the private function while keeping all public names stable would break this test without any real backward-compatibility regression. Remove the `_get_public_method_names` identity assertion from this test. If coverage of the private helper is important, move it to a dedicated focused test with a comment explaining why.

python/packages/jumpstarter/jumpstarter/client/introspect.py
raballew: [HIGH] `_describe_param` uses `"type"` and `"opts"` as dict keys, but callers expecting `"type_name"` and `"option_names"` get a `KeyError`. The tests in `introspect_test.py` (e.g. lines 444, 446) also use the wrong keys, so this mismatch is not caught by the suite. Additionally, for `click.Argument`, `param.opts` returns the declaration name (e.g. `["PORT"]`) rather than `[]`. To fix: rename `"type"` to `"type_name"` and `"opts"` to `"option_names"` in `_describe_param`, emit `list(param.opts)` for `click.Option` and `[]` for `click.Argument`, and update all test assertions accordingly.

python/packages/jumpstarter/jumpstarter/client/introspect.py
raballew: [MEDIUM] The `choices` key is entirely absent from the parameter dict for non-Choice parameters rather than being set to `null`. Any caller that does `param_desc["choices"]` on a non-Choice parameter gets a `KeyError`. The fix is to unconditionally initialize `described["choices"] = None` in the base dict, then overwrite it inside the `isinstance(param.type, click.Choice)` branch. A test assertion like `assert params["port"]["choices"] is None` should also be added.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+            "path": f"client.{'.'.join(keys)}" if keys else "client",
+            "driver_path": keys,
+            "class": f"{cls.__module__}.{cls.__qualname__}",
+            "description": getattr(client, "description", None),
+            "methods": _get_public_method_names(client),
+        }
+    ]
+    children = getattr(client, "children", {})
+    for name, child in children.items():
+        results.extend(list_drivers(child, keys + [name]))
```
raballew: [MEDIUM] `list_drivers` recurses into `client.children` with no visited-object guard and no depth limit. A driver that creates a cycle (`client.children["self"] = client`) causes a `RecursionError` that can crash the MCP server process and drop all active connections. Consider threading a `visited: frozenset[int]` (keyed by `id(client)`) through the recursion, skipping any node already seen, or adding a `max_depth` parameter defaulting to 50.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+
+    Uses getmembers_static to avoid triggering property descriptors, which can
+    make gRPC calls that fail when invoked from the event loop thread.
+    """
+    names = []
+    try:
+        members = inspect.getmembers_static(obj)
+    except Exception:
+        logger.debug("inspect.getmembers_static failed for %s", type(obj).__name__, exc_info=True)
+        return names
```
raballew: [MEDIUM] When `inspect.getmembers_static` raises, `_get_public_method_names` returns `[]` and logs at DEBUG. Callers receive a `"methods": []` entry that is indistinguishable from a driver that genuinely has no public methods, and production failures are invisible unless debug logging is enabled. The same silent handler exists in `get_driver_methods` (line 235). Consider adding an `"introspection_error"` field to the driver dict when the exception path is taken, elevating the log to WARNING, and adding a test with an object whose `__dir__` raises `RuntimeError`.

python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/connections.py
```diff
@@ -8,8 +8,8 @@
 import anyio.to_thread  # noqa: F401
```
raballew: [MEDIUM] `import anyio.to_thread` at line 8 is suppressed with `# noqa: F401` but is genuinely unused at module scope. The only `anyio` usage inside `connect()` comes from a separate `import anyio` at line 48. Removing the noqa comment and running ruff confirms the F401 violation. Remove `import anyio.to_thread # noqa: F401` and move the `import anyio` inside `connect()` to module level to keep imports visible and PEP 8 compliant.

python/packages/jumpstarter/jumpstarter/client/introspect_test.py
```diff
@@ -0,0 +1,464 @@
...
```
raballew: [MEDIUM] `test_describes_arguments_and_options` checks `multiple` but has no assertion on `nargs`, even though `nargs` is part of the parameter metadata that `_describe_param` emits. Removing `"nargs": param.nargs` from `_describe_param` would leave the test suite green, meaning the field has no coverage. Add assertions like `assert params["port"]["nargs"] == 1` and `assert params["tag"]["nargs"] == 1` to close that gap.

python/packages/jumpstarter/jumpstarter/client/introspect.py
```diff
@@ -0,0 +1,339 @@
...
+async def describe_devices_async(config: ClientConfigV1Alpha1, lease_name: str) -> dict[str, Any]:
+    """Async wrapper around describe_devices, run in a worker thread.
+
+    The driver clients cannot be driven from an event loop thread, so the
+    whole attach-and-introspect flow runs off-loop.
+    """
+    if not lease_name:
+        raise ValueError("lease_name must be a non-empty existing lease name")
+
+    return await to_thread.run_sync(describe_devices, config, lease_name)
```
raballew: [MEDIUM] `describe_devices_async` delegates to `anyio.to_thread.run_sync` with the default `cancellable=False`. When the outer coroutine is cancelled (e.g. by `fail_after` or an HTTP timeout), the worker thread is not interrupted. Inside that thread, `describe_devices` holds a `start_blocking_portal()` context and a live gRPC lease connection via `_connect_lease`. Because the default `acquisition_timeout` is 7200 seconds, a slow lease can keep the worker thread and its resources open for up to two hours after the caller has given up. Consider passing a shorter `acquisition_timeout` when constructing the Lease inside `_connect_lease`, or documenting clearly that this function does not support prompt cancellation.

python/packages/jumpstarter/jumpstarter/client/introspect.py
raballew: [LOW] The `KeyError` raised in `get_driver_methods` includes the full list of sibling child driver names in the message: `Available: {list(children.keys())}`. An authenticated caller who does not know the driver tree can enumerate child names by submitting unknown paths one level at a time. Consider dropping the `Available:` list from the public error and logging it at DEBUG instead: `raise KeyError(f"Driver path component '{key}' not found")`.

python/packages/jumpstarter-mcp/jumpstarter_mcp/server_test.py
raballew: [LOW] `TestIntrospectReexports.test_reexports_core_implementation` checks only 4 of the 7 names declared in `jumpstarter_mcp/introspect.__all__`: `walk_click_tree`, `list_drivers`, `get_driver_methods`, and `_get_public_method_names`. The names `_BASE_METHODS`, `_INTERNAL_METHODS`, and `_inspect_method` are not verified. An accidental re-definition of any of those three in the shim would go undetected. Extend the test to cover all names in `__all__`, for example by iterating: `for name in jumpstarter_mcp.introspect.__all__: assert getattr(shim, name) is getattr(core, name)`.

python/packages/jumpstarter-mcp/jumpstarter_mcp/tools/commands.py
```diff
@@ -12,7 +12,8 @@
...
 
 logger = logging.getLogger(__name__)
```
raballew: [LOW] There is an `import os` statement deferred inside the `run_command` function body (around line 43 in this file). `os` is a zero-cost standard library module, so there is no benefit to the deferred placement. It obscures the module's dependencies and goes against PEP 8 convention that imports belong at the top of the file. Move `import os` into the module-level imports section here alongside the other stdlib imports.

raballew (approved):
some merge conflicts, but it looks okay to me

mangelajo:
yup, some conflicts from the streaming flash client changes

mangelajo (approved):
approval waiting on rebase, then feel free to merge.

PR #1043, opened 2026-08-30
feat(cli): add jmp describe for exporters, leases, and clients

kirkbrauer:
`jmp get` answers what exists; there was no way to ask about one of them in detail short of reading the raw object.

Adds `jmp describe exporter|lease|client <name>`, following `kubectl describe`: a readable summary by default, `-o json|yaml` for a machine. Describing an exporter reports the lease holding it, describing a lease reports its conditions, and describing a client reads the local config — so it works without a connection and says whether the token is still valid. The client description never emits token material.

Documented in the distributed mode guide, where a user first needs to know which exporters exist and what to select on, and from the lease troubleshooting entry in the testing guide. The man page picks the commands up from click.

### DIFF
--- a/docs/source/getting-started/guides/examples/testing.md
+++ b/docs/source/getting-started/guides/examples/testing.md
@@ -277,7 +277,10 @@ hardware-test:
 **Lease acquisition times out**
 : Verify that an {term}`exporter` matching your `selector` labels is running and
   registered with the {term}`controller`. Check available {term}`exporter`s with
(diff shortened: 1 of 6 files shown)


mangelajo (commented):
nice. can we add it to documentation? :)

kirkbrauer:
@mangelajo Added some docs and fixed a few additional issues I noticed from review.

bennyz (approved):

PR #1045, opened 2026-08-30
fix(admin): handle resources the controller has not reconciled yet

kirkbrauer:
A Client or Exporter exists the moment it is created, but its credentials and endpoint arrive later — or never, in a namespace nothing is watching. Reading one in that state failed with a traceback rather than an explanation:

```
jmp admin import client fresh   -> AttributeError: 'NoneType' object has no attribute 'credential'
jmp admin get exporter          -> KeyError: 'status'
```

Makes the exporter's status optional the way the client's already is, so a status-less exporter lists and renders, and raises `CredentialNotReadyError` from `get_client_config` / `get_exporter_config`, which `jmp admin import` reports as a message naming the likely cause.

### DIFF
--- a/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/import_res.py
+++ b/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/import_res.py
@@ -13,6 +13,7 @@ from jumpstarter_cli_common.opt import (
     opt_output_path_only,
 )
 from jumpstarter_kubernetes import ClientsV1Alpha1Api, ExportersV1Alpha1Api
+from jumpstarter_kubernetes.exceptions import JumpstarterKubernetesError
 from kubernetes_asyncio.client.exceptions import ApiException
 from kubernetes_asyncio.config.config_exception import ConfigException
 
@@ -92,6 +93,8 @@ async def import_client(
                 click.echo(f"Client configuration successfully saved to {config_path}")
             else:
                 click.echo(config_path)
+    except JumpstarterKubernetesError as e:
+        raise click.ClickException(str(e)) from e
     except ApiException as e:
         handle_k8s_api_exception(e)
     except ConfigException as e:
@@ -141,6 +144,8 @@ async def import_exporter(
                 click.echo(f"Exporter configuration successfully saved to {config_path}")
             else:
                 click.echo(config_path)
+    except JumpstarterKubernetesError as e:
+        raise click.ClickException(str(e)) from e
     except ApiException as e:
         handle_k8s_api_exception(e)
     except ConfigException as e:
--- a/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/clients_test.py
+++ b/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/clients_test.py
@@ -7,6 +7,7 @@ from kubernetes_asyncio.client.models import V1ConfigMap, V1ObjectMeta, V1Secret
 
 from jumpstarter_kubernetes import V1Alpha1Client, V1Alpha1ClientStatus
 from jumpstarter_kubernetes.clients import ClientsV1Alpha1Api
+from jumpstarter_kubernetes.exceptions import CredentialNotReadyError
 
 TEST_CLIENT = V1Alpha1Client(
     api_version="jumpstarter.dev/v1alpha1",
@@ -424,3 +425,30 @@ def test_client_from_dict_keeps_labels():
     )
     assert client.metadata.labels == {"team": "platform"}
     assert '"team": "platform"' in client.dump_json()
+
+
+@pytest.mark.asyncio
+async def test_get_client_config_without_credentials():
+    """A client whose credentials the controller has not issued yet is reported, not crashed on"""
+    api = ClientsV1Alpha1Api(namespace="test-namespace")
+    api.api = AsyncMock()
+    api.core_api = AsyncMock()
+    api.api.get_namespaced_custom_object = AsyncMock(
+        return_value={
+            "apiVersion": "jumpstarter.dev/v1alpha1",
+            "kind": "Client",
+            "metadata": {
+                "creationTimestamp": "2021-10-01T00:00:00Z",
+                "generation": 1,
+                "name": "fresh-client",
+                "namespace": "test-namespace",
+                "resourceVersion": "1",
+                "uid": "test-uid",
+            },
+        }
+    )
+
+    with pytest.raises(CredentialNotReadyError, match="fresh-client"):
+        await api.get_client_config("fresh-client", allow=[], unsafe=False)
+
+    api.core_api.read_namespaced_secret.assert_not_awaited()
--- a/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/exceptions.py
+++ b/python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/exceptions.py
@@ -11,6 +11,24 @@ class JumpstarterKubernetesError(Exception):
     pass
 
 
+class CredentialNotReadyError(JumpstarterKubernetesError):
+    """Raised when a client or exporter exists but has no credentials yet.
+
+    The controller issues credentials asynchronously, so a resource created a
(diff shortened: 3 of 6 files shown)


bennyz (approved):

kirkbrauer:
@bennyz I added a few new tweaks if you could review :)

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/import_res.py
```diff
@@ -141,6 +144,8 @@ async def import_exporter(
                 click.echo(f"Exporter configuration successfully saved to {config_path}")
             else:
                 click.echo(config_path)
+    except JumpstarterKubernetesError as e:
+        raise click.ClickException(str(e)) from e
```
raballew: [HIGH] The new `except JumpstarterKubernetesError` handler in `import_exporter` (lines 147-148) and the equivalent in `import_client` (lines 96-97) have no test coverage. `import_res_test.py` was not modified in this PR and contains no assertions about this error path. A regression that miswires either handler would produce an unhandled traceback instead of a clean CLI message with no automated test catching it. Consider adding `test_import_client_reports_credential_not_ready` (patch `get_client_config` to raise `CredentialNotReadyError`, invoke via `CliRunner`, assert `result.exit_code != 0` and the resource name in `result.output`) and a matching `test_import_exporter_reports_credential_not_ready`.

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/import_res.py
```diff
@@ -92,6 +93,8 @@ async def import_client(
                 click.echo(f"Client configuration successfully saved to {config_path}")
             else:
                 click.echo(config_path)
+    except JumpstarterKubernetesError as e:
+        raise click.ClickException(str(e)) from e
```
raballew, line 97: The except is broader than necessary. It covers the entire `JumpstarterKubernetesError` hierarchy rather than just `CredentialNotReadyError`. Not an error today but maybe something we could design more defensively.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/exceptions.py
```diff
@@ -11,6 +11,24 @@ class JumpstarterKubernetesError(Exception):
...
+    """
+
+    def __init__(self, kind: str, name: str):
+        self.kind = kind
+        self.name = name
+        super().__init__(
+            f"The {kind} '{name}' has no credentials yet. "
+            "The Jumpstarter controller issues them shortly after the resource is created; "
+            "check that it is running and watching this namespace, then try again."
+        )
```
raballew: [MEDIUM] `CredentialNotReadyError.__init__` stores `self.kind` and `self.name` as public attributes, but no test accesses them directly. All existing tests use `pytest.raises(CredentialNotReadyError, match="...")`, which only checks the string representation. A typo such as `self.knd = kind` would go undetected. Consider adding a unit test that constructs `CredentialNotReadyError("client", "my-client")` and asserts `err.kind == "client"`, `err.name == "my-client"`, and `"my-client" in str(err)`.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/clients_test.py
```diff
@@ -424,3 +425,30 @@ def test_client_from_dict_keeps_labels():
...
+                "generation": 1,
+                "name": "fresh-client",
+                "namespace": "test-namespace",
+                "resourceVersion": "1",
+                "uid": "test-uid",
+            },
+        }
+    )
+
+    with pytest.raises(CredentialNotReadyError, match="fresh-client"):
```
raballew: [MEDIUM] This `pytest.raises` call uses `match="fresh-client"`, which checks only that the resource name appears in the exception message. A refactor that strips the guidance text (e.g. instructions to verify the controller is running) would not be caught. Consider extending the `match` pattern to cover the guidance phrase, e.g. `match=r"fresh-client.*controller.*running"`, or adding a `str(exc_info.value)` assertion for the guidance text.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/clients_test.py
```diff
@@ -424,3 +425,30 @@ def test_client_from_dict_keeps_labels():
...
+        await api.get_client_config("fresh-client", allow=[], unsafe=False)
+
+    api.core_api.read_namespaced_secret.assert_not_awaited()
```
raballew: [MEDIUM] This test mocks a response with no `"status"` key at all, exercising only the `client.status is None` branch of the guard at `clients.py:152`. The partially-reconciled case (status present but credential absent -- the other branch, `status.credential is None`) has no test. Consider adding an async test that returns a payload with a `status` field but no `credential` sub-key and asserts that `CredentialNotReadyError` is raised and `read_namespaced_secret` is not awaited.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/exporters_test.py
```diff
@@ -384,3 +385,93 @@ def test_exporter_from_dict_keeps_labels():
...
+                "resourceVersion": "1",
+                "uid": "test-uid",
+            },
+        }
+    )
+
+    with pytest.raises(CredentialNotReadyError, match="fresh-exporter"):
+        await api.get_exporter_config("fresh-exporter")
+
+    api.core_api.read_namespaced_secret.assert_not_awaited()
```
raballew: [MEDIUM] This test mocks a response with no `"status"` key, exercising only the `exporter.status is None` branch of the guard at `exporters.py:209`. The partially-reconciled case (status present, credential absent -- the `status.credential is None` branch) is untested. Consider adding an async test that returns a payload with a `status` field but no `credential` key and asserts `CredentialNotReadyError` is raised without calling `read_namespaced_secret`.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/exporters.py
```diff
@@ -57,13 +60,16 @@ def from_dict(dict: dict):
...
-                endpoint=dict["status"]["endpoint"],
+                endpoint=dict["status"].get("endpoint", ""),
                 devices=[V1Alpha1ExporterDevice(labels=d["labels"], uuid=d["uuid"]) for d in dict["status"]["devices"]]
                 if "devices" in dict["status"]
                 else [],
                 exporter_status=dict["status"].get("exporterStatus"),
                 status_message=dict["status"].get("statusMessage"),
-            ),
+            )
+            # An exporter the controller has not reconciled yet has no status.
```
raballew: [LOW] This inline comment restates context already provided by the class-level comment a few lines above. The conditional `if "status" in dict else None` is self-explanatory without it. Consider removing this comment.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/exporters.py
```diff
@@ -188,6 +206,8 @@ async def create_exporter(
     async def get_exporter_config(self, name: str) -> ExporterConfigV1Alpha1:
         """Get an exporter config for a specified exporter name"""
         exporter = await self.get_exporter(name)
+        if exporter.status is None or exporter.status.credential is None:
+            raise CredentialNotReadyError("exporter", name)
```
raballew: [LOW] `V1Alpha1ExporterStatus.endpoint` now defaults to `""`. This guard checks `status is None or status.credential is None`, but not whether the endpoint is set. If the controller writes the credential secret before setting the endpoint, the guard passes, the secret is fetched, and a config file with `endpoint=""` is saved to disk -- valid for authentication but unusable for connecting, with no warning to the user. Consider extending the guard: `if exporter.status is None or exporter.status.credential is None or not exporter.status.endpoint: raise CredentialNotReadyError("exporter", name)` (and the equivalent for the client in `clients.py`).

raballew (commented):
I could only find a nit. So looks good to me but I let @bennyz if this is merge-worthy.

bennyz (approved):

PR #1046, opened 2026-08-30
feat(admin): add jmp admin apply for Jumpstarter manifests

kirkbrauer:
`jmp admin create` builds a Client or an Exporter from flags, which stops where a resource has more shape than flags can carry: an ExporterSet's selector, template and replica bounds have to be written as YAML, and there was no way to send that YAML to the cluster without kubectl.

Adds `jmp admin apply -f <file|->`. Resources are sent as a server-side apply, so the same manifest creates a resource and later updates it, and each is reported the way kubectl does — created, configured or unchanged. Kind plurals and scope come from the cluster's discovery data rather than a table in this client, so kinds added later apply without another release. Only the jumpstarter.dev API groups are accepted; a manifest naming any other group is refused before anything reaches the cluster.

Conflicts are reported rather than won. An apply that would take a field another manager owns fails and names the field; `--force-conflicts` takes it deliberately, as in kubectl.

Also fixes `handle_k8s_api_exception`, which raised `KeyError: 'reason'` on any Status without one — which is what a schema or admission failure returns, so those surfaced as tracebacks instead of the server's message.

### DIFF
--- a/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply.py
+++ b/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply.py
@@ -1,3 +1,4 @@
+from http import HTTPStatus
 from typing import IO, Optional
 
 import click
@@ -31,6 +32,12 @@ from .k8s import (
     help="Manifest to apply, or '-' to read from stdin. Can be set multiple times.",
 )
 @click.option(
+    "--force-conflicts",
+    is_flag=True,
+    default=False,
+    help="Take ownership of fields another manager holds, instead of reporting the conflict.",
+)
+@click.option(
     "--dry-run",
     is_flag=True,
     default=False,
@@ -43,6 +50,7 @@ from .k8s import (
 @blocking
 async def apply(
     filenames: tuple[IO, ...],
+    force_conflicts: bool,
     dry_run: bool,
     namespace: str,
     kubeconfig: Optional[str],
@@ -64,11 +72,17 @@ async def apply(
 
     try:
         async with ApplyV1Alpha1Api(namespace, kubeconfig, context) as api:
-            applied = await api.apply_all(manifests, dry_run=dry_run)
+            applied = await api.apply_all(manifests, dry_run=dry_run, force_conflicts=force_conflicts)
     except ManifestError as e:
         raise click.ClickException(str(e)) from e
     except ApiException as e:
-        handle_k8s_api_exception(e)
+        try:
+            handle_k8s_api_exception(e)
+        except click.ClickException as error:
+            if e.status == HTTPStatus.CONFLICT and not force_conflicts:
+                # The conflict names the fields; say how to win them on purpose.
+                error.message += "\nRe-run with --force-conflicts to take ownership of those fields."
+            raise
     except ConfigException as e:
         handle_k8s_config_exception(e)
 
--- a/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply.py
+++ b/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply.py
@@ -1,0 +1,81 @@
+from typing import IO, Optional
+
+import click
+from jumpstarter_cli_common.blocking import blocking
+from jumpstarter_cli_common.opt import (
+    OutputType,
+    opt_context,
+    opt_kubeconfig,
+    opt_namespace,
+    opt_output_all,
+)
+from jumpstarter_cli_common.print import model_print
+from jumpstarter_kubernetes import ApplyV1Alpha1Api, ManifestError, load_manifests
+from kubernetes_asyncio.client.exceptions import ApiException
+from kubernetes_asyncio.config.config_exception import ConfigException
+
+from .k8s import (
+    handle_k8s_api_exception,
+    handle_k8s_config_exception,
+)
+
+
+@click.command("apply")
+@click.option(
+    "-f",
+    "--filename",
+    "filenames",
+    type=click.File("r"),
+    multiple=True,
+    required=True,
+    help="Manifest to apply, or '-' to read from stdin. Can be set multiple times.",
+)
+@click.option(
+    "--dry-run",
+    is_flag=True,
+    default=False,
+    help="Send the manifests to the server for validation without persisting them.",
+)
+@opt_namespace
+@opt_kubeconfig
+@opt_context
+@opt_output_all
+@blocking
+async def apply(
+    filenames: tuple[IO, ...],
+    dry_run: bool,
+    namespace: str,
+    kubeconfig: Optional[str],
+    context: Optional[str],
+    output: OutputType,
+):
+    """Apply Jumpstarter manifests to a Kubernetes cluster
+
+    Creates the resources in a manifest, or updates them in place when they
+    already exist. Only Jumpstarter resources (clients, exporters, exporter
+    sets, virtual target classes) can be applied.
+    """
+    try:
+        manifests = []
+        for file in filenames:
+            manifests.extend(load_manifests(file.read(), file.name))
+    except ManifestError as e:
+        raise click.ClickException(str(e)) from e
+
+    try:
+        async with ApplyV1Alpha1Api(namespace, kubeconfig, context) as api:
+            applied = await api.apply_all(manifests, dry_run=dry_run)
+    except ManifestError as e:
+        raise click.ClickException(str(e)) from e
+    except ApiException as e:
+        handle_k8s_api_exception(e)
+    except ConfigException as e:
+        handle_k8s_config_exception(e)
+
+    if output is None:
+        # Mirror kubectl: one line per resource saying what happened to it.
+        suffix = " (dry run)" if dry_run else ""
+        for resource in applied.items:
+            click.echo(f"{resource.qualified_name} {resource.action}{suffix}")
+    else:
+        model_print(applied, output, namespace=namespace)
--- a/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply_test.py
+++ b/python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply_test.py
@@ -1,0 +1,150 @@
+import json
+from unittest.mock import AsyncMock, patch
+
+from click.testing import CliRunner
+from jumpstarter_kubernetes import (
+    ApplyV1Alpha1Api,
+    V1Alpha1AppliedResource,
+    V1Alpha1AppliedResourceList,
+)
+
+from .apply import apply
+
+CLIENT_MANIFEST = """\
+apiVersion: jumpstarter.dev/v1alpha1
+kind: Client
+metadata:
+  name: hello
+"""
+
+EXPORTER_SET_MANIFEST = """\
+apiVersion: virtualtarget.jumpstarter.dev/v1alpha1
+kind: ExporterSet
+metadata:
+  name: pool
+"""
+
+
+def applied(kind: str, name: str, action: str, api_version: str = "jumpstarter.dev/v1alpha1"):
+    return V1Alpha1AppliedResource(
+        apiVersion=api_version,
+        kind=kind,
+        name=name,
+        namespace="default",
+        action=action,
+        resource={"apiVersion": api_version, "kind": kind, "metadata": {"name": name}},
+    )
+
+
+def run(args, files: dict[str, str], apply_all: AsyncMock, input: str | None = None):
+    runner = CliRunner()
+    with runner.isolated_filesystem():
+        for filename, contents in files.items():
+            with open(filename, "w") as f:
+                f.write(contents)
+        with (
+            patch.object(ApplyV1Alpha1Api, "_load_kube_config", AsyncMock()),
+            patch.object(ApplyV1Alpha1Api, "apply_all", apply_all),
+        ):
+            return runner.invoke(apply, args, input=input)
+
+
+def test_apply_reports_what_happened_to_each_resource():
+    apply_all = AsyncMock(
+        return_value=V1Alpha1AppliedResourceList(
+            items=[
+                applied("Client", "hello", "created"),
+                applied("ExporterSet", "pool", "configured", "virtualtarget.jumpstarter.dev/v1alpha1"),
+            ]
+        )
+    )
+
+    result = run(
+        ["-f", "client.yaml", "-f", "set.yaml"],
+        {"client.yaml": CLIENT_MANIFEST, "set.yaml": EXPORTER_SET_MANIFEST},
+        apply_all,
+    )
+
+    assert result.exit_code == 0
+    assert "client.jumpstarter.dev/hello created" in result.output
+    assert "exporterset.virtualtarget.jumpstarter.dev/pool configured" in result.output
+    # Both files reach the cluster in the order they were given.
+    assert [m["kind"] for m in apply_all.call_args.args[0]] == ["Client", "ExporterSet"]
+    assert apply_all.call_args.kwargs["dry_run"] is False
+
+
+def test_apply_reads_a_multi_document_manifest():
+    apply_all = AsyncMock(
+        return_value=V1Alpha1AppliedResourceList(
+            items=[
+                applied("Client", "hello", "created"),
+                applied("ExporterSet", "pool", "created", "virtualtarget.jumpstarter.dev/v1alpha1"),
+            ]
+        )
+    )
+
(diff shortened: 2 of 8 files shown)


python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+class V1Alpha1AppliedResourceList(V1Alpha1List[V1Alpha1AppliedResource]):
+    kind: Literal["List"] = Field(default="List")
+
+    @classmethod
+    def rich_add_columns(cls, table, **kwargs):
+        V1Alpha1AppliedResource.rich_add_columns(table, **kwargs)
+
+    def rich_add_rows(self, table, **kwargs):
+        for applied in self.items:
+            applied.rich_add_rows(table, **kwargs)
```
raballew: [HIGH] `rich_add_columns` and `rich_add_rows` on `V1Alpha1AppliedResource` and `V1Alpha1AppliedResourceList` have no test coverage. The CLI tests only exercise `-o name` and `-o json` paths, so the table rendering code is completely untested. Consider adding a unit test that constructs a `V1Alpha1AppliedResource`, creates a `rich.table.Table`, and calls both methods, asserting the expected column headers (NAME, KIND, NAMESPACE, ACTION) and correct cell values. The pattern used in `clients_test.py` and `exporters_test.py` is a good template.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply_test.py
```diff
@@ -0,0 +1,305 @@
...
+
+@pytest.mark.asyncio
+async def test_apply_all_looks_up_each_kind_once():
+    api = make_api()
+
+    applied = await api.apply_all(load_manifests(f"{CLIENT_MANIFEST}\n---\n{CLIENT_MANIFEST}"))
+
+    assert [item.action for item in applied.items] == ["created", "created"]
+    discoveries = [call for call in api._client.call_api.await_args_list if call.args[1] == "GET"]
+    assert len(discoveries) == 1
```
raballew: [HIGH] `apply_all` only has a single happy-path test. There is no coverage for what happens when `apply()` fails partway through the list (does the exception propagate?) or when the cluster does not serve a referenced kind (should raise `ManifestError`). Worth adding two tests: one where a manifest references an unknown kind, and one where `apply()` raises `ApiException` on the second item, to verify both failure modes behave as expected.

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply.py
```diff
@@ -0,0 +1,95 @@
...
+        for file in filenames:
+            manifests.extend(load_manifests(file.read(), file.name))
+    except ManifestError as e:
+        raise click.ClickException(str(e)) from e
+
+    try:
+        async with ApplyV1Alpha1Api(namespace, kubeconfig, context) as api:
+            applied = await api.apply_all(manifests, dry_run=dry_run, force_conflicts=force_conflicts)
+    except ManifestError as e:
+        raise click.ClickException(str(e)) from e
```
raballew: [HIGH] The `ManifestError` catch here converts a cluster error into a `ClickException`, but there is no test for this branch. None of the CLI tests set `apply_all.side_effect = ManifestError(...)`, so if this path regresses it would surface as an unhandled exception rather than a clean error message. Adding a test that sets `apply_all.side_effect = ManifestError("the cluster does not serve kind Foo")` and asserts `result.exit_code != 0` and the error text in `result.output` would close the gap.

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply.py
```diff
@@ -0,0 +1,95 @@
...
+    except ApiException as e:
+        try:
+            handle_k8s_api_exception(e)
+        except click.ClickException as error:
+            if e.status == HTTPStatus.CONFLICT and not force_conflicts:
+                # The conflict names the fields; say how to win them on purpose.
+                error.message += "\nRe-run with --force-conflicts to take ownership of those fields."
+            raise
+    except ConfigException as e:
+        handle_k8s_config_exception(e)
```
raballew: [HIGH] The `ConfigException` handler at these lines has zero test coverage. The test helper always patches `_load_kube_config` to succeed, so a missing or broken kubeconfig would currently surface as an unhandled traceback instead of a clean error message. A test that patches `_load_kube_config` to raise `ConfigException("no kubeconfig found")` and asserts `result.exit_code != 0` with the message in `result.output` would cover this path.

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/k8s_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+    with pytest.raises(click.ClickException, match="Server error: "):
+        handle_k8s_api_exception(error)
+
+
+def test_passes_through_a_missing_body():
+    error = api_exception(500, None)
+
+    with pytest.raises(click.ClickException, match="Server error: None"):
+        handle_k8s_api_exception(error)
```
raballew: [HIGH] This file tests `handle_k8s_api_exception` but has nothing for its sibling `handle_k8s_config_exception`, which is called from multiple CLI commands including the new apply command. Its body (`raise click.ClickException(e.args[0]) from e`) could silently regress. A simple test calling it with `ConfigException("bad config")` inside `pytest.raises(click.ClickException, match="bad config")` would provide the missing coverage.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+        which is what lets one command apply Clients, ExporterSets and whatever
+        the operator adds next.
+        """
+        cached = self._resources.get((group, version, kind))
+        if cached is not None:
+            return cached
+
+        try:
+            listing = await self._client.call_api(
+                f"/apis/{group}/{version}",
```
raballew: [MEDIUM] `_resolve_resource` builds the discovery URL as `f"/apis/{group}/{version}"` without validating the `version` segment. A manifest with `apiVersion: jumpstarter.dev/v1alpha1/../../../api/v1` passes `validate_manifest` and causes the GET to target an unintended path. Adding a format check in `validate_manifest` that rejects values not matching `v[0-9]+(alpha[0-9]+|beta[0-9]+)?`, or passing group and version through the kubernetes_asyncio client's URL template so they are percent-encoded, would prevent this.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+
+    api_version = document.get("apiVersion")
+    if not isinstance(api_version, str) or api_version == "":
+        raise ManifestError(f"{origin}: missing apiVersion")
+    group, _, version = api_version.partition("/")
+    if version == "":
+        raise ManifestError(
+            f"{origin}: apiVersion '{api_version}' is a core Kubernetes resource, "
+            "only Jumpstarter resources can be applied"
+        )
```
raballew: [MEDIUM] When `apiVersion` has a trailing slash but an empty version component (e.g. `"jumpstarter.dev/"`), the `partition` call produces `version == ""`, which falls into the "core Kubernetes resource" branch and emits a factually wrong error: `jumpstarter.dev` is not a core group. The fix is to distinguish the two cases: use `if sep == "":` for the genuine no-slash case, then separately handle the empty-version-after-slash case with a distinct message such as "apiVersion '...' is missing a version component".

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply_test.py
```diff
@@ -0,0 +1,305 @@
...
+    assert ("force", "true") in patch_call(api).kwargs["query_params"]
+
+
+@pytest.mark.asyncio
+async def test_apply_passes_a_dry_run_through_to_the_server():
+    api = make_api()
+
+    await api.apply(load_manifests(CLIENT_MANIFEST)[0], dry_run=True)
+
+    assert ("dryRun", "All") in patch_call(api).kwargs["query_params"]
```
raballew: [MEDIUM] `test_apply_passes_a_dry_run_through_to_the_server` asserts the `dryRun=All` query parameter is sent but discards the returned `applied` object. For a non-existent resource, the code correctly returns `"created"` via `_action_taken(None, applied, dry_run=True)`, but there is no assertion verifying this. Capturing the return value and adding `assert applied.action == "created"` (or a dedicated `test_a_dry_run_on_a_new_resource_reports_created`) would close this coverage gap.

python/packages/jumpstarter-cli-admin/jumpstarter_cli_admin/apply_test.py
```diff
@@ -0,0 +1,203 @@
+import json
```
raballew: [MEDIUM] The CLI tests cover two cases separately: all-valid multi-document manifests and single-document non-Jumpstarter manifests. There is no test for a mixed multi-document file containing one valid Jumpstarter resource and one non-Jumpstarter resource (e.g. `apps/v1 Deployment`). The expected behavior is to reject the whole file without calling `apply_all`. Adding a test for this combination would confirm the validation is applied consistently across all documents in a file.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+            if "/" in resource.get("name", ""):
+                continue
+            if resource.get("kind") == kind:
+                found = (resource["name"], bool(resource.get("namespaced", True)))
+                self._resources[(group, version, kind)] = found
+                return found
+
+        raise ManifestError(f"the cluster does not serve kind {kind} in {group}/{version}")
+
+    async def _read(self, group, version, plural, name, namespace) -> Optional[dict]:
```
raballew: [MEDIUM] `_read` has a return type annotation but all five positional parameters (`group`, `version`, `plural`, `name`, `namespace`) are unannotated. This is inconsistent with the rest of the codebase. Annotating them as `group: str, version: str, plural: str, name: str, namespace: Optional[str]` would bring the signature in line with the surrounding code.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+            return await self.api.get_namespaced_custom_object(
+                group=group, version=version, plural=plural, name=name, namespace=namespace
+            )
+        except ApiException as e:
+            if e.status == 404:
+                return None
+            raise
+
+    async def _server_side_apply(
+        self, group, version, plural, name, namespace, manifest: dict, *, dry_run: bool, force: bool
```
raballew: [MEDIUM] `_server_side_apply` has its keyword-only parameters (`manifest`, `dry_run`, `force`) annotated, but the five positional parameters (`group`, `version`, `plural`, `name`, `namespace`) are bare. Annotating them as `group: str, version: str, plural: str, name: str, namespace: Optional[str]` would make the signature consistent with `_read` and with call sites.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+    except yaml.YAMLError as e:
+        raise ManifestError(f"{origin} is not valid YAML: {e}") from e
+
+    resources = []
+    for index, document in enumerate(documents):
+        if document is None:
+            continue
+        # Only number the documents when there is more than one to point at.
+        label = origin if len(documents) == 1 else f"{origin} (document {index + 1})"
+        resources.append(validate_manifest(document, label))
```
raballew: [LOW] `load_manifests` uses `len(documents)` to decide whether to number documents, but `documents` includes `None` entries from empty YAML separator blocks. A file with a leading `---` and one real document produces `len(documents) == 2`, labeling the only real document as "(document 2)" instead of using the bare origin. Filtering out `None` entries before the length check, or using a separate counter for non-None documents, would fix the label.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+    kind = document.get("kind")
+    if not isinstance(kind, str) or kind == "":
+        raise ManifestError(f"{origin}: missing kind")
+
+    metadata = document.get("metadata")
+    if not isinstance(metadata, dict):
+        raise ManifestError(f"{origin}: missing metadata")
+    name = metadata.get("name")
+    if not isinstance(name, str) or name == "":
+        raise ManifestError(f"{origin}: missing metadata.name")
```
raballew: [LOW] `validate_manifest` only checks that `metadata.name` is a non-empty string. Names with slashes or control sequences pass locally. While `_server_side_apply` URL-encodes the name via path params (preventing path traversal), the name appears verbatim in CLI output through `resource.qualified_name`. Adding a pre-flight check against `^[a-z0-9]([a-z0-9\-]{0,251}[a-z0-9])?$` and raising `ManifestError` for non-matching values would provide a clearer error than waiting for the server to reject the request.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+                _return_http_data_only=True,
+            )
+        except ApiException as e:
+            if e.status == 404:
+                raise ManifestError(
+                    f"the cluster does not serve {group}/{version}, check that the Jumpstarter CRDs are installed"
+                ) from e
+            raise
+
+        for resource in listing.get("resources", []):
```
raballew: [LOW] `_resolve_resource` does not guard against a non-dict discovery response. `response_types_map={200: "object"}` can produce a list or `None` if a proxy returns unexpected JSON, causing `listing.get("resources", [])` to raise `AttributeError`. Adding a check after `call_api` such as `if not isinstance(listing, dict): raise ManifestError(f"unexpected discovery response for {group}/{version}: expected an object, got {type(listing).__name__}")` would surface the problem with a clean error message.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def _without_server_fields(obj: dict) -> dict:
+    trimmed = {key: value for key, value in obj.items() if key != "status"}
+    trimmed["metadata"] = {
+        key: value
+        for key, value in (obj.get("metadata") or {}).items()
+        if key not in _SERVER_OWNED_METADATA
+    }
+    return trimmed
```
raballew: [LOW] `_action_taken`, `_same_content`, `_without_server_fields`, and `_SERVER_OWNED_METADATA` are defined at module scope but are only used inside `ApplyV1Alpha1Api.apply()`. At module scope they look like reusable utilities, but they are not. The rest of the package (`clients.py`, `exporters.py`, `leases.py`) does not follow this pattern. Moving them inside `ApplyV1Alpha1Api` as `@staticmethod`s and a class-level `frozenset` would match the encapsulation style used everywhere else.

python/packages/jumpstarter-kubernetes/jumpstarter_kubernetes/apply.py
```diff
@@ -0,0 +1,327 @@
...
+
+class V1Alpha1AppliedResource(JsonBaseModel):
+    """One resource as the cluster stored it, and what applying it did."""
+
+    api_version: str = Field(alias="apiVersion")
+    kind: str
+    name: str
+    namespace: Optional[str] = None
+    action: Literal["created", "configured", "unchanged"]
+    resource: dict
```
raballew: [LOW] `V1Alpha1AppliedResource.resource` is typed as bare `dict` and holds the full server response including `managedFields`, `selfLink`, `uid`, and other noise. Every other model in the package exposes only relevant fields via typed Pydantic projections. For `-o json` output this means all server-side metadata is visible. Since apply handles arbitrary Kinds a full typed projection is not feasible, but stripping `managedFields` before storage (mirroring `_without_server_fields` already used for dry-run comparison) and annotating `resource: dict[str, object]` with a note about verbatim content would reduce the noise and make the intent explicit.

mangelajo (approved):

PR #1047, opened 2026-08-30
feat(exporterset): label pool members with the set that owns them

kirkbrauer:
An exporter provisioned by an ExporterSet carried nothing saying which set it came from, so a client looking at a list of exporters could not tell a pool member from a standalone exporter, or one pool's members from another's. Working that out meant reading the ExporterSet's `ownerReferences`, which needs a kubeconfig.

The reconciler now labels each member with the set, class and provisioner that own it:

```
exporterset.jumpstarter.dev/name
exporterset.jumpstarter.dev/class
exporterset.jumpstarter.dev/provisioner
```

Existing members are backfilled on reconcile, not only newly provisioned ones. These pass the controller's label denylist, so clients see them over the normal exporter listing with no protocol change.

### DIFF
--- a/controller/internal/exporterset/reconciler.go
+++ b/controller/internal/exporterset/reconciler.go
@@ -67,9 +67,12 @@ const (
 	// lives only in ownerReferences, which the client API does not expose.
 	labelExporterSetName = "exporterset.jumpstarter.dev/name"
 
-	// The VirtualTargetClass backing the pool, so clients can tell how an
-	// exporter is provisioned without cluster access.
+	// The VirtualTargetClass backing the pool, and the provisioner that
+	// class names, so clients can tell how an exporter is provisioned without
+	// cluster access. The provisioner is a property of the class, which a
+	// client cannot read, so it has to be carried here.
 	labelVirtualTargetClass = "exporterset.jumpstarter.dev/class"
+	labelProvisioner        = "exporterset.jumpstarter.dev/provisioner"
 
 	defaultScaleDownCooldown = 5 * time.Minute
 
@@ -347,7 +350,7 @@ func (r *ExporterSetReconciler) scaleUp(
 			ObjectMeta: metav1.ObjectMeta{
 				GenerateName: es.Name + "-",
 				Namespace:    es.Namespace,
-				Labels:       exporterLabels(es),
+				Labels:       r.exporterLabels(es),
 				Annotations:  maps.Clone(es.Spec.Template.Metadata.Annotations),
 			},
 			Spec: jumpstarterdevv1alpha1.ExporterSpec{
@@ -931,18 +934,32 @@ func (r *ExporterSetReconciler) clearSurplusAnnotation(ctx context.Context, es *
 	}
 }
 
+// identityLabels mark which pool an exporter belongs to and how it is
+// provisioned. Reconcile has already established that the referenced class
+// names this reconciler's provisioner, so it is the provisioner in effect.
+func (r *ExporterSetReconciler) identityLabels(
+	es *virtualtargetv1alpha1.ExporterSet,
+) map[string]string {
+	labels := map[string]string{labelExporterSetName: es.Name}
+	if es.Spec.VirtualTargetClassName != "" {
+		labels[labelVirtualTargetClass] = es.Spec.VirtualTargetClassName
+	}
+	if r.Provisioner != nil {
+		labels[labelProvisioner] = r.Provisioner.Name()
+	}
+	return labels
+}
+
 // exporterLabels are the labels an Exporter of this set carries: the set's
-// template labels plus the identity labels that let a client tell which pool
-// an exporter belongs to and how it is provisioned.
-func exporterLabels(es *virtualtargetv1alpha1.ExporterSet) map[string]string {
+// template labels plus the identity labels above.
+func (r *ExporterSetReconciler) exporterLabels(
+	es *virtualtargetv1alpha1.ExporterSet,
+) map[string]string {
 	labels := maps.Clone(es.Spec.Template.Metadata.Labels)
 	if labels == nil {
 		labels = map[string]string{}
 	}
-	labels[labelExporterSetName] = es.Name
-	if es.Spec.VirtualTargetClassName != "" {
-		labels[labelVirtualTargetClass] = es.Spec.VirtualTargetClassName
-	}
+	maps.Copy(labels, r.identityLabels(es))
 	return labels
 }
 
@@ -956,13 +973,10 @@ func (r *ExporterSetReconciler) reconcileExporterLabels(
 ) error {
 	logger := log.FromContext(ctx)
 
+	desired := r.identityLabels(es)
+
 	for i := range owned {
 		exporter := &owned[i]
-		desired := map[string]string{labelExporterSetName: es.Name}
-		if es.Spec.VirtualTargetClassName != "" {
-			desired[labelVirtualTargetClass] = es.Spec.VirtualTargetClassName
-		}
-
 		missing := map[string]string{}
 		for key, value := range desired {
 			if exporter.Labels[key] != value {
--- a/controller/internal/exporterset/reconciler.go
+++ b/controller/internal/exporterset/reconciler.go
@@ -62,8 +62,14 @@ import (
 const (
 	annotationSurplusSince = "exporterset.jumpstarter.dev/surplus-since"
 
-	// Bridges the grandparent lookup (ExporterSet -> Exporter -> Pod).
+	// Bridges the grandparent lookup (ExporterSet -> Exporter -> Pod), and
+	// tells clients which pool an exporter came from: membership otherwise
+	// lives only in ownerReferences, which the client API does not expose.
 	labelExporterSetName = "exporterset.jumpstarter.dev/name"
+
+	// The VirtualTargetClass backing the pool, so clients can tell how an
+	// exporter is provisioned without cluster access.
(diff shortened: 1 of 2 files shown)


controller/internal/exporterset/reconciler.go
```diff
@@ -62,9 +62,18 @@ import (
...
+	// tells clients which pool an exporter came from: membership otherwise
+	// lives only in ownerReferences, which the client API does not expose.
 	labelExporterSetName = "exporterset.jumpstarter.dev/name"
 
+	// The VirtualTargetClass backing the pool, and the provisioner that
+	// class names, so clients can tell how an exporter is provisioned without
+	// cluster access. The provisioner is a property of the class, which a
+	// client cannot read, so it has to be carried here.
+	labelVirtualTargetClass = "exporterset.jumpstarter.dev/class"
+	labelProvisioner        = "exporterset.jumpstarter.dev/provisioner"
```
mangelajo, line 75: These three labels will be visible to every client by default. The PR description says they "pass the controller's label denylist", but that's because `hiddenLabels.keys` defaults to empty , nothing is hidden unless an admin explicitly configures it.

In practice, every ExporterSet-managed exporter shown via `jmp get exporters` or the MCP `jmp_list_exporters` tool will render all three labels in the LABELS column:

```
exporterset.jumpstarter.dev/class=qemu-class,exporterset.jumpstarter.dev/name=demo-set,exporterset.jumpstarter.dev/provisioner=qemu.jumpstarter.dev,exporterset=demo-set
```

That's a lot of noise compared to a standalone exporter with just `board=rpi4`. A few options to consider:

I would add these to `hiddenLabels.keys` by default in the operator's default config, so they're hidden unless requested via `--show-hidden-labels`. Clients that want to filter by pool can still use them in `--filter` selectors even when hidden.

controller/internal/exporterset/reconciler.go
```diff
@@ -62,9 +62,18 @@ import (
...
 	annotationSurplusSince = "exporterset.jumpstarter.dev/surplus-since"
 
-	// Bridges the grandparent lookup (ExporterSet -> Exporter -> Pod).
+	// Bridges the grandparent lookup (ExporterSet -> Exporter -> Pod), and
+	// tells clients which pool an exporter came from: membership otherwise
+	// lives only in ownerReferences, which the client API does not expose.
 	labelExporterSetName = "exporterset.jumpstarter.dev/name"
 
+	// The VirtualTargetClass backing the pool, and the provisioner that
+	// class names, so clients can tell how an exporter is provisioned without
```
mangelajo, line 71: Nit: the comment seems like cut in the middle?  "The VirtualTargetClass backing the pool, and the provisioner that class names"

controller/internal/exporterset/reconciler.go
```diff
@@ -919,6 +934,72 @@ func (r *ExporterSetReconciler) clearSurplusAnnotation(ctx context.Context, es *
...
+			exporter.Labels = map[string]string{}
+		}
+		maps.Copy(exporter.Labels, missing)
+		if err := r.Patch(ctx, exporter, patch); err != nil {
+			return fmt.Errorf("unable to label Exporter %s: %w", exporter.Name, err)
+		}
+		logger.Info("stamped exporter set labels", "exporter", exporter.Name, "labels", missing)
+	}
+	return nil
+}
```
mangelajo: The backfill logic looks correct — it patches only exporters that are missing or have stale labels. One consideration: if a cluster has many ExporterSets with many exporters, each reconcile will iterate all owned exporters and issue a Patch for each one missing labels. This is fine as a one-time migration cost, but on a busy cluster the first reconcile after upgrade could produce a burst of API calls.

Since this only fires for exporters with stale labels and converges quickly (subsequent reconciles skip the `len(missing) == 0` path), this is acceptable. Just noting it for awareness.

---

controller/internal/exporterset/reconciler.go
```diff
@@ -919,6 +934,72 @@ func (r *ExporterSetReconciler) clearSurplusAnnotation(ctx context.Context, es *
...
+func (r *ExporterSetReconciler) exporterLabels(
+	es *virtualtargetv1alpha1.ExporterSet,
+) map[string]string {
+	labels := maps.Clone(es.Spec.Template.Metadata.Labels)
+	if labels == nil {
+		labels = map[string]string{}
+	}
+	maps.Copy(labels, r.identityLabels(es))
+	return labels
+}
```
mangelajo: Identity labels take precedence over template labels via `maps.Copy(labels, r.identityLabels(es))` — if a user's template happened to contain a label with key `exporterset.jumpstarter.dev/name`, it would be silently overwritten. This is probably fine (users shouldn't use the `exporterset.jumpstarter.dev/` prefix), but worth a brief note in the comment or a log warning if there's a collision.

---

controller/internal/exporterset/reconciler.go
```diff
@@ -919,6 +934,72 @@ func (r *ExporterSetReconciler) clearSurplusAnnotation(ctx context.Context, es *
...
+		if len(missing) == 0 {
+			continue
+		}
+
+		patch := client.MergeFrom(exporter.DeepCopy())
+		if exporter.Labels == nil {
+			exporter.Labels = map[string]string{}
+		}
+		maps.Copy(exporter.Labels, missing)
+		if err := r.Patch(ctx, exporter, patch); err != nil {
```
raballew, line 995: This returns on the first patch error, leaving all remaining exporters in the slice unlabeled for that reconcile cycle. Replace with log-and-continue, accumulate errors, and return a combined error after the loop using `errors.Join`.

controller/internal/exporterset/reconciler_test.go
```diff
@@ -2014,3 +2014,74 @@ func TestMergeImages_esOverridesVtc(t *testing.T) {
...
+	if got.Labels[labelExporterSetName] != "demo-set" {
+		t.Errorf("existing exporter not labelled: %v", got.Labels)
+	}
+	if got.Labels[labelVirtualTargetClass] != "qemu-class" {
+		t.Errorf("existing exporter missing class label: %v", got.Labels)
+	}
+	if got.Labels[labelProvisioner] != qemu.ProvisionerName {
+		t.Errorf("existing exporter missing provisioner label: %v", got.Labels)
+	}
+}
```
raballew: [MEDIUM] `TestReconcile_backfillsIdentityLabelsOnExistingExporters` only creates a single pre-existing exporter. The backfill logic is not exercised against a slice of exporters in mixed label states (e.g., one missing all identity labels, one with a stale class label, one already correct). Add a multi-exporter variant of this test to confirm all exporters reach the correct label state in a single reconcile cycle.

controller/internal/exporterset/reconciler_test.go
```diff
@@ -2014,3 +2014,74 @@ func TestMergeImages_esOverridesVtc(t *testing.T) {
...
+
+func TestReconcile_backfillsIdentityLabelsOnExistingExporters(t *testing.T) {
+	es := makeExporterSet(func(es *virtualtargetv1alpha1.ExporterSet) {
+		es.Spec.MinReplicas = 1
+		es.Spec.MinAvailableReplicas = 0
+	})
+	// An exporter from before these labels existed.
+	existing := makeExporter("demo-set-old", true, false, true)
+	delete(existing.Labels, labelExporterSetName)
```
raballew: [MEDIUM] The `delete(existing.Labels, labelExporterSetName)` call here is a no-op: `makeExporter` initializes labels with the key `"exporterset"`, not `labelExporterSetName` (`"exporterset.jumpstarter.dev/name"`), so there is nothing to delete. The test passes identically with or without this line. The comment above already conveys the intent; remove the dead delete to avoid misleading readers.

controller/internal/exporterset/reconciler.go
```diff
@@ -919,6 +934,72 @@ func (r *ExporterSetReconciler) clearSurplusAnnotation(ctx context.Context, es *
...
 }
 
+// identityLabels mark which pool an exporter belongs to and how it is
+// provisioned. Reconcile has already established that the referenced class
+// names this reconciler's provisioner, so it is the provisioner in effect.
+func (r *ExporterSetReconciler) identityLabels(
+	es *virtualtargetv1alpha1.ExporterSet,
+) map[string]string {
+	labels := map[string]string{labelExporterSetName: es.Name}
+	if es.Spec.VirtualTargetClassName != "" {
```
raballew: [LOW] `VirtualTargetClassName` has only a `MinLength=1` constraint in the CRD. Kubernetes label values are capped at 63 characters, but DNS subdomain names can be up to 253 characters. A VirtualTargetClass name between 64 and 253 characters passes ExporterSet admission but causes every `r.Patch` call in `reconcileExporterLabels` to return a 422 error, putting the reconciler in a permanent requeue loop. Add `+kubebuilder:validation:MaxLength=63` to the `VirtualTargetClassName` field in the CRD and regenerate.

controller/internal/exporterset/reconciler.go
raballew: [LOW] The new `labelVirtualTargetClass` and `labelProvisioner` labels are stamped on every managed Exporter. Any principal with `exporters/get` or `exporters/list` access can now read the VirtualTargetClass name and provisioner identifier without needing RBAC access to VirtualTargetClass objects. If RBAC restricts VTC visibility to cluster-admins, that restriction is bypassed through Exporter labels. Worth documenting this RBAC implication and deciding whether the provisioner name should appear on Exporter objects at all.

raballew:
@kirkbrauer this seems to be a quick win. Can you address the review feedback?

PR #1048, opened 2026-08-30
fix(lease): let a pool provision for a lease that matches nothing yet

kirkbrauer:
An ExporterSet with `minReplicas: 0` has no members until something asks for one, so a lease against it matches zero exporters. The lease controller called that unsatisfiable, and `reconcileStatusEnded` ends an unsatisfiable lease in the same reconcile. The set's demand rule only counts leases that are pending and not ended, so it never saw the request — a pool that keeps nothing warm could never provision on demand, which is the case it exists for.

The controller now asks whether an exporter set could provision a match before deciding. If one could, the lease becomes `Pending`/`Provisioning` and is requeued, which is the state the set scales on. If none could — no pool matches the selector, or the only one that does is at `maxReplicas` — the lease is unsatisfiable as before, so a typo in a selector still fails fast rather than hanging.

This is the first thing the lease controller reads outside `jumpstarter.dev`, so it needs the virtualtarget scheme registered and a read on exportersets in the Role the operator builds for it. Where that read is refused — a controller newer than its operator, or a cluster without the exporter set CRDs — the check reports "no pool" and behaviour is exactly what it was before.

Three specs cover it. With the fix reverted, the one describing the bug fails and the other two (pool at ceiling, no matching pool) still pass.

### DIFF
--- a/controller/internal/controller/lease_controller.go
+++ b/controller/internal/controller/lease_controller.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	jumpstarterdevv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/api/v1alpha1"
+	virtualtargetv1alpha1 "github.com/jumpstarter-dev/jumpstarter/controller/api/virtualtarget/v1alpha1"
 	jmpmetrics "github.com/jumpstarter-dev/jumpstarter/controller/internal/metrics"
 	corev1 "k8s.io/api/core/v1"
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
@@ -63,6 +64,7 @@ type ApprovedExporter struct {
 // +kubebuilder:rbac:groups=jumpstarter.dev,resources=leases,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=jumpstarter.dev,resources=leases/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=jumpstarter.dev,resources=leases/finalizers,verbs=update
+// +kubebuilder:rbac:groups=virtualtarget.jumpstarter.dev,resources=exportersets,verbs=get;list;watch
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 // TODO(user): Modify the Reconcile function to compare the state specified by
@@ -288,6 +290,27 @@ func (r *LeaseReconciler) reconcileStatusExporterRef(
 			listed, err := r.ListMatchingExporters(ctx, lease, selector)
 			if err != nil {
 				return fmt.Errorf("reconcileStatusExporterRef: failed to list matching exporters: %w", err)
+			}
+			if len(listed.Items) == 0 {
+				// Nothing matches yet, which is the normal state of a pool that
+				// keeps nothing warm. An exporter set that could provision a
+				// match makes this lease pending rather than unsatisfiable —
+				// and pending is what the set counts as demand to scale up on.
+				// Calling it unsatisfiable here ends the lease in this same
+				// reconcile, so the set would never see it.
+				set, err := r.exporterSetCanProvision(ctx, lease, selector)
+				if err != nil {
+					return fmt.Errorf("reconcileStatusExporterRef: %w", err)
+				}
+				if set != "" {
+					lease.SetStatusPending(
+						"Provisioning",
+						"No exporter matches the selector yet; exporter set %s can provision one",
+						set,
+					)
+					result.RequeueAfter = pendingRequeueAfter(lease)
+					return nil
+				}
 			}
 			// Filter out disabled exporters from selector-based listing
 			matchingExporters = filterOutDisabledExporters(listed.Items)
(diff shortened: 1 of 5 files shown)


controller/internal/controller/lease_controller.go
```diff
@@ -207,6 +209,98 @@ func (r *LeaseReconciler) reconcileStatusBeginEndTimes(
...
+//
+// It reports decided=true when it has already settled the lease's status and
+// the caller should stop — the named exporter is missing, does not match, or is
+// disabled; every match is disabled; or nothing matches yet and a pool can
+// provide one.
+func (r *LeaseReconciler) candidateExporters(
+	ctx context.Context,
+	result *ctrl.Result,
+	lease *jumpstarterdevv1alpha1.Lease,
+	selector labels.Selector,
```
mangelajo, line 223: don't we have the selector in lease already?

controller/internal/controller/lease_controller.go
```diff
@@ -245,60 +339,12 @@ func (r *LeaseReconciler) reconcileStatusExporterRef(
 			return nil
 		}
```
mangelajo, line 340: nit: you can move this block inside the function as well, and then no need to pass selector. It's not used anywhere later :)

controller/internal/controller/lease_controller.go
```diff
@@ -207,6 +209,98 @@ func (r *LeaseReconciler) reconcileStatusBeginEndTimes(
...
+	matchingExporters := filterOutDisabledExporters(listed.Items)
+	if len(matchingExporters) == 0 && len(listed.Items) > 0 {
+		lease.SetStatusUnsatisfiable(
+			"AllDisabled",
+			"All %d exporters matching the selector are disabled",
+			len(listed.Items),
+		)
+		return nil, true, nil
+	}
+	return matchingExporters, false, nil
```
mangelajo, line 301: When `len(listed.Items) == 0` and no pool can provision (`set == ""`), the function falls through to the disabled-exporter filter and returns `(nil, false, nil)` — an empty slice with `decided=false`. The caller then passes zero exporters into `attachMatchingPolicies`, which produces zero approved exporters, and the lease is marked unsatisfiable with reason `"NoAccess"` and a message about policies.

That message is misleading when the real cause is "no exporters match the selector at all." The old code had the same behavior (this is pre-existing), but since you're refactoring this path anyway, it might be a good time to handle the `len(listed.Items) == 0 && set == ""` case explicitly with a clearer `SetStatusUnsatisfiable("NoMatch", ...)` before falling through.

---

controller/internal/controller/lease_controller.go
```diff
@@ -557,6 +603,51 @@ func (r *LeaseReconciler) ListMatchingExporters(ctx context.Context, lease *jump
...
+	var sets virtualtargetv1alpha1.ExporterSetList
+	if err := r.List(ctx, &sets, client.InNamespace(lease.Namespace)); err != nil {
+		if k8serrors.IsForbidden(err) || meta.IsNoMatchError(err) {
+			// A controller running against an operator that has not granted
+			// this read, or a cluster without the exporter set CRDs at all.
+			// Neither is transient, and neither is a reason to fail a lease:
+			// fall back to the behaviour from before pools existed.
+			log.FromContext(ctx).V(1).Info(
+				"cannot read exporter sets, treating the selector as unsatisfiable",
+				"lease", lease.Name, "reason", err.Error())
```
mangelajo, line 632: Nit: the log message says "treating the selector as unsatisfiable" but returning `"", nil` actually means "no pool can help, so let the normal flow decide." This is accurate behavior-wise (the existing flow will mark it unsatisfiable), but the message could be slightly more precise — something like "no pool check available, falling back to normal selector resolution" — to avoid confusion when reading logs, since the function itself isn't the one setting unsatisfiable.

---

controller/internal/controller/lease_controller_test.go
```diff
@@ -2654,3 +2655,115 @@ var _ = Describe("pendingRequeueAfter", func() {
...
+
+			updated := getLease(ctx, lease.Name)
+			Expect(meta.IsStatusConditionTrue(
+				updated.Status.Conditions,
+				string(jumpstarterdevv1alpha1.LeaseConditionTypeUnsatisfiable),
+			)).To(BeTrue())
+			Expect(updated.Status.Ended).To(BeTrue())
+		})
+	})
+})
```
mangelajo, line 2769: Good test coverage for the three key scenarios. One edge case that might be worth adding: a pool with `maxReplicas: 0` (unlimited) and zero current replicas should also keep the lease pending (verifying the `maxReplicas > 0` guard in `exporterSetCanProvision` works correctly for the unlimited case).

---

controller/internal/controller/lease_controller.go
```diff
@@ -557,6 +603,51 @@ func (r *LeaseReconciler) ListMatchingExporters(ctx context.Context, lease *jump
...
+		if !selector.Matches(labels.Set(set.Spec.Template.Metadata.Labels)) {
+			continue
+		}
+		if set.Spec.MaxReplicas > 0 && set.Status.Replicas >= set.Spec.MaxReplicas {
+			continue
+		}
+		return set.Name, nil
+	}
+	return "", nil
+}
```
mangelajo, line 649: The lease controller now reads ExporterSets, but `SetupWithManager` doesn't watch them. When a pool that was at ceiling loses a replica (dropping below `MaxReplicas`), pending "Provisioning" leases won't be re-evaluated until their next periodic requeue (up to 30s).

The existing `Watches(&Exporter{}, ...)` already covers the happy path (new exporter appears → lease reconciles), so the gap mainly affects the "at ceiling → capacity freed" transition. Not a blocker but worth considering a `Watches(&ExporterSet{}, ...)` with a predicate on `Status.Replicas` changes, mapping to pending leases whose selectors match the set's template labels.

---

controller/internal/controller/lease_controller.go
mangelajo: inside the candidateExporters function I mean.

PR #1030, opened 2026-08-27
Update default python installation source to release-0.9

mangelajo:

PR #1053, opened 2026-08-31
feat(cli): add self update command to jmp

engelmi:
The recommended installation and update mechanism is to use the provided install script and run it with the desired source.

Adds a new CLI command `self update` to `jmp, which fetches the script from the repository and executes it. This is similar to the self update command from uv.

### DIFF
--- a/python/packages/jumpstarter-cli/jumpstarter_cli/self_update.py
+++ b/python/packages/jumpstarter-cli/jumpstarter_cli/self_update.py
@@ -1,0 +1,43 @@
+import shutil
+import subprocess
+import urllib.request
+from pathlib import Path
+
+import click
+from jumpstarter_cli_common.exceptions import handle_exceptions
+
+
+def _determine_install_dir() -> Path | None:
+    jmp_path = shutil.which("jmp")
+    if jmp_path:
+        return Path(jmp_path).parent.parent
+    return None
+
+def _fetch_install_script() -> str:
+    INSTALL_SCRIPT_URL = "https://raw.githubusercontent.com/jumpstarter-dev/jumpstarter/main/python/install.sh"
+    with urllib.request.urlopen(INSTALL_SCRIPT_URL) as response:
+        return response.read().decode("utf-8")
+
+
+@click.command("update")
+@click.option(
+    "--source",
+    type=str,
+    help="Overwrites the current installation source. Available: latest, rc, main and release-x.x",
+    default=None,
+)
+@handle_exceptions
+def self_update(source: str):
+    """
+    Update jumpstarter
+    """
+    install_dir = _determine_install_dir()
+    script = _fetch_install_script()
+
+    cmd = ["bash", "-s", "-"]
+    if install_dir:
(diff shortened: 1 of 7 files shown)


python/packages/jumpstarter-cli/jumpstarter_cli/self_update.py
```diff
@@ -0,0 +1,30 @@
...
+import urllib.request
+
+import click
+from jumpstarter_cli_common.exceptions import handle_exceptions
+
+
+@click.command("update")
+@click.option(
+    "--source",
+    type=click.Choice(["release-0.8", "latest", "rc", "main"]),
```
bennyz: latest release is 0.9, and we probably need to guard against drift here

python/packages/jumpstarter-cli/jumpstarter_cli/self_update.py
```diff
@@ -0,0 +1,30 @@
...
+    help="Installation source",
+    default="main",
```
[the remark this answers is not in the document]
bennyz: i agree, while internally we recommend using main it's not our "official" policy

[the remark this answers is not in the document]
engelmi: ~Yes, the script sets the default to `release-0.8` at the moment. In practice, however, the installer was mostly run with `main` AFAIK. This might need an update.~

Makes sense. Lets change it to the release-xx then.

python/packages/jumpstarter-cli/jumpstarter_cli/self_update.py
engelmi: I was using it from 
https://github.com/jumpstarter-dev/jumpstarter/blob/main/python/install.sh#L55

How about using only `release` as possible choice? Then we would need to determine the latest release (0.9.0, 0.8.1 etc.) - is there a better way than using the git tags for it?

bennyz: yeah, that sounds good to me

mangelajo: @bennyz can you have an eye to this one: https://github.com/jumpstarter-dev/jumpstarter/pull/1030 :)

mangelajo: @engelmi @bennyz I think it would be cool if install.sh stored the preferred branch (main/release/xxx) in ~/.local/jumpstarter/install_branch or similar

and if that file is present, stick to that version instead. So that we continue to update on the branch the user installed.

engelmi: @mangelajo @bennyz Added another commit: the install.sh creates the `install_source` file and uses the order CLI -s option > install_source file > default source. The `self update` is now agnostic to the possible sources (validity is handled in the script anyway), but still provides the `--source` parameter to support overwriting/switching the current source.

bennyz (approved):

mangelajo:
Thank you @engelmi !! :)

PR #1054, opened 2026-09-01
ci: add go fix check to controller lint job

bennyz:
in go 1.26 `go fix` has been improved https://go.dev/blog/gofix
it can convert loops, use new `new` semantics instead of `ptr.To`, etc


controller/internal/config/types.go
```diff
@@ -15,9 +15,9 @@ type Config struct {
 	Authentication   Authentication   `json:"authentication" yaml:"authentication"`
 	Provisioning     Provisioning     `json:"provisioning" yaml:"provisioning"`
 	Grpc             Grpc             `json:"grpc" yaml:"grpc"`
-	LeasePolicy      LeasePolicy      `json:"leasePolicy,omitempty" yaml:"leasePolicy,omitempty"`
```
mangelajo, line 18: why does it remove the omitempty? is it the default now? I guess...

bennyz: so apparently encoding/json does not actually omit empty structs making this a noop

in 1.24 omitzero was introduced for this https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/modernize#hdr-Analyzer_omitzero

.github/workflows/lint.yaml
```diff
@@ -57,6 +57,12 @@ jobs:
           go-version-file: .go-version
           cache-dependency-path: "**/go.sum"
 
+      - name: Run go fix
+        working-directory: controller
+        run: |
+          go fix ./...
+          git diff --exit-code
```
raballew, line 65: Does this cover `controller/deploy/operator/` which has its own `go.mod` as well? I am worried this could mean `interface` usages in the sub-module will silently pass the check.

controller/internal/metrics/lease.go
```diff
@@ -192,7 +192,7 @@ func truncateRunes(s string, maxRunes int) string {
 		return s
 	}
 	i := 0
-	for n := 0; n < maxRunes; n++ {
+	for range maxRunes {
```
raballew: [LOW] `truncateRunes` is unexported, handles multi-byte UTF-8, and has no direct unit test. `grep -rn "truncateRunes" controller/ --include="*_test.go"` returns nothing; coverage is only indirect via `TestRecordAcquisitionTruncatesOverBudgetExemplars`, which does not exercise multi-byte rune truncation.

It would be worth adding a `TestTruncateRunes` in `controller/internal/metrics/lease_test.go` covering: ASCII strings at/under/over the limit, multi-byte UTF-8 input (e.g. Cyrillic, emoji), `maxRunes == 0`, and `maxRunes >= len(runes)`. This pins the behaviour and makes the `for range` migration verifiably correct without relying on e2e tests.

.github/workflows/lint.yaml
raballew: [LOW] When `go fix` rewrites files, `git diff --exit-code` exits non-zero and dumps a raw unified diff to the Actions log, but there is no accompanying message telling contributors what went wrong or how to fix it locally. A contributor seeing this failure has to infer what command to run.

Consider replacing the step body with something like:
```yaml
run: |
  go fix ./...
  if ! git diff --quiet; then
    echo "go fix found patterns to modernize. Run 'go fix ./...' in the controller/ directory and commit the result."
    git diff
    exit 1
  fi
```
This makes the failure self-explanatory in the Actions UI.

bennyz: right, added

mangelajo (approved):

PR #1060, opened 2026-09-02
JEP-0013 Phase 3 - MetricsStream

RoddieKieley:
## Summary

JEP-0013 Phase 3 PR A**: the `MetricsStream` contract and the telemetry reverse-scrape hub. This is the merge-base for the rest of the Phase 3 series.

Related PRs will be linked here as they are opened (B, C, D, E). This PR stays draft until those links are filled in.

- Add `MetricsStream` to `telemetry.proto` (register + scrape request/response) and regenerate Go + Python stubs.
- Telemetry fans out scrapes over connected streams, merges OpenMetrics, remaps unknown `driver_type` to `other`, and serves `GET /metrics` on a dedicated HTTP port (not gRPC `:9093`).
- `/healthz` and `/readyz` as in DD-7. `jumpstarter_scrape_timeouts_total` on scrape timeout.
- **Lab follow-up in the same PR:** stop silently dropping unparseable exporter snapshots. Log `exporter metrics snapshot omitted` and increment `jumpstarter_metrics_parse_errors_total{exporter}` on the same `/metrics` response. Python `prometheus_client` OpenMetrics exemplars (`# {lease_id=...}`) still fail `parseMetricFamilies`; the snapshot is omitted so one exporter cannot 500 the hub. Fixing OpenMetrics/exemplar parse is a later pass — not this PR.

No exporter client, no operator/image/CRD, no Loki in this PR.

## DEMO

An asciinema demo for this combined Phase 3 work is available in the[ jep-0013-phase3-demo branch](https://github.com/RoddieKieley/jumpstarter/tree/jep-0013-phase3-demo) in my repository. Best to [read the description in the DEMO.md there](https://github.com/RoddieKieley/jumpstarter/blob/jep-0013-phase3-demo/DEMO.md) and then watch (manually for now) the asciinema file contained at the end -  [jep-0013-phase3-demo.cast](https://github.com/RoddieKieley/jumpstarter/blob/jep-0013-phase3-demo/jep-0013-phase3-demo.cast)

### IMPORTANT NOTES

* See [MetricsStream during the lease](https://github.com/RoddieKieley/jumpstarter/blob/jep-0013-phase3-demo/DEMO.md#metricsstream-during-the-lease) for an important caveat about the implementation about a potential limitation pointed out in the original JEP-0013 DD-3 regarding potential limitations for exemplar support.
* Register identity must be exporter_name after #1058 (not Metadata.name / "unknown"). That is in PR C.

## How this PR fits the series

```mermaid
flowchart TB
  A["PR A this PR: proto + hub"]
  B["PR B: image + metrics port + scrape CR"]
  C["PR C: exporter MetricsStream client"]
  D["PR D: Loki HTTP push"]
  E["PR E: jmp PushLogs"]
  A --> B
  A --> C
  B --> D
  D --> E
```

| PR | Branch | Status |
|----|--------|--------|
| **A** | `jep-0013-phase3-metricsstream` | **This PR** |
| B | `jep-0013-phase3-operator-image` | #1061  |
| C | `jep-0013-phase3-exporter-metricsstream` | #1062  |
| D | `jep-0013-phase3-loki-push` | #1063  |
| E | `jep-0013-phase3-client-pushlogs` | #1064  |

Reverse-scrape is useful only after **A+B+C**. This PR is still reviewable alone with mock streams.

## Data flow (this PR)

```mermaid
sequenceDiagram
  participant Prom as Prometheus
  participant Tel as jumpstarter-telemetry
  participant Exp as Exporter stream (PR C)
  Prom->>Tel: GET /metrics
  Tel->>Exp: MetricsScrapeRequest
  Exp-->>Tel: OpenMetrics snapshot
  alt parse OK
    Tel-->>Prom: merged families + scrape_timeouts
  else OpenMetrics exemplar parse fail
    Tel-->>Prom: snapshot omitted + parse_errors_total
  end
```

## Out of scope / later passes

- Exporter client (C), operator image/metrics port (B), Loki (D), `jmp` logs (E)
- OpenMetrics exemplar parse (hub still omits those snapshots; now visible)
- `ServiceMonitor` (Phase 5), driver telemetry API (Phase 4), multi-replica sticky streams (DD-8)
- Out-of-cluster telemetry Route

### Known merge work (not new features)

* #1062 conflicts with main on exporter.py because of [#1059](https://github.com/jumpstarter-dev/jumpstarter/pull/1059). When C rebases, keep identity=exporter_name, _session_labels(), and #1059’s exporter_name= on Session(...).
* [#1027](https://github.com/jumpstarter-dev/jumpstarter/pull/1027) still overlaps operator telemetry files with B/D.
The series is still based on #1058. After A lands, rebase the rest onto current main (#1059) as a set.

### Explicitly later (already called out as out of scope)

* OpenMetrics exemplar parse (A omits those snapshots; JEP DD-3)
* ServiceMonitor (Phase 5), driver telemetry API (Phase 4)
* Out-of-cluster telemetry Route (demo-only)
* Changing PushLogs logger.Error(nil, …) for exporter TFTP errors
* Cosmetic only: #1060 still has the[ jep-0013-phase3-demo.


protocol/proto/jumpstarter/v1/telemetry.proto
```diff
@@ -6,13 +6,44 @@ package jumpstarter.v1;
...
 
+// Exporter → Telemetry
+message MetricsStreamRequest {
+  oneof msg {
+    MetricsRegister register = 1;                    // First message: identify this exporter.
+    MetricsScrapeResponse scrape_response = 2;       // Subsequent: reply to a scrape.
+  }
+}
+
+message MetricsRegister {
```
raballew: [CRITICAL] `MetricsRegister` message is missing a doc comment, which will fail `buf lint` with the STANDARD ruleset (COMMENT_MESSAGE rule). Add a comment like `// MetricsRegister is the first message an exporter sends on a MetricsStream to identify itself to the telemetry server.` immediately before the `message MetricsRegister` declaration. Reproduction: `buf lint protocol/proto/jumpstarter/v1/telemetry.proto` -> `telemetry.proto:28:1:Message "MetricsRegister" should have a comment`.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+	// before hub metrics are gathered into the same /metrics response.
+	families := mergeSnapshots(snaps, nil, s.mergeConfigFor, s.recordMetricsParseError)
+
+	if s.metricsRegistry != nil {
+		gathered, err := s.metricsRegistry.Gather()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		families = mergeSnapshots(nil, append(gathered, families...), s.mergeConfigFor, nil)
```
raballew: Flip the append order to `append(families, gathered...)` so hub metrics always win. Or is it intentional that exporters can override?

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+	var snaps []exporterSnapshot
+	var wg sync.WaitGroup
+	for _, c := range conns {
+		wg.Add(1)
+		go func(c *metricsConn) {
+			defer wg.Done()
+			text, err := c.scrape(ctx, timeout)
+			if err != nil {
+				if errors.Is(err, errScrapeTimeout) || errors.Is(err, context.DeadlineExceeded) {
+					s.scrapeTimeouts.Inc()
```
raballew: Change `scrapeTimeouts` to `*prometheus.CounterVec` with `[]string{"exporter"}` and increment with `s.scrapeTimeouts.WithLabelValues(c.id.name).Inc()` so a user could sum it.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+		c.mu.Unlock()
+	}()
+
+	err := c.send(&pb.MetricsStreamResponse{
+		Msg: &pb.MetricsStreamResponse_ScrapeRequest{
+			ScrapeRequest: &pb.MetricsScrapeRequest{},
+		},
+	})
+	if err != nil {
+		return nil, err
```
raballew: Add a test constructing a `metricsConn` with `send` returning a sentinel error, calling `scrape`, and asserting the returned error matches the sentinel.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+		if resp == nil {
+			return nil, errScrapeTimeout
+		}
+		return resp.GetMetricsText(), nil
+	case <-timer.C:
+		return nil, errScrapeTimeout
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-c.done:
+		return nil, context.Canceled
```
raballew: No test closes the `done` channel while a scrape is in flight.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+	}
```
raballew, line 223: `c.send()` is called before the `select` that checks `ctx.Done()` and `timer.C`. `stream.Send` will block in-definitively on stalled exporter connections.

controller/cmd/telemetry/main.go
```diff
@@ -55,9 +57,36 @@ var (
...
+	parts := strings.Split(s, ",")
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p != "" {
+			out = append(out, p)
+		}
+	}
+	return out
+}
```
raballew: [MEDIUM] `splitCSV` has no test file. There is no `*_test.go` anywhere under `controller/cmd/telemetry/`. `splitCSV` controls `--driver-type-enum` and `--exemplar-keys` flag parsing. Add `controller/cmd/telemetry/main_test.go` with table-driven tests covering: empty string, single value, whitespace-padded entries, entries that collapse to empty after trimming, and multi-value CSV.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+
+	select {
+	case err := <-errCh:
+		if err != nil && err != io.EOF && status.Code(err) != codes.Canceled && status.Code(err) != codes.Unavailable {
+			t.Fatalf("stream goroutine: %v", err)
+		}
+	default:
+	}
+}
```
raballew: [MEDIUM] There is no test for `GET /metrics` when zero exporters are connected. All integration tests register at least one exporter before calling the endpoint. The `fanoutScrapes` early-return path that produces only hub counters is untested through the HTTP stack. Add `TestMetrics_EmptyHubReturnsValidOpenMetrics` that calls `startTestHub` with no exporters, calls `GET /metrics`, and asserts HTTP 200 with a body containing `# EOF` and `jumpstarter_scrape_timeouts_total`.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+			defer wg.Done()
+			text, err := c.scrape(ctx, timeout)
+			if err != nil {
+				if errors.Is(err, errScrapeTimeout) || errors.Is(err, context.DeadlineExceeded) {
+					s.scrapeTimeouts.Inc()
+				}
+				log.FromContext(ctx).WithName("telemetry").V(1).Info("exporter scrape omitted",
+					"exporter", c.id.name,
+					"error", err.Error(),
+				)
```
raballew: [MEDIUM] Scrape timeout events are logged at `V(1).Info(...)` which is suppressed at the default log level. JEP-0013 requires warning severity for scrape timeout log entries. Split the log: emit a warning-level entry for timeout errors (drop the `V(1)` call, add `"severity", "warning"` key-value) and keep `V(1)` for other cancellation paths.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,469 @@
...
+	}
+	value := labeledCounterValue(t, mfs, metricsParseErrorsMetric, labelExporter, "sidekick")
+	if value < 1 {
+		t.Fatalf("%s{exporter=sidekick} = %v, want >= 1 body:\n%s", metricsParseErrorsMetric, value, body)
+	}
+
+	select {
+	case err := <-errCh:
+		if err != nil && err != io.EOF && status.Code(err) != codes.Canceled && status.Code(err) != codes.Unavailable {
+			t.Fatalf("stream goroutine: %v", err)
```
raballew: [MEDIUM] Multi-exporter concurrent fan-out is not tested through the HTTP endpoint. All end-to-end tests register exactly one exporter. Add `TestFanout_TwoExportersMergedOnSingleScrape` that registers two exporters with distinct identities and verifies a single `GET /metrics` response contains metric families with both `exporter="a"` and `exporter="b"` labels.

controller/internal/service/telemetry_service.go
```diff
@@ -276,12 +294,24 @@ func (s *TelemetryService) Start(ctx context.Context) error {
 
 	select {
 	case <-ctx.Done():
+		s.grpcReady.Store(false)
+		if httpShutdown != nil {
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			_ = httpShutdown(shutdownCtx)
+		}
 		srv.GracefulStop()
```
raballew: [MEDIUM] `srv.GracefulStop()` is called with no deadline. gRPC's `GracefulStop` blocks until all active RPCs complete but does not cancel existing stream contexts. A `MetricsStream` handler waiting on `stream.Recv()` from a network-partitioned exporter blocks indefinitely, preventing pod shutdown within `terminationGracePeriodSeconds`. Run `GracefulStop` in a goroutine with a `time.After(30 * time.Second)` fallback that calls `srv.Stop()` if the grace period expires.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      writeTimeout,
+		IdleTimeout:       5 * time.Minute,
+	}
+	go func() {
+		if err := srv.Serve(ln); err != nil && err != http.ErrServerClosed {
+			ctrl.Log.WithName("telemetry").Error(err, "metrics HTTP server stopped unexpectedly")
+		}
+	}()
```
raballew: [MEDIUM] `startMetricsHTTP` launches the HTTP server in a goroutine and drops non-`ErrServerClosed` errors after logging them. `Start()`'s select loop only watches the gRPC error channel, so an HTTP crash leaves `/metrics`, `/healthz`, and `/readyz` unreachable while gRPC stays alive, silently failing Kubernetes liveness and readiness probes. Thread an error channel back to `Start()` and handle it alongside the gRPC `errCh`.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,469 @@
...
+
+	mfs, err := svc.metricsRegistry.Gather()
+	if err != nil {
+		t.Fatalf("Gather: %v", err)
+	}
+	value := labeledCounterValue(t, mfs, metricsParseErrorsMetric, labelExporter, "sidekick")
+	if value < 1 {
+		t.Fatalf("%s{exporter=sidekick} = %v, want >= 1 body:\n%s", metricsParseErrorsMetric, value, body)
+	}
+
```
raballew: [MEDIUM] There is no test verifying that `/metrics` is not reachable on the gRPC port. Tests use an in-memory `bufconn` listener for gRPC, so a refactor that accidentally multiplexed the ports would pass all existing tests. Add a test binding the gRPC server on an ephemeral TCP port via `net.Listen("tcp", "127.0.0.1:0")`, making `GET /metrics` to that address, and asserting a connection error or non-200 response.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+	select {
+	case resp := <-reply:
+		if resp == nil {
+			return nil, errScrapeTimeout
+		}
+		return resp.GetMetricsText(), nil
+	case <-timer.C:
+		return nil, errScrapeTimeout
+	case <-ctx.Done():
+		return nil, ctx.Err()
```
raballew: [MEDIUM] The `case <-ctx.Done(): return nil, ctx.Err()` arm (HTTP request cancellation path) is never tested. No test cancels the context passed to `scrape` while a reply is pending. When this arm fires via `context.WithCancel`, the returned `context.Canceled` is not matched by `fanoutScrapes` and no counter is incremented. Add a unit test that creates a cancellable context, starts a `scrape` goroutine, cancels the context before the reply channel receives, and asserts `scrape` returns `context.Canceled`.

controller/internal/service/telemetry_service.go
```diff
@@ -260,9 +266,21 @@ func (s *TelemetryService) Start(ctx context.Context) error {
...
+		return fmt.Errorf("telemetry: metrics HTTP listen %s: %w", s.MetricsBindAddr, err)
+	}
+
 	srv := grpc.NewServer(
 		grpc.Creds(creds),
 		grpc.ChainUnaryInterceptor(recovery.UnaryServerInterceptor()),
+		grpc.ChainStreamInterceptor(recovery.StreamServerInterceptor()),
 	)
 	pb.RegisterTelemetryServiceServer(srv, s)
 	reflection.Register(srv)
```
raballew: [MEDIUM] `reflection.Register(srv)` enables gRPC server reflection on the production TLS listener without authentication. The bearer-token interceptor protecting `PushLogs` and `MetricsStream` does not gate reflection calls, so any TLS client that reaches port :9093 can enumerate all registered services, methods, and full protobuf schemas without a valid token. Remove `reflection.Register` from the production server, or gate it behind an explicit opt-in flag protected by the same auth interceptor.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+func (s *TelemetryService) recordMetricsParseError(exporter string, err error) {
+	if s.parseErrors != nil {
+		s.parseErrors.WithLabelValues(exporter).Inc()
+	}
+	ctrl.Log.WithName("telemetry").Error(err, "exporter metrics snapshot omitted", logFieldExporter, exporter)
+}
+
+func metricsHTTPEnabled(addr string) bool {
+	return addr != "" && addr != "0"
+}
```
raballew, line 58: Add `TestMetricsHTTPEnabled` with cases: `""` -> false, `"0"` -> false, `":8080"` -> true.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	var buf bytes.Buffer
+	if err := encodeMetricFamilies(&buf, families); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	w.Header().Set("Content-Type", string(expfmt.NewFormat(expfmt.TypeOpenMetrics)))
+	w.WriteHeader(http.StatusOK)
```
raballew: [MEDIUM] The two HTTP 500 error paths in `handleMetrics` (failed `Gather()` and failed `encodeMetricFamilies`) are not covered by any test. Add tests using `httptest.NewRecorder` with a deliberately broken registry to trigger the Gather 500 path, and an `io.Writer` that returns an error to trigger the encode 500 path.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,469 @@
...
+	var value float64
+	var found bool
+	for _, mf := range mfs {
+		if mf.GetName() == scrapeTimeoutsMetric || mf.GetName()+"_total" == scrapeTimeoutsMetric {
+			found = true
+			if len(mf.Metric) > 0 {
+				value = mf.Metric[0].GetCounter().GetValue()
+			}
+		}
+	}
```
raballew: Timeout counter assertions use `value < 1` (equivalent to >= 1) instead of `value != 1`. A double-increment bug would pass undetected.

controller/internal/service/telemetry_service.go
```diff
@@ -276,12 +294,24 @@ func (s *TelemetryService) Start(ctx context.Context) error {
...
 		if err := <-errCh; err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 			return err
 		}
 		return nil
 	case err := <-errCh:
+		s.grpcReady.Store(false)
+		if httpShutdown != nil {
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			_ = httpShutdown(shutdownCtx)
```
raballew: [LOW] Both shutdown exit paths in `Start()` discard the return value of `httpShutdown(shutdownCtx)` with `_ = ...`. If active `/metrics` connections are present at shutdown, `Shutdown` may return `context.DeadlineExceeded` silently. Log the error instead: `if err := httpShutdown(shutdownCtx); err != nil { logger.Error(err, "HTTP metrics server shutdown error") }`.

controller/internal/service/telemetry_identity.go
```diff
@@ -0,0 +1,62 @@
...
+	token, err := authentication.BearerTokenFromContext(ctx)
+	if err != nil {
+		return exporterIdentity{}, err
+	}
+	if s.Signer == nil {
+		return exporterIdentity{}, status.Error(codes.Internal, "telemetry signer is not configured")
+	}
+	subject, err := s.Signer.ParseSubject(token)
+	if err != nil {
+		return exporterIdentity{}, status.Errorf(codes.Unauthenticated, "invalid token: %v", err)
```
raballew, line 59: Could this expose JWT details that should stay private?

protocol/proto/jumpstarter/v1/telemetry.proto
```diff
@@ -6,13 +6,44 @@ package jumpstarter.v1;
...
+  string identity = 1; // Exporter CRD name (verified against the auth token by the server).
+}
+
+message MetricsScrapeResponse {
+  bytes metrics_text = 1;                // generate_latest() OpenMetrics output.
+  google.protobuf.Timestamp timestamp = 2;
+}
+
+// Telemetry → Exporter
```
raballew: [LOW] Lines 20 and 37 use Unicode right-arrow characters (U+2192) in `// Exporter -> Telemetry` and `// Telemetry -> Exporter`. The project requires textual representation; these characters also propagate to generated Go and Python stubs. Replace with ASCII `->` and regenerate stubs.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+func (s *TelemetryService) unregisterConn(c *metricsConn) {
+	s.stateMu.Lock()
+	defer s.stateMu.Unlock()
+	if s.conns == nil {
+		return
+	}
+	if cur, ok := s.conns[c.id.key()]; ok && cur == c {
+		delete(s.conns, c.id.key())
+	}
+}
```
raballew, line 89: Add a test that registers two `metricsConn` objects with the same identity sequentially, call `unregisterConn` on the first, and asserts the second remains registered.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+type metricsConn struct {
+	id       exporterIdentity
+	scrapeMu sync.Mutex
+	mu       sync.Mutex
+	send     func(*pb.MetricsStreamResponse) error
+	// pending is non-nil only while a scrape is waiting for a response.
+	pending chan *pb.MetricsScrapeResponse
+	done    <-chan struct{}
+	// started is closed when the Recv loop is running.
+	started chan struct{}
```
raballew: [LOW] `metricsConn.started` is created and closed on every live connection but has no production reader. Only the test helper `waitRegistered` consumes it, so every connection incurs a channel allocation and close with no production benefit. Remove `started` from `metricsConn` and replace `waitRegistered` with a polling loop on `svc.snapshotConns()`, or move the channel behind a test-only hook.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+	"io"
+	"sort"
+	"strings"
+
+	dto "github.com/prometheus/client_model/go"
+	"github.com/prometheus/common/expfmt"
+)
+
+const (
+	labelExporter   = "exporter"
```
raballew: [LOW] `const labelExporter = "exporter"` (Prometheus label name) and `const logFieldExporter = "exporter"` in `telemetry_service.go` both resolve to the identical string. A future rename that misses one would silently diverge the Prometheus label from the log field key. Define a single shared constant and reference it in both packages, or at minimum add a comment pairing them.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+	keys := s.ExemplarKeys
+	if len(keys) == 0 {
+		keys = DefaultExemplarKeys
+	}
+	return mergeConfig{
+		exporterName: name,
+		driverTypes:  setToMap(types),
+		exemplarKeys: setToMap(keys),
+	}
+}
```
raballew, line 95: Please test the fallback paths (`len(types) == 0` and `len(keys) == 0`).

raballew (changes requested):

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+	s.stateMu.Lock()
+	defer s.stateMu.Unlock()
+	if s.metricsRegistry != nil {
+		return
+	}
+	s.metricsRegistry = prometheus.NewRegistry()
+	s.scrapeTimeouts = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: scrapeTimeoutsMetric,
+		Help: "Exporter MetricsStream scrapes that exceeded scrapeTimeout.",
+	})
```
raballew: [CRITICAL] `scrapeTimeouts` is declared as a bare `prometheus.Counter` with no label vector, so the `{exporter}` label is missing. Every scrape timeout increments a single unlabeled counter, making per-exporter alerting queries return no data. Compare `parseErrors` on line 42, which correctly uses `prometheus.NewCounterVec` with `[]string{labelExporter}`. Change `scrapeTimeouts` to `*prometheus.CounterVec`, register it with `prometheus.NewCounterVec`, replace `s.scrapeTimeouts.Inc()` with `s.scrapeTimeouts.WithLabelValues(c.id.name).Inc()`, update the field type in `telemetry_service.go` to `*prometheus.CounterVec`, and update the timeout tests to assert the labeled counter value.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+	s.initMetricsState()
+	id, err := s.authenticateExporter(stream.Context())
+	if err != nil {
+		return err
+	}
+
+	first, err := stream.Recv()
+	if err != nil {
+		return err
+	}
```
raballew: [HIGH] When the exporter calls `CloseSend()`, `stream.Recv()` returns `io.EOF`. Returning that directly from the gRPC server handler causes the runtime to emit `codes.Unknown` instead of a clean `codes.OK`. The metrics merge path at `metrics_merge.go:96` already handles this correctly with `errors.Is(err, io.EOF)`. Add the same guard at both `Recv()` sites (lines 103 and 137): check `errors.Is(err, io.EOF)` and return `nil` before the general `return err`.

controller/internal/service/telemetry_identity.go
```diff
@@ -0,0 +1,62 @@
...
+	}
```
raballew: [HIGH] The raw JWT parse error is formatted with `%v` directly into the gRPC status message, exposing internal strings like `"token is expired by 5m0s"` or `"crypto/rsa: verification error"` to callers. Log the raw error server-side at debug level and return a fixed `"invalid token"` message with no additional detail:
```go
if err != nil {
    ctrl.Log.WithName("telemetry").V(1).Info("token validation failed", "error", err)
    return exporterIdentity{}, status.Error(codes.Unauthenticated, "invalid token")
}
```

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+	families := mergeSnapshots(snaps, nil, s.mergeConfigFor, s.recordMetricsParseError)
+
+	if s.metricsRegistry != nil {
+		gathered, err := s.metricsRegistry.Gather()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		families = mergeSnapshots(nil, append(gathered, families...), s.mergeConfigFor, nil)
+	}
```
raballew: [HIGH] The second `mergeSnapshots` call passes `append(gathered, families...)` where `families` is exporter data. Inside `mergeSnapshots`, entries are stored with last-write-wins, so any exporter metric family whose name matches a hub-internal metric (e.g., `jumpstarter_scrape_timeouts_total`) completely replaces the hub's real value in the output. Reverse the slice order to `append(families, gathered...)` so hub metrics win, or strip exporter families whose names match hub-owned metric names before the second merge pass.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+	metricsParseErrorsMetric = "jumpstarter_metrics_parse_errors_total"
+)
+
+// DefaultDriverTypeEnum is the JEP-0013 default allowlist for driver_type.
+var DefaultDriverTypeEnum = []string{
+	"power", "storage", "network", "serial", "console", "video", "composite",
+}
+
+// DefaultExemplarKeys is the JEP-0013 default exemplar allowlist.
+var DefaultExemplarKeys = []string{"client", "lease_id"}
```
raballew: [MEDIUM] `DefaultExemplarKeys` is declared independently here and in `controller/internal/metrics` with identical values and doc comments. A future change to one will silently diverge from the other. `cmd/telemetry/main.go` imports `service.DefaultExemplarKeys` with no reference to the pre-existing `metrics.DefaultExemplarKeys`. Remove this declaration and import from `controller/internal/metrics` instead; if an import cycle prevents that, keep this as a thin re-export referencing the metrics package value.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+		if len(m.Histogram.Exemplars) > 0 {
+			filtered := make([]*dto.Exemplar, 0, len(m.Histogram.Exemplars))
+			for _, ex := range m.Histogram.Exemplars {
+				if fe := filterExemplar(ex, cfg.exemplarKeys); fe != nil {
+					filtered = append(filtered, fe)
+				}
+			}
+			m.Histogram.Exemplars = filtered
+		}
+	}
```
raballew: [MEDIUM] The histogram bucket and native exemplar filter paths in `applyMetric` have no test coverage. The existing `TestApplyMetric_*` test only constructs a `Counter` metric; the `Histogram` branch here is never reached. Add a test that passes a `dto.Metric` with `Histogram` set, including a bucket with both an allowed and a disallowed exemplar key and a populated `Exemplars` slice, and assert that disallowed keys are removed while allowed keys are retained.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+	for _, f := range extra {
+		if f == nil || f.GetName() == "" {
+			continue
+		}
+		byName[f.GetName()] = f
+	}
+	for _, snap := range snapshots {
+		if len(snap.text) == 0 {
+			continue
+		}
```
raballew: [MEDIUM] The empty-snapshot skip path in `mergeSnapshots` is untested. The spec states empty bytes should be treated as valid (zero metrics from that exporter), not as a parse error, but no test passes `exporterSnapshot{text: []byte{}}`. Add a test with an empty-text snapshot alongside a healthy one and assert `onParseError` is never called, no families from the empty exporter are returned, and the healthy exporter's families are present.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+
+	pb "github.com/jumpstarter-dev/jumpstarter/controller/internal/protocol/jumpstarter/v1"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var errScrapeTimeout = errors.New("metrics scrape timeout")
+
+const defaultScrapeTimeout = 7 * time.Second
```
raballew: [MEDIUM] `defaultScrapeTimeout` is unexported, so `main.go` repeats the `7*time.Second` literal as a magic number for the flag default. `DefaultDriverTypeEnum` and `DefaultExemplarKeys` from the same package are exported and used directly in `main.go`. Export this constant as `DefaultScrapeTimeout` and update `main.go` to use `service.DefaultScrapeTimeout` as the flag default.

controller/internal/service/metrics_stream.go
raballew: [MEDIUM] The `started` field on `metricsConn` is written only in production code and read only in the `waitRegistered` test helper. Every live exporter connection allocates a synchronization object that no production path reads. The underlying issue is that `registerConn` (line 122) runs before `close(conn.started)` (line 134), meaning a scrape arriving between those two points sends a request before the Recv loop is running. Moving `registerConn` to after `close(conn.started)` fixes the race; once fixed, `started` has no purpose and can be deleted.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+		pending := conn.pending
+		conn.pending = nil
+		conn.mu.Unlock()
+		if pending == nil {
+			continue
+		}
+		select {
+		case pending <- resp:
+		case <-done:
+			return nil
```
raballew: [MEDIUM] The `case <-done:` arm in this select is unreachable. `done` is only closed by `defer finish()`, which runs when `MetricsStream` returns, meaning execution cannot be inside the select at that point. Stream termination is already covered by `case <-stream.Context().Done()`. Remove the dead arm.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+		id:      id,
+		send:    stream.Send,
+		done:    done,
+		started: make(chan struct{}),
+	}
+	s.registerConn(conn)
+	defer s.unregisterConn(conn)
+
+	logger := log.FromContext(stream.Context()).WithName("telemetry").WithValues(
+		"exporter", id.name,
```
raballew: [MEDIUM] The string literal `"exporter"` is used here and on line 229, bypassing the `logFieldExporter` constant defined in `telemetry_service.go`. The constant is used correctly in `telemetry_service.go` and `telemetry_http.go`. Replace both occurrences with `logFieldExporter` to keep the log field name consistent.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+	case resp := <-reply:
+		if resp == nil {
+			return nil, errScrapeTimeout
+		}
+		return resp.GetMetricsText(), nil
+	case <-timer.C:
+		return nil, errScrapeTimeout
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-c.done:
```
raballew: [MEDIUM] The `case <-c.done:` branch returns `context.Canceled`, but no test closes `c.done` while a scrape is in-flight to verify this path. Add a test that starts `c.scrape(...)` in a goroutine, closes `done` immediately, and asserts the returned error satisfies `errors.Is(err, context.Canceled)`.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,469 @@
+/*
```
raballew: [MEDIUM] There is no test for `GET /metrics` with zero exporters connected. `fanoutScrapes` returns `nil` early when `len(conns) == 0`, but all existing fanout tests register at least one exporter before calling `/metrics`. Add `TestGetMetrics_ZeroExporters` using `startTestHub` with no exporters registered; assert the response is 200 OK, `Content-Type` contains `openmetrics-text`, and the body contains `# EOF` and is parseable.

controller/internal/service/metrics_stream_test.go
raballew: [MEDIUM] `TestFanout_UnparseableSnapshotIncrementsParseErrorsOnSameResponse` registers only one exporter ("sidekick") which sends malformed data. The spec requires that `GET /metrics` returns 200 with data from healthy exporters when one has a parse error. With a single exporter, the test cannot verify this coexistence requirement. Register a second healthy exporter and assert that its metric families appear in the response alongside the incremented parse-error counter.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+func metricsHTTPEnabled(addr string) bool {
+	return addr != "" && addr != "0"
+}
+
+func (s *TelemetryService) startMetricsHTTP() (func(context.Context) error, error) {
+	addr := s.MetricsBindAddr
+	if !metricsHTTPEnabled(addr) {
+		return nil, nil
+	}
+	s.initMetricsState()
```
raballew: [MEDIUM] The `metricsHTTPEnabled` helper and the early-return `(nil, nil)` branch when `MetricsBindAddr` is empty or `"0"` have no test coverage. Every test supplies `MetricsBindAddr: "127.0.0.1:0"`. Add tests for empty and `"0"` addresses that assert both the shutdown function and error are nil.

controller/cmd/telemetry/main.go
```diff
@@ -14,9 +14,10 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// jumpstarter-telemetry receives structured log entries from exporters and clients
-// via the PushLogs gRPC RPC and writes them to structured stdout for downstream
-// log shippers (Promtail, Grafana Alloy, Vector) to forward to Loki.
+// jumpstarter-telemetry reverse-scrapes exporter metrics via MetricsStream and
+// receives structured log entries via PushLogs. Logs are written to structured
+// stdout for downstream log shippers (Promtail, Grafana Alloy, Vector).
+// Loki push is a later Phase 3 PR.
```
raballew: [LOW] This sentence ("Loki push is a later Phase 3 PR.") is a project-management note about future work rather than a description of what the binary currently does. It will become either stale or misleading once shipping decisions are finalized. Remove the sentence; a GitHub issue link is more durable than inline prose for tracking planned work.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+	labelDriverType = "driver_type"
+	driverTypeOther = "other"
+
+	// scrapeTimeoutsMetric is incremented once per exporter that does not
+	// answer a MetricsStream scrape within scrapeTimeout (JEP-0013).
+	scrapeTimeoutsMetric = "jumpstarter_scrape_timeouts_total"
+
+	// metricsParseErrorsMetric is incremented once per exporter snapshot that
+	// MetricsStream delivered but OpenMetrics parse rejected (JEP-0013).
+	metricsParseErrorsMetric = "jumpstarter_metrics_parse_errors_total"
```
raballew: [LOW] The comment for `scrapeTimeoutsMetric` describes counter mechanics ("is incremented once per exporter...") that the Prometheus `Help` string in `telemetry_http.go` already covers. Trim to the domain fact, for example: `// scrapeTimeoutsMetric tracks per-exporter MetricsStream scrape timeouts (JEP-0013).`

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+		return nil, err
+	}
+
+	timer := time.NewTimer(timeout)
+	defer timer.Stop()
+	select {
+	case resp := <-reply:
+		if resp == nil {
+			return nil, errScrapeTimeout
+		}
```
raballew: [LOW] The `if resp == nil` guard is permanently false. The sender in `MetricsStream` guarantees a non-nil value before sending to `reply` (the nil-guard at line 141 in `MetricsStream` ensures this). Remove the dead branch.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,469 @@
...
+				break
+			}
+		}
+		if ready {
+			return
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+	t.Fatalf("timed out waiting for %d MetricsStream connections, have %d", n, len(svc.snapshotConns()))
+}
```
raballew: [LOW] `waitRegistered` polls `snapshotConns()` with 10 ms `time.Sleep` calls, but `conn.started` is already a deterministic signal closed by the server goroutine. Replace the polling loop with `<-c.started` (with a deadline context) for each connection and remove the `time.Sleep` calls to make the helper deterministic and faster.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+		Help: "Exporter MetricsStream snapshots omitted because OpenMetrics parse failed.",
+	}, []string{labelExporter})
+	s.metricsRegistry.MustRegister(s.scrapeTimeouts, s.parseErrors)
+}
+
+func (s *TelemetryService) recordMetricsParseError(exporter string, err error) {
+	if s.parseErrors != nil {
+		s.parseErrors.WithLabelValues(exporter).Inc()
+	}
+	ctrl.Log.WithName("telemetry").Error(err, "exporter metrics snapshot omitted", logFieldExporter, exporter)
```
raballew: [LOW] `recordMetricsParseError` logs "exporter metrics snapshot omitted" at error severity, but no test injects a test logger and asserts this string appears in the log output. Inject a test logger via `logf.IntoContext` in an integration-style test and assert the expected log message is emitted after delivering an unparseable snapshot.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+		http.Error(w, "grpc not ready\n", http.StatusServiceUnavailable)
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte("ok\n"))
+}
+
+func (s *TelemetryService) handleMetrics(w http.ResponseWriter, r *http.Request) {
+	s.initScrapeTimeouts()
+	snaps := s.fanoutScrapes(r.Context())
```
raballew: [LOW] The `/metrics` endpoint has no authentication. Any client that can reach `MetricsBindAddr` triggers `fanoutScrapes`, which fans out a scrape request to every connected exporter and blocks up to `scrapeTimeout` (default 7 s). Rapid unauthenticated requests cause continuous fan-out traffic with no throttling. Document that `MetricsBindAddr` must be bound to a non-public interface or protected by network policy. Optionally add a per-endpoint rate limit or request serialization.

controller/internal/service/telemetry_service.go
```diff
@@ -69,42 +69,48 @@ var reservedExtraFieldKeys = map[string]struct{}{
...
-	// BindAddr is the TCP address to listen on (e.g. ":9093").
+	// BindAddr is the TCP address to listen on for gRPC (e.g. ":9093").
 	BindAddr string
 
-	// Signer is used to validate bearer tokens on every PushLogs call.
-	// Tokens are issued by the controller from the same CONTROLLER_KEY seed,
-	// so the telemetry binary can verify them locally without a k8s client.
+	// MetricsBindAddr is the TCP address for HTTP GET /metrics, /healthz, and
+	// /readyz. Empty or "0" disables the HTTP server (tests). Production
+	// default is :8080 — a dedicated port, not multiplexed onto gRPC :9093.
```
raballew: [LOW] This comment contains a non-ASCII em-dash (U+2014) in `"// default is :8080 \u2014 a dedicated port..."`). Replace the em-dash with `--` or a plain comma to keep the source file ASCII-clean.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+		mf := new(dto.MetricFamily)
+		err := dec.Decode(mf)
+		if errors.Is(err, io.EOF) {
+			return out, nil
+		}
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, mf)
+	}
```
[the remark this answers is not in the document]
RoddieKieley: As the parser in question isn't available as per [prometheus common #812](https://github.com/prometheus/common/issues/812#issue-3318753457), JEP-0013 DD-3 sidecar implemented. As per commit [df81340c](https://github.com/RoddieKieley/jumpstarter/commit/df81340c5cc07e3780209819c071aeb258073f8c)

controller/internal/service/telemetry_http.go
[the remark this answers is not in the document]
RoddieKieley: Addressed via commit [b1e162ab](https://github.com/RoddieKieley/jumpstarter/commit/b1e162ab0709592d7b6bbd2ab9e60551ed2f5650)

controller/internal/service/metrics_stream.go
RoddieKieley: Addressed via commit [3d4ad505](https://github.com/RoddieKieley/jumpstarter/commit/3d4ad50575932d57f750fdc6bbccef6fc5094049)

controller/internal/service/metrics_stream.go
RoddieKieley: Addressed via commit [3d4ad505](https://github.com/RoddieKieley/jumpstarter/commit/3d4ad50575932d57f750fdc6bbccef6fc5094049/)

controller/internal/service/telemetry_http.go
RoddieKieley: Addressed via commit [3d4ad505](https://github.com/RoddieKieley/jumpstarter/commit/3d4ad50575932d57f750fdc6bbccef6fc5094049/)

controller/internal/service/metrics_stream.go
RoddieKieley: Addressed via commit [3d4ad505](https://github.com/RoddieKieley/jumpstarter/commit/3d4ad50575932d57f750fdc6bbccef6fc5094049/)

controller/internal/service/metrics_merge.go
RoddieKieley: Addressed via commit [3d4ad505](https://github.com/RoddieKieley/jumpstarter/commit/3d4ad50575932d57f750fdc6bbccef6fc5094049/)

controller/internal/service/telemetry_http.go
RoddieKieley: This was not intentional. Addressed via commit [f2dbbd6f](https://github.com/RoddieKieley/jumpstarter/commit/f2dbbd6f6b876feee3adbdf76eea5be5e56d229f/)

controller/internal/service/metrics_stream.go
RoddieKieley: Addressed via commit [8fc0cf6c](https://github.com/RoddieKieley/jumpstarter/commit/8fc0cf6c71b9104ac4a1ba17d544407dd9c81499/)

controller/internal/service/metrics_stream.go
RoddieKieley: Addressed via commit [a8d42561](https://github.com/RoddieKieley/jumpstarter/commit/a8d4256107d647054b53adf95b9c2272c75ceb45/)

controller/internal/service/telemetry_identity.go
RoddieKieley: There was a return passing the actual token - addressed via [cfac8bfa](https://github.com/RoddieKieley/jumpstarter/commit/cfac8bfa712dc17a8f50f9e71eeb5c5d231296c5/)

controller/internal/service/metrics_stream_test.go
RoddieKieley: Addressed via commit [3c1e2179](https://github.com/RoddieKieley/jumpstarter/commit/3c1e2179d99e210ca8840229943d6d3fbeb79e1d/)

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,261 @@
...
+		return resp, nil
```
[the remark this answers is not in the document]
RoddieKieley: This would require a JEP-0013 related change and is thus not in scope for this pass.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+
+func (c *metricsConn) scrape(ctx context.Context, timeout time.Duration) (*pb.MetricsScrapeResponse, error) {
+	c.scrapeMu.Lock()
+	defer c.scrapeMu.Unlock()
+
+	// stream.Send has no context and can block indefinitely if the exporter
+	// is not reading. Run it in the background so timeout, ctx, and conn
+	// teardown can still finish the scrape.
+	if !c.sendMu.TryLock() {
+		return nil, errScrapeTimeout
```
bkhizgiy, line 177: `Send` is synchronous, so if the exporter stops reading, could it block past `scrapeTimeout`? Maybe it would be safer to run it in a goroutine with a `sendMu` (or something similar), so timeout/teardown can still win.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+			if err != nil {
+				if errors.Is(err, errScrapeTimeout) || errors.Is(err, context.DeadlineExceeded) {
+					if s.scrapeTimeouts != nil {
+						s.scrapeTimeouts.WithLabelValues(c.id.name).Inc()
+					}
+				}
+				log.FromContext(ctx).WithName("telemetry").V(1).Info("exporter scrape omitted",
+					"exporter", c.id.name,
+					"error", err.Error(),
+				)
```
bkhizgiy, line 268: Should this be logged at the default level instead of `V(1)`? With the current verbosity, scrape timeouts/omissions won’t show up in the default pod logs, even though we increment the timeout counter. The JEP also mentions emitting a warning with the exporter when a scrape times out, so it might be useful to make this visible without increasing the log level.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+				return
+			}
+			mu.Lock()
+			snaps = append(snaps, exporterSnapshot{
+				name:     c.id.name,
+				text:     resp.GetMetricsText(),
+				families: resp.GetFamilies(),
```
bkhizgiy, line 275: Do we want to put any limit on the snapshot size here? IIRC we talked about limits in the JEP, and a buggy exporter could return a very large `metrics_text` or `families` response that we keep in memory before parsing/merging. probably not a blocker for now, but might be worth adding a per exporter limit, can be done as a follow up as well.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,148 @@
...
+		ReadHeaderTimeout: 10 * time.Second,
+		ReadTimeout:       30 * time.Second,
+		WriteTimeout:      writeTimeout,
+		IdleTimeout:       5 * time.Minute,
+	}
+	go func() {
+		if err := srv.Serve(ln); err != nil && err != http.ErrServerClosed {
+			ctrl.Log.WithName("telemetry").Error(err, "metrics HTTP server stopped unexpectedly")
+		}
+	}()
```
raballew: [HIGH] `srv.Serve(ln)` runs in a goroutine and any error other than `http.ErrServerClosed` is only logged. `Start()` has no way to know the HTTP server crashed, so `grpcReady` stays true while `/metrics`, `/healthz`, and `/readyz` silently stop responding. The gRPC path correctly propagates failures via `errCh`; the HTTP path has no equivalent. Consider returning an `httpErrCh chan error` from `startMetricsHTTP` and selecting on it alongside `errCh` and `ctx.Done` in `Start()`, calling `stopGRPCServer` and returning a wrapped error on unexpected HTTP exit.

controller/internal/service/metrics_stream.go
raballew: [MEDIUM] When `sendMu.TryLock()` fails because a prior send goroutine is still running, `scrape()` returns `errScrapeTimeout`. `doFanoutScrapes` (around line 260) treats this sentinel as a real timeout and increments `scrapeTimeouts`. A busy send is a distinct failure mode from an exporter not responding within the timeout, so this double-counts the counter. Consider introducing a separate sentinel (e.g. `errScrapeSendBusy`) and only incrementing `scrapeTimeouts` for genuine timeout errors, logging the busy case at a lower verbosity without incrementing.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,148 @@
...
+	ln, err := net.Listen("tcp", addr)
+	if err != nil {
+		return nil, err
+	}
+	s.metricsAddr = ln.Addr().String()
+
+	mux := http.NewServeMux()
+	mux.HandleFunc("/metrics", s.handleMetrics)
+	mux.HandleFunc("/healthz", s.handleHealthz)
+	mux.HandleFunc("/readyz", s.handleReadyz)
```
raballew: [MEDIUM] The HTTP server on `:8080` has no authentication middleware. Any pod with TCP access can hit `/metrics` and enumerate all connected exporter names and operational status via per-exporter labels. The gRPC server enforces TLS and bearer-token auth, but this endpoint is wide open. Consider adding a lightweight bearer token middleware, or at minimum document a required NetworkPolicy restricting port 8080 to authorized scrapers.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+			})
+			mu.Unlock()
+		}(c)
```
raballew: [MEDIUM] There is no size limit on `MetricsScrapeResponse` payloads beyond gRPC's default 4 MB per-message ceiling, and no explicit `grpc.MaxRecvMsgSize` is configured. During a fanout scrape all N exporter payloads are buffered in `snaps` simultaneously before any merge. An authenticated exporter sending near-max-size payloads can cause significant memory pressure. Consider adding `grpc.MaxRecvMsgSize(maxScrapePayloadBytes)` to `grpc.NewServer` and a length guard in `scrape()` that drops oversized payloads and increments a dedicated counter.

controller/internal/service/telemetry_service.go
```diff
@@ -260,9 +273,21 @@ func (s *TelemetryService) Start(ctx context.Context) error {
 		return fmt.Errorf("telemetry: listen %s: %w", s.BindAddr, err)
 	}
 
+	s.initMetricsState()
+	s.initScrapeTimeouts()
```
raballew: [MEDIUM] `initMetricsState` and `initScrapeTimeouts` are called defensively with nil-checks at five separate sites across three files. `handleMetrics` calls `initScrapeTimeouts` then `doFanoutScrapes` which calls it again, causing unnecessary `stateMu` contention. `initScrapeTimeouts` is defined in `telemetry_http.go` but called from `metrics_stream.go`. Consider initializing all runtime state once inside `Start()` using `sync.Once` and removing all other defensive call sites, adding a constructor or `Init()` method for tests that bypass `Start()`.

controller/internal/service/telemetry_identity.go
```diff
@@ -0,0 +1,67 @@
...
+func parseExporterSubject(subject string) (exporterIdentity, error) {
+	parts := strings.SplitN(subject, ":", 4)
+	if len(parts) != 4 || parts[0] != "exporter" {
+		return exporterIdentity{}, status.Errorf(codes.PermissionDenied, "token is not an exporter token")
+	}
+	if parts[1] == "" || parts[2] == "" {
+		return exporterIdentity{}, status.Errorf(codes.PermissionDenied, "token has incomplete exporter identity")
+	}
+	return exporterIdentity{namespace: parts[1], name: parts[2]}, nil
+}
```
raballew: [MEDIUM] `parseExporterSubject` has three distinct outcomes (PermissionDenied for wrong kind, PermissionDenied for incomplete identity, and success) but no direct unit test. All paths are exercised only indirectly through integration tests. This is security-sensitive token-parsing code where a regression in edge-case handling would be high-impact. A small table-driven test in `telemetry_identity_test.go` covering a valid subject, a client-kind subject, an empty-namespace subject, and an empty-name subject would close this gap.

controller/internal/service/telemetry_service.go
```diff
@@ -260,9 +273,21 @@ func (s *TelemetryService) Start(ctx context.Context) error {
...
+	s.grpcReady.Store(true)
```
raballew: [LOW] `s.grpcReady.Store(true)` is set here before the goroutine that calls `srv.Serve(lis)` is launched. Between these two points `/readyz` returns HTTP 200 but the gRPC server has not yet called `Accept()`. The OS-level listen queue mitigates this in practice, but the window is real. Moving `s.grpcReady.Store(true)` to just after the `go func()` launch would tighten the guarantee.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+	// streams are not concurrent-safe for Send.
+	sendMu sync.Mutex
+	mu     sync.Mutex
+	send   func(*pb.MetricsStreamResponse) error
+	// pending is non-nil only while a scrape is waiting for a response.
+	pending chan *pb.MetricsScrapeResponse
+	done    <-chan struct{}
+	// started is closed when the Recv loop is running.
+	started chan struct{}
+}
```
raballew: [LOW] `metricsConn` holds three mutexes (`scrapeMu`, `sendMu`, `mu`) with non-obvious interaction. `sendMu` is acquired via `TryLock()` in `scrape()` but released by `defer` inside a spawned goroutine, making the cross-goroutine unlock implicit. The locking invariant comment is several lines above the goroutine that relies on it. A struct-level comment documenting the locking order and cross-goroutine invariant would help, and renaming `mu` to something like `pendingMu` would clarify its scope.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+func (s *TelemetryService) doFanoutScrapes() []exporterSnapshot {
+	s.initMetricsState()
+	s.initScrapeTimeouts()
+	timeout := s.scrapeTimeoutDuration()
+	conns := s.snapshotConns()
+	if len(conns) == 0 {
+		return nil
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
raballew: [LOW] `doFanoutScrapes` receives the HTTP handler's request context but immediately discards it by creating a new `context.WithTimeout(context.Background(), timeout)`. The trade-off is correct because singleflight coalesces requests from multiple callers and the work cannot be tied to any single caller's lifetime, but there is no comment explaining the intentional drop. A brief inline comment would prevent future readers from thinking this is an oversight.

protocol/proto/jumpstarter/v1/telemetry.proto
```diff
@@ -6,13 +6,86 @@ package jumpstarter.v1;
...
+  }
+}
+
+message MetricsRegister {
+  string identity = 1; // Exporter CRD name (verified against the auth token by the server).
+}
+
+message MetricsScrapeResponse {
+  bytes metrics_text = 1;                // Optional generate_latest() OpenMetrics text (legacy / debug).
+  google.protobuf.Timestamp timestamp = 2; // When the snapshot was taken.
```
raballew: [LOW] `MetricsScrapeResponse` defines a `google.protobuf.Timestamp timestamp` field (proto field 2), but `GetTimestamp()` is never called in any production service code. `exporterSnapshot` has no timestamp field. If this field has no planned use, consider removing it and regenerating stubs. If it is reserved for future staleness detection, a comment in the proto saying so would prevent reviewers from flagging it as dead code.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,936 @@
...
+	if !strings.Contains(body, `driver_type="other"`) {
+		t.Errorf("unknown driver_type must remap to other:\n%s", body)
+	}
+	select {
+	case err := <-errCh:
+		if err != nil && err != io.EOF && status.Code(err) != codes.Canceled && status.Code(err) != codes.Unavailable {
+			t.Fatalf("stream goroutine: %v", err)
+		}
+	default:
+	}
```
raballew: [LOW] The exporter goroutine error drain at the end of integration tests uses a `default` arm, which fires immediately if the goroutine has not sent yet. Errors that arrive after cleanup runs are silently dropped. A deliberately injected bug that makes the exporter goroutine return a non-EOF error would still pass these tests. Replacing the `default` arm with a `t.Cleanup` that blocks on `errCh` with a short timeout (e.g. 500ms) would catch post-cleanup errors. The same pattern applies at lines 614-627, 680-694, and 777-787.

controller/internal/service/metrics_stream_test.go
```diff
@@ -0,0 +1,936 @@
...
+	start := time.Now()
+	stopGRPCServer(gs, grace)
+	elapsed := time.Since(start)
+	if elapsed < grace/2 {
+		t.Fatalf("stop returned in %v; stuck stream should wait for the grace deadline", elapsed)
+	}
+	if elapsed > time.Second {
+		t.Fatalf("stopGRPCServer hung for %v", elapsed)
+	}
+}
```
raballew: [LOW] `TestStopGRPCServer_UnblocksStuckMetricsStream` only checks server-side shutdown timing. It does not assert that `stream.Recv()` on the client side returned an error after `stopGRPCServer` returned. A server that crashed without cleanly signaling the client would also satisfy the timing assertions. Adding an assertion that `stream.Recv()` returns a non-nil error after shutdown would make the test verify the full behavior.

controller/cmd/telemetry/main.go
```diff
@@ -55,9 +57,36 @@ var (
...
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p != "" {
+			out = append(out, p)
+		}
+	}
+	return out
+}
+
```
raballew: [LOW] `splitCSV` has non-trivial edge-case behavior (empty string returns nil, whitespace-only tokens are dropped, consecutive commas produce no empty entries) but lives in `package main`, making it untestable without a binary-level integration test. There are no `*_test.go` files in `controller/cmd/telemetry/`. Moving this function (or an equivalent `ParseCSV`) into `package service` alongside `DefaultDriverTypeEnum` and `DefaultExemplarKeys` would allow a table-driven unit test covering the edge cases.

python/packages/jumpstarter/jumpstarter/metrics/metrics_test.py
```diff
@@ -200,6 +200,72 @@ def test_exemplars_include_client_and_lease_id():
...
+        "jumpstarter_operations_total",
+        "jumpstarter_operation_duration_seconds",
+    }
+
+
+def test_families_from_collector_normalizes_info_and_enum():
+    from prometheus_client import CollectorRegistry, Enum, Info
+    from jumpstarter_protocol import telemetry_pb2
+
+    from jumpstarter.metrics.families import families_from_collector
```
raballew: [LOW] The import block here is unsorted (ruff I001). `from jumpstarter_protocol import telemetry_pb2` appears before `from jumpstarter.metrics.families import families_from_collector`, reversing the expected third-party after local order. Running `ruff check --fix` on the file will sort these automatically.

controller/internal/service/metrics_families_test.go
```diff
@@ -0,0 +1,345 @@
...
+	}
+	if len(families[0].Metric) != 1 {
+		t.Fatalf("created sample should be dropped, got %d metrics", len(families[0].Metric))
+	}
+	if families[0].Metric[0].GetCounter().GetExemplar() == nil {
+		t.Fatal("exemplar dropped during canonicalize")
+	}
+}
+
+func TestMergeSnapshots_PrefersFamiliesOverUnparseableText(t *testing.T) {
```
raballew: [LOW] `TestMergeSnapshots_PrefersFamiliesOverUnparseableText` contains a typo: "Unparseable" should be "Unparsable". The `typos` checker flags this. Since this file is new on the branch, renaming to `TestMergeSnapshots_PrefersFamiliesOverUnparsableText` is the right fix.

controller/internal/service/metrics_families.go
```diff
@@ -0,0 +1,348 @@
...
+
+func skipCreatedSample(sample string) bool {
+	return strings.HasSuffix(sample, "_created")
+}
+
+func dtoLabels(in []*pb.MetricsLabel) []*dto.LabelPair {
+	if len(in) == 0 {
+		return nil
+	}
+	out := make([]*dto.LabelPair, 0, len(in))
```
raballew: [LOW] `dtoLabels` accepts label names from the `families` proto field verbatim and the expfmt encoder writes them literally. A label name containing `}`, `"`, or `\n` produces corrupt OpenMetrics output that downstream Prometheus may misparse. The raw-text path is protected (malformed input is caught by `parseMetricFamilies` and the snapshot is dropped), but the proto path lacks this defense. Adding a validation step in `dtoLabels` rejecting names not matching `^[a-zA-Z_][a-zA-Z0-9_]*$` and returning an error from `dtoFromProtoFamilies` for non-conforming labels would close the gap.

controller/cmd/telemetry/main.go
raballew: [LOW] The package comment includes `// Loki push is a later Phase 3 PR.` This is a roadmap note that will become stale once the follow-up PR ships. Removing the line keeps the package doc accurate long-term. If the absence of Loki push needs to be signaled at the implementation point, a `// TODO(jep-0013-phase3-d):` inline in the `PushLogs` body would be more precise.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,148 @@
...
+	)
+	return srv.Shutdown, nil
+}
+
+func (s *TelemetryService) handleHealthz(w http.ResponseWriter, _ *http.Request) {
+	w.WriteHeader(http.StatusOK)
+	_, _ = w.Write([]byte("ok\n"))
+}
+
+// handleReadyz is 200 once the gRPC listener is bound.
```
raballew: [LOW] The first sentence of the `handleReadyz` comment (`// handleReadyz is 200 once the gRPC listener is bound.`) restates what the function body already makes obvious. The subsequent sentences explaining why Loki reachability and exporter checks are intentionally omitted are genuinely useful context worth keeping. Removing only the first sentence would leave the meaningful part intact.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+
+var errScrapeTimeout = errors.New("metrics scrape timeout")
+
+const defaultScrapeTimeout = 7 * time.Second
+
+// scrapeFlightKey is the singleflight key for overlapping /metrics fan-outs.
+const scrapeFlightKey = "metricsstream-fanout"
+
+// metricsConn is one authenticated MetricsStream. Scrapes are serialized
+// because MetricsScrapeRequest has no scrape id (JEP-0013).
```
raballew: [LOW] The first sentence of the `metricsConn` doc comment (`// metricsConn is one authenticated MetricsStream.`) adds nothing beyond what the type name already conveys. The second sentence (`Scrapes are serialized because MetricsScrapeRequest has no scrape id (JEP-0013).`) is a genuine non-obvious WHY explanation worth keeping. Dropping the first sentence would leave only the informative part.

controller/internal/service/telemetry_service.go
```diff
@@ -69,42 +74,50 @@ var reservedExtraFieldKeys = map[string]struct{}{
...
+	// default is :8080 — a dedicated port, not multiplexed onto gRPC :9093.
+	MetricsBindAddr string
+
+	// ScrapeTimeout is the fan-out wait for MetricsStream responses (JEP default 7s).
+	ScrapeTimeout time.Duration
+
+	// DriverTypeEnum allowlist; unknown driver_type values are remapped to "other".
+	DriverTypeEnum []string
+
+	// ExemplarKeys allowlist applied on the telemetry merge path.
```
raballew: [LOW] The `ExemplarKeys` field comment (`// ExemplarKeys allowlist applied on the telemetry merge path.`) is vague and does not document the key behavior: exemplar labels not in this list are stripped during metrics merge. Something like `// Exemplar labels not in this list are stripped during metrics merge.` would make the field's effect self-evident.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,281 @@
...
+	// MetricsStream delivered but OpenMetrics parse rejected (JEP-0013).
+	metricsParseErrorsMetric = "jumpstarter_metrics_parse_errors_total"
+)
+
+// hubMetricNames are reserved telemetry-hub series. Exporter snapshots of the
+// same name are dropped so a CounterVec with no observations yet cannot be
+// replaced by a spoofed family.
+var hubMetricNames = []string{scrapeTimeoutsMetric, metricsParseErrorsMetric}
+
+// DefaultDriverTypeEnum is the JEP-0013 default allowlist for driver_type.
```
raballew: [LOW] The `DefaultDriverTypeEnum` comment (`// DefaultDriverTypeEnum is the JEP-0013 default allowlist for driver_type.`) is a near-verbatim expansion of the identifier and omits the key behavior: values not present in the list are remapped to "other" during metrics merge. Replacing with `// Values not present are remapped to "other" during metrics merge.` would document the actual effect.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,281 @@
...
+var DefaultDriverTypeEnum = []string{
+	"power", "storage", "network", "serial", "console", "video", "composite",
+}
+
+// DefaultExemplarKeys is the JEP-0013 default exemplar allowlist.
```
raballew: [LOW] The `DefaultExemplarKeys` comment (`// DefaultExemplarKeys is the JEP-0013 default exemplar allowlist.`) restates the identifier and omits the fallback behavior: this value is used when `ExemplarKeys` on `TelemetryService` is nil or empty. Something like `// Exemplar labels not in this list are stripped during metrics merge. Used when ExemplarKeys is nil or empty.` would document both the effect and when this default applies.

controller/internal/service/telemetry_service.go
```diff
@@ -69,42 +74,50 @@ var reservedExtraFieldKeys = map[string]struct{}{
...
 
-	// Signer is used to validate bearer tokens on every PushLogs call.
-	// Tokens are issued by the controller from the same CONTROLLER_KEY seed,
-	// so the telemetry binary can verify them locally without a k8s client.
+	// MetricsBindAddr is the TCP address for HTTP GET /metrics, /healthz, and
+	// /readyz. Empty or "0" disables the HTTP server (tests). Production
+	// default is :8080 — a dedicated port, not multiplexed onto gRPC :9093.
+	MetricsBindAddr string
+
+	// ScrapeTimeout is the fan-out wait for MetricsStream responses (JEP default 7s).
```
raballew: [LOW] The `ScrapeTimeout` field comment (`// ScrapeTimeout is the fan-out wait for MetricsStream responses (JEP default 7s).`) rephrases "timeout" as "fan-out wait" without adding information, and the 7s default is already documented at the call site in `main.go`. The field type (`time.Duration`) and name make the intent clear; removing the comment would reduce noise.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+	id       exporterIdentity
+	scrapeMu sync.Mutex
+	// sendMu is held for the lifetime of one stream.Send. A scrape that
+	// times out while Send is blocked must not start another Send: gRPC
+	// streams are not concurrent-safe for Send.
+	sendMu sync.Mutex
+	mu     sync.Mutex
+	send   func(*pb.MetricsStreamResponse) error
+	// pending is non-nil only while a scrape is waiting for a response.
+	pending chan *pb.MetricsScrapeResponse
```
raballew: [LOW] The `pending` channel field requires a comment to explain when it is nil vs non-nil (`// pending is non-nil only while a scrape is waiting for a response.`). The need for this comment signals that the name is not expressive enough on its own. Renaming to something like `activeScrapeReply` would encode the lifecycle constraint in the name and make the comment unnecessary.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+	done    <-chan struct{}
+	// started is closed when the Recv loop is running.
+	started chan struct{}
```
raballew: [LOW] The `started` channel field requires a comment to explain the channel-closed-as-signal idiom (`// started is closed when the Recv loop is running.`). The name does not encode the closure semantics. Renaming to something like `recvLoopRunning` would make the semantics self-evident and eliminate the need for the comment. Usages in `MetricsStream` and the `waitRegistered` test helper would also need updating.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,282 @@
...
+	"google.golang.org/grpc/status"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var errScrapeTimeout = errors.New("metrics scrape timeout")
+
+const defaultScrapeTimeout = 7 * time.Second
+
+// scrapeFlightKey is the singleflight key for overlapping /metrics fan-outs.
+const scrapeFlightKey = "metricsstream-fanout"
```
raballew: [LOW] The `scrapeFlightKey` constant has a comment (`// scrapeFlightKey is the singleflight key for overlapping /metrics fan-outs.`) that restates what the constant name and its value `"metricsstream-fanout"` together already communicate. The call site `s.scrapeGroup.DoChan(scrapeFlightKey, ...)` makes the purpose obvious; the comment adds no additional information and can be removed.

raballew (approved):

raballew:
@RoddieKieley linters are still failing otherwise its good to go.

mangelajo (approved):

mangelajo:
fixed the linter/typos on top of @RoddieKieley as he's traveling.

mangelajo:
Rebase was necessary to catch go modernization changes, then applied new go linter fixes.

PR #1061, opened 2026-09-02
JEP-0013 Phase 3 - operator image

RoddieKieley:
## Summary

JEP-0013 Phase 3 PR B**: make the MetricsStream hub from PR A runnable in-cluster.

Depends on **PR A**: https://github.com/jumpstarter-dev/jumpstarter/pull/1060

Do not merge until A is on `main`, then rebase this branch onto `main`. Related PRs C–E will be linked here as they are opened.

- Build and copy `/telemetry` in `controller/Containerfile` and `make build` / `docker-build-ci` (operator already `Command: ["/telemetry"]`; `main` still only ships `manager` + `router`).
- Expose telemetry HTTP metrics on Service/container port **8080**; `/healthz` and `/readyz` probes on that port (replacing TCP probes on gRPC `:9093`).
- CR: `spec.telemetry.metrics.scrapeTimeout`, `driverTypeEnum`, `exemplarKeys`. Defaults match JEP-0013 (`7s`, the driver-type enum, `client,lease_id`). **No** `ServiceMonitor` (Phase 5).
- Set `GRPC_TELEMETRY_ENDPOINT` so the telemetry process advertises the in-cluster Service DNS.

**TLS:** [#1023](https://github.com/jumpstarter-dev/jumpstarter/pull/1023) already landed telemetry TLS on `main`. This PR keeps that path (cert-manager **or** manual `spec.telemetry.grpc.tls.certSecret`, rolling-restart hash, non-fatal missing CA). It does **not** reimplement TLS.

**Lab:** in-cluster reverse-scrape of `GET /metrics` on Service port 8080 worked (port-forward). After driver ops, Python OpenMetrics exemplars still fail Go `parseMetricFamilies`; the hub from A omits that snapshot and increments `jumpstarter_metrics_parse_errors_total`. `spec.telemetry.metrics.exemplarKeys` only allowlists keys on the merge path **after** a successful parse — it does not fix exemplar decode. That remains A / JEP DD-3, not this PR.

## DEMO

The lab walkthrough for the stacked Phase 3 work (A+B+C+D+E plus lab-only Route) is on the fork demo branch, not this PR:

- [DEMO.md](https://github.com/RoddieKieley/jumpstarter/blob/jep-0013-phase3-demo/DEMO.md)
- [jep-0013-phase3-demo.cast](https://github.com/RoddieKieley/jumpstarter/blob/jep-0013-phase3-demo/jep-0013-phase3-demo.cast)

See [MetricsStream during the lease](https://github.com/RoddieKieley/jumpstarter/blob/jep-0013-phase3-demo/DEMO.md#metricsstream-during-the-lease) for the DD-3 exemplar limitation observed on `:8080`.

## How this PR fits the series

```mermaid
flowchart TB
  A["PR A #1060: proto + hub"]
  B["PR B this PR: image + metrics port + scrape CR"]
  C["PR C: exporter MetricsStream client"]
  D["PR D: Loki HTTP push"]
  E["PR E: jmp PushLogs"]
  A --> B
  A --> C
  B --> D
  D --> E
```

| PR | Branch | Status |
|----|--------|--------|
| A | `jep-0013-phase3-metricsstream` | #1060  |
| **B** | `jep-0013-phase3-operator-image` | **This PR** |
| C | `jep-0013-phase3-exporter-metricsstream` | #1062  |
| D | `jep-0013-phase3-loki-push` | #1063  |
| E | `jep-0013-phase3-client-pushlogs` | #1064  |

Unique work vs PR A: the operator/image commit on this branch (`435eebee`). If GitHub shows A's commits, that is stacking against `main`; review that unique commit only.

## Out of scope

- MetricsStream protocol and parse-error omit (A / #1060)
- Exporter MetricsStream client and Register `identity` (C) — after `#1058`, identity must be `exporter_name`, not `Metadata.name`
- Loki / PushLogs (D, E), including exporter TFTP errors that only show up on the log path
- Advertising a Route/hostname for out-of-cluster exporters (lab-only on the demo branch)
- `ServiceMonitor` (Phase 5)

## NOTE

Open [#1027](https://github.com/jumpstarter-dev/jumpstarter/pull/1027) also touches telemetry operator files (log-ingest e2e).


controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+			return nil, errScrapeTimeout
+		}
+		return resp.GetMetricsText(), nil
+	case <-timer.C:
+		return nil, errScrapeTimeout
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-c.done:
+		return nil, context.Canceled
+	}
```
raballew: [HIGH] The `metricsConn` struct holds a single `pending` field with no generation counter or scrape ID. If scrape A times out and clears `c.pending = nil`, and scrape B then sets `c.pending = replyB`, a late in-flight response from scrape A arrives at the recv loop, reads `conn.pending = replyB`, and delivers stale A data into B's channel with no error logged or returned. The protocol comment itself acknowledges there is no scrape ID. To fix this, add a `uint64 generation` field incremented atomically when setting `pending`, and in the recv loop only deliver if the stored generation still matches.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -39,14 +40,23 @@ import (
...
 	grpcPortName               = "grpc"
+	metricsPortName            = "metrics"
 	telemetryCARequeueInterval = 30 * time.Second
 )
 
+// JEP-0013 default allowlists used when spec.telemetry.metrics omits a field.
+var (
+	defaultTelemetryDriverTypeEnum = []string{"power", "storage", "network", "serial", "console", "video", "composite"}
+	defaultTelemetryExemplarKeys   = []string{"client", "lease_id"}
+	defaultTelemetryScrapeTimeout  = 7 * time.Second
```
raballew: [HIGH] The three private variables at lines 55-57 mirror exported symbols in the controller module's `metrics_merge.go:44-50` and `metrics_stream.go:33`. Because those symbols live in an `internal` package they cannot be imported directly, but the duplication means a future change to the default driver-type allowlist requires coordinated edits in two packages across two modules. Moving the canonical defaults to a non-internal package such as `controller/api/telemetrydefaults` would give both sides a single source of truth.

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+	// before hub metrics are gathered into the same /metrics response.
+	families := mergeSnapshots(snaps, nil, s.mergeConfigFor, s.recordMetricsParseError)
+
+	if s.metricsRegistry != nil {
+		gathered, err := s.metricsRegistry.Gather()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+		families = mergeSnapshots(nil, append(gathered, families...), s.mergeConfigFor, nil)
```
raballew: [MEDIUM] In `mergeSnapshots`, the `extra` loop uses unconditional assignment `byName[f.GetName()] = f`. `handleMetrics` calls `mergeSnapshots(nil, append(gathered, families...), ...)` with hub metrics first and exporter metrics last. When `snapshots` is nil only the `extra` loop runs, so any exporter metric whose name collides with a hub metric (e.g. `jumpstarter_scrape_timeouts_total`) silently overwrites the hub counter. Reversing the order to `append(families, gathered...)` would make hub metrics win on collision.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+		for _, f := range families {
+			applyFamily(f, cfg)
+			existing, ok := byName[f.GetName()]
+			if !ok {
+				byName[f.GetName()] = f
+				continue
+			}
+			existing.Metric = append(existing.Metric, f.Metric...)
+		}
+	}
```
raballew: [MEDIUM] When `handleMetrics` calls `mergeSnapshots(nil, append(gathered, families...), s.mergeConfigFor, nil)`, the `cfgFor` callback is silently ignored because `snapshots` is nil and `cfgFor` is only invoked inside the snapshots loop. The `s.mergeConfigFor` argument is passed but never reached on this code path. Splitting `mergeSnapshots` into a snapshot-based variant and a simpler `mergeFamilySlices(a, b []*dto.MetricFamily)` for the hub-plus-exporter merge case would make the two call sites explicit about which logic they actually use.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+	for _, c := range s.conns {
+		out = append(out, c)
+	}
+	return out
+}
+
+// MetricsStream reverse-scrapes one exporter. The first message must be
+// MetricsRegister whose identity matches the bearer token.
+func (s *TelemetryService) MetricsStream(stream pb.TelemetryService_MetricsStreamServer) error {
+	s.initMetricsState()
```
raballew: [MEDIUM] `initMetricsState()` and `initScrapeTimeouts()` are called defensively from 8 call sites across 3 files. Each hot-path call acquires `stateMu` to check whether fields are already non-nil, and a single `/metrics` request incurs two redundant lock acquisitions. Reading `Start()` alone cannot reveal the authoritative initialization moment. A `NewTelemetryService(cfg TelemetryConfig)` constructor that eagerly initializes `conns`, `metricsRegistry`, `scrapeTimeouts`, and `parseErrors` would eliminate both the defensive calls and the repeated locking.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_test.go
```diff
@@ -1066,3 +1083,39 @@ var _ = Describe("telemetryCANeedsRequeue", func() {
...
 })
+
+func hasEnv(c corev1.Container, name string) bool {
+	for _, e := range c.Env {
+		if e.Name == name {
+			return true
+		}
+	}
+	return false
+}
```
raballew: [MEDIUM] `hasEnv` is defined at line 1087 but has zero callers in the entire test package. All env-presence checks use `Expect(envNames).To(HaveKey(...))` directly. `envValue` (defined at line 1096) is called in `telemetry_phase3_test.go`, suggesting `envValue` superseded `hasEnv` without the original being removed. Lines 1087-1094 can be deleted.

controller/cmd/telemetry/main.go
```diff
@@ -55,9 +57,36 @@ var (
...
+	out := make([]string, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		if p != "" {
+			out = append(out, p)
+		}
+	}
+	return out
+}
+
```
raballew: [MEDIUM] `splitCSV` is the only function that converts `--driver-type-enum` and `--exemplar-keys` flag strings into the slices `TelemetryService` consumes, and `controller/cmd/telemetry/` contains no `*_test.go` file at all. Edge cases like empty string returning nil vs. empty slice, trailing comma, and whitespace-only entries are untested. A table-driven `TestSplitCSV` in `controller/cmd/telemetry/main_test.go` covering those cases would close this gap.

controller/deploy/operator/internal/controller/jumpstarter/telemetry.go
```diff
@@ -293,6 +309,29 @@ func (r *JumpstarterReconciler) reconcileTelemetryService(ctx context.Context, j
...
+	if len(keys) == 0 {
+		keys = defaultTelemetryExemplarKeys
+	}
+	return []string{
+		fmt.Sprintf("--grpc-bind=:%d", telemetryPort),
+		fmt.Sprintf("-metrics-bind-address=:%d", telemetryMetricsPort),
+		fmt.Sprintf("-scrape-timeout=%s", timeout),
+		fmt.Sprintf("-driver-type-enum=%s", strings.Join(enum, ",")),
+		fmt.Sprintf("-exemplar-keys=%s", strings.Join(keys, ",")),
+	}
```
raballew, line 332: Mixed use of dash and double-dash

controller/internal/service/telemetry_http.go
```diff
@@ -0,0 +1,145 @@
...
+func (s *TelemetryService) recordMetricsParseError(exporter string, err error) {
+	if s.parseErrors != nil {
+		s.parseErrors.WithLabelValues(exporter).Inc()
+	}
+	ctrl.Log.WithName("telemetry").Error(err, "exporter metrics snapshot omitted", logFieldExporter, exporter)
+}
+
+func metricsHTTPEnabled(addr string) bool {
+	return addr != "" && addr != "0"
+}
```
raballew: [MEDIUM] `metricsHTTPEnabled` returns false for `""` and `"0"`, where `"0"` is the documented sentinel to disable the HTTP server. All existing tests that set `MetricsBindAddr` use `"127.0.0.1:0"`, so the disabled-path guard is never exercised. If the guard were accidentally deleted or inverted, all current tests would still pass. A `TestMetricsHTTPEnabled` covering `""`, `"0"`, `"127.0.0.1:0"`, and `":8080"` would catch regressions, and an integration test setting `MetricsBindAddr = "0"` then calling `startMetricsHTTP` and asserting both return values are nil would complete the coverage.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+			return nil
+		case <-stream.Context().Done():
+			return stream.Context().Err()
+		}
+	}
+}
+
+func (c *metricsConn) scrape(ctx context.Context, timeout time.Duration) ([]byte, error) {
+	c.scrapeMu.Lock()
+	defer c.scrapeMu.Unlock()
```
raballew: [LOW] `c.scrapeMu.Lock()` is called unconditionally. The `case <-ctx.Done()` branch is only reachable after the mutex is already held, so a goroutine waiting for the lock cannot exit when the HTTP request context expires, causing `wg.Wait()` in `fanoutScrapes` to block for the full first scrape duration. Adding `select { case <-ctx.Done(): return nil, ctx.Err(); default: }` immediately before the lock acquisition would allow early exit.

controller/internal/service/telemetry_service.go
```diff
@@ -276,12 +294,24 @@ func (s *TelemetryService) Start(ctx context.Context) error {
...
 		if err := <-errCh; err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 			return err
 		}
 		return nil
 	case err := <-errCh:
+		s.grpcReady.Store(false)
+		if httpShutdown != nil {
+			shutdownCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+			defer cancel()
+			_ = httpShutdown(shutdownCtx)
```
raballew: [LOW] Both shutdown paths use `_ = httpShutdown(shutdownCtx)`. The shutdown context has a 5-second deadline, but the HTTP `WriteTimeout` is `scrapeTimeoutDuration() + 15*time.Second` (22 s at default). An in-flight scrape lasting more than 5 s causes `http.Server.Shutdown` to return `context.DeadlineExceeded`, which is silently dropped. Logging the error and setting the shutdown timeout to at least `s.scrapeTimeoutDuration() + 2*time.Second` would surface this failure mode.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -325,6 +329,22 @@ type TelemetryGRPCConfig struct {
...
+	ExemplarKeys []string `json:"exemplarKeys,omitempty"`
+
+	// Allowed driver_type label values. Unlisted types are remapped to "other".
+	// +kubebuilder:default={"power","storage","network","serial","console","video","composite"}
+	DriverTypeEnum []string `json:"driverTypeEnum,omitempty"`
+
+	// Max wait for parallel exporter MetricsStream responses during a /metrics fan-out.
+	// Should be lower than the Prometheus scrape_timeout.
+	// +kubebuilder:default="7s"
+	ScrapeTimeout *metav1.Duration `json:"scrapeTimeout,omitempty"`
```
raballew, line 345: Also set an upper limit to avoid keeping connections open for very long times.

controller/internal/service/telemetry_service.go
```diff
@@ -260,9 +266,21 @@ func (s *TelemetryService) Start(ctx context.Context) error {
...
+		s.grpcReady.Store(false)
+		_ = lis.Close()
+		return fmt.Errorf("telemetry: metrics HTTP listen %s: %w", s.MetricsBindAddr, err)
+	}
+
 	srv := grpc.NewServer(
 		grpc.Creds(creds),
 		grpc.ChainUnaryInterceptor(recovery.UnaryServerInterceptor()),
+		grpc.ChainStreamInterceptor(recovery.StreamServerInterceptor()),
 	)
```
raballew, line 284: Adding `grpc.MaxRecvMsgSize` to the server options would provide an explicit limit especially if you are running 1000s of exporters that are scraped concurrently (each with up to 4 MB per message). Otherwise this might cause some OOM killed processes.

controller/internal/service/metrics_merge.go
```diff
@@ -0,0 +1,246 @@
...
+			continue
+		}
+		if strings.TrimSpace(lp.GetValue()) == "" {
+			continue
+		}
+		kept = append(kept, lp)
+	}
+	if len(kept) == 0 {
+		return nil
+	}
```
raballew: [LOW] `filterExemplar` returns `nil` when `len(kept) == 0`, but no test reaches this branch. The only exemplar filtering test provides a mix of allowlisted and non-allowlisted keys so `len(kept)` is always positive. A subtest that constructs a metric whose exemplar contains only non-allowlisted keys and asserts `m.Counter.GetExemplar() == nil` after `applyMetric` would close this gap.

controller/internal/service/metrics_stream.go
```diff
@@ -0,0 +1,241 @@
...
+		if resp == nil {
+			return nil, errScrapeTimeout
+		}
+		return resp.GetMetricsText(), nil
+	case <-timer.C:
+		return nil, errScrapeTimeout
+	case <-ctx.Done():
+		return nil, ctx.Err()
+	case <-c.done:
+		return nil, context.Canceled
```
raballew: [LOW] The `case <-c.done` branch returns `context.Canceled` when the exporter stream terminates while a scrape reply is pending, but `done` is never closed in any existing test. A `TestMetricsConn_ScrapeReturnsCanceledOnDisconnect` that constructs a `metricsConn`, starts `c.scrape` in a goroutine, closes `done`, and asserts the returned error satisfies `errors.Is(err, context.Canceled)` would cover this path.

controller/internal/service/telemetry_service.go
```diff
@@ -69,42 +69,48 @@ var reservedExtraFieldKeys = map[string]struct{}{
...
-	claimedNamespace := parts[1]
-	claimedName := parts[2]
-	if claimedNamespace == "" || claimedName == "" {
-		return nil, status.Errorf(codes.PermissionDenied, "token has incomplete exporter identity")
-	}
+	claimedNamespace := id.namespace
+	claimedName := id.name
 
 	// Use context-based logger so tests can inject their own via logf.IntoContext.
 	logger := log.FromContext(ctx).WithName("telemetry")
```
raballew, line 116: `entry.Message`, `entry.Component`, `entry.Severity`, `entry.Lease`, `entry.Client`, `entry.Operation`, `entry.Result`, `entry.DriverType` have no size bound. Should we truncate them too?

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
```diff
@@ -325,6 +329,22 @@ type TelemetryGRPCConfig struct {
...
+// TelemetryMetricsConfig configures telemetry /metrics reverse-scrape behavior.
+type TelemetryMetricsConfig struct {
+	// Allowlist of keys to include in Prometheus exemplars. Unlisted keys are omitted.
+	// +kubebuilder:default={"client","lease_id"}
+	ExemplarKeys []string `json:"exemplarKeys,omitempty"`
+
+	// Allowed driver_type label values. Unlisted types are remapped to "other".
+	// +kubebuilder:default={"power","storage","network","serial","console","video","composite"}
+	DriverTypeEnum []string `json:"driverTypeEnum,omitempty"`
+
```
raballew, line 341: A user with CR write access can set either list to an arbitrarily large array for `ExemplarKeys []string` and `DriverTypeEnum []string`. Adding `// +kubebuilder:validation:MaxItems=<insert a good limit here>` and `// +kubebuilder:validation:MaxLength=<insert a good limit here>` per item would enforce a bound.

controller/deploy/operator/internal/controller/jumpstarter/telemetry_phase3_test.go
```diff
@@ -0,0 +1,158 @@
...
+			},
+			Data: map[string][]byte{"tls.crt": []byte(testPEM)},
+		})).To(Succeed())
+
+		r := &JumpstarterReconciler{Client: k8sClient, Scheme: k8sClient.Scheme()}
+		cfg, err := r.buildConfig(ctx, js)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(cfg.Telemetry).NotTo(BeNil())
+		Expect(cfg.Telemetry.Enabled).To(BeTrue())
+		Expect(cfg.Telemetry.Endpoint).To(Equal(telemetryEndpointFor(crNamespace)))
```
raballew: [LOW] Both assertions compare against `telemetryEndpointFor(...)`, the same internal function the production code calls. If `telemetryEndpointFor` produced a malformed DNS name, both the actual and expected values would be wrong in the same way and the test would pass. The separate literal test in `telemetry_test.go:632-637` partially mitigates this, but replacing the expected value on line 82 with `"jumpstarter-telemetry.jumpstarter-lab.svc:9093"` and on line 136 with `"jumpstarter-telemetry." + crNamespace + ".svc:9093"` would make each assertion genuinely independent.

raballew (commented):
Just a few comments for now. Waiting for merge of the first PR.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
mangelajo: Could be worth limiting to 16 for example? /8 default + 8 custom, more than enough.

controller/deploy/operator/api/v1alpha1/jumpstarter_types.go
mangelajo: 2 minutes max? (and sounds excessive)

PR #1067, opened 2026-09-03
feat(controller): have exporter request a new token when about to expire

bennyz:
- Add RotateToken RPC to ControllerService in protocol/proto/jumpstarter/v1/jumpstarter.proto.
- Implement RotateToken in ControllerService to generate a new signed token and update the exporter Secret.
- Regenerate jumpstarter.v1 protobuf bindings for Go and Python.
- Add token_refresh helper to decode JWT claims and calculate renewal lead times and sleep intervals.
- Add _token_refresh_loop in Exporter to automatically request a new token before expiry.


controller/internal/service/controller_service.go
```diff
@@ -342,6 +342,65 @@ func (s *ControllerService) GetServiceEndpoints(
...
+	var expiry *timestamppb.Timestamp
+	if claims.ExpiresAt != nil {
+		expiry = timestamppb.New(claims.ExpiresAt.Time)
+	}
+
+	return &pb.RotateTokenResponse{
+		Token:  token,
+		Expiry: expiry,
+	}, nil
+}
```
raballew: [HIGH] `RotateToken` never fires a `CredentialCreated` Kubernetes event. `ControllerService` has no `Recorder` field, and the reconciler only emits this event when `prevCredential == nil && exporter.Status.Credential != nil`, which is the initial creation path only. Every successful rotation goes unrecorded in cluster event history. Fix: add a `Recorder record.EventRecorder` field to `ControllerService`, wire it in `main.go`, and call `s.Recorder.Eventf(exporter, corev1.EventTypeNormal, "CredentialCreated", "Token rotated for exporter %s/%s", exporter.Namespace, exporter.Name)` after the successful `Patch`. Guard with `if s.Recorder != nil` to keep unit tests passing without a recorder.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -601,16 +620,103 @@ async def _setup_telemetry(self) -> None:
...
+            response: jumpstarter_pb2.RotateTokenResponse = await controller.RotateToken(
+                jumpstarter_pb2.RotateTokenRequest(),
+                timeout=_RPC_TIMEOUT,
+            )
+
+        new_token = response.token
+        self.token = new_token
+
+        if self._telemetry_handler is not None:
+            self._telemetry_handler.token = new_token
```
raballew: [HIGH] `rotate_token()` sets `self._telemetry_handler.token = new_token`, but `TelemetryLogHandler` stores its credential as `self._token` (private) and reads it via `self._token` when building `Bearer` headers. The public `.token` attribute set here is a shadow attribute that is never read; all subsequent `PushLogs` calls continue using the pre-rotation token. Fix: expose a `@token.setter` on `TelemetryLogHandler` that writes to `self._token`, or change the assignment to `self._telemetry_handler._token = new_token` with an explanatory comment.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -601,16 +620,103 @@ async def _setup_telemetry(self) -> None:
...
+                    break
+                except grpc.aio.AioRpcError as e:
+                    if e.code() == grpc.StatusCode.UNIMPLEMENTED:
+                        logger.warning(
+                            "Controller does not support token rotation (UNIMPLEMENTED); disabling automatic refresh"
+                        )
+                        return
+                    elif e.code() in (grpc.StatusCode.UNAUTHENTICATED, grpc.StatusCode.PERMISSION_DENIED):
+                        logger.error("Failed to rotate token (%s): %s", e.code(), e.details())
+                        return
```
raballew: [HIGH] When `RotateToken` returns `UNAUTHENTICATED` or `PERMISSION_DENIED`, `_token_refresh_loop` does a bare `return`. The anyio task group treats this as successful completion and does not cancel peer tasks. The exporter keeps running with an expired in-memory token indefinitely, acting as a zombie. Fix: replace the bare `return` with `raise RuntimeError("Token expired and rotation denied by controller; exporter cannot recover")` so the task group propagates the failure and terminates the process.

controller/internal/service/controller_service_test.go
```diff
@@ -2383,3 +2386,116 @@ func TestStatusResponseIncludesLeaseContext(t *testing.T) {
...
+
+		_, err := svc.RotateToken(context.Background(), &pb.RotateTokenRequest{})
+		if err == nil {
+			t.Fatal("expected error with missing secret")
+		}
+		if status.Code(err) != codes.Internal {
+			t.Fatalf("expected Internal error, got %v", err)
+		}
+	})
+}
```
raballew: [HIGH] Every other exporter-authenticated endpoint in this service (`Register`, `Dial`, `Status`) has a companion test that asserts authentication failure returns `codes.Unauthenticated`. `TestControllerServiceRotateToken` calls `authenticateExporter` as its first step but only exercises valid authenticators. Fix: add a sub-test using `authFailureServiceCtx` and assert `status.Code(err) == codes.Unauthenticated`.

controller/internal/controller/exporter_controller.go
```diff
@@ -125,24 +147,28 @@ func (r *ExporterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
...
 }
 
 func (r *ExporterReconciler) reconcileStatusCredential(
 	ctx context.Context,
 	exporter *jumpstarterdevv1alpha1.Exporter,
-) error {
-	secret, _, err := ensureSecret(ctx, client.ObjectKey{
+) (time.Time, error) {
+	secret, expiry, err := ensureSecret(ctx, client.ObjectKey{
 		Name:      exporter.Name + "-exporter",
```
raballew: [HIGH] The credential secret name is constructed as `exporter.Name + "-exporter"` independently here and also in `controller_service.go` (inside `RotateToken`). A naming scheme change requires two-site updates with no compiler enforcement to catch mismatches. Fix: define a method on the `Exporter` CRD type, e.g., `func (e *ExporterV1Alpha1) CredentialSecretName() string { return e.Name + "-exporter" }`, and call it from both the reconciler and the `RotateToken` handler.

controller/internal/controller/exporter_controller_test.go
```diff
@@ -125,5 +135,110 @@ var _ = Describe("Exporter Controller", func() {
...
+
+			// Verify condition is Valid
+			exporter := &jumpstarterdevv1alpha1.Exporter{}
+			Expect(k8sClient.Get(ctx, typeNamespacedName, exporter)).To(Succeed())
+			cond := meta.FindStatusCondition(exporter.Status.Conditions, string(jumpstarterdevv1alpha1.ExporterConditionTypeTokenExpiring))
+			Expect(cond).NotTo(BeNil())
+			Expect(cond.Status).To(Equal(metav1.ConditionFalse))
+			Expect(cond.Reason).To(Equal("Valid"))
+		})
 	})
```
raballew: [MEDIUM] This PR adds two event emission branches in `exporter_controller.go`, but all new reconciler tests create the reconciler without a `Recorder` field. `emitEventf` silently no-ops when `Recorder == nil`, so event regressions are completely invisible. Fix: inject `record.NewFakeRecorder(10)` as the reconciler's `Recorder` in the relevant tests, then assert `<-recorder.Events` contains strings matching `"CredentialCreated"` and `"TokenExpiringSoon"` after the appropriate reconcile cycles.

controller/internal/controller/exporter_controller_test.go
```diff
@@ -125,5 +135,110 @@ var _ = Describe("Exporter Controller", func() {
...
+			secret.Data[TokenKey] = []byte("invalid-token")
+			Expect(k8sClient.Update(ctx, secret)).To(Succeed())
+
+			// Reconcile
+			_, err = controllerReconciler.Reconcile(ctx, reconcile.Request{
+				NamespacedName: typeNamespacedName,
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			// Verify secret was recreated with a valid token
```
raballew: [MEDIUM] The new test steps in this block use `// ...` inline comments to narrate what is happening, while the rest of this file uses `By("...")` calls for step descriptions. Fix: replace each `// step description` comment with a matching `By("step description")` call to stay consistent with the established pattern.

controller/internal/service/controller_service.go
```diff
@@ -342,6 +342,65 @@ func (s *ControllerService) GetServiceEndpoints(
...
+	}{}
+	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
+	if _, _, err := parser.ParseUnverified(token, claims); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to parse token claims: %s", err)
+	}
+
+	var expiry *timestamppb.Timestamp
+	if claims.ExpiresAt != nil {
+		expiry = timestamppb.New(claims.ExpiresAt.Time)
+	}
```
raballew: [MEDIUM] After `s.Signer.Token(...)` returns the signed token, `RotateToken` immediately re-parses it with `jwt.NewParser(jwt.WithoutClaimsValidation())` to extract `ExpiresAt`. This duplicates exactly what `Signer.TokenExpiry` already does. Fix: replace the inline parse block with `expiry, err := s.Signer.TokenExpiry(token)` placed before the Patch call. This removes duplication, keeps expiry extraction encapsulated, and eliminates a small window where expiry extraction fails after the Patch has already been applied.

python/packages/jumpstarter/jumpstarter/exporter/token_refresh.py
```diff
@@ -0,0 +1,132 @@
...
+
+    try:
+        payload = decode_jwt_payload(token)
+    except ValueError:
+        return None
+
+    exp = payload.get("exp")
+    if exp is None or not isinstance(exp, (int, float)):
+        return None
+    exp_f = float(exp)
```
raballew: [MEDIUM] In the `else` branch (no `iat` claim in token), the 50%-cap on lead time is applied before checking whether `remaining <= min_lead_time`. This violates the documented guarantee that `min_lead_time` is always honoured. Fix: add a short-circuit before the cap check: `if remaining <= min_lead_time: return 0.0`.

python/packages/jumpstarter/jumpstarter/exporter/token_refresh_test.py
```diff
@@ -0,0 +1,120 @@
...
+    get_token_expiry,
+    get_token_remaining_seconds,
+)
+
+
+def _make_jwt(payload: dict) -> str:
+    header = {"alg": "ES256", "typ": "JWT"}
+    h_b64 = base64.urlsafe_b64encode(json.dumps(header).encode()).rstrip(b"=").decode()
+    p_b64 = base64.urlsafe_b64encode(json.dumps(payload).encode()).rstrip(b"=").decode()
+    return f"{h_b64}.{p_b64}.dummy_signature"
```
raballew: [MEDIUM] `_make_jwt` is defined with an identical body in both `token_refresh_test.py` (lines 14-18) and `exporter_token_refresh_test.py` (lines 17-21). Any change to the helper must be made in two places. Fix: extract `_make_jwt` into a shared `conftest.py` or `tests/helpers.py` module and import it from both test files.

python/packages/jumpstarter/jumpstarter/exporter/token_refresh.py
```diff
@@ -0,0 +1,132 @@
...
+    remaining = exp_f - now
+    if remaining <= 0:
+        return 0.0
+
+    iat = payload.get("iat")
+    if iat is not None and isinstance(iat, (int, float)) and float(iat) < exp_f:
+        total_lifetime = exp_f - float(iat)
+        lead = min(lead_time, max(total_lifetime * fraction, min_lead_time))
+        # Ensure lead time never exceeds half the token's total lifetime
+        lead = min(lead, total_lifetime * 0.5)
```
raballew: [MEDIUM] The comment `# Ensure lead time never exceeds half the token's total lifetime` immediately precedes `lead = min(lead, total_lifetime * 0.5)`. The code is self-explanatory. Fix: remove the comment. If additional clarity is wanted, extract `total_lifetime / 2` into a named intermediate variable.

protocol/proto/jumpstarter/v1/jumpstarter.proto
```diff
@@ -280,4 +283,13 @@ message TelemetryEndpoint {
...
+}
+
+// Request to rotate the authentication token for an exporter.
+message RotateTokenRequest {}
+
+// Response containing the newly rotated token.
+message RotateTokenResponse {
+  string token = 1; // The new authentication token.
+  google.protobuf.Timestamp expiry = 2; // The expiration time of the new token.
 }
```
raballew: [MEDIUM] The file ends at the closing `}` of `RotateTokenResponse` with no trailing newline. `buf lint` (STANDARD ruleset) enforces `FILE_NEWLINE` for `.proto` files and will fail CI when buf is available in the pipeline. Fix: add a single trailing newline after the closing brace.

controller/api/v1alpha1/exporter_types.go
```diff
@@ -52,13 +52,16 @@ type ExporterStatus struct {
 	ExporterStatusValue string `json:"exporterStatus,omitempty"`
 	// StatusMessage is an optional human-readable message describing the current state
 	StatusMessage string `json:"statusMessage,omitempty"`
+	// TokenExpiresAt is the expiry time of the internal credential token.
```
raballew: [LOW] `TokenExpiresAt *metav1.Time` is written to the `Exporter` CRD status and is readable by anyone with `get`/`list` on `Exporter` objects. Status condition messages also include exact RFC3339 expiry timestamps. Fix: replace exact timestamps in event and condition messages with coarse descriptions like "expires within 7 days". If precise expiry metrics are needed, expose them via a Prometheus endpoint with its own RBAC policy.

controller/internal/service/controller_service.go
```diff
@@ -342,6 +342,65 @@ func (s *ControllerService) GetServiceEndpoints(
...
+
+	log.FromContext(ctx).Info("token rotated", "exporter", exporter.Name, "namespace", exporter.Namespace)
+
+	claims := &struct {
+		jwt.RegisteredClaims
+	}{}
+	parser := jwt.NewParser(jwt.WithoutClaimsValidation())
+	if _, _, err := parser.ParseUnverified(token, claims); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to parse token claims: %s", err)
+	}
```
raballew: [LOW] Both `RotateToken` here and `Signer.TokenExpiry` in `oidc/op.go` use `ParseUnverified` to extract expiry, establishing a codebase-wide pattern of trusting unverified claims. Fix: update `Signer.Token` to return the expiry alongside the string as `(string, time.Time, error)` so callers never need to re-parse a token received directly from the signer. Add a doc comment to `Signer.TokenExpiry` warning that its output is unverified.

protocol/proto/jumpstarter/v1/jumpstarter.proto
```diff
@@ -280,4 +283,13 @@ message TelemetryEndpoint {
...
   string min_severity = 3;  // Minimum log severity to forward (e.g. "info", "warning").
+}
+
+// Request to rotate the authentication token for an exporter.
+message RotateTokenRequest {}
+
+// Response containing the newly rotated token.
+message RotateTokenResponse {
+  string token = 1; // The new authentication token.
+  google.protobuf.Timestamp expiry = 2; // The expiration time of the new token.
```
raballew: [LOW] `string token = 1; // The new authentication token.` and `google.protobuf.Timestamp expiry = 2; // The expiration time of the new token.` both restate what the field name already communicates. Fix: remove both inline comments. If documentation beyond the field name is needed, use leading `//` block comments above each field.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -601,16 +620,103 @@ async def _setup_telemetry(self) -> None:
...
+            The newly rotated token string.
+        """
+        async with self._controller_stub() as controller:
+            response: jumpstarter_pb2.RotateTokenResponse = await controller.RotateToken(
+                jumpstarter_pb2.RotateTokenRequest(),
+                timeout=_RPC_TIMEOUT,
+            )
+
+        new_token = response.token
+        self.token = new_token
```
raballew: [LOW] `self.token = new_token` sets the rotated token without any validation. An empty token causes `calculate_token_refresh_sleep("")` to raise `ValueError`, silently exiting the refresh loop with no diagnostic. Fix: validate before assignment with `if not new_token: raise ValueError("RotateToken returned an empty token")`. Optionally also verify that `"exp"` is present in the decoded payload.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
```diff
@@ -601,16 +620,103 @@ async def _setup_telemetry(self) -> None:
...
+                except Exception as e:
+                    attempt += 1
+                    backoff = min(_RPC_BACKOFF_BASE * (2 ** min(attempt - 1, 5)), _RPC_BACKOFF_CAP)
+                    logger.warning(
+                        "Unexpected failure rotating token (%s); retrying in %.1fs (attempt %d)",
+                        e,
+                        backoff,
+                        attempt,
+                    )
+                    await sleep(backoff)
```
raballew: [LOW] The `except Exception as e:` retry block has no iteration limit or escape condition. A persistent non-transient error will retry indefinitely with backoff capped at 30 seconds, never surfacing as a process failure. Fix: track a retry counter and `raise` after a configurable maximum (e.g., 10 attempts), or log at `ERROR` level after a threshold to make the failure visible.

python/packages/jumpstarter/jumpstarter/exporter/exporter_token_refresh_test.py
```diff
@@ -0,0 +1,245 @@
...
+    now = time.time()
+    token = _make_jwt({"iat": now - 100, "exp": now + 1})
+
+    mock_channel = MagicMock(spec=grpc.aio.Channel)
+    mock_channel.close = AsyncMock()
+
+    mock_ctrl = AsyncMock()
+    mock_ctrl.RotateToken.side_effect = FakeRpcError(grpc.StatusCode.UNIMPLEMENTED)
+
+    async def channel_factory():
```
raballew: [LOW] The `async def channel_factory():` inner functions at lines 44, 78, 111, 148, and 179 are missing `-> grpc.aio.Channel` return type annotations. Fix: add `-> grpc.aio.Channel` to each closure definition.

python/packages/jumpstarter/jumpstarter/exporter/exporter_token_refresh_test.py
```diff
@@ -0,0 +1,245 @@
...
+    loaded = ExporterConfigV1Alpha1.load_path(config_file)
+    assert loaded.token == "initial-token"
+
+    # Simulate create_exporter wiring
+    async with loaded.create_exporter() as exporter:
+        assert exporter.on_token_rotated is not None
+        await exporter.on_token_rotated("new-saved-token")
+
+    assert loaded.token == "new-saved-token"
+    # Verify persisted to disk
```
raballew: [LOW] Comments like `# Normal rotation result is preserved and token is updated` and `# Token was not updated` appear immediately before self-explanatory `assert` calls in multiple test bodies. Fix: remove the comments. Encode intent in assertion messages or split into more narrowly named tests whose names make the comment redundant.

python/packages/jumpstarter/jumpstarter/exporter/token_refresh.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Raises:
+        ValueError: If the token format is invalid or cannot be decoded
+    """
+    parts = token.split(".")
+    if len(parts) != 3:
+        raise ValueError(f"Invalid JWT format: expected 3 parts, got {len(parts)}")
+
+    payload_b64 = parts[1]
+    # Handle base64 padding
```
raballew: [LOW] `# Handle base64 padding` precedes `padded = payload_b64 + "=" * (-len(payload_b64) % 4)`. The expression already describes what it does. Fix: extract `_restore_base64_padding(s: str) -> str: return s + "=" * (-len(s) % 4)` as a named helper and remove the comment.

python/packages/jumpstarter/jumpstarter/exporter/token_refresh_test.py
```diff
@@ -0,0 +1,120 @@
...
+    token = _make_jwt({"iat": iat, "exp": exp})
+
+    # Expected sleep: 50s
+    sleep_s = calculate_token_refresh_sleep(token, now=now)
+    assert sleep_s is not None
+    assert pytest.approx(sleep_s, rel=1e-3) == 50.0
+
+
+def test_calculate_token_refresh_sleep_near_expiry():
+    # Token expiring in 3 days (less than 7 days lead time on a 365-day token)
```
raballew: [LOW] Comment blocks like `# Lead time is capped by DEFAULT_TOKEN_REFRESH_LEAD_TIME (7 days = 604,800s)` immediately precede assertions that encode the same arithmetic. Fix: delete the explanation comments. Where a particular case is non-obvious, encode the intent in the test function name instead.

python/packages/jumpstarter/jumpstarter/exporter/exporter.py
raballew: [LOW] If the controller issues a freshly rotated token that itself expires within the configured lead time, `calculate_token_refresh_sleep` returns `0.0` and the outer loop immediately rotates again with no pause. This can produce a continuous stream of `RotateToken` RPCs until the controller issues a token with sufficient lifetime. Backoff only applies inside the inner retry loop for failures, not between successive successful rotations. Fix: after a successful rotation where `sleep_duration == 0.0`, insert a minimum pause (e.g., `await sleep(1.0)`) before recalculating.

controller/internal/service/controller_service.go
```diff
@@ -342,6 +342,65 @@ func (s *ControllerService) GetServiceEndpoints(
...
+		return nil, status.Errorf(codes.Internal, "failed to get credential secret: %s", err)
+	}
+
+	original := client.MergeFrom(secret.DeepCopy())
+	if secret.Data == nil {
+		secret.Data = map[string][]byte{}
+	}
+	secret.Data[controller.TokenKey] = []byte(token)
+	if err := s.Client.Patch(ctx, &secret, original); err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to update credential secret: %s", err)
```
raballew: [LOW] `status.Errorf(codes.Internal, "failed to get credential secret: %s", err)` and `status.Errorf(codes.Internal, "failed to update credential secret: %s", err)` embed raw Kubernetes API errors in the gRPC response. An authenticated exporter receiving a not-found error learns the expected secret naming convention and namespace from the error message. Fix: return a generic message without the underlying error detail, and log the full error server-side: `log.FromContext(ctx).Error(err, "failed to get credential secret", "exporter", exporter.Name)` followed by `return nil, status.Errorf(codes.Internal, "failed to get credential secret")`.

PR #1068, opened 2026-09-04
feat(dut-network): add VLAN sub-interface and policy-based routing support

mangelajo:
## Summary

Add VLAN tagging and policy-based routing (PBR) to the `dut-network` driver, enabling DUTs to reach public networks through VLAN-tagged uplinks or untagged source-IP PBR.

## New Features

- **VLAN sub-interfaces**: `AddressEntry` gains `vlan_id` and `public_gateway` fields. When both are set, the driver creates a VLAN sub-interface on the upstream interface (e.g. `eth0.905`), assigns the public IP, and installs PBR rules so traffic from that DUT is routed through the VLAN gateway.

- **Untagged source-IP PBR**: Setting `public_gateway` without `vlan_id` installs PBR using `int(IPv4Address(dut_ip))` as the routing table ID — useful when no VLAN tag is needed but a non-default gateway is required.

- **Validation & safety**:
  - Warning emitted when `vlan_id` is set without `public_gateway` (VLAN created but no routing — probably misconfiguration)
  - Reserved kernel routing tables (0, 253, 254, 255) rejected at validation time for both tagged and untagged PBR
  - Invalid `public_gateway` IPs rejected

- **Runtime support**: `add_address` / `remove_address` properly tear down and rebuild VLAN interfaces, PBR rules, masquerade/1:1 NAT rules, and Docker FORWARD ACCEPT handles.

- **nftables**: Masquerade and 1:1 NAT extended with per-VLAN `nat_interfaces` so traffic egresses the correct interface.

### Unit tests (258 passed)
- `AddressEntry` validation (VLAN range, reserved tables, IPv4 requirement, gateway format)
- VLAN setup/teardown for masquerade and 1:1 modes
- Untagged PBR table-ID derivation and cleanup
- Runtime `_sync_nat` refreshes forward-chain handles and masquerade rules
- `iproute` helpers: create/delete VLAN, policy routes, IP rules
- `nftables`: `nat_interfaces` in masquerade and 1:1 rule generation

### E2E tests (data-plane verification)
- TCP echo through VLAN PBR
- TCP echo through untagged source-IP PBR
- Negative: VLAN-only peer unreachable without `public_gateway`

## Documentation
- README updated with configuration guide and YAML examples
- `exporter-vlan.yaml` example added
- E2E README updated with new test descriptions


python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/test_driver_integration.py
bennyz, line 1: we have both driver_test.py and test_driver.py - what is the difference?

mangelajo: true, that's a good question. Since we had different files I had not realized that we had the same in two forms.

so driver_tests.py seems to be doing integration tests (veth pairs, etc..) and test_driver.py has only unit tests.

I will rename this to test_driver_integration.py

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/iproute.py
```diff
@@ -81,20 +102,111 @@ def remove_ip_alias(interface: str, ip: str, prefix_len: int) -> None:
...
+    )
+
+
+def add_ip_rule(from_ip: str, table: int, priority: int = 100) -> None:
+    """Add a policy routing rule sending traffic from *from_ip* to *table*."""
+    logger.info("Adding ip rule from %s table %d priority %d", from_ip, table, priority)
+    _run_priv(
+        ["ip", "rule", "add", "from", from_ip, "table", str(table), "priority", str(priority)],
+        check=False,
+    )
```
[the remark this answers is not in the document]
mangelajo: Fixed in 6b1ad469. `add_policy_route` and `add_ip_rule` now check the return code and raise `RuntimeError` on failure. "File exists" is tolerated for idempotency. The driver's `_setup_vlans_and_pbr` rolls back the routing table if `add_ip_rule` fails after `add_policy_route` succeeded. Added tests for both failure and idempotent-exists paths.

e2e/test/dut_network_test.go
```diff
@@ -255,39 +270,60 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+				pbrDutIP, "--public-gateway", extIP)
+			Expect(err).NotTo(HaveOccurred(), out)
+
+			addDutAddr(pbrDutIP)
+			defer func() {
+				delDutAddr(pbrDutIP)
+				_, _ = jmpShell("j", "dut-network", "remove-address", pbrDutIP)
+			}()
+
+			expectTCPEcho(dutNs, extNs, pbrDutIP, extIP, 9998)
```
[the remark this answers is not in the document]
mangelajo: Acknowledged — this is a valid observation: the test proves connectivity works *with* PBR configured but does not prove traffic actually used the PBR route vs. the main table. Making this truly verifiable would require an additional netns with a destination reachable only through the PBR gateway, which adds significant test plumbing. The current test still catches regressions where PBR *breaks* connectivity (e.g. routing loops or misconfigured rules). I'll leave this as a follow-up improvement.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/test_driver.py
```diff
@@ -548,3 +550,277 @@ def test_unresolvable_hostname_raises_during_setup(self, tmp_path: Path):
...
+        }]
+        _, mock_ip, mock_nft, _ = _make_driver(tmp_path, nat_mode="masquerade", addresses=addrs)
+        mock_ip.create_vlan_interface.assert_called_once_with("eth-up", 905)
+        mock_ip.set_interface_forwarding.assert_any_call("eth-up.905", True)
+        mock_ip.set_interface_rp_filter.assert_called_once_with("eth-up.905", 2)
+        mock_ip.add_ip_alias.assert_called_once_with("eth-up.905", "10.6.183.1", 24)
+        mock_ip.add_policy_route.assert_called_once_with("10.6.183.254", "eth-up.905", 905)
+        mock_ip.add_ip_rule.assert_called_once_with("192.168.100.125", 905, priority=100)
+        call_kwargs = mock_nft.apply_masquerade_rules.call_args[1]
+        assert call_kwargs["nat_interfaces"] == ["eth-up.905"]
```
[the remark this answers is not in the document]
mangelajo: This is a false positive. The test at line 633 covers a **single VLAN-tagged address**. In that scenario `nat_interfaces` should only contain the VLAN sub-interface (`eth-up.905`) because the base upstream (`eth-up`) already receives its masquerade rule through the standard `apply_masquerade_rules(dut_iface, upstream, subnet)` positional args. Adding `eth-up` to `nat_interfaces` too would double-masquerade the untagged path.

The `_outbound_interfaces()` method in the driver already handles the mixed case correctly — it includes `eth-up` only when at least one address has no VLAN. The existing `test_no_vlan_does_not_pass_nat_interfaces` test at line 636 verifies the untagged-only path, and the combination is tested at the integration level.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/iproute.py
```diff
@@ -81,20 +102,134 @@ def remove_ip_alias(interface: str, ip: str, prefix_len: int) -> None:
...
+    result = _run_priv(
+        ["ip", "route", "add", "default", "via", gateway, "dev", device,
+         "table", str(table), "onlink"],
+        check=False,
+    )
+    if result.returncode != 0:
+        stderr = result.stderr.strip()
+        if "File exists" in stderr:
+            logger.info("Policy route table %d already exists, skipping", table)
+        else:
```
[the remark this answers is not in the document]
mangelajo: Good catch — fixed in 06325674. Switched from `ip route add` to `ip route replace`, which is idempotent by design: identical routes are a no-op, and conflicting stale defaults are overwritten with the requested gateway/device. This eliminates the ambiguity of "File exists" entirely. Updated tests accordingly.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,185 @@ def _setup_network(self) -> None:
...
+
+        VLAN mappings use the VLAN ID so tagged traffic shares one table.
+        Untagged mappings use the DUT private IPv4 address as a unique
+        32-bit table ID (``from <ip> lookup <int(ip)>``).
+        """
+        if entry.vlan_id is not None:
+            return entry.vlan_id
+        return int(ipaddress.IPv4Address(entry.ip))
+
+    def _outbound_interfaces(self) -> list[str]:
```
bennyz, line 467: is it possible we are losing the upstream iface? 

in nftables.py there's:
    if upstream not in outbound:
        outbound.append(upstream)

mangelajo: I think we could specify or have auto detection, I will check if the logic is still valid or got perversed

mangelajo: So I looked at the code, and, in _outbound_interfaces(), a VLAN-only config intentionally returns just the VLAN sub-interfaces since there are no untagged DUTs, in this case no traffic needs to masquerade on the upstream. 

Then on the 1to1 path in nftables.py has the "if upstream not in outbound: outbound.append(upstream)" because 1:1 allows mixed mapped/unmapped entries, where unmapped DUTs still need the untagged upstream for fallback port for masquerade.

This makes me think though that it could make sense to declare 1to1 / snat / disabled per entry instead of globally.

mangelajo: I have documented this a bit better, and added some support comments in code.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,188 @@ def _setup_network(self) -> None:
...
+        """Create VLAN sub-interfaces, sysctls, IP aliases, and PBR rules."""
+        parent = self._vlan_parent()
+        if not parent:
+            return
+        for entry in self.addresses:
+            nat_if = self._nat_iface_for(entry)
+            if entry.vlan_id is not None:
+                name = nat_if
+                iproute.create_vlan_interface(parent, entry.vlan_id)
+                self._created_vlans.add(name)
```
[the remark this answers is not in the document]
mangelajo: This is ok, if such an interface would exist in an exporter it's very likely we created it, this is not general purpose software to run on a server.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,188 @@ def _setup_network(self) -> None:
...
+                        "VLAN interface %s is configured without policy-based "
+                        "routing, so DUT traffic may not egress via the VLAN",
+                        entry.ip,
+                        entry.vlan_id,
+                        name,
+                    )
+            if entry.public_gateway:
+                gateway = self._resolve_ip(entry.public_gateway)
+                table = self._pbr_table_id(entry)
+                iproute.add_policy_route(gateway, nat_if, table)
```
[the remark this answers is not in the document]
mangelajo: Same answer as for the other comment, this is not general software on a server, if such rules exist, it's probably because we created them. We should not over engineer this.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/iproute.py
```diff
@@ -81,20 +102,134 @@ def remove_ip_alias(interface: str, ip: str, prefix_len: int) -> None:
...
+            raise RuntimeError(
+                f"Failed to add ip rule (from {from_ip} table {table}): {stderr}"
+            )
+
+
+def delete_ip_rule(from_ip: str, table: int) -> None:
+    """Delete a policy routing rule previously added by :func:`add_ip_rule`."""
+    logger.info("Deleting ip rule from %s table %d", from_ip, table)
+    _run_priv(["ip", "rule", "del", "from", from_ip, "table", str(table)], check=False)
```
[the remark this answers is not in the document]
mangelajo: Same answer as for the other comment, this is not general software on a server, if such rules exist, it's probably because we created them. We should not over engineer this.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
bennyz: So do we want to have a drop rule to the untagged iface traffic?

mangelajo: No, we need to use the untagged iface traffic under some conditions. (SNAT).

mangelajo:
@bennyz now it will keep the upstream interface on the masquerade interface list, and added an E2E test for that corner case we discussed (VLAN 1:1 mapping for a dut, but an extra device, or ... the device with unexpected mac address..) . Also handled a coderabbit cleanup/rollback request on setup failures.

bennyz:
> @bennyz now it will keep the upstream interface on the masquerade interface list, and added an E2E test for that corner case we discussed (VLAN 1:1 mapping for a dut, but an extra device, or ... the device with unexpected mac address..) . Also handled a coderabbit cleanup/rollback request on setup failures.

cool, thanks!

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+
+    def _teardown_vlans_and_pbr(self) -> None:
+        """Reverse VLAN sub-interfaces, PBR rules, and custom routing tables."""
+        for from_ip, table in self._pbr_rules:
+            iproute.delete_ip_rule(from_ip, table)
+        self._pbr_rules.clear()
+        for table in list(self._pbr_tables):
+            iproute.flush_routing_table(table)
+        self._pbr_tables.clear()
+        for name in list(self._created_vlans):
```
bennyz, line 573: is it possible we delete stuff we didn't create?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+            if entry.vlan_id is not None:
+                name = self._vlan_name(entry.vlan_id)
+                if name not in result:
+                    result.append(name)
+        return result
+
+    def _setup_vlan_interface(self, parent: str, entry: "AddressEntry", name: str) -> None:
+        """Create a single VLAN sub-interface with sysctls, alias, and warnings."""
+        assert entry.vlan_id is not None
+        iproute.create_vlan_interface(parent, entry.vlan_id)
```
bennyz, line 492: maybe worth returning if we create the interface or it was already there so we clean only those we created?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
bennyz: addressed in code

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: not a bad idea,

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: it could be, I will implement the extra flag you mentioned. But I think it would be quite a corner case, this is normally software to be run on a exporter. Perhaps makes sense for the use case of laptop-enabled exporter.

bennyz: yeah, i was thinking in case we have multiple exporters on the same sidekick for example, or is it not affected?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Ok, I was thinking about this. I think it's better to cleanup anyways.

For example, if an exporter crashes or it's forcefully killed, the interfaces will remain, then we never delete them.

I think I will document this behavior as warning on the README.md

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -113,6 +113,83 @@ def from_dict(cls, data: dict) -> "FilterConfig":
...
+            if not _VLAN_ID_MIN <= self.vlan_id <= _VLAN_ID_MAX:
+                raise ValueError(
+                    f"vlan_id must be between {_VLAN_ID_MIN} and {_VLAN_ID_MAX}, got {self.vlan_id}"
+                )
+        if self.public_gateway is not None:
+            try:
+                ipaddress.ip_address(self.public_gateway)
+            except ValueError as exc:
+                raise ValueError(
+                    f"public_gateway is not a valid IP address: {self.public_gateway!r}"
```
raballew, line 166: Does IPv6 cause `add_policy_route` to fail at runtime thus making NAT unusable for all DUTs until driver restart?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+        except Exception:
+            # Roll back: undo alias bookkeeping and delete the VLAN
+            # interface (deleting it also removes any addresses on it).
+            for alias_ip in list(self._added_aliases):
+                if self._alias_ifaces.get(alias_ip) == name:
+                    self._added_aliases.discard(alias_ip)
+                    self._alias_ifaces.pop(alias_ip, None)
+            self._created_vlans.discard(name)
+            iproute.delete_vlan_interface(name)
+            raise
```
raballew, line 520: iproute.create_vlan_interface(parent, entry.vlan_id)
        self._created_vlans.add(name)
        
 Can those fail and leave orphaned resources in the kernel?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+                gateway = self._resolve_ip(entry.public_gateway)
+                table = self._pbr_table_id(entry)
+                iproute.add_policy_route(gateway, nat_if, table)
+                try:
+                    iproute.add_ip_rule(entry.ip, table, priority=_PBR_PRIORITY)
+                except RuntimeError:
+                    iproute.flush_routing_table(table)
+                    raise
+                self._pbr_rules.append((entry.ip, table))
+                self._pbr_tables.add(table)
```
raballew: [HIGH] When `add_policy_route` fails inside `_setup_vlans_and_pbr`, the exception propagates through `_sync_nat` and `add_address` without removing the partially-created VLAN interface. The VLAN name is already in `_created_vlans` at that point but no kernel teardown occurs. Wrap the `_setup_vlans_and_pbr()` call in `_sync_nat` with a try/except that calls `_teardown_vlans_and_pbr()` before re-raising, mirroring the `try: _setup_network() except: self.cleanup()` pattern used in `__post_init__`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/iproute.py
```diff
@@ -81,20 +102,134 @@ def remove_ip_alias(interface: str, ip: str, prefix_len: int) -> None:
...
-    _run_priv(["sysctl", "-w", f"net.ipv4.conf.{iface}.forwarding={value}"])
+    key = _sysctl_conf_key(iface, "forwarding")
+    logger.info("Setting %s=%s", key, value)
+    _run_priv(["sysctl", "-w", f"{key}={value}"])
+
+
+def get_interface_rp_filter(iface: str) -> str:
+    """Return the current per-interface rp_filter value."""
+    result = _run(["sysctl", "-n", _sysctl_conf_key(iface, "rp_filter")], check=False)
+    return result.stdout.strip() or "0"
```
raballew: [HIGH] `get_interface_rp_filter` is a new public function with no unit tests. The companion `set_interface_rp_filter` is tested but this read function is not. The function also appears to be dead code with no call sites in the driver, meaning the sysctl key construction including the VLAN dot-to-slash translation is entirely unvalidated. Add at least `test_get_rp_filter_returns_current_value` and `test_get_rp_filter_translates_vlan_dots`, mirroring the `get_interface_forwarding` test suite.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -429,81 +701,123 @@ def _get_leases_list(self) -> list[dict]:
...
+                self._added_aliases.add(ip)
+                self._alias_ifaces[ip] = iface
 
         nftables.flush_rules(self._table_name)
         nftables.apply_1to1_rules(
-            self.interface, upstream_for_alias, mappings, self.subnet,
+            self.interface, parent, mappings, self.subnet,
             table_name=self._table_name,
             filter_config=self.filter,
         )
```
raballew: [HIGH] `_sync_nat` in 1:1 NAT mode has no unit tests for runtime operations. No test creates a 1:1 NAT driver and calls `add_address` or `remove_address` against it. The alias migration branch (lines 768-772, moving an alias from one interface to another) has zero coverage. Add `TestSync1to1NatRuntime` with tests for alias removal, alias addition, and alias migration between untagged and VLAN interfaces.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/test_driver.py
```diff
@@ -424,30 +426,30 @@ def test_add_address_with_mac(self, tmp_path: Path):
...
         addrs = [{"mac": "AA:BB:CC:DD:EE:FF", "ip": "192.168.100.10"}]
         driver, _, _, _ = _make_driver(tmp_path, addresses=addrs)
         with patch(f"{_DRIVER_MODULE}.dnsmasq"):
             driver.add_address("192.168.100.10", mac="11:22:33:44:55:66")
             assert len(driver.addresses) == 1
-            assert driver.addresses[0]["mac"] == "11:22:33:44:55:66"
+            assert driver.addresses[0].mac == "11:22:33:44:55:66"
 
     def test_remove_address(self, tmp_path: Path):
         addrs = [{"mac": "aa:bb:cc:dd:ee:ff", "ip": "192.168.100.10"}]
```
raballew: [HIGH] The only `remove_address` unit test uses a plain address with no `vlan_id` and only asserts `driver.addresses == []`. It never patches `iproute` and never verifies that `delete_vlan_interface`, `delete_ip_rule`, or `flush_routing_table` are called when removing a VLAN-configured entry. Add `TestVlanRemoveAddress` that creates a driver with a VLAN entry, calls `remove_address`, and asserts those three iproute functions were called with correct arguments, and that `driver._created_vlans` is empty afterwards.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/nftables.py
```diff
@@ -73,9 +79,20 @@ def _render_filter_rule(
...
 
+def _normalize_upstreams(upstream: str | list[str]) -> list[str]:
+    """Return a de-duplicated list of upstream/NAT interface names."""
+    if isinstance(upstream, str):
+        return [upstream]
+    seen: list[str] = []
+    for name in upstream:
+        if name not in seen:
+            seen.append(name)
+    return seen
```
raballew: [HIGH] `_normalize_upstreams` is a new function with two code paths (string to singleton list; list to deduplicated list) and empty-string filtering, but it has zero test coverage. `grep -n '_normalize_upstreams' test_nftables.py` returns no matches. Add `TestNormalizeUpstreams` covering: string becomes singleton list, list with duplicates is deduplicated, empty strings are filtered, and order is preserved.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -160,6 +241,7 @@ def client(cls) -> str:
         return "jumpstarter_driver_dut_network.client.DutNetworkClient"
 
     def __post_init__(self):
+        """Initialise the driver: validate config, set up the network stack."""
```
raballew: [HIGH] British English spellings in new code will fail the `typos` spell-check. The project enforces `en-us` locale. Running `typos python/packages/jumpstarter-driver-dut-network/ --format brief` exits with code 2, reporting `Initialise` at driver.py:244, and `behaviour` at driver.py:438, README.md:105, README.md:223, and exporter-vlan.yaml:28. Replace `Initialise` with `Initialize` and all occurrences of `behaviour` with `behavior`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -113,6 +113,83 @@ def from_dict(cls, data: dict) -> "FilterConfig":
...
+                if table in _RESERVED_ROUTE_TABLES:
+                    raise ValueError(
+                        f"DUT IP {self.ip!r} maps to routing table {table} which "
+                        "is reserved by the kernel (tables 0, 253, 254, 255)"
+                    )
+            elif self.vlan_id in _RESERVED_ROUTE_TABLES:
+                raise ValueError(
+                    f"vlan_id {self.vlan_id} cannot be used as a routing table ID "
+                    "(tables 0, 253, 254, 255 are reserved by the kernel)"
+                )
```
raballew, line 185: When a `vlan_id` is provided, the `ip` field flows unchecked into `iproute.add_ip_rule`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+                self._setup_vlan_interface(parent, entry, nat_if)
+            if entry.public_gateway:
+                gateway = self._resolve_ip(entry.public_gateway)
+                table = self._pbr_table_id(entry)
+                iproute.add_policy_route(gateway, nat_if, table)
+                try:
+                    iproute.add_ip_rule(entry.ip, table, priority=_PBR_PRIORITY)
+                except RuntimeError:
+                    iproute.flush_routing_table(table)
+                    raise
```
raballew: [HIGH] The rollback path inside `_setup_vlans_and_pbr` where `add_ip_rule` fails after `add_policy_route` succeeds has no unit test. The implementation calls `iproute.flush_routing_table(table)` on `RuntimeError` here, but nothing verifies this behavior. Add a test in `TestVlanCreationRollback` that sets `mock_ip.add_ip_rule.side_effect = RuntimeError("rule failed")`, constructs a driver with a VLAN+gateway entry, catches the raised error, and asserts `mock_ip.flush_routing_table` was called with the expected table ID.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -113,6 +113,83 @@ def from_dict(cls, data: dict) -> "FilterConfig":
...
+            result["mac"] = self.mac
+        if self.hostname:
+            result["hostname"] = self.hostname
+        if self.public_ip:
+            result["public_ip"] = self.public_ip
+        if self.vlan_id is not None:
+            result["vlan_id"] = self.vlan_id
+        if self.public_gateway:
+            result["public_gateway"] = self.public_gateway
+        return result
```
raballew: [HIGH] `AddressEntry.to_dict()` is a new public method with seven conditional `if` branches determining which fields appear in the output, but it has zero unit tests. The output feeds directly into dnsmasq config generation for DHCP static leases. Add `TestAddressEntryToDict` covering: all fields populated, mac-less entry, hostname-less entry, `vlan_id` set vs. absent, `public_gateway` set, verifying the exact dict contents for each combination.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+        """Reverse VLAN sub-interfaces, PBR rules, and custom routing tables."""
+        for from_ip, table in self._pbr_rules:
+            iproute.delete_ip_rule(from_ip, table)
+        self._pbr_rules.clear()
+        for table in list(self._pbr_tables):
+            iproute.flush_routing_table(table)
+        self._pbr_tables.clear()
+        for name in list(self._created_vlans):
+            iproute.delete_vlan_interface(name)
+        self._created_vlans.clear()
```
raballew: [MEDIUM] `_teardown_vlans_and_pbr` clears `_created_vlans`, `_pbr_rules`, and `_pbr_tables` but never touches `_added_aliases` or `_alias_ifaces`. In masquerade mode, `_sync_nat` calls `_teardown_vlans_and_pbr` without reconciling aliases. Stale alias entries linger until `cleanup()`, where `remove_ip_alias` silently fails with `check=False`, masking potential real failures. A re-added entry with the same `public_ip` on a different VLAN interface may behave unexpectedly. After `_teardown_vlans_and_pbr` in `_sync_nat`, purge alias entries whose interface is no longer in `_created_vlans`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -113,6 +113,83 @@ def from_dict(cls, data: dict) -> "FilterConfig":
         return cls(**{k: v for k, v in data.items() if k in ("egress", "ingress")})
 
 
+_VLAN_ID_MIN = 1
+_VLAN_ID_MAX = 4094
+_RESERVED_ROUTE_TABLES = frozenset({0, 253, 254, 255})
```
raballew: [MEDIUM] `_RESERVED_ROUTE_TABLES = frozenset({0, 253, 254, 255})` is defined independently in both `driver.py` and `iproute.py`. A future change to one copy will not automatically propagate to the other. Remove the definition from `driver.py` and import it from `iproute`: `from .iproute import _RESERVED_ROUTE_TABLES`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -113,6 +113,83 @@ def from_dict(cls, data: dict) -> "FilterConfig":
...
+                )
+
+    @classmethod
+    def from_dict(cls, data: "AddressEntry | dict[str, Any]") -> "AddressEntry":
+        """Create an AddressEntry from a YAML/config dictionary."""
+        if isinstance(data, AddressEntry):
+            return data
+        return cls(**{k: v for k, v in data.items() if k in _ADDRESS_FIELDS})
+
+    def to_dict(self) -> dict[str, Any]:
```
raballew: The actual values are bounded to `str | int | None`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -113,6 +113,83 @@ def from_dict(cls, data: dict) -> "FilterConfig":
...
+                        "is reserved by the kernel (tables 0, 253, 254, 255)"
+                    )
+            elif self.vlan_id in _RESERVED_ROUTE_TABLES:
+                raise ValueError(
+                    f"vlan_id {self.vlan_id} cannot be used as a routing table ID "
+                    "(tables 0, 253, 254, 255 are reserved by the kernel)"
+                )
+
+    @classmethod
+    def from_dict(cls, data: "AddressEntry | dict[str, Any]") -> "AddressEntry":
```
raballew: [MEDIUM] `AddressEntry.from_dict()` accepts `dict[str, Any]` without static type checking on the values. The valid set of keys is already enumerated in `_ADDRESS_FIELDS`. Replace `dict[str, Any]` with `Mapping[str, str | int | None]` or an `AddressEntryDict` TypedDict to make the contract explicit.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -429,81 +701,123 @@ def _get_leases_list(self) -> list[dict]:
...
+            public_gateway=public_gateway,
+        )
+
+        self.addresses = [entry for entry in self.addresses if entry.ip != ip]
         self.addresses.append(new_entry)
         self._reload_dnsmasq_config()
-        if self.nat_mode == "1to1":
-            self._sync_1to1_nat()
+        self._sync_nat()
         self.logger.info("Added address: ip=%s mac=%s hostname=%s", ip, mac, hostname)
```
raballew: [MEDIUM] In `add_address`, `self.addresses.append(new_entry)` and `self._reload_dnsmasq_config()` execute before `self._sync_nat()`. If `_sync_nat` raises (for example, an OS-level VLAN creation failure), the new entry is permanently in `self.addresses` and dnsmasq has been updated, but no VLAN/PBR/NAT kernel state was created for it. The driver's in-memory model then diverges from kernel reality for all subsequent operations. Snapshot `self.addresses` before the mutation and restore it (plus rollback dnsmasq config) in an except block if `_sync_nat` raises.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/iproute.py
```diff
@@ -81,20 +102,134 @@ def remove_ip_alias(interface: str, ip: str, prefix_len: int) -> None:
...
+    if not interface_exists(name):
+        logger.info("Creating VLAN interface %s on %s id %d", name, parent, vlan_id)
+        _run_priv(
+            ["ip", "link", "add", "link", parent, "name", name, "type", "vlan", "id", str(vlan_id)]
+        )
+    else:
+        logger.info("VLAN interface %s already exists", name)
+    nm_set_unmanaged(name)
+    _run_priv(["ip", "link", "set", name, "up"])
+    return name
```
raballew, line 162: A pre-existing VLAN from a persistent network configuration or another management tool would be destroyed on driver exit.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -228,15 +312,19 @@ def _resolve_ip(value: str) -> str:
         return address
```
raballew, line 329: [MEDIUM] `_resolve_ip` calls `socket.getaddrinfo()` at runtime for `public_ip` and `public_gateway`. An authenticated client supplying a hostname causes the exporter host to perform outbound DNS resolution. If DNS is intercepted, the resolved IP can be manipulated so the exporter NATs to an unintended destination. Require `public_ip` and `public_gateway` to be valid IP address literals at the API boundary by adding an `ipaddress.ip_address()` pre-check before calling `_resolve_ip`. DNS resolution at startup from a config file is lower risk.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
             self.local_ntp,
         )
```
raballew: [MEDIUM] `_setup_network` is 63 lines sequentially configuring six distinct subsystems (upstream detection, state directory, bridge, IP forwarding and VLAN/PBR, nftables, dnsmasq, NAT, NTP), each with its own conditional logic. The method cannot be unit-tested without mocking all six subsystems simultaneously. Extract `_setup_routing()` (upstream detection, forwarding, VLAN/PBR, nftables FORWARD) and `_setup_services()` (dnsmasq, NAT, NTP) as sub-methods so that `_setup_network` becomes a sequencing stub of around 10 lines.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -160,6 +241,7 @@ def client(cls) -> str:
         return "jumpstarter_driver_dut_network.client.DutNetworkClient"
 
     def __post_init__(self):
+        """Initialise the driver: validate config, set up the network stack."""
         if hasattr(super(), "__post_init__"):
             super().__post_init__()
         self._table_name = nftables._table_name_for(self.interface)
```
raballew: [MEDIUM] `driver.py` calls `nftables._table_name_for(self.interface)` in `__post_init__`, reaching directly into a module-internal function (leading underscore). Callers outside `nftables.py` should not depend on module internals. Either rename `_table_name_for` to `table_name_for` to promote it to the module's public API, or derive the table name only within nftables-layer calls, removing the need for the driver to know the derivation rule.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/test_nftables.py
```diff
@@ -273,6 +273,58 @@ def test_without_filter_backward_compatible(self):
...
+             patch.object(nftables, "_load_ruleset") as mock_load:
+            nftables.apply_masquerade_rules(
+                "enp1s0u1", "end0", "192.168.100.0/24",
+                nat_interfaces=["end0.905"],
+            )
+            ruleset = mock_load.call_args[0][0]
+            assert 'oifname "end0.905" ip saddr 192.168.100.0/24 masquerade' in ruleset
+            assert 'iifname "enp1s0u1" oifname "end0.905" accept' in ruleset
+            assert 'iifname "end0.905" oifname "enp1s0u1" ct state related,established accept' in ruleset
+            assert 'oifname "end0" ip saddr' not in ruleset
```
raballew: [MEDIUM] `TestVlanNatRules.test_masquerade_uses_vlan_interface` calls `apply_masquerade_rules` with `nat_interfaces=["end0.905"]` (VLAN-only). In practice the driver always passes both the upstream interface and the VLAN interface together. No test verifies that when both are present, masquerade and forward rules are generated for both. Add `test_masquerade_with_upstream_and_vlan` that supplies a two-element list and asserts both interface names appear in the generated masquerade and forward rules.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+
+    def _setup_vlans_and_pbr(self) -> None:
+        """Create VLAN sub-interfaces, sysctls, IP aliases, and PBR rules."""
+        parent = self._vlan_parent()
+        if not parent:
+            return
+        for entry in self.addresses:
+            nat_if = self._nat_iface_for(entry)
+            if entry.vlan_id is not None:
+                nat_if = nat_if
```
raballew: `nat_if = nat_if` == no-op?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/test_driver.py
```diff
@@ -548,3 +550,408 @@ def test_unresolvable_hostname_raises_during_setup(self, tmp_path: Path):
...
+                    "vlan_id": 254,
+                    "public_gateway": "203.0.113.254",
+                }],
+            )
+
+    def test_invalid_public_gateway(self, tmp_path: Path):
+        with pytest.raises(ValueError, match="public_gateway"):
+            _make_driver(
+                tmp_path,
+                addresses=[{
```
raballew: [LOW] The reserved routing table validation tests cover only table ID 254. `_RESERVED_ROUTE_TABLES` contains `{0, 253, 254, 255}`, and the other three values are untested at both the driver level (test_driver.py:602-608) and the iproute level (test_iproute.py:208-210). Parameterize both tests with `@pytest.mark.parametrize("bad_id", [0, 253, 254, 255])` to ensure all reserved IDs are rejected.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/client.py
```diff
@@ -71,6 +79,7 @@ def tcpdump(self, args: list[str] | None = None) -> Generator[str, None, None]:
             yield line
 
     def cli(self):  # noqa: C901
```
raballew: Why does this have no return type annotation?

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/nftables.py
```diff
@@ -102,50 +123,69 @@ def _build_forward_chain(
...
+        if filter_config is None
+        else _filtered_forward_body(interface, upstreams, filter_config, extras)
+    )
+    lines = [
         "    chain forward {",
         "        type filter hook forward priority filter; policy accept;",
+        *body,
+        "    }",
     ]
+    return "\n".join(lines)
```
raballew: [LOW] `_build_forward_chain` is 47 lines that mix dispatch logic (`_legacy_forward_body` vs. `_filtered_forward_body`) with `chain forward { ... }` header/footer assembly inline. Move the chain wrapper into a `_wrap_chain(name, body_lines)` helper so `_build_forward_chain` becomes a dispatch call plus wrap, reducing it to under 10 lines.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
```diff
@@ -318,14 +399,195 @@ def _setup_network(self) -> None:
...
+    def _pbr_table_id(entry: AddressEntry) -> int:
+        """Routing-table ID for policy-based routing.
+
+        VLAN mappings use the VLAN ID so tagged traffic shares one table.
+        Untagged mappings use the DUT private IPv4 address as a unique
+        32-bit table ID (``from <ip> lookup <int(ip)>``).
+        """
+        if entry.vlan_id is not None:
+            return entry.vlan_id
+        return int(ipaddress.IPv4Address(entry.ip))
```
raballew, line 465: The way I read this is that when two address entries share the same `vlan_id` they also share the same routing table. Which can cause problems when either entry is removed.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: true, multiple exporters on the same sidekick would be affected by this and the cleanup in general. Perhaps we should add a flag to disable cleanup. But multiple exporters in a sidekick was not taken in account when designing some of the other parts of this. May be we should document that you should only use one instance of this driver on a host at once, until any time is spent in making sure such thing is possible.

mangelajo: we could add a cleanup (default to True) flag at some point
and play around to make sure we can run multiple at once...

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: why? it would work ipaddress.ip_address( knows about IPv6

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: we have a failback to delete what we created if it happens. there is a try/catch that will delete those if something happens later or at cleanup (that's why we add them to that list/set

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: This is based on our infrastructure, and made on purpose. If you use a host, you know that risk exists. I don't want to add another layer to be maintained in terms of host->IP, that's where I let DNS be our database.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: yep? 🤣

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Documented in README under "Known Limitations": only one driver instance per host is supported, and cleanup is unconditional (not ownership-tracked). A `cleanup`-disable flag for multi-instance/sidekick scenarios is left as a follow-up.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Documented in README under "Known Limitations": the driver always cleans up every VLAN interface it's configured for on exit (even pre-existing ones), so that crash recovery stays reliable.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Confirmed no runtime failure: untagged PBR (`public_gateway` without `vlan_id`) already validates `ip` as IPv4 in `AddressEntry.__post_init__` via `ipaddress.IPv4Address(self.ip)`, so an IPv6 DUT address is rejected at config-validation time (driver startup / `add_address`), not at PBR-setup time. VLAN-tagged PBR keys on `vlan_id` instead, so it has no IPv4 requirement. Added a note to the README documenting the IPv4 requirement for untagged PBR.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Confirmed — handled by the try/except rollback in `_setup_vlan_interface` (undoes alias bookkeeping and deletes the VLAN interface on any failure before re-raising). No further action needed here.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Fixed: `AddressEntry.__post_init__` now validates `ip` is a real IP address before it can reach `iproute.add_ip_rule` (skipped only for the untagged-PBR case, which already enforces the stricter IPv4 requirement with a more precise error). Added regression tests.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Fixed: tightened `AddressEntry.to_dict()` return type from `dict[str, Any]` to `dict[str, str | int | None]`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/iproute.py
mangelajo: By design — in the intended exporter deployment, the driver owns every VLAN interface matching its configuration. Documented explicitly in README under "Known Limitations" so this expectation is no longer implicit.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Agreed, keeping this as intentional design — no code change. `public_gateway` is already required to be an IP literal via `AddressEntry.__post_init__`, so this DNS behavior only applies to `public_ip`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Fixed: removed the dead self-assignment.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/client.py
mangelajo: Fixed: added `-> click.Group` return type annotation to `cli()`.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Correct catch. Fixed by validating at config time: `_validate_config` now rejects address entries that share a `vlan_id` but specify different `public_gateway` values, since they'd otherwise silently share/overwrite one PBR table's default route. Documented as a known limitation in the README (use separate VLAN IDs, or untagged PBR, for independent per-DUT gateways). Added regression tests for both the conflict-rejected and same-gateway-allowed cases.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: Right, I see w have no IPv6 support yet here. We can open an RFE when we need it. I know it's 2026... :-/

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/client.py
mangelajo: This part of the code was untouched by this patch, but since we are touching very close.. why not.

python/packages/jumpstarter-driver-dut-network/jumpstarter_driver_dut_network/driver.py
mangelajo: One could argue that then public_gateway needs to behave on the same way, but I will leave as a bug/RFE if somebody ever needs this. For the other one, I need it and I use it in the lab.

mangelajo:
Ok, I addressed a some of the comments, some others are limitations that got documented. The IPv6 support is out of scope for this patch. The general DUTNET base driver didn't support IPv6 at all... we can RFE this if we need it.

bennyz (approved):

PR #1069, opened 2026-09-05
JEP-0015: Multi-Exporter Leases and Inter-Exporter Port Forwarding

kirkbrauer:
Multi-device tests currently need separate exporter leases and manually configured connections. This JEP extends the existing `Lease` so a test can acquire a complete bench, such as a phone and head unit, and connect its devices even when their exporters run on different hosts.

An optional `spec.members[]` assigns roles such as `phone` and `headunit` to exporter selections. The controller selects distinct exporters and commits all required member claims in one status update, so a waiting lease holds no partial bench. Members share expiry and release, and each must pass the client's existing access policy. If a member fails after binding, the lease reports degradation and retains the surviving devices for diagnostics.

Drivers declare named ports as `provides` or `requires`. An optional `spec.forwards[]` connects those ports—for example, `headunit.rootcanal` to `phone.controller`—while addresses stay in exporter configuration. Exporters establish authenticated connections directly within a network zone when possible, falling back to the existing router. The client is outside the forwarding data path. The controller validates port direction and compares protocol tags when both endpoints supply them. Direct mode uses mTLS; Kubernetes 1.37 Pod Certificates and ClusterTrustBundles are recorded as a future workload-identity source, along with broader mTLS-based exporter authentication.

The proposal adds fields and one controller RPC without introducing a new lease resource or changing the `RouterService` protobuf. Router token validation is strengthened to pair only reciprocal endpoint claims under a unique per-forward subject. Existing single-exporter workflows remain supported; multi-member clients select devices by role through the CLI and Python API. A Mobly testbed export is also proposed for existing test suites.

Delivery is phased:

1. Bluetooth pairing between virtual devices in separate Pods.
2. Virtual phone projection over a forwarded TCP connection.
3. Physical phone and head unit testing across exporter hosts, with radio peers still within RF range.
4. Simulated Wi-Fi association and the full Bluetooth-to-Wi-Fi projection handover.

The early projection phase exercises the projection session but does not simulate the radio handover. Mixed physical/virtual radio benches are deferred. Atomic member claims do not fix the existing scheduling race between separate leases, and reconnecting a forward can still require protocol-specific recovery. Optional-member forwards now become explicitly `Disabled` when their endpoint is omitted, and an explicit one-member lease uses member status and the role-mapping client shape. Remaining review questions include launch/readiness ordering, listener allocation, synchronization, and Mobly endpoint lifetime.

The JEP includes design decisions, Mermaid diagrams, recorded prototype results, and acceptance criteria for these phases.

Validation: spelling, whitespace, YAML/Python example syntax, and local links checked. Sphinx/MyST parsed and rendered the JEP successfully; the repository-wide warnings-as-errors build remains blocked by an unrelated pre-existing broken authentication-document link.

AI Generated, Human Reviewed

### DIFF
--- a/docs/source/contributing/jeps/JEP-0015-multi-exporter-leases-port-forwarding.md
+++ b/docs/source/contributing/jeps/JEP-0015-multi-exporter-leases-port-forwarding.md
@@ -9,7 +9,7 @@
 | **Type**          | Standards Track                                          |
 | **Created**       | 2026-09-01                                               |
 | **Updated**       | 2026-09-04                                               |
-| **Discussion**    | *TBD (PR link)*                                          |
+| **Discussion**    | [PR #1069](https://github.com/jumpstarter-dev/jumpstarter/pull/1069) |
 | **Requires**      | JEP-0014                                                 |
 | **Supersedes**    |                                                          |
 | **Superseded-By** |                                                          |
@@ -1582,6 +1582,8 @@ These are outside the initial scope:
   Updated `Auto` to prefer direct connections for same-zone peers.
 - 2026-09-04: Consolidated rationale and prototype notes; clarified scope,
   implementation requirements, and remaining verification.
+- 2026-09-04: Submitted for discussion in
+  [PR #1069](https://github.com/jumpstarter-dev/jumpstarter/pull/1069).
 
 ## References
 
--- a/docs/source/contributing/jeps/JEP-0015-multi-exporter-leases-port-forwarding.md
+++ b/docs/source/contributing/jeps/JEP-0015-multi-exporter-leases-port-forwarding.md
@@ -5,10 +5,10 @@
 | **JEP**           | 0015                                                     |
 | **Title**         | Multi-Exporter Leases and Inter-Exporter Port Forwarding |
 | **Author(s)**     | @kirkbrauer (Kirk Brauer)                                |
-| **Status**        | Draft                                                    |
+| **Status**        | Discussion                                               |
 | **Type**          | Standards Track                                          |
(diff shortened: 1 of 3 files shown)


docs/source/contributing/jeps/JEP-0015-multi-exporter-leases-port-forwarding.md
```diff
@@ -0,0 +1,1860 @@
...
+1. **Reuse the `RouterService.Stream` RPC and forwarding path**, while
+   strengthening its JWT validation for peer-pair claims.
+2. **Add a peer-specific RPC** to `router.proto` with explicit A/B roles.
+3. **Reuse the current shared exporter subject unchanged**, relying only on
+   controller-side token issuance.
+
+**Decision:** Option 1 — no `router.proto` change, with authorization changes
+inside `RouterService.Stream`.
+
+**Rationale:** The current router uses the JWT `sub` as its pending-stream key;
```
bennyz, line 860: I assume we want something like this:
1. requires side accepts local socket
3. controller creates fresh connection UUID X
4. controller tells provide "open connection X"
5. both sides get router tokens with sub=X
6. router pairs those two streams

so we probably also want a connection/stream id like we have today,

 it seems the code comment in router.proto is wrong and sub is stream id

kirkbrauer: Yeah, I think we need to look into this more deeply, I was also looking at inter-pod mTLS, but that requires newer k8s versions. Maybe we can implement it though for this as our first trial run of using mTLS?

PR #1071, opened 2026-09-05
feat(cli): add driver configuration schema discovery

kirkbrauer:
## Summary

Add `jmp driver schema [NAMES...] [-o json|yaml|name]` for environment-aware exporter configuration editing. It reports locally installed driver config keys, types, required fields, referenced definitions, client class paths, and package metadata without instantiating drivers or opening a lease itself.

## Contract and review fixes

- JSON/YAML return `{drivers: [...]}`; tables and name output are also supported.
- Driver loading/schema errors are per-entry. Failed JSON Schema generation retains best-effort dataclass constructor keys, with `error` set.
- Keep Python-level import/schema-hook output on stderr rather than tolerating prefixes before JSON.
- Preserve nested definitions and handle recursive schemas whose root is a `$ref`.
- Exclude common Driver fields and runtime-only (`init=False`) fields from config keys.
- Reject unknown filters before loading, including mixed valid/invalid requests; only selected drivers are loaded.
- Replace environment-dependent/vacuous tests with hermetic entry-point fixtures and direct stdout parsing.
- Document that imports and metadata/schema hooks execute trusted local package code; discovery is not a sandbox.

The VS Code extension keeps this API behind a dedicated default-off `experimentalDriverSchemas` flag in addition to its general experimental CLI opt-in. Static YAML schemas and manual allow-list entry remain available without it.

## Validation

- `make pkg-test-jumpstarter-cli-driver`: 22 passed
- `make pkg-test-jumpstarter-cli`: 257 passed
- `make lint-fix`: passed
- `make pkg-ty-jumpstarter-cli-driver`: passed
- Local installed-driver smoke: TcpNetwork and Xcp produce typed schemas; Qemu produces constructor keys plus its expected unsupported-type warning. Stdout parses directly as JSON.
- Extension: 227 unit tests, 73 integration tests, type-check and lint pass.

---
AI generated, human reviewed


python/packages/jumpstarter-cli-driver/jumpstarter_cli_driver/driver.py
```diff
@@ -36,6 +41,174 @@ def rich_add_names(self, names):
...
+        cls = entry_point.load()
+    except Exception as e:
+        # One uninstallable driver must not sink the whole listing.
+        entry.error = f"{type(e).__name__}: {e}"
+        return entry
+
+    entry.description = _first_docstring_line(cls)
+    entry.client = _client_class_path(cls)
+    try:
+        schema = TypeAdapter(cls).json_schema()
```
bennyz, line 158: so it seems that it would fail generating for qemu, becasue it has a TemporaryDirectory field

kirkbrauer: Oh yeah, that might be an issue...I will take a look, when testing, it did seem to work OK for most drivers, but I wasn't able to try them all, maybe I can make an audit script

PR #1072, opened 2026-09-06
feat: add cuttlefish provisioner for dynamic exporters

bennyz:
Adds `cuttlefish.jumpstarter.dev` as an ExporterSet provisioner, using the VirtualTargetClass and ExporterSet lifecycle defined by JEP-0014. Users lease a virtual Android target through the existing Jumpstarter workflow, then boot and control it through the Cuttlefish driver.

Each Pod contains the Jumpstarter exporter, a Cuttlefish Host Orchestrator runtime sidecar, and socat relays for netsim and Bluetooth HCI. The provisioner enriches Cuttlefish, netsim, and bt-peer configuration, supports fetched images or a prewarmed PVC, and inherits class scheduling and image overrides. Runtime privilege configuration is explicit. The exporter is the main container with `restartPolicy: Never`, supporting `ExitAndReplace` teardown.


kirkbrauer:
@bennyz @mangelajo 

Could we consider a follow-up exec mode based on [Podcvd’s approach](https://github.com/google/android-cuttlefish/blob/main/container/src/podcvd/internal/main.go), which controls devices by executing `cvd` inside their container? This would follow our [QEMU provisioner pattern](https://github.com/jumpstarter-dev/jumpstarter/blob/main/controller/internal/exporterset/provisioners/qemu/qemu.go): inject `launcher_socket` and have the Cuttlefish driver invoke commands through `jumpstarter-exec`.

That would give us consistent runtime control and a place to enforce one CVD per Pod, while preserving the existing HTTP mode for externally managed hosts. We could retain this PR’s image preparation, scheduling and lifecycle work. Podcvd currently still uses the orchestration image, so removing Host Orchestrator could be a separate step after validating the CLI backend.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+			Image:           runtimeImage,
+			ImagePullPolicy: runtimePullPolicy,
+			Command: []string{
+				"bash", "-c", permissionCommand,
+			},
+			VolumeMounts: volumeMounts,
+		},
+		corev1.Container{
+			Name:            "cuttlefish",
+			Image:           runtimeImage,
```
kirkbrauer: Restrict access to the runtime API

The driver's loopback address does not restrict Host Orchestrator or nginx listeners. Without ingress restrictions, other Pods can bypass Jumpstarter leases and control the device. Could we add a NetworkPolicy or bind all API/proxy listeners to loopback? Verify the actual image's listeners against [upstream defaults](https://github.com/google/android-cuttlefish/blob/6d0d2284d06c86c0bc578d595b7fbef2f6d75388/frontend/debian/cuttlefish-orchestration.cuttlefish-host_orchestrator.default).

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+			ImagePullPolicy: runtimePullPolicy,
+			RestartPolicy:   &runtimeRestart,
+			SecurityContext: &corev1.SecurityContext{Privileged: boolPtr(runtimePrivileged), RunAsUser: &runAsRoot},
```
kirkbrauer: Document the initial OpenShift SCC requirements

A privileged runtime is fine initially, but `runtime_privileged: true` does not grant SCC admission. Please document a scoped SCC grant, preferably for a dedicated workload service account, and test init containers plus runtime boot. Reject `false` until its device permissions and capabilities are supported. [SCC reference](https://docs.redhat.com/en/documentation/openshift_container_platform/4.19/html/authentication_and_authorization/managing-pod-security-policies).

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+		},
+		corev1.Container{
+			Name:            "cuttlefish-relay",
+			Image:           relayImage,
+			ImagePullPolicy: corev1.PullIfNotPresent,
+			RestartPolicy:   &runtimeRestart,
+			Command: []string{
+				"sh", "-c",
+				fmt.Sprintf(
+					"socat TCP-LISTEN:%d,bind=127.0.0.1,fork,reuseaddr TCP:127.0.0.1:7681 & socat TCP-LISTEN:%d,bind=127.0.0.1,fork,reuseaddr TCP:127.0.0.1:7300 & wait",
```
kirkbrauer: Validate relay ports

`hci_relay_port: 17681` makes both relays bind the same port; `netsim_relay_port: 7681` forwards to itself and occupies netsim's port. Both configurations are accepted, and the API health check misses relay failures. Could we reject duplicate, out-of-range, and conflicting service ports?

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+		Spec: corev1.PodSpec{
+			RestartPolicy:  corev1.RestartPolicyNever,
+			InitContainers: initContainers,
+			Containers:     []corev1.Container{exporterContainer},
+			Volumes: []corev1.Volume{
+				imageVolume,
+				{Name: "cvd-state", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{SizeLimit: &stateSize}}},
+				{Name: "android-tmp", VolumeSource: corev1.VolumeSource{EmptyDir: &corev1.EmptyDirVolumeSource{SizeLimit: &tmpSize}}},
+				deviceVolume("kvm", "/dev/kvm"),
+				deviceVolume("vhost-vsock", "/dev/vhost-vsock"),
```
kirkbrauer: Use private userspace VSOCK

Could we explicitly enable `vhost_user_vsock=true` with a compatible runtime/guest/VMM and private per-Pod Unix sockets? Kernel VSOCK in a shared namespace can collide when every exporter defaults to CID 3; distinct CIDs alone do not isolate tenants. Test two colocated Pods with identical CIDs. Retaining HTTP is fine; private sockets do not contain a compromised privileged runtime. [Upstream backend](https://github.com/google/android-cuttlefish/blob/6d0d2284d06c86c0bc578d595b7fbef2f6d75388/base/cvd/cuttlefish/host/commands/run_cvd/launch/vhost_device_vsock.cpp).

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+		corev1.Container{
+			Name:            "cuttlefish-relay",
+			Image:           relayImage,
+			ImagePullPolicy: corev1.PullIfNotPresent,
+			RestartPolicy:   &runtimeRestart,
+			Command: []string{
+				"sh", "-c",
+				fmt.Sprintf(
+					"socat TCP-LISTEN:%d,bind=127.0.0.1,fork,reuseaddr TCP:127.0.0.1:7681 & socat TCP-LISTEN:%d,bind=127.0.0.1,fork,reuseaddr TCP:127.0.0.1:7300 & wait",
+					netsimPort, hciPort,
```
kirkbrauer: Verify simulator VSOCK compatibility

[Netsim](https://github.com/google/android-cuttlefish/blob/6d0d2284d06c86c0bc578d595b7fbef2f6d75388/base/cvd/cuttlefish/host/commands/run_cvd/launch/netsim_server.cpp#L160) propagates the userspace VSOCK flag to its proxies; [standalone RootCanal](https://github.com/google/android-cuttlefish/blob/6d0d2284d06c86c0bc578d595b7fbef2f6d75388/base/cvd/cuttlefish/host/commands/run_cvd/launch/root_canal.cpp#L68) currently omits it. Please verify the packaged runtime and test supported BT modes across two colocated Pods before relying on private sockets for all simulator traffic.

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+	if envConfig == nil {
+		envConfig = map[string]interface{}{}
+	}
+	common, _ := envConfig["common"].(map[string]interface{})
+	if common == nil {
+		common = map[string]interface{}{}
+	}
+	setDefault(common, "host_package", fetchPath)
+	envConfig["common"] = common
+	instances, _ := envConfig["instances"].([]interface{})
```
kirkbrauer: Enforce one CVD per Pod

`env_config.instances` accepts multiple entries, but only the first is enriched and tracked by the driver. Could we reject multiple instances and account for multiple Cuttlefish drivers and the exported `create_cvd()` operation when enforcing one CVD per Pod?

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+			pod.Spec.NodeSelector = maps.Clone(vtc.Spec.Scheduling.NodeSelector)
+		}
+		if vtc.Spec.Scheduling.Tolerations != nil {
+			pod.Spec.Tolerations = append([]corev1.Toleration(nil), vtc.Spec.Scheduling.Tolerations...)
+		}
+		if vtc.Spec.Scheduling.Resources != nil {
+			// Cuttlefish runs in the runtime sidecar; the exporter and relay remain unbounded.
+			for i := range pod.Spec.InitContainers {
+				if pod.Spec.InitContainers[i].Name == "cuttlefish" {
+					pod.Spec.InitContainers[i].Resources = *vtc.Spec.Scheduling.Resources.DeepCopy()
```
kirkbrauer: Validate runtime memory allocation

The default 8192 MB guest accepts both missing CPU/memory requests and a 1 GiB memory limit. Please require or derive suitable requests and reject incompatible memory limits, including VMM/simulator overhead. CPU overcommit can remain configurable.

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+	// Run in the exporter image so the check uses the same network namespace and Python runtime as jmp.
+	healthURL := fmt.Sprintf("http://127.0.0.1:%d/_debug/statusz", parameterInt(mergedParameters, "host_orchestrator_port", hostOrchestratorPort))
+	healthCheck := "import urllib.request; urllib.request.urlopen(" + fmt.Sprintf("%q", healthURL) + ", timeout=3).close()"
+	pod.Spec.InitContainers = append(pod.Spec.InitContainers, corev1.Container{
+		Name: "wait-for-cuttlefish", Image: exporterImage, ImagePullPolicy: exporterPullPolicy,
+		SecurityContext: exporterContainer.SecurityContext.DeepCopy(),
+		Command:         []string{"python3", "-c", "import time, urllib.request\nfor attempt in range(60):\n try:\n  " + healthCheck + "\n  break\n except Exception:\n  time.sleep(5)\nelse:\n raise SystemExit('Host Orchestrator did not become ready')"},
+	})
+	// With restartPolicy Never, a failed liveness check ends the exporter and lets ExitAndReplace recycle the Pod.
+	pod.Spec.Containers[0].LivenessProbe = &corev1.Probe{
```
kirkbrauer: Handle guest and simulator failures

[`/_debug/statusz`](https://github.com/google/android-cuttlefish/blob/6d0d2284d06c86c0bc578d595b7fbef2f6d75388/frontend/src/host_orchestrator/orchestrator/controller.go#L900) always returns 200 while the API is serving, so this probe misses VMM/netsim failures. Please test those failures and runtime restarts during a lease, and define recovery behavior. Health checks must still allow intentional power-off.

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+					pod.Spec.InitContainers[i].Resources = *vtc.Spec.Scheduling.Resources.DeepCopy()
+					break
+				}
+			}
+		}
+	}
+
+	if imageClaim != "" {
+		pod.Spec.Volumes = append(pod.Spec.Volumes, corev1.Volume{Name: "image-source", VolumeSource: corev1.VolumeSource{
+			PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{ClaimName: imageClaim, ReadOnly: true},
```
kirkbrauer: Document image PVC access modes

A read-only mount does not make a single-node PVC usable across nodes, and the source remains mounted after copying. Please document compatible shared storage or placement requirements for multi-node pools. [Access modes](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#access-modes).

AI Generated, Human reviewed

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,557 @@
...
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	ProvisionerName = "cuttlefish.jumpstarter.dev"
+
+	DefaultExporterImage = "quay.io/jumpstarter-dev/jumpstarter:latest"
+	DefaultRuntimeImage  = "us-docker.pkg.dev/android-cuttlefish-artifacts/cuttlefish-orchestration/cuttlefish-orchestration:stable"
```
kirkbrauer, line 45: Pin a reproducible runtime/build pair

Nonblocking: Maybe we can provide an example pinning the runtime digest and Android build ID. Mutable `stable` and build-branch defaults can produce different replacement exporters and make VSOCK compatibility failures difficult to reproduce.

AI Generated, Human reviewed

bennyz:
> @bennyz @mangelajo
> 
> Could we consider a follow-up exec mode based on [Podcvd’s approach](https://github.com/google/android-cuttlefish/blob/main/container/src/podcvd/internal/main.go), which controls devices by executing `cvd` inside their container? This would follow our [QEMU provisioner pattern](https://github.com/jumpstarter-dev/jumpstarter/blob/main/controller/internal/exporterset/provisioners/qemu/qemu.go): inject `launcher_socket` and have the Cuttlefish driver invoke commands through `jumpstarter-exec`.
> 
> That would give us consistent runtime control and a place to enforce one CVD per Pod, while preserving the existing HTTP mode for externally managed hosts. We could retain this PR’s image preparation, scheduling and lifecycle work. Podcvd currently still uses the orchestration image, so removing Host Orchestrator could be a separate step after validating the CLI backend.

I'll look into that

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: Added a deny-all ingress NetworkPolicy. curl from Pods in two other namespaces to 2081, 2080 and 15037 times out, loopback still works

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: relayPorts now rejects out-of-range, reserved runtime ports (2080/2081, 6520-6620, 7300-7303, 7681, 15550-15560 and others) and duplicates. Verified hci_relay_port: 17681 and netsim_relay_port: 7681 are both rejected with a clear error

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: vm.crosvm.vhost_user_vsock is forced to "true". Tested two Pods on one kind node, both guests CID 3, both booted concurrently. Each crosvm uses a private vhost.socket in its own container, no collision and no liveness failures. Bluetooth traffic flowed on both at the same time without cross-talk

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: netsim_bt is forced true and standalone RootCanal is rejected. Verified at runtime: start_netsim=true, start_rootcanal=false in the generated config, a bumble peer on the HCI relay connected to the guest with BR/EDR ACL up in logcat, and netsim packet counters rose on both chips. Repeated on a second colocated Pod, and neither netsim instance saw the other Pod's peer.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: done

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: Memory budget is now guest memory plus runtime_memory_overhead_mb (default 2048). Requests are derived when missing, and a request or limit below the budget is rejected

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: liveness now runs health.py, which checks the HO API, a per-start runtime ID, CVD inventory and status, and simulator/relay listeners. Intentional power off keeps the Pod healthy. The failed Pod and Exporter stay while leased and are replaced under ExitAndReplace

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: Added a digest-pinned example (runtime, relay, exporter images and Android build ID) to the README

controller/deploy/operator/config/rbac/role.yaml
```diff
@@ -171,6 +171,17 @@ rules:
...
+  - networking.k8s.io
+  resources:
+  - networkpolicies
+  verbs:
+  - create
+  - get
+  - list
+  - patch
+  - update
+  - watch
```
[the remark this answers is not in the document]
bennyz: The grant is required. The operator creates a namespaced Role for the ExporterSet controller that includes networkpolicies, and Kubernetes RBAC escalation prevention requires the granter to hold the permission

bennyz:
> > @bennyz @mangelajo
> > Could we consider a follow-up exec mode based on [Podcvd’s approach](https://github.com/google/android-cuttlefish/blob/main/container/src/podcvd/internal/main.go), which controls devices by executing `cvd` inside their container? This would follow our [QEMU provisioner pattern](https://github.com/jumpstarter-dev/jumpstarter/blob/main/controller/internal/exporterset/provisioners/qemu/qemu.go): inject `launcher_socket` and have the Cuttlefish driver invoke commands through `jumpstarter-exec`.
> > That would give us consistent runtime control and a place to enforce one CVD per Pod, while preserving the existing HTTP mode for externally managed hosts. We could retain this PR’s image preparation, scheduling and lifecycle work. Podcvd currently still uses the orchestration image, so removing Host Orchestrator could be a separate step after validating the CLI backend.
> 
> I'll look into that

made draft in https://github.com/jumpstarter-dev/jumpstarter/pull/1082
i can't stack on top since i've already created this PR from my fork

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -125,7 +156,51 @@ def _wait_for_operation(self, op_name: str, timeout: float = 300) -> dict:
...
+                self._write_health()
+                raise
+            if isinstance(result, dict) and path == "/cvds":
+                cvds = result.get("cvds", [])
+                if len(cvds) == 1:
+                    self._cvd_group = cvds[0].get("group")
+                    self._cvd_name = cvds[0].get("name")
+            self._health.update(
+                state=target_state, group=self._cvd_group or self.group, name=self._cvd_name or self.name,
+            )
```
raballew: [CRITICAL] Four Python files in this package fail the ruff format check: `driver.py`, `driver_test.py`, `health.py`, and `health_test.py`. Running `ruff format --check --config python/pyproject.toml python/packages/jumpstarter-driver-cuttlefish/` exits with "4 files would be reformatted". A representative violation is the multi-argument `_health.update(state=..., group=..., name=...)` call here on a single line where ruff expects expansion. Run `ruff format --config python/pyproject.toml python/packages/jumpstarter-driver-cuttlefish/` and commit the result.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+func (p *Provisioner) resolveImage(image string) string {
+	if p.Version == "" || p.Version == "dev" || strings.Contains(p.Version, "-g") {
+		return image
+	}
+	version := strings.TrimPrefix(p.Version, "v")
+	if base, ok := strings.CutSuffix(image, ":latest"); ok {
+		return base + ":" + version
+	}
+	return image
+}
```
raballew: [CRITICAL] The release version substitution path in `resolveImage` has no test coverage. Every test in `cuttlefish_test.go` uses `New("dev")`, which short-circuits at the `p.Version == "dev"` check. The branch that strips the `v` prefix and replaces `:latest` with the release version (lines 118-120) determines the container images deployed in every production release and is never executed by any test. The QEMU provisioner has six `TestResolveImage_*` tests for equivalent logic. Consider adding table-driven tests covering: `"dev"` passthrough, `"v1.2.3"` with `:latest` image, `"v1.2.3"` with a pinned digest, and a git-hash version containing `-g`.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+			ProbeHandler: corev1.ProbeHandler{Exec: &corev1.ExecAction{
+				Command: []string{"python3", "-m", "jumpstarter_driver_cuttlefish.health", healthStatePath},
+			}},
+			PeriodSeconds: 10, TimeoutSeconds: 10, FailureThreshold: 6,
+		},
+	}
+}
+
+// initContainers stages images, fixes ownership, starts the runtime as a
+// native sidecar and gates the exporter on Host Orchestrator readiness.
```
raballew: [HIGH] The liveness probe can kill the container before `boot_timeout` elapses. The probe is configured with `FailureThreshold:6, PeriodSeconds:10`, giving a 60-second kill budget, but `boot_timeout` defaults to 300 seconds. `CvdPower._on()` writes `state="running"` before `_wait_boot(300)` returns, so the probe sees `state="running"`, queries `/cvds`, and while the Android guest is still `status="Starting"`, `check()` raises. After 60 seconds of probe failures the container is killed 240 seconds before the driver's own timeout fires. One fix: keep the state as `"transition"` until `_wait_boot()` succeeds. Another: set `FailureThreshold` to `ceil(boot_timeout / PeriodSeconds)` (30 for defaults) and document the coupling explicitly.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -125,7 +156,51 @@ def _wait_for_operation(self, op_name: str, timeout: float = 300) -> dict:
...
+            # Probes allow bounded transitions, then fail closed if the operation
+            # hangs or leaves the target's state unknown.
+            self._health.update(state="transition", deadline=time.monotonic() + timeout + 30)
+            self._write_health()
+            try:
+                result = self._perform_operation(method, path, data, timeout)
+            except Exception:
+                self._health["state"] = "failed"
+                self._write_health()
+                raise
```
raballew: [HIGH] If `_write_health()` raises inside this `except` block (e.g. disk full or `os.replace` failure), the bare `raise` below never runs and the original exception is swallowed. The caller receives the I/O error from `_write_health()` instead of the original `CuttlefishError`, and the health file is left un-updated. Wrapping the `_write_health()` call in a nested `try/except Exception: pass` ensures the original exception always propagates regardless of what happens during health state persistence.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish_test.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+	if exporter.Command[6] != hostOrchestratorURL {
+		t.Errorf("exporter endpoint = %q", exporter.Command[6])
+	}
+	if pod.Spec.RestartPolicy != corev1.RestartPolicyNever {
+		t.Error("Pod must not restart the exporter in place")
+	}
+	if !hasVolume(pod.Spec.Volumes, "kvm", "/dev/kvm") || !hasVolume(pod.Spec.Volumes, "tun", "/dev/net/tun") {
+		t.Fatalf("device volumes missing: %#v", pod.Spec.Volumes)
+	}
```
raballew: [MEDIUM] `TestRenderPod` asserts nine distinct behaviors under a single name: init-container order, sidecar restart policy, privileged security context, UID-0, single main container, HOME env, endpoint URL argument position, pod restart policy, and device volume. A failure anywhere stops the rest from running and does not identify which invariant broke. Consider splitting into focused tests with behavioral names like `TestRenderPod_initContainerOrder`, `TestRenderPod_runtimeContainer_isNativeSidecar`, `TestRenderPod_restartPolicy_isNever`, etc., or at least using `t.Run` subtests with descriptive labels.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+		},
+	}
+	if vtc.Spec.Scheduling != nil {
+		if vtc.Spec.Scheduling.NodeSelector != nil {
+			pod.Spec.NodeSelector = maps.Clone(vtc.Spec.Scheduling.NodeSelector)
+		}
+		if vtc.Spec.Scheduling.Tolerations != nil {
+			pod.Spec.Tolerations = append([]corev1.Toleration(nil), vtc.Spec.Scheduling.Tolerations...)
+		}
+	}
```
raballew: [MEDIUM] The VTC scheduling propagation here (copying `NodeSelector` and `Tolerations` to the Pod spec) has no test coverage. `grep -n 'NodeSelector\|Toleration'` in `cuttlefish_test.go` returns no output; every test passes a VTC with nil or resources-only Scheduling. A bug in this copy would silently place Pods on wrong nodes. Consider adding a test that passes a VTC with `NodeSelector: {"disktype": "ssd"}` and `Tolerations: [{Key: "gpu", Effect: NoSchedule}]` and asserts those values appear verbatim on the Pod spec.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -125,7 +156,51 @@ def _wait_for_operation(self, op_name: str, timeout: float = 300) -> dict:
...
+            if isinstance(result, dict) and path == "/cvds":
+                cvds = result.get("cvds", [])
+                if len(cvds) == 1:
+                    self._cvd_group = cvds[0].get("group")
+                    self._cvd_name = cvds[0].get("name")
+            self._health.update(
+                state=target_state, group=self._cvd_group or self.group, name=self._cvd_name or self.name,
+            )
+            self._write_health()
+            return result
```
raballew: [MEDIUM] State-machine transition logic is inlined here inside `_do_operation`, which also manages locks, health I/O, and error recovery. `target_state` is determined by matching `path` and `method` strings in place: adding a new operation requires editing this block, and omitting it silently leaves state wrong. Consider extracting `_infer_target_state(method: str, path: str) -> str` to own the transition table and make it independently testable.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+
+def wait_ready(url: str, attempts: int = 60, interval: float = 5) -> None:
+    """Startup gate: block until Host Orchestrator answers, so the exporter never registers early."""
+    for _ in range(attempts):
+        try:
+            with urllib.request.urlopen(f"{url}/_debug/statusz", timeout=3):
+                return
+        except Exception:
+            time.sleep(interval)
+    raise RuntimeError(f"Host Orchestrator at {url} did not become ready")
```
raballew: [MEDIUM] `wait_ready` calls `time.sleep(interval)` on every loop iteration including after the last failed attempt, before raising `RuntimeError`. This adds one full interval (default 5 seconds) of unnecessary delay between the final probe failure and the container exiting with an error. Moving the sleep inside an `if i < attempts - 1:` guard would skip it on the last iteration.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+
+func (p *Provisioner) RenderNetworkPolicy(es *virtualtargetv1alpha1.ExporterSet) *networkingv1.NetworkPolicy {
+	return &networkingv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{Name: "cuttlefish-" + string(es.UID), Namespace: es.Namespace},
+		Spec: networkingv1.NetworkPolicySpec{
+			PodSelector: metav1.LabelSelector{MatchLabels: map[string]string{isolationLabel: string(es.UID)}},
+			PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress},
+		},
+	}
+}
```
raballew: [MEDIUM] The NetworkPolicy here sets `PolicyTypes: []networkingv1.PolicyType{networkingv1.PolicyTypeIngress}` but omits `PolicyTypeEgress`. The Cuttlefish pod runs a sidecar with `Privileged: true` mounting `/dev/kvm`, `/dev/vhost-net`, and `/dev/net/tun`. Without an egress policy, the pod can reach the Kubernetes API server and cloud metadata endpoints. Consider adding `PolicyTypeEgress` with explicit allow rules for DNS (UDP 53) and the Jumpstarter controller endpoint, and denying all other egress.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -125,7 +156,51 @@ def _wait_for_operation(self, op_name: str, timeout: float = 300) -> dict:
...
+                self._validate_creation(data)
+            state = self._health.get("state", "off")
+            target_state = state
+            if path == "/reset" or method == "DELETE" or path.endswith("/:stop"):
+                target_state = "off"
+            elif path == "/cvds" or path.endswith(("/:start", "/:restart", "/:powerwash")):
+                target_state = "running"
+            # Probes allow bounded transitions, then fail closed if the operation
+            # hangs or leaves the target's state unknown.
+            self._health.update(state="transition", deadline=time.monotonic() + timeout + 30)
```
raballew: [MEDIUM] `time.monotonic()` is stored in the health JSON by the driver process and compared by the liveness probe process. Python's documentation states the reference point of `monotonic()` is undefined across processes. On CPython/Linux this works because `CLOCK_MONOTONIC` is shared, but any container runtime or kernel configuration that resets the monotonic clock would cause the deadline to appear already expired at operation start. Using `time.time()` (wall-clock) for the cross-process deadline would give a well-defined reference. The same issue affects the comparison in `health.py:31`.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+def listening_ports() -> set[int]:
+    # Opening an HCI connection can create a simulator peer. Inspect the shared
+    # Pod network namespace instead of disturbing active Bluetooth sessions.
+    listeners = set()
+    for table in ("/proc/net/tcp", "/proc/net/tcp6"):
+        for line in Path(table).read_text().splitlines()[1:]:
+            fields = line.split()
+            if fields[3] == "0A":
+                listeners.add(int(fields[1].rsplit(":", 1)[1], 16))
+    return listeners
```
raballew: [MEDIUM] `listening_ports()` reads both `/proc/net/tcp` and `/proc/net/tcp6` unconditionally with `Path(table).read_text()`. On a kernel with `ipv6.disable=1`, `/proc/net/tcp6` does not exist and `read_text()` raises `FileNotFoundError`. This propagates through `check()` and exits 1, failing the liveness probe on every invocation. After 6 failures (60 s) the Pod is killed and replaced, failing identically on the same node. Wrapping the read in `try/except FileNotFoundError: continue` would fix this.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+			},
+		})
+	}
+	if storage.fetchImages {
+		containers = append(containers, corev1.Container{
+			Name: "fetch-images", Image: img.runtime, ImagePullPolicy: img.runtimePull,
+			Command:      []string{"cvd", "fetch", "--default_build=" + storage.build, "--target_directory=" + fetchPath},
+			Resources:    storageReservation(storage.budget),
+			VolumeMounts: []corev1.VolumeMount{{Name: "cvd-images", MountPath: fetchPath}},
+		})
```
raballew: [MEDIUM] The `copy-images` and `fetch-images` init containers here have no `SecurityContext` field. Sibling containers explicitly set `RunAsUser: &root`. Without a SecurityContext these containers inherit image defaults (typically root) with no capability restrictions and no `AllowPrivilegeEscalation: false`. Both containers only need to read and write emptyDir volumes. Consider adding an explicit SecurityContext with `AllowPrivilegeEscalation: boolPtr(false)` and `Capabilities: {Drop: ["ALL"]}`, plus a non-root UID if the underlying tool supports it.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+		meta.Labels = map[string]string{}
+	}
+	meta.Labels[isolationLabel] = string(exporterSet.UID)
+	return meta
+}
+
+func exporterSecurityContext() *corev1.SecurityContext {
+	uid := exporterNonRootUID
+	return &corev1.SecurityContext{RunAsUser: &uid, RunAsNonRoot: boolPtr(true)}
+}
```
raballew: [MEDIUM] `exporterSecurityContext()` sets `RunAsUser` and `RunAsNonRoot: true` but leaves `AllowPrivilegeEscalation` unset, which defaults to `true`. This function is used for both the `exporter` main container and the `wait-for-cuttlefish` gate container. In a privileged namespace, any setuid binary in the exporter image could allow the non-root process to escalate to root. Adding `AllowPrivilegeEscalation: boolPtr(false)` and `Capabilities: {Drop: ["ALL"]}` would close this.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+	}
+	return append(containers,
+		corev1.Container{
+			Name: "fix-cuttlefish-permissions", Image: img.runtime, ImagePullPolicy: img.runtimePull,
+			Command: []string{"bash", "-c", "mkdir -p " + cvdStatePath + " " + androidTmpPath +
+				" && chown -R httpcvd:httpcvd " + cvdStatePath + " " + androidTmpPath + " " + fetchPath},
+			// chown needs UID 0 regardless of the runtime image's default user.
+			SecurityContext: &corev1.SecurityContext{RunAsUser: &root},
+			VolumeMounts:    stateMounts,
+		},
```
raballew: [MEDIUM] The `fix-cuttlefish-permissions` init container runs as `RunAsUser: 0` but has no `AllowPrivilegeEscalation` restriction and no `Capabilities.Drop`. A root process with the full default capability set (including `CAP_NET_RAW`, `CAP_SYS_PTRACE`, `CAP_DAC_READ_SEARCH`) in a privileged namespace is a wider attack surface than needed. The `chown -R` and `mkdir -p` operations require only `CAP_CHOWN` and `CAP_DAC_OVERRIDE`. Consider setting `AllowPrivilegeEscalation: false` and `Capabilities: {Drop: ["ALL"], Add: ["CHOWN", "DAC_OVERRIDE"]}`.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish_test.go
```diff
@@ -0,0 +1,465 @@
...
+			t.Fatalf("accepted %v", params)
+		}
+	}
+	es.Spec.RecycleStrategy = virtualtargetv1alpha1.RecycleStrategyInPlaceReuse
+	if _, err := New("dev").RenderPod(context.Background(), es, &virtualtargetv1alpha1.VirtualTargetClass{}, map[string]interface{}{
+		"fetch_images": true, "runtime_privileged": true, "service_account_name": "cuttlefish-runtime",
+	}, nil, nil); err == nil {
+		t.Fatal("accepted InPlaceReuse")
+	}
+}
```
raballew: [MEDIUM] `TestPodIsolation` chains five independent behaviors in a single test function: NetworkPolicy selector/ingress, service account name and token automount, absence of `/dev/vhost-vsock` hostPath volume, service account configuration rejection cases, and `InPlaceReuse` recycle strategy rejection. A failure in the first assertion stops the rest from running, making it hard to tell which invariant broke. Consider splitting into `TestNetworkPolicySpec`, `TestServiceAccountIsolation`, `TestNoKernelVsockDevice`, `TestServiceAccountValidation`, and `TestInPlaceReuseRejected`.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -295,7 +370,7 @@ def create_cvd(self, config_json: str) -> str:
```
raballew: [MEDIUM] `get_webrtc_url()` is an exported method with two branches (explicit URL override and constructed default), but neither branch is tested. `grep -n 'get_webrtc_url\|webrtc' driver_test.py` returns no output. Consider adding `test_get_webrtc_url_returns_default` and `test_get_webrtc_url_returns_explicit_override` to cover both code paths.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish_test.go
```diff
@@ -0,0 +1,465 @@
+package cuttlefish
```
raballew: [MEDIUM] This file begins directly with a `package` declaration and is missing the Apache-2.0 license header. All comparable existing Go test files (`qemu_test.go`, `lease_controller_test.go`) carry the standard header block. `controller/internal/exporterset/networkpolicy_test.go` has the same issue. Please prepend the Apache-2.0 header matching `controller/internal/exporterset/provisioners/qemu/qemu_test.go` to both files.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -426,12 +508,21 @@ def on(self) -> None:  # noqa: C901
                         )
                     break
 
+        if self.parent.managed:
+            with self.parent._operation_lock:
+                self.parent._health.update(state="running", group=self.parent._cvd_group or self.parent.group,
+                                           name=self.parent._cvd_name or self.parent.name)
+                self.parent._write_health()
```
raballew: [MEDIUM] The block that writes `state="running"` after a successful managed `power.on()` call has no test. Existing managed `power.on()` tests only exercise rejection paths that raise before reaching this code. No test verifies that a successful managed `power.on()` leaves the health file with `state="running"` and the correct `group` and `name`. Consider adding a test that patches `_request`/`_perform_operation` to simulate successful CVD creation and asserts the resulting health file contents.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -39,16 +42,44 @@ class Cuttlefish(Driver):
...
+    _health: dict = field(default_factory=dict, init=False, repr=False)
     _cvd_group: str | None = field(default=None, init=False, repr=False)
     _cvd_name: str | None = field(default=None, init=False, repr=False)
 
     def __post_init__(self):
         if hasattr(super(), "__post_init__"):
             super().__post_init__()
+        if self.managed:
+            self._validate_managed_config()
+            self._health = json.loads(Path(self.health_state_path).read_text())
```
raballew: [MEDIUM] `_validate_managed_config()` only checks that `health_state_path` is a non-empty string; it does not verify the file exists. If `health.initialize()` was skipped, `read_text()` here raises a bare `FileNotFoundError` rather than a typed error with an actionable message. Wrapping the read in `try/except OSError` and raising a descriptive `CuttlefishError` (e.g. `"managed Cuttlefish health state not found at ...; ensure --run-exporter ran before jmp"`) would make the failure much easier to diagnose.

controller/internal/exporterset/reconciler.go
```diff
@@ -173,6 +175,10 @@ func (r *ExporterSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		return ctrl.Result{}, nil
 	}
 
+	if err := r.syncNetworkPolicy(ctx, &exporterSet); err != nil {
+		return ctrl.Result{}, err
+	}
```
raballew: [LOW] `syncNetworkPolicy` is called here without checking `exporterSet.DeletionTimestamp`. When an ExporterSet is terminating, `controllerutil.CreateOrUpdate` runs on the NetworkPolicy while owner-reference GC has already started, producing a stale API round-trip. Adding an early return when `DeletionTimestamp` is non-zero would avoid this.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+    url = state["url"]
+    with urllib.request.urlopen(f"{url}/_debug/statusz", timeout=2):
+        pass
+    if state["state"] == "off":
+        return
+    if state["state"] == "transition" and time.monotonic() < state["deadline"]:
+        return
+    if state["state"] != "running":
+        raise RuntimeError("Cuttlefish operation failed or timed out")
+    with urllib.request.urlopen(f"{url}/cvds", timeout=2) as response:
```
raballew: [LOW] The URL read from the health state file is passed directly to `urllib.request.urlopen()`, which also accepts `file://`, `ftp://`, and `data://` schemes. If the file is writable by the exporter process, a compromised write could redirect the liveness probe to an unintended target. Validating that `urlparse(url).scheme in ("http", "https")` before calling `urlopen` would prevent this.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish_test.go
```diff
@@ -0,0 +1,465 @@
...
+		t.Fatal("missing runtime failure detection")
+	}
+	if !hasMount(pod.Spec.Containers[0].VolumeMounts, "cvd-state", runtimeIDMount) {
+		t.Fatal("exporter cannot read the runtime marker")
+	}
+}
+
+func TestRenderPod_storageBudgets(t *testing.T) {
+	pod := renderTestPod(t, map[string]interface{}{"fetch_images": true, "storage": map[string]interface{}{"imageSize": "8Gi", "stateSize": "4Gi", "tmpSize": "2Gi"}})
+	total := resource.MustParse("1Gi")
```
raballew: [LOW] `total := resource.MustParse("1Gi")` in this test mirrors the production overhead literal in `resolveStorageConfig`. If the overhead constant changes, both the production code and this test must be updated independently, and the mismatch may not be obvious. Computing the expected sum from explicit volume sizes and verifying the overhead separately via a constant reference would make the test more resilient to future changes.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
raballew: [LOW] The `copy-images`, `fetch-images`, and runtime sidecar containers all carry explicit `storageReservation` fields, but `fix-cuttlefish-permissions` here has no `Resources` field. Without a reservation the Kubernetes scheduler does not account for this container's storage use. `TestRenderPod_storageBudgets` only checks `fetch-images` and the runtime container. Consider adding `Resources: storageReservation(storage.budget)` here and extending the test to assert its storage reservation.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
raballew: [LOW] `DefaultRuntimeImage` ends with `:stable`. The `resolveImage` function only substitutes the version for images ending with `:latest`, so `:stable` is always returned unchanged across releases. The `:stable` tag is mutable; a supply-chain compromise of that tag would give an attacker full node access through the `Privileged: true` sidecar. Either extend `resolveImage` to also replace `:stable` with the provisioner version, or change `DefaultRuntimeImage` to end in `:latest` so the existing substitution logic applies.

controller/internal/exporterset/reconciler.go
```diff
@@ -288,6 +294,34 @@ func (r *ExporterSetReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
+			return err
+		}
+		policy.Spec = *desired.Spec.DeepCopy()
+		return nil
+	})
+	if err != nil {
+		return fmt.Errorf("ensure runtime network isolation: %w", err)
+	}
+	return nil
+}
```
raballew: [LOW] The `NetworkPolicyProvisioner` interface only declares `RenderNetworkPolicy`, but the caller `syncNetworkPolicy` adds the owner reference via `ctrl.SetControllerReference`. A future provisioner implementation could pre-set a different owner reference, causing an error, or assume garbage collection happens automatically without realizing the reconciler sets it. Adding a comment to the interface clarifying that the returned NetworkPolicy must have no `OwnerReferences` would prevent this confusion.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+			VolumeMounts:    append(stateMounts, deviceMounts...),
+		},
+		corev1.Container{
+			// Runs in the exporter image so the check shares the network namespace and Python runtime with jmp.
+			Name: gateContainerName, Image: img.exporter, ImagePullPolicy: img.exporterPull,
+			Command:         []string{"python3", "-m", "jumpstarter_driver_cuttlefish.health", "--wait", hostOrchestratorURL},
+			SecurityContext: exporterSecurityContext(),
+		},
+	)
+}
```
raballew: [LOW] `storageReservation(storage.budget)` applies the full multi-volume budget (sum of `imageSize + stateSize + tmpSize + 1Gi`) to `copy-images` and `fetch-images`, each of which only mounts the `cvd-images` emptyDir. On clusters with tight ephemeral-storage quotas this inflates per-Pod resource usage beyond what is actually needed. Defining a separate `imageBudget` sized for a single volume and applying it to image-staging init containers would give a more accurate reservation.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver_test.py
```diff
@@ -517,3 +520,102 @@ def test_cvd_power_on_ignores_other_groups(requests_mock, drv):
...
+    initialize(str(state_path), str(runtime_id), BASE)
+    config = {"managed": True, "runtime_id_path": str(runtime_id), "health_state_path": str(state_path),
+              "env_config": {"instances": [{}]}, "health_ports": [7681]}
+    instance = Cuttlefish(**config)
+    assert instance._health["runtime_id"] == "runtime-1"
+    assert instance._health["ports"] == [7681]
+    instance.close()
+    runtime_id.write_text("runtime-2")
+    with pytest.raises(CuttlefishError, match="restarted"):
+        Cuttlefish(**config)
```
raballew: [LOW] `test_managed_initialization_preserves_startup_runtime_id` chains two independent behaviors: successful initialization preserving the runtime ID (lines 606-618) and rejection of a changed runtime ID (lines 619-621). A failure in the first part prevents the second from running, making it harder to identify the root cause. Splitting into `test_managed_initialization_preserves_startup_runtime_id` and `test_managed_initialization_rejects_runtime_restart` (with a fresh `tmp_path`) would isolate the failure point.

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+func (p *Provisioner) Cleanup(
+	ctx context.Context,
+	exporterSet *virtualtargetv1alpha1.ExporterSet,
+	exporter *jumpstarterdevv1alpha1.Exporter,
+) error {
+	// Cuttlefish state and fetched images are Pod-scoped. Kubernetes removes
+	// the Pod-owned emptyDir volumes, while a claimed image tree is external
+	// and must outlive the exporter, so there is nothing for the provisioner
+	// to clean up here.
+	return nil
```
raballew: [LOW] `Cleanup()` intentionally returns `nil` but has no test pinning that contract. If a future refactor accidentally introduces an error return, no test catches it. Adding a `TestCleanup_isNoOp` that calls `New("dev").Cleanup(ctx, exporterSet, exporter)` and asserts `nil` would lock in the intended no-op behavior.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -39,16 +42,44 @@ class Cuttlefish(Driver):
...
+    def _write_health(self):
+        if not self.managed:
+            return
+        temporary = Path(self.health_state_path + ".tmp")
+        temporary.write_text(json.dumps(self._health))
+        os.replace(temporary, self.health_state_path)
+
     @classmethod
     def client(cls) -> str:
         return "jumpstarter_driver_cuttlefish.client.CuttlefishClient"
```
raballew: [LOW] The `result` parameter of `_fmt` (defined near line 99) lacks a type annotation. The function accepts values of type `dict | list | str` but the parameter carries no annotation while the return type is correctly annotated. Adding `def _fmt(self, result: dict | list | str) -> str:` would complete the signature and make the expected input types explicit.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
mangelajo: The `_on()` method acquires `_operation_lock` at line 437 (via `on()`), then calls `_do_operation()` which re-acquires it, and then at line 512 acquires it a third time. The inner `with self.parent._operation_lock:` here is redundant since `on()` already holds it. While `RLock` prevents deadlock, the nested acquisition adds confusion.

Consider removing this redundant lock acquisition — `_on()` is only ever called from `on()` which already holds the lock:

```python
        if self.parent.managed:
            self.parent._health.update(state="running", group=self.parent._cvd_group or self.parent.group,
                                       name=self.parent._cvd_name or self.parent.name)
            self.parent._write_health()
```

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -426,12 +508,21 @@ def on(self) -> None:  # noqa: C901
                         )
                     break
 
+        if self.parent.managed:
```
mangelajo: The `_do_operation` wrapper already writes `state=running` to health (via `target_state`) after a successful `/cvds` POST. Then `_on()` writes `state=running` again here. This double-write is harmless but means the health state is written 3+ times during a single `on()` call (`transition` → `running` from `_do_operation` → `running` again here). The second write at line 511-515 is needed only because it sets `group`/`name` — but `_do_operation` already sets those at lines 192-195 for the `/cvds` path.

Consider either:
1. Removing this second write entirely if `_do_operation` already covers the `group`/`name` update, or
2. Adding a comment explaining why it's needed despite `_do_operation` already transitioning to `running`.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+    if not runtime_id:
+        raise RuntimeError("Cuttlefish runtime ID is empty")
+    Path(state_path).write_text(json.dumps({
+        "runtime_id_path": runtime_id_path, "runtime_id": runtime_id,
+        "url": url, "ports": [], "state": "off",
+    }))
+
+
+def check(state_path: str) -> None:
+    state = json.loads(Path(state_path).read_text())
```
mangelajo: The `check()` function reads the health state file and the runtime ID file, then calls `urlopen` — all without any exception handling for file I/O or JSON parsing errors. If the health state file is mid-write (even with the atomic rename in `_write_health`, the read could land between the `write_text` and `os.replace` on the `.tmp` file during a race), the probe could get a `FileNotFoundError` or `JSONDecodeError`.

In practice the liveness probe's `failureThreshold: 6` provides tolerance, but consider wrapping the `Path(...).read_text()` + `json.loads()` in a try/except to give a clearer error message than a raw traceback.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+    runtime_id = Path(state["runtime_id_path"]).read_text().strip()
+    if not runtime_id or runtime_id != state["runtime_id"]:
+        raise RuntimeError("Cuttlefish runtime restarted; release the lease to replace this Pod")
+    url = state["url"]
+    with urllib.request.urlopen(f"{url}/_debug/statusz", timeout=2):
+        pass
+    if state["state"] == "off":
+        return
+    if state["state"] == "transition" and time.monotonic() < state["deadline"]:
```
mangelajo: `time.monotonic()` is process-local — it cannot be meaningfully compared across processes. The `deadline` is set in the driver process (`_do_operation`) using `time.monotonic() + timeout + 30`, but `check()` runs in a separate liveness probe process (invoked via `ExecAction`). The monotonic clocks of these two processes have unrelated epochs, so the comparison `time.monotonic() < state["deadline"]` is not valid.

Consider using `time.time()` instead (wall clock), or storing the deadline as an absolute wall-clock timestamp (e.g., `time.time() + timeout + 30`). Wall clock has its own issues (NTP jumps), but at least the comparison is meaningful across processes.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+
+
+def check(state_path: str) -> None:
+    state = json.loads(Path(state_path).read_text())
+    runtime_id = Path(state["runtime_id_path"]).read_text().strip()
+    if not runtime_id or runtime_id != state["runtime_id"]:
+        raise RuntimeError("Cuttlefish runtime restarted; release the lease to replace this Pod")
+    url = state["url"]
+    with urllib.request.urlopen(f"{url}/_debug/statusz", timeout=2):
+        pass
```
mangelajo: The `urlopen` call for `/_debug/statusz` discards the response body but doesn't check the HTTP status code. The context manager exits cleanly for any 2xx response, but `urlopen` raises `HTTPError` for 4xx/5xx, so the error case is implicitly handled. However, the comment in kirkbrauer's review (#12) correctly notes that `/_debug/statusz` always returns 200 as long as nginx/HO are serving, even if the VMM or simulator has crashed. The health module addresses this via the `/cvds` inventory check and port scanning when `state == "running"`, which is a reasonable approach.

No action needed — noting for reviewers that the `statusz` limitation is mitigated by the CVD and port checks.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+func resolveServiceAccount(parameters map[string]interface{}) (string, error) {
+	if privileged, _ := parameters["runtime_privileged"].(bool); !privileged {
+		return "", fmt.Errorf("cuttlefish requires runtime_privileged=true; unprivileged device access is not supported")
+	}
+	name, _ := parameters["service_account_name"].(string)
+	if name == "" || name == "default" || len(validation.IsDNS1123Subdomain(name)) != 0 {
+		return "", fmt.Errorf("service_account_name must name a dedicated workload service account")
+	}
+	return name, nil
+}
```
mangelajo: The `resolveServiceAccount` function rejects `""`, `"default"`, and DNS-invalid names, but it doesn't reject the empty/missing `runtime_privileged` key gracefully. When `runtime_privileged` is absent, `parameters["runtime_privileged"].(bool)` returns `false` (the zero value), which triggers the error. This works but the error message "cuttlefish requires runtime_privileged=true" could be confusing when the user simply forgot to set the parameter vs. explicitly setting it to false.

Consider distinguishing "missing" vs "explicitly false" in the error message.

Minor/nit — no action required if you prefer to keep it simple.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+
+func (p *Provisioner) RenderPod(
+	ctx context.Context,
+	exporterSet *virtualtargetv1alpha1.ExporterSet,
+	vtc *virtualtargetv1alpha1.VirtualTargetClass,
+	mergedParameters map[string]interface{},
+	overrides *virtualtargetv1alpha1.ImageOverrides,
+	exporter *jumpstarterdevv1alpha1.Exporter,
+) (*corev1.Pod, error) {
+	_ = ctx
```
mangelajo: The `_ = ctx` suppression suggests the context parameter is unused. If the provisioner doesn't need the context today, consider removing it from the signature and adjusting the interface, or document why it's reserved for future use (e.g., for API calls during rendering). Currently the QEMU provisioner also receives `ctx` — is it used there?

If the interface requires `ctx` for other provisioners that need it, this suppression is fine — just noting it.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+	request := resources.Requests[corev1.ResourceEphemeralStorage]
+	if request.Cmp(resources.Limits[corev1.ResourceEphemeralStorage]) > 0 {
+		return fmt.Errorf("ephemeral-storage request exceeds limit")
+	}
+	return nil
+}
+
+func (p *Provisioner) RenderNetworkPolicy(es *virtualtargetv1alpha1.ExporterSet) *networkingv1.NetworkPolicy {
+	return &networkingv1.NetworkPolicy{
+		ObjectMeta: metav1.ObjectMeta{Name: "cuttlefish-" + string(es.UID), Namespace: es.Namespace},
```
mangelajo: The `RenderNetworkPolicy` uses `es.UID` as part of the policy name: `"cuttlefish-" + string(es.UID)`. Kubernetes UIDs are 36-character UUIDs (e.g., `a1b2c3d4-e5f6-...`), making the full name ~47 characters. This is well within the 253-char DNS subdomain limit, but the name is opaque/unfriendly in `kubectl get netpol` output.

Consider using `es.Name` instead (which is already DNS-safe), e.g., `"cuttlefish-" + es.Name`. The UID is already embedded in the `PodSelector` match label, so uniqueness per ExporterSet is maintained either way. The UID-based name just makes debugging harder.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+	}
+	serviceAccount, err := resolveServiceAccount(mergedParameters)
+	if err != nil {
+		return nil, err
+	}
+	// The reconciler persists the enriched drivers itself; rendering needs the
+	// validation and the effective guest size for the runtime budget.
+	_, guest, err := enrichDrivers(exporterSet.Spec.Template.Spec.Drivers, mergedParameters)
+	if err != nil {
+		return nil, err
```
mangelajo: The `enrichCuttlefishDriver` and `RenderPod` both call `enrichDrivers` — once from `RenderPod` (line 231) and once from `EnrichExporterExport` (line 403). The reconciler calls `EnrichExporterExport` to persist the enriched drivers, then later calls `RenderPod` which enriches again. This means the driver config is validated and enriched twice per reconcile cycle.

This is safe because enrichment is idempotent (pinned values match on re-application), but it's redundant work. The comment at line 229-230 acknowledges this: "The reconciler persists the enriched drivers itself; rendering needs the validation and the effective guest size for the runtime budget."

Consider extracting the guest size computation from `enrichDrivers` so `RenderPod` can get the guest spec without re-running the full enrichment. This would also reduce the risk of the two code paths diverging in the future.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -358,9 +433,16 @@ def client(cls) -> str:
         return "jumpstarter_driver_cuttlefish.client.CvdPowerClient"
 
     @export
-    def on(self) -> None:  # noqa: C901
+    def on(self) -> None:
+        with self.parent._operation_lock:
+            self._on()
+
+    def _on(self) -> None:  # noqa: C901
```
mangelajo: In managed mode, `_do_operation` wraps `_perform_operation` with health state tracking. However, `_on()` calls `_do_operation` for sub-operations like DELETE (line 455) and POST `/cvds` (line 481), and each of these individually transitions health through `transition` → `running`/`off`. Then at line 511-515, `_on()` overwrites the health state again to `running`.

This means during a `power.on()` call that needs to delete a stale CVD first, the health state sequence is:
1. `transition` (from DELETE's `_do_operation`)
2. `off` (DELETE completes → target_state=off)
3. `transition` (from POST `/cvds`'s `_do_operation`)
4. `running` (POST completes → target_state=running)
5. `running` again (line 511-515)

The liveness probe could see `state=off` between steps 2 and 3, which is technically correct (no CVD exists), but it means a brief window where the probe thinks the system is intentionally powered off rather than in a multi-step operation. This is benign because `off` state is liveness-safe, but worth noting in a comment.

---

controller/internal/exporterset/reconciler.go
mangelajo: The `syncNetworkPolicy` runs before any pool state computation or Pod creation, which correctly ensures network isolation is in place before workloads start. The ownership check (`!metav1.IsControlledBy(policy, es)`) prevents hijacking an existing policy not owned by this ExporterSet.

One edge case: if two ExporterSets happen to produce the same policy name (e.g., both using `"cuttlefish-" + string(es.UID)` — unlikely since UIDs are unique), the second one would fail with the ownership error. This is the correct behavior. Just noting that the `CreateOrUpdate` + ownership pattern is well-applied here.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+			SecurityContext: &corev1.SecurityContext{Privileged: boolPtr(true), RunAsUser: &root},
+			VolumeMounts:    append(stateMounts, deviceMounts...),
+		},
+		corev1.Container{
+			// Runs in the exporter image so the check shares the network namespace and Python runtime with jmp.
+			Name: gateContainerName, Image: img.exporter, ImagePullPolicy: img.exporterPull,
+			Command:         []string{"python3", "-m", "jumpstarter_driver_cuttlefish.health", "--wait", hostOrchestratorURL},
+			SecurityContext: exporterSecurityContext(),
+		},
+	)
```
mangelajo: The `gate` init container (`wait-for-cuttlefish`) runs in the **exporter** image, not the runtime image — this is intentional so it shares the Python health module. However, it blocks the exporter container from starting until Host Orchestrator responds.

What happens if HO never starts? The gate will retry for `60 * 5 = 300 seconds` (5 minutes) and then raise `RuntimeError`, which fails the init container. With `RestartPolicy: Never`, the Pod will transition to `Failed`. The ExporterSet controller's `cleanupTerminalExporters` will then delete the Exporter and scale-up will create a replacement.

This is reasonable behavior, but the 5-minute gate timeout is not configurable. If HO startup is slow (e.g., large images, slow storage), the gate could fail before HO is ready. Consider either making the timeout configurable via parameters or using a longer default.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish_test.go
mangelajo: The unit tests are comprehensive for the provisioner logic (Pod rendering, driver enrichment, security constraints, storage budgets, isolation). 

However, there are no E2E tests for the cuttlefish provisioner. Given that this provisioner requires privileged containers, KVM device access, and significant memory, a full E2E test in standard CI may not be feasible. Consider adding E2E tests labeled as `openshift` or `fullblown` that cover:

1. ExporterSet creation with the cuttlefish provisioner → Pod comes up with correct sidecars
2. NetworkPolicy is created and blocks ingress
3. Lease acquisition → CVD boot → liveness probe passes
4. Lease release → Pod recycled via ExitAndReplace

Even a smoke test that verifies Pod spec correctness (without actually booting a CVD) would add value in the `controller/deploy/operator/test/e2e` directory.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+        if sys.argv[1] == "--run-exporter":
+            initialize(sys.argv[2], sys.argv[3], sys.argv[4])
+            os.execvp("jmp", ["jmp", "run", "--exporter-config", sys.argv[5]])
+        elif sys.argv[1] == "--wait":
+            wait_ready(sys.argv[2])
+        else:
+            check(sys.argv[1])
+    except Exception as exc:
+        print(f"Cuttlefish unhealthy: {exc}", file=sys.stderr)
+        raise SystemExit(1) from exc
```
mangelajo: The `__main__` block uses positional argument parsing via `sys.argv` indices. This works but is fragile — if the argument order changes or a new mode is added, all the indices shift. The three modes (`--run-exporter`, `--wait`, bare check) have different arities (5, 2, 1 args after the script name).

Consider using `argparse` with subcommands for clarity and better error messages when arguments are missing. This is a minor style point — the current approach is fine for a module that's only invoked from container commands the provisioner controls.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+	ctx context.Context,
+	exporterSet *virtualtargetv1alpha1.ExporterSet,
+	exporter *jumpstarterdevv1alpha1.Exporter,
+) error {
+	// Cuttlefish state and fetched images are Pod-scoped. Kubernetes removes
+	// the Pod-owned emptyDir volumes, while a claimed image tree is external
+	// and must outlive the exporter, so there is nothing for the provisioner
+	// to clean up here.
+	return nil
+}
```
mangelajo: The `Cleanup` method is a no-op, which is documented. This is correct — emptyDir volumes are Pod-scoped and cleaned up by kubelet, and the PVC (if used) is external. 

However, if a CVD is running when the Pod is deleted, there's no graceful shutdown of the Cuttlefish runtime or cleanup of the CVD via the HO API. The Pod's `RestartPolicy: Never` means the containers just terminate. For `ExitAndReplace`, this is fine (the entire Pod is disposable), but it means any in-flight guest work (e.g., a test running on the Android VM) is abruptly terminated.

This is likely acceptable for the current use case, but worth documenting that lease release implies immediate VM termination with no graceful shutdown.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,750 @@
...
+		VolumeMounts:    []corev1.VolumeMount{{Name: "cvd-state", MountPath: runtimeIDMount, ReadOnly: true}},
+		// With restartPolicy Never, a failed liveness check ends the exporter and lets ExitAndReplace recycle the Pod.
+		LivenessProbe: &corev1.Probe{
+			ProbeHandler: corev1.ProbeHandler{Exec: &corev1.ExecAction{
+				Command: []string{"python3", "-m", "jumpstarter_driver_cuttlefish.health", healthStatePath},
+			}},
+			PeriodSeconds: 10, TimeoutSeconds: 10, FailureThreshold: 6,
+		},
+	}
+}
```
mangelajo: The `exporterContainer` mounts `cvd-state` at `/run/cuttlefish-runtime` as `ReadOnly: true`, which lets it read the runtime ID marker. However, the exporter also needs to write the health state file at `/tmp/jumpstarter-cuttlefish-health.json`. This path is not on any mounted volume — it lives in the container's writable layer.

This is fine for ephemeral storage, but the health state file will be lost if the container restarts (which won't happen with `RestartPolicy: Never`). The liveness probe reads it via `ExecAction` which runs in the same container, so there's no cross-container access concern.

Just confirming the design is consistent — no issue here.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,79 @@
...
+        except Exception:
+            time.sleep(interval)
+    raise RuntimeError(f"Host Orchestrator at {url} did not become ready")
+
+
+def listening_ports() -> set[int]:
+    # Opening an HCI connection can create a simulator peer. Inspect the shared
+    # Pod network namespace instead of disturbing active Bluetooth sessions.
+    listeners = set()
+    for table in ("/proc/net/tcp", "/proc/net/tcp6"):
```
kirkbrauer: `/proc/net/tcp6` does not exist when the kernel is booted with `ipv6.disable=1`, so this read raises `FileNotFoundError` and `__main__` turns it into `SystemExit(1)`. That is persistent, not transient, so the liveness probe kills a healthy IPv4-only exporter. It is only reached in the `running` state — exactly the case the probe exists to protect.

```python
    try:
    except FileNotFoundError:
        if required:
            raise
        continue
    ...
```

---

kirkbrauer (commented):
One blocking issue in the Cuttlefish liveness check — see inline comment on `health.py`. The rest of the change (Go provisioner, NetworkPolicy reconciliation, RBAC trio, driver/README updates) reviewed clean.

---

_AI generated, human reviewed._

kirkbrauer (approved):

PR #1077, opened 2026-09-07
docs: document traffic filter for dut-network driver + E2E tests

mangelajo:
## Summary

Adds documentation and E2E tests for the **traffic filter** feature of the `dut-network` driver.

### Documentation (README.md)
- New **Traffic Filtering** section with examples for:
  - Egress allowlist (drop everything except specific destinations)
  - Egress denylist (allow everything except specific destinations)
  - Ingress filtering (restrict inbound connections)
- Filter rule fields reference table
- `filter` parameter added to the main config reference table

### E2E Tests
- New exporter config (`exporter-dut-network-filter.yaml`) with egress `policy: drop` and a single TCP accept rule
- Three new E2E test cases:
  - **Allowed traffic**: TCP to the permitted port passes through the filter
  - **Blocked TCP**: TCP to a non-allowed port is dropped
  - **Blocked ICMP**: ping is dropped when egress policy is `drop`
- Refactored TCP server/client into shared `startTCPServer`/`tcpConnect`/`killCmd` helpers (used by both base and filter tests)
- Updated `e2e/README.md` with the filter sub-lane documentation

### Testing
- Go `vet` passes
- Filter tests use the same veth/netns topology as the base dut-network tests but with a separate exporter on port 19092

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -136,6 +136,18 @@ offering the `dut-network` driver (nftables NAT/masquerade + DHCP/DNS). CI insta
 | should add, list, and remove DNS entries via CLI | `add-dns e2e-test.lab.local 10.0.0.42`, `dns-entries`, `remove-dns`, `dns-entries` | entry appears then disappears |
 | should allow TCP connections from DUT to external via NAT | start Python TCP server in ext ns, connect from DUT ns via NAT | client receives "E2E_OK" |
 
+### Filter sub-lane (`exporter-dut-network-filter.yaml`)
+
+Same netns topology, separate exporter on port 19092 with an egress filter
+(`policy: drop`, one `accept` rule for TCP port 9997 to `10.99.0.1/32`).
+
+| Test Name | Steps | Pass Check |
+|---|---|---|
+| should show filter rules in nftables output | `j dut-network nat-rules` | output contains "drop" and `dport 9997` |
+| should allow TCP to the permitted port | TCP server on allowed port 9997, client connects from DUT ns | client receives "FILTER_OK" |
+| should block TCP to a non-allowed port | TCP server on blocked port 9998, client connects from DUT ns | connection fails (timeout/reset) |
+| should block ICMP ping when egress policy is drop | ping from DUT ns to ext IP | ping fails (`Consistently`) |
+
 ---
 
 ## Lane: `exit-on-lease-end` (`exit_on_lease_end_test.go`)
--- a/e2e/exporters/exporter-dut-network-filter.yaml
+++ b/e2e/exporters/exporter-dut-network-filter.yaml
@@ -1,0 +1,34 @@
+apiVersion: jumpstarter.dev/v1alpha1
+kind: ExporterConfig
+metadata:
+  name: test-exporter-dut-network-filter
+  namespace: default
+export:
+  dut-network:
+    type: jumpstarter_driver_dut_network.driver.DutNetwork
+    config:
+      interface: "jmp-vhost"
+      subnet: "192.168.200.0/24"
+      gateway_ip: "192.168.200.1"
+      upstream_interface: "jmp-vup"
+      nat_mode: "masquerade"
+      dhcp_enabled: true
+      dhcp_range_start: "192.168.200.100"
+      dhcp_range_end: "192.168.200.200"
+      addresses:
+        - mac: "02:00:00:00:00:01"
+          ip: "192.168.200.10"
+          hostname: "test-dut"
+      dns_servers: ["8.8.8.8"]
+      state_dir: "/tmp/jmp-e2e-dut-network-filter"
+      filter:
+        egress:
+          policy: drop
+          rules:
+            # Allow only the blocked-check port (9997) to a specific
+            # destination, and deny everything else.  The E2E test verifies
+            # that traffic to a non-allowed destination (port 9998) is dropped.
+            - action: accept
+              destination: "10.99.0.1/32"
+              port: 9997
+              protocol: tcp
--- a/e2e/test/dut_network_test.go
+++ b/e2e/test/dut_network_test.go
@@ -266,29 +266,61 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
 })
 
 // startTCPServer starts a one-shot TCP listener in the given network namespace.
-// It accepts a single connection, sends payload, and exits.  Returns the
+// It accepts a single connection, sends payload, and exits.  The function
+// waits for a "READY" line on stdout (emitted after listen()) so callers
+// know the port is actually open before sending traffic.  Returns the
 // *exec.Cmd so the caller can clean up via process-group kill.
 func startTCPServer(ns string, port int, payload string) *exec.Cmd {
 	serverScript := fmt.Sprintf(
-		"import socket; "+
+		"import socket,sys; "+
 			"s=socket.socket(); "+
 			"s.setsockopt(socket.SOL_SOCKET,socket.SO_REUSEADDR,1); "+
 			"s.bind(('', %d)); "+
 			"s.listen(1); "+
+			"print('READY',flush=True); "+
 			"s.settimeout(15); "+
 			"conn,_=s.accept(); "+
 			"conn.sendall(b'%s'); "+
 			"conn.close(); "+
 			"s.close()", port, payload)
-	fullArgs := []string{"ip", "netns", "exec", ns, "python3", "-c", serverScript}
+	fullArgs := []string{"ip", "netns", "exec", ns, "python3", "-u", "-c", serverScript}
 	bin, cmdArgs := sudoArgs(fullArgs...)
 	cmd := exec.Command(bin, cmdArgs...) //nolint:gosec
 	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+
+	stdout, err := cmd.StdoutPipe()
+	ExpectWithOffset(1, err).NotTo(HaveOccurred())
 	ExpectWithOffset(1, cmd.Start()).To(Succeed())
-	time.Sleep(500 * time.Millisecond)
+
+	readyCh := make(chan error, 1)
+	go func() {
+		buf := make([]byte, 64)
+		n, readErr := stdout.Read(buf)
+		if readErr != nil {
+			readyCh <- fmt.Errorf("server stdout read failed: %w", readErr)
+			return
+		}
+		if !strings.Contains(string(buf[:n]), "READY") {
+			readyCh <- fmt.Errorf("unexpected server output: %s", string(buf[:n]))
+			return
+		}
+		readyCh <- nil
+	}()
+
+	select {
+	case readyErr := <-readyCh:
+		ExpectWithOffset(1, readyErr).NotTo(HaveOccurred(),
+			fmt.Sprintf("TCP server on port %d failed to become ready", port))
+	case <-time.After(5 * time.Second):
+		_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+		_ = cmd.Wait()
+		Fail(fmt.Sprintf("TCP server on port %d did not become ready within 5s", port))
+	}
+
 	return cmd
 }
 
+// killCmd sends SIGKILL to the process group and waits for exit.
 func killCmd(cmd *exec.Cmd) {
 	_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
 	_ = cmd.Wait()
--- a/e2e/test/dut_network_test.go
+++ b/e2e/test/dut_network_test.go
@@ -255,39 +255,200 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
 
 	Context("TCP connectivity", func() {
 		It("should allow TCP connections from DUT to external via NAT", func() {
-			serverScript := "import socket; " +
-				"s=socket.socket(); " +
-				"s.setsockopt(socket.SOL_SOCKET,socket.SO_REUSEADDR,1); " +
-				"s.bind(('',9998)); " +
-				"s.listen(1); " +
-				"s.settimeout(10); " +
-				"conn,_=s.accept(); " +
-				"conn.sendall(b'E2E_OK'); " +
-				"conn.close(); " +
-				"s.close()"
+			srv := startTCPServer(extNs, 9998, "E2E_OK")
+			defer killCmd(srv)
 
-			fullArgs := []string{"ip", "netns", "exec", extNs, "python3", "-c", serverScript}
-			bin, cmdArgs := sudoArgs(fullArgs...)
-			listener := exec.Command(bin, cmdArgs...) //nolint:gosec
-			listener.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
-			Expect(listener.Start()).To(Succeed())
-			defer func() {
-				_ = syscall.Kill(-listener.Process.Pid, syscall.SIGKILL)
-				_ = listener.Wait()
-			}()
-
-			time.Sleep(500 * time.Millisecond)
-
-			clientScript := fmt.Sprintf(
-				"import socket; "+
-					"s=socket.create_connection(('%s',9998),timeout=5); "+
-					"data=s.recv(10); "+
-					"s.close(); "+
-					"print(data.decode())",
-				extIP)
-			out, err := runInNsCapture(dutNs, "python3", "-c", clientScript)
+			out, err := tcpConnect(dutNs, extIP, 9998, 5)
 			Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("TCP connection failed: %s", out))
 			Expect(out).To(ContainSubstring("E2E_OK"))
+		})
+	})
+})
+
+// startTCPServer starts a one-shot TCP listener in the given network namespace.
+// It accepts a single connection, sends payload, and exits.  Returns the
+// *exec.Cmd so the caller can clean up via process-group kill.
+func startTCPServer(ns string, port int, payload string) *exec.Cmd {
+	serverScript := fmt.Sprintf(
+		"import socket; "+
+			"s=socket.socket(); "+
+			"s.setsockopt(socket.SOL_SOCKET,socket.SO_REUSEADDR,1); "+
+			"s.bind(('', %d)); "+
+			"s.listen(1); "+
+			"s.settimeout(15); "+
+			"conn,_=s.accept(); "+
+			"conn.sendall(b'%s'); "+
+			"conn.close(); "+
+			"s.close()", port, payload)
+	fullArgs := []string{"ip", "netns", "exec", ns, "python3", "-c", serverScript}
+	bin, cmdArgs := sudoArgs(fullArgs...)
+	cmd := exec.Command(bin, cmdArgs...) //nolint:gosec
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	ExpectWithOffset(1, cmd.Start()).To(Succeed())
+	time.Sleep(500 * time.Millisecond)
+	return cmd
+}
+
+func killCmd(cmd *exec.Cmd) {
+	_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
+	_ = cmd.Wait()
+}
+
+// tcpConnect attempts a TCP connection from the given namespace to host:port
+// and returns whatever the server sends.
+func tcpConnect(ns string, host string, port int, timeoutSec int) (string, error) {
+	clientScript := fmt.Sprintf(
+		"import socket; "+
+			"s=socket.create_connection(('%s',%d),timeout=%d); "+
+			"data=s.recv(64); "+
+			"s.close(); "+
+			"print(data.decode())",
+		host, port, timeoutSec)
+	return runInNsCapture(ns, "python3", "-c", clientScript)
+}
+
+// Serial: reuses the same veth topology as the base dut-network tests but
+// starts the exporter with a filter config that restricts egress to a single
+// TCP port.  Verifies that allowed traffic passes and everything else is dropped.
+var _ = Describe("DUT Network Filter E2E Tests", Label("dut-network"), Ordered, ContinueOnFailure, Serial, func() {
+	var (
+		tracker      *ProcessTracker
+		listenerPort = 19092
+		exporterDir  string
+	)
+
+	const (
+		dutNs      = "jmp-e2e-dut"
+		extNs      = "jmp-e2e-ext"
+		vethHost   = "jmp-vhost"
+		vethDut    = "jmp-vdut"
+		vethUp     = "jmp-vup"
+		vethExt    = "jmp-vext"
+		nftTable   = "jumpstarter_jmp_vhost"
+		dutIP      = "192.168.200.10"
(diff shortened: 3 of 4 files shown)


e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,204 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+			"conn,_=s.accept(); "+
+			"conn.sendall(b'%s'); "+
+			"conn.close(); "+
+			"s.close()", port, payload)
+	fullArgs := []string{"ip", "netns", "exec", ns, "python3", "-c", serverScript}
+	bin, cmdArgs := sudoArgs(fullArgs...)
+	cmd := exec.Command(bin, cmdArgs...) //nolint:gosec
+	cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}
+	ExpectWithOffset(1, cmd.Start()).To(Succeed())
+	time.Sleep(500 * time.Millisecond)
```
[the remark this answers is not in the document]
mangelajo: Good catch — fixed in 3aaeb117. `startTCPServer` now reads a `READY` line from stdout (emitted after `listen()` succeeds) instead of using a fixed 500ms sleep. If the child exits or fails to bind, the test fails immediately with a clear error.

mangelajo:
ok, this is looking good

e2e/exporters/exporter-dut-network-filter.yaml
```diff
@@ -0,0 +1,34 @@
...
+      nat_mode: "masquerade"
+      dhcp_enabled: true
+      dhcp_range_start: "192.168.200.100"
+      dhcp_range_end: "192.168.200.200"
+      addresses:
+        - mac: "02:00:00:00:00:01"
+          ip: "192.168.200.10"
+          hostname: "test-dut"
+      dns_servers: ["8.8.8.8"]
+      state_dir: "/tmp/jmp-e2e-dut-network-filter"
```
raballew: [MEDIUM] The state directory is created at a fixed path under `/tmp` (world-writable), which creates a TOCTOU race. Between the `rm -rf` in teardown and the driver's `mkdir`, any local user can plant a symlink at that path. Since the driver runs as root, it follows the symlink when writing dnsmasq config, potentially overwriting privileged files such as `/etc/shadow`. Replace the fixed `/tmp/jmp-e2e-dut-network-filter` path with `os.mkdtemp(prefix='jmp-e2e-dut-network-filter-', dir='/run/user/...')` or a root-owned non-world-writable directory. At minimum, verify ownership after `mkdir(exist_ok=True)`.

e2e/exporters/exporter-dut-network-filter.yaml
```diff
@@ -0,0 +1,34 @@
...
+      upstream_interface: "jmp-vup"
+      nat_mode: "masquerade"
+      dhcp_enabled: true
+      dhcp_range_start: "192.168.200.100"
+      dhcp_range_end: "192.168.200.200"
+      addresses:
+        - mac: "02:00:00:00:00:01"
+          ip: "192.168.200.10"
+          hostname: "test-dut"
+      dns_servers: ["8.8.8.8"]
```
raballew, line 22: Arent we dropping everything (including DNS) except 9997

e2e/exporters/exporter-dut-network-filter.yaml
```diff
@@ -0,0 +1,34 @@
...
+      state_dir: "/tmp/jmp-e2e-dut-network-filter"
+      filter:
+        egress:
+          policy: drop
+          rules:
+            # Allow only the blocked-check port (9997) to a specific
+            # destination, and deny everything else.  The E2E test verifies
```
raballew, line 29: Port 9997 has multiple meaning throughout this change. I guess a clear blocked and allowed port would make it easier to understand.

e2e/README.md
```diff
@@ -136,6 +136,18 @@ offering the `dut-network` driver (nftables NAT/masquerade + DHCP/DNS). CI insta
...
+
+Same netns topology, separate exporter on port 19092 with an egress filter
+(`policy: drop`, one `accept` rule for TCP port 9997 to `10.99.0.1/32`).
+
+| Test Name | Steps | Pass Check |
+|---|---|---|
+| should show filter rules in nftables output | `j dut-network nat-rules` | output contains "drop" and `dport 9997` |
+| should allow TCP to the permitted port | TCP server on allowed port 9997, client connects from DUT ns | client receives "FILTER_OK" |
+| should block TCP to a non-allowed port | TCP server on blocked port 9998, client connects from DUT ns | connection fails (timeout/reset) |
+| should block ICMP ping when egress policy is drop | ping from DUT ns to ext IP | ping fails (`Consistently`) |
```
raballew: [LOW] The filter sub-lane section does not state which Ginkgo label triggers it. The filter tests use `Label("dut-network")` (confirmed at `dut_network_test.go` line 345), the same label as the parent lane. Adding a sentence such as "These tests run under the same `dut-network` Ginkgo label and require the same `nftables` and `dnsmasq-base` packages as the parent lane" would make the triggering label explicit.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		if readErr != nil {
+			readyCh <- fmt.Errorf("server stdout read failed: %w", readErr)
+			return
+		}
+		if !strings.Contains(string(buf[:n]), "READY") {
+			readyCh <- fmt.Errorf("unexpected server output: %s", string(buf[:n]))
+			return
+		}
+		readyCh <- nil
+	}()
```
raballew: [MEDIUM] The readiness check reads at most 64 bytes in a single `stdout.Read(buf)` call. `io.Read` on a pipe does not guarantee atomic delivery of "READY\n" in one call. If `sudo ip netns exec` emits any preamble to stdout before Python runs, the buffer is consumed before "READY" arrives, causing a spurious readiness failure. Replace with a `bufio.Scanner` loop that reads lines until "READY" is found or EOF/timeout occurs. (`grep -n "bufio\|Scanner\|ReadString" e2e/test/dut_network_test.go` returns no matches.)

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+			"s=socket.socket(); "+
+			"s.setsockopt(socket.SOL_SOCKET,socket.SO_REUSEADDR,1); "+
+			"s.bind(('', %d)); "+
+			"s.listen(1); "+
+			"print('READY',flush=True); "+
+			"s.settimeout(15); "+
+			"conn,_=s.accept(); "+
+			"conn.sendall(b'%s'); "+
+			"conn.close(); "+
+			"s.close()", port, payload)
```
raballew: [LOW] The TCP server Python one-liner has no `try/finally` around the accept-and-serve block. If `s.accept()` raises `socket.timeout` after 15 seconds, `s.close()` is never reached. `killCmd` normally terminates the process via SIGKILL before the timeout fires so the OS reclaims the fd, but adding `try/finally: s.close()` or using `with socket.socket() as s:` would be cleaner and safer.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		_ = cmd.Wait()
+		Fail(fmt.Sprintf("TCP server on port %d did not become ready within 5s", port))
+	}
+
+	return cmd
+}
+
+// killCmd sends SIGKILL to the process group and waits for exit.
+func killCmd(cmd *exec.Cmd) {
+	_ = syscall.Kill(-cmd.Process.Pid, syscall.SIGKILL)
```
raballew: [LOW] Both `syscall.Kill` and `cmd.Wait` errors are silently discarded with `_ = ...`. If `cmd.Process` is nil (process never started), the Kill call panics. Non-ExitError returns from `cmd.Wait` can indicate unexpected failures. Guard with `if cmd.Process != nil` before Kill, and log non-ExitError returns from `cmd.Wait` via `GinkgoWriter.Printf`.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+func tcpConnect(ns string, host string, port int, timeoutSec int) (string, error) {
+	clientScript := fmt.Sprintf(
+		"import socket; "+
+			"s=socket.create_connection(('%s',%d),timeout=%d); "+
+			"data=s.recv(64); "+
+			"s.close(); "+
+			"print(data.decode())",
+		host, port, timeoutSec)
+	return runInNsCapture(ns, "python3", "-c", clientScript)
+}
```
raballew: [LOW] The inline Python client script has no `try/finally` or context manager around the socket. If `s.recv(64)` raises (the expected path when testing a blocked port), `s.close()` is skipped. The OS reclaims the fd on process exit so there is no cross-test leak under current usage, but using `with socket.create_connection(('HOST', PORT), timeout=SEC) as s: data = s.recv(64)` is cleaner and more robust.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
-					"s=socket.create_connection(('%s',9998),timeout=5); "+
-					"data=s.recv(10); "+
-					"s.close(); "+
-					"print(data.decode())",
-				extIP)
-			out, err := runInNsCapture(dutNs, "python3", "-c", clientScript)
+			srv := startTCPServer(extNs, 9998, "E2E_OK")
+			defer killCmd(srv)
+
+			out, err := tcpConnect(dutNs, extIP, 9998, 5)
```
raballew: [LOW] Port 9998 is used here for a successful TCP connectivity assertion, but the filter Describe block at line 366 declares `blockedPort = 9998` and asserts that a connection to it fails. The same port number carries opposite semantics depending on which Describe block is being read. Use `blockedPort = 9999` in the filter test so each port number has a single, consistent semantic role.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		vethExt    = "jmp-vext"
+		nftTable   = "jumpstarter_jmp_vhost"
+		dutIP      = "192.168.200.10"
+		gatewayIP  = "192.168.200.1"
+		extIP      = "10.99.0.1"
+		upstreamIP = "10.99.0.2"
+		subnet     = "192.168.200.0/24"
+		allowedPort  = 9997
+		blockedPort  = 9998
+	)
```
raballew: [MEDIUM] `gofmt -l` flags this file. The newly added `allowedPort` and `blockedPort` constants are the longest identifiers in the const block, so all 14 constants need their `=` signs realigned to the column after the longest name. Run `gofmt -w e2e/test/dut_network_test.go` to fix.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		vethHost   = "jmp-vhost"
+		vethDut    = "jmp-vdut"
+		vethUp     = "jmp-vup"
+		vethExt    = "jmp-vext"
+		nftTable   = "jumpstarter_jmp_vhost"
+		dutIP      = "192.168.200.10"
+		gatewayIP  = "192.168.200.1"
+		extIP      = "10.99.0.1"
+		upstreamIP = "10.99.0.2"
+		subnet     = "192.168.200.0/24"
```
raballew: [MEDIUM] The same 11-12 topology constants (network name, VLAN CIDR, gateway IP, etc.) are declared verbatim in both Describe blocks. A topology change requires two synchronized edits (lines 63-75 and 352-364). Hoist the shared constants to package scope; keep `allowedPort` and `blockedPort` local to the filter Describe.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+	}
+
+	teardownNetworkNamespaces := func() {
+		runIgnoreErr("ip", "link", "del", vethHost)
+		runIgnoreErr("ip", "link", "del", vethUp)
+		runIgnoreErr("ip", "netns", "del", dutNs)
+		runIgnoreErr("ip", "netns", "del", extNs)
+		runIgnoreErr("nft", "delete", "table", "ip", nftTable)
+		runIgnoreErr("rm", "-rf", "/tmp/jmp-e2e-dut-network-filter")
+	}
```
raballew: [MEDIUM] `teardownNetworkNamespaces` is implemented twice with five identical `runIgnoreErr` calls; the only difference between the two closures is the `rm -rf` path (compare lines 104-111 and 389-396). Extract a package-level `teardownNetworkNamespaces(stateDir string)` to eliminate the duplication.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+			if err != nil {
+				return nil
+			}
+			conn.Close()
+			return fmt.Errorf("port %d is still open", listenerPort)
+		}, 10*time.Second, 500*time.Millisecond).Should(Succeed(),
+			"port %d should be closed after stopping exporter", listenerPort)
+
+		tracker.Cleanup()
+	})
```
raballew: [MEDIUM] Both BeforeAll and AfterAll blocks are structurally identical across the two Describe closures, differing only in `configPath` and `listenerPort`. The port-wait polling block is duplicated at lines 134-142 and 419-427. Extract `waitForPortClosed(port int)` as a package-level helper, and consider extracting helpers for the BeforeAll/AfterAll boilerplate to make future test additions less error-prone.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		setupNetworkNamespaces()
+
+		configPath := filepath.Join(exporterDir, "exporter-dut-network-filter.yaml")
+		tracker.StartDirectExporter(configPath, listenerPort, "", false)
+		WaitForDirectExporterReady(listenerPort, "")
+	})
+
+	AfterAll(func() {
+		tracker.StopAll()
+		teardownNetworkNamespaces()
```
raballew: [MEDIUM] The `jumpstarter_jmp_vhost` nftables table (with egress `policy: drop`) is removed only in BeforeAll and AfterAll. No signal handler exists for SIGKILL (`grep -rn "signal.Notify|os.Interrupt" e2e/test/` returns nothing). If the test process is killed mid-run, the egress drop table persists on the host indefinitely, silently blocking outbound traffic on shared CI runners until the next BeforeAll teardown. Add a `TestMain` or Ginkgo `SynchronizedAfterSuite` with `signal.Notify` for SIGTERM/SIGINT that calls `nft delete table ip jumpstarter_jmp_vhost` before exiting.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		runInNs(dutNs, "ip", "route", "add", "default", "via", gatewayIP)
+	}
+
+	teardownNetworkNamespaces := func() {
+		runIgnoreErr("ip", "link", "del", vethHost)
+		runIgnoreErr("ip", "link", "del", vethUp)
+		runIgnoreErr("ip", "netns", "del", dutNs)
+		runIgnoreErr("ip", "netns", "del", extNs)
+		runIgnoreErr("nft", "delete", "table", "ip", nftTable)
+		runIgnoreErr("rm", "-rf", "/tmp/jmp-e2e-dut-network-filter")
```
raballew: [LOW] `ensure_filter_forward()` inserts per-handle ACCEPT rules into the host's `ip filter FORWARD` chain and stores handle IDs only in memory. The Go teardown here removes only the `jumpstarter_jmp_vhost` table and has no commands targeting `ip filter FORWARD`. If the driver process is SIGKILLed, these ACCEPT rules persist. On hosts where Docker uses the same `ip filter FORWARD` chain, accumulated ACCEPT rules for `jmp-vhost`/`jmp-vup` iifname/oifname can create policy holes if Docker containers later receive veth interfaces with matching names. Extend `teardownNetworkNamespaces` to flush jumpstarter-related ACCEPT rules from `ip filter FORWARD`, or persist handle IDs to a file in `state_dir` so teardown can recover and remove them.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+			tracker.DumpLogs(250)
+		}
+	})
+
+	jmpShell := func(args ...string) (string, error) {
+		shellArgs := []string{"shell", "--tls-grpc", fmt.Sprintf("127.0.0.1:%d", listenerPort),
+			"--tls-grpc-insecure", "--"}
+		shellArgs = append(shellArgs, args...)
+		return Jmp(shellArgs...)
+	}
```
raballew: [MEDIUM] The `jmpShell` closure is defined identically at lines 157-162 and 442-447; the only difference is the captured `listenerPort` variable. Extract a package-level `func jmpShellOn(port int, args ...string) (string, error)` to remove the duplication.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+
+	Context("Filter rules visible in NAT output", func() {
+		It("should show filter rules in nftables output", func() {
+			out, err := jmpShell("j", "dut-network", "nat-rules")
+			Expect(err).NotTo(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("drop"))
```
raballew, line 453: Use `ContainSubstring("policy drop")` to specifically pin the chain policy declaration.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		})
+	})
+
+	Context("Blocked traffic is dropped by filter", func() {
+		It("should block TCP to a non-allowed port", func() {
+			srv := startTCPServer(extNs, blockedPort, "SHOULD_NOT_ARRIVE")
+			defer killCmd(srv)
+
+			_, err := tcpConnect(dutNs, extIP, blockedPort, 3)
+			Expect(err).To(HaveOccurred(), "connection to blocked port should fail")
```
raballew: [MEDIUM] With `ContinueOnFailure` set at line 345, this TCP block test can pass for the wrong reason. If the allowed-port connectivity test fails due to missing NAT, `tcpConnect` to the blocked port also fails, so the `Expect(err).To(HaveOccurred())` assertion passes even though the filter was never exercised. Remove `ContinueOnFailure` or add an explicit precondition confirming the allowed-port test passed before asserting the blocked port fails.

e2e/test/dut_network_test.go
```diff
@@ -255,43 +255,236 @@ var _ = Describe("DUT Network E2E Tests", Label("dut-network"), Ordered, Continu
...
+		})
+
+		It("should block ICMP ping when egress policy is drop", func() {
+			Consistently(func() error {
+				_, err := runInNsCapture(dutNs, "ping", "-c", "1", "-W", "1", extIP)
+				return err
+			}, 3*time.Second, 1*time.Second).Should(HaveOccurred(),
+				"ping should be blocked by egress drop policy")
+		})
```
raballew: [MEDIUM] The ICMP `Consistently` assertion (`Should(HaveOccurred())`) passes whenever `ping` encounters any network error, including missing NAT, not only when the egress filter policy drops it. `WaitForDirectExporterReady` confirms only gRPC reachability, not nftables state. Under `ContinueOnFailure`, an infrastructure failure satisfies the assertion for the wrong reason. Add a positive precondition (ICMP reachable before the filter is applied, or verify nftables shows `policy drop`) before the Consistently block.

python/packages/jumpstarter-driver-dut-network/README.md
```diff
@@ -107,6 +107,75 @@ export:
...
+```yaml
+      filter:
+        ingress:
+          policy: drop
+          rules:
+            - action: accept
+              source: "198.51.100.0/24"
+              port: 22
+              protocol: tcp
+```
```
raballew, line 162: Why is this intended? Easier to copy paste at the right level?

python/packages/jumpstarter-driver-dut-network/README.md
```diff
@@ -107,6 +107,75 @@ export:
...
+
+| Field | Required | Description |
+|-------|----------|-------------|
+| `policy` | no | Default action: `accept` (default) or `drop` |
+| `rules` | no | Ordered list of rules evaluated before the policy |
+| `rules[].action` | yes | `accept` or `drop` |
+| `rules[].destination` | no | CIDR for egress matching (e.g. `10.0.0.0/8`) |
+| `rules[].source` | no | CIDR for ingress matching |
+| `rules[].port` | no | Destination port (requires `protocol`) |
+| `rules[].protocol` | no | `tcp` or `udp` (required when `port` is set) |
```
raballew: [MEDIUM] The "Filter Rule Fields" table uses three columns (Field | Required | Description) while the main Reference table immediately below uses four columns including a Type column. Peer driver READMEs follow the four-column pattern. Add a `Type` column: `policy->str`, `rules->list`, `rules[].action->str`, `rules[].destination/source->str (CIDR)`, `rules[].port->int`, `rules[].protocol->str`.

e2e/exporters/exporter-dut-network-filter.yaml
mangelajo: It would be worth checking. For example it could be that the dnsmasq responder between the DUT and the sidekick is before filtering, (that happens at routing time), and you can still get the DNS to answer.

python/packages/jumpstarter-driver-dut-network/README.md
mangelajo: exactly

e2e/exporters/exporter-dut-network-filter.yaml
mangelajo: Yes — with egress `policy: drop`, everything except TCP 9997 to 10.99.0.1 is dropped, including UDP/53. However, the dnsmasq responder sits on the gateway (192.168.200.1), which is a host-local IP, so queries and answers never traverse the FORWARD filter chain. Added an E2E test verifying DNS entries still resolve under the drop policy (f326c2df).

e2e/exporters/exporter-dut-network-filter.yaml
mangelajo: Fixed in f326c2df — the comment now explicitly distinguishes the allowed port (9997) from the blocked port (9998), matching the test constants.

e2e/test/dut_network_test.go
mangelajo: Done in f326c2df — the assertion now pins `policy drop` specifically.

raballew (approved):

PR #1080, opened 2026-09-09
test(driver-snmp): add protocol-level tests for SNMPv3 wire traffic

mangelajo:
## Summary

Adds protocol-level (integration) tests for the SNMP driver that exercise **real in-process UDP SNMPv3 traffic** against a local agent, complementing the existing mock-only unit tests.

## Changes

- **New `driver_protocol_test.py`** (12 tests) with an in-process SNMPv3 USM agent (`SnmpTestAgent` + recording `SetCommandResponder`) listening on an ephemeral UDP port:
  - SET varbind wire encoding: on → `1`, off → `0`; default and custom OID/plug paths
  - USM auth (MD5/SHA, no privacy) and auth+priv (DES, AES) round-trips
  - wrong auth key → silent USM drop → driver timeout
  - agent errorStatus (`noSuchName`) propagation to `SNMPError`
  - no-response timeout (blackhole UDP port)
  - unresolvable hostname → `SNMPError` at construction
  - full client round-trip through `jumpstarter.common.utils.serve()` (gRPC over UNIX socket → session → driver → SNMP agent)
- **`pyproject.toml`**: declare `cryptography>=43.0.3` as a direct dependency. DES/AES privacy requires it at runtime (pysnmp only ships it as an optional dev-extra); it was previously pulled transitively via `jumpstarter` and was missing from the isolated `make pkg-test` env, breaking the priv tests there. Specifier matches the core `jumpstarter` package.
- **`uv.lock`**: updated — new `cryptography` edge plus the previously-missing `jumpstarter-driver-netsim` workspace member (stale lock); no version changes.

## Verification

- `make pkg-test-jumpstarter-driver-snmp`: **17 passed** (12 new protocol + 5 pre-existing mock)
- `make pkg-ty-jumpstarter-driver-snmp`, `make lint`, `ruff format --check`: clean
- existing `driver_test.py` untouched

### DIFF
--- a/python/packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver_protocol_test.py
+++ b/python/packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver_protocol_test.py
@@ -1,0 +1,319 @@
+"""Protocol-level (integration) tests for the SNMP power driver.
+
+These tests run a real in-process pysnmp SNMPv3 agent on a loopback UDP port
+so the driver performs actual SNMPv3 SET traffic on the wire (USM
+authentication, privacy, error-status propagation, and timeouts), instead of
+mocking the pysnmp internals as in driver_test.py.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import socket
+import threading
+import time
+from collections.abc import Generator
+from contextlib import contextmanager
+from typing import Any
+
+import pytest
+from pysnmp.carrier.asyncio.dgram import udp
+from pysnmp.entity import config, engine
+from pysnmp.entity.rfc3413 import cmdrsp, context
+from pysnmp.proto.api import v2c
+
+from jumpstarter_driver_snmp.driver import AuthProtocol, PrivProtocol, SNMPError, SNMPServer
+
+from jumpstarter.common.utils import serve
+
+
+@pytest.fixture(autouse=True)
+def _reset_event_loop():
+    """The driver creates, sets and closes its own event loop per operation,
+    leaving a closed loop registered on the main thread. Replace it with a
+    fresh open loop after each test so tests running later (e.g. the mock
+    based driver_test.py) do not hit a closed loop via asyncio.get_event_loop()."""
+    yield
+    asyncio.set_event_loop(asyncio.new_event_loop())
+
+
+class _RecordingSetResponder(cmdrsp.SetCommandResponder):
+    """SET responder that records received varbinds and answers with a
+    configurable SNMP error status."""
+
+    def __init__(self, snmpEngine, snmpContext, agent: SnmpTestAgent):
+        super().__init__(snmpEngine, snmpContext)
+        self._agent = agent
+
+    def handle_management_operation(self, snmpEngine, stateReference, contextName, PDU):
+        varBinds = v2c.apiPDU.get_varbinds(PDU)
+        self._agent.record([(tuple(int(component) for component in oid), int(value)) for oid, value in varBinds])
+        self.send_varbinds(
+            snmpEngine,
+            stateReference,
+            self._agent.error_status,
+            self._agent.error_index,
+            varBinds,
+        )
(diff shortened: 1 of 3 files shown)


python/packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver_protocol_test.py
```diff
@@ -0,0 +1,319 @@
...
+        server = SNMPServer(
+            host="127.0.0.1",
+            port=agent.port,
+            user="agent-user",
+            plug=7,
+            oid="1.3.6.1.2.1.95",
+        )
+        server.off()
+
+    assert agent.requests == [[((1, 3, 6, 1, 2, 1, 95, 7), 0)]]
```
raballew: [HIGH] `test_custom_oid_and_plug_sent_on_wire` only calls `server.off()` and asserts the value=0 varbind. The value=1 (on) case through a custom OID is never exercised, so a regression in the `on()` path with a non-default OID would go undetected.

Suggested fix:
```python
server.on()
server.off()
assert agent.requests == [
    [((1, 3, 6, 1, 2, 1, 95, 7), 1)],
    [((1, 3, 6, 1, 2, 1, 95, 7), 0)],
]
```

python/packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver_protocol_test.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _reset_event_loop():
+    """The driver creates, sets and closes its own event loop per operation,
+    leaving a closed loop registered on the main thread. Replace it with a
+    fresh open loop after each test so tests running later (e.g. the mock
+    based driver_test.py) do not hit a closed loop via asyncio.get_event_loop()."""
+    yield
+    asyncio.set_event_loop(asyncio.new_event_loop())
```
raballew: [MEDIUM] The `_reset_event_loop` fixture replaces the global event loop after each test but never closes the old one. Python 3.12+ emits `ResourceWarning: unclosed event loop` for each test. The root cause is in `driver.py`: `_setup_event_loop` calls `asyncio.set_event_loop(loop)` and `_snmp_set` closes that loop in its `finally` block, leaving the process-wide event loop closed and unusable after any non-async driver call.

Short-term fix: close the old loop before replacing it:
```python
@pytest.fixture(autouse=True)
def _reset_event_loop():
    yield
    old = asyncio.get_event_loop_policy().get_event_loop()
    old.close()
    asyncio.set_event_loop(asyncio.new_event_loop())
```
Long-term fix: rewrite `_setup_event_loop` to use `asyncio.run()`, which manages its own isolated loop without touching the global, eliminating the need for this fixture entirely.

python/packages/jumpstarter-driver-snmp/jumpstarter_driver_snmp/driver_protocol_test.py
```diff
@@ -0,0 +1,319 @@
...
+    def __enter__(self) -> SnmpTestAgent:
+        return self.start()
+
+    def __exit__(self, *exc_info):
+        self.stop()
+
+
+@contextmanager
+def _blackhole_port() -> Generator[int, None, None]:
+    """Yield a UDP port that is bound but never answers."""
```
raballew: [LOW] The docstring `"Yield a UDP port that is bound but never answers."` restates what the function name and its three-line body already show clearly. Consider removing it. If the name alone feels insufficient, `_bound_unresponsive_udp_port` makes the intent equally clear without the redundant prose.

raballew:
This actually looks pretty solid but there seems to be a merge conflict so CI can not run.

mangelajo:
I had to fix the await calls from tests since this broke for python 3.14

raballew (approved):

raballew:
This looks good.

PR #1082, opened 2026-09-09
feat(cuttlefish): add exec backend running cvd over jumpstarter-exec

bennyz:
Host Orchestrator is a thin wrapper over the same cvd subcommands, so with parameters.backend=exec the exporter runs them in the runtime container through jumpstarter-exec instead of calling HTTP. In exec mode Host
Orchestrator and nginx are not started at all; only cuttlefish-host-resources and the WebRTC operator run next to `jumpstarter-exec serve`, which is PID 1 so lease teardown ends the container. http stays the default for externally managed hosts.

The launcher serves from / because children inherit its working directory and cvd aborts when it cannot read it. cvd_user defaults to httpcvd, the owner of the state directories, so guest processes stay non-root.

The driver splits lifecycle operations behind HostOrchestratorBackend and CvdCliBackend; the health probe and the `--wait` startup gate accept either an http:// or an exec:// endpoint.

depends on https://github.com/jumpstarter-dev/jumpstarter/pull/1072


kirkbrauer (approved):
I think this looks like a great start, we should merge this to start testing it and see if there are any changes needed.

kirkbrauer:
@mangelajo Let's cancel the merge for now, I just realized this is stacked on the other PR https://github.com/jumpstarter-dev/jumpstarter/pull/1072 which needs to be fully reviewed before we merge this.

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,115 @@
...
+        except Exception:
+            time.sleep(interval)
+    raise RuntimeError(f"Cuttlefish runtime at {endpoint} did not become ready")
+
+
+def listening_ports() -> set[int]:
+    # Opening an HCI connection can create a simulator peer. Inspect the shared
+    # Pod network namespace instead of disturbing active Bluetooth sessions.
+    listeners = set()
+    for table in ("/proc/net/tcp", "/proc/net/tcp6"):
```
kirkbrauer: Originates in #1072 and is unchanged here, so it affects both the `http` and `exec` backends.

`/proc/net/tcp6` does not exist when the kernel is booted with `ipv6.disable=1`, so this read raises `FileNotFoundError` and `__main__` turns it into `SystemExit(1)`. That is persistent, not transient, so the liveness probe kills a healthy IPv4-only exporter. It is only reached in the `running` state — exactly the case the probe exists to protect.

```python
    try:
    except FileNotFoundError:
        if required:
            raise
        continue
    ...
```

---

kirkbrauer (changes requested):
Requesting changes for the `health.py` liveness issue flagged inline — it originates in #1072 and `listening_ports()` is unchanged in this PR, so it affects both the `http` and `exec` backends here.

This supersedes my earlier approval, per my comment about this being stacked on #1072.

---

_AI generated, human reviewed._

kirkbrauer (approved):

PR #1083, opened 2026-09-10
feat(cuttlefish): serve the WebRTC display over the lease

bennyz:
Media is UDP addressed to the runtime Pod's own interfaces, and the isolation policy denies all ingress, so a browser outside the Pod has no candidate it can reach. With parameters.webrtc_turn=true a coturn sidecar on loopback gives it
one: the browser allocates a relay over the forwarded TCP connection and the streamer sends media to that relay in its own network namespace.

An added nginx vhost overrides /infra_config, whose ICE server list is otherwise a public STUN server compiled into the operator binary. The client side is `j cuttlefish webrtc --forward`, which pins the local TURN port because the advertised URL names it.

depends on https://github.com/jumpstarter-dev/jumpstarter/pull/1082


python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
```diff
@@ -0,0 +1,115 @@
...
+        except Exception:
+            time.sleep(interval)
+    raise RuntimeError(f"Cuttlefish runtime at {endpoint} did not become ready")
+
+
+def listening_ports() -> set[int]:
+    # Opening an HCI connection can create a simulator peer. Inspect the shared
+    # Pod network namespace instead of disturbing active Bluetooth sessions.
+    listeners = set()
+    for table in ("/proc/net/tcp", "/proc/net/tcp6"):
```
kirkbrauer: Originates in #1072 and is unchanged here, so it affects both the `http` and `exec` backends.

`/proc/net/tcp6` does not exist when the kernel is booted with `ipv6.disable=1`, so this read raises `FileNotFoundError` and `__main__` turns it into `SystemExit(1)`. That is persistent, not transient, so the liveness probe kills a healthy IPv4-only exporter. It is only reached in the `running` state — exactly the case the probe exists to protect.

```python
    try:
    except FileNotFoundError:
        if required:
            raise
        continue
    ...
```

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,1091 @@
...
+
+// runtimePortReserved reports whether a port is taken by a service in the
+// orchestration image or by the TURN relay range, so no configurable listener
+// can be placed on top of one.
+func runtimePortReserved(port int) bool {
+	switch port {
+	case 80, 443, 1080, 1443, 2080, hostOrchestratorPort, 2443, hciPort, 7301, 7302, 7303, netsimPort, 15037, 19531:
+		return true
+	}
+	return (port >= 6520 && port <= 6620) || (port >= 15550 && port <= 15560) || (port >= turnRelayMin && port <= turnRelayMax)
```
kirkbrauer: `runtimePortReserved` stops at 15560, but line 111 declares the streamer's range as 15550-15599. A `turn_port` of 15561-15599 therefore passes validation and can occupy a port the streamer wants (coturn's UDP listener is not disabled).

Not a guaranteed outage — the streamer may pick another candidate — but keeping these disjoint is precisely this function's job, and `turnRelayMin`/`turnRelayMax` show the intent.

**Ask:** cover the full declared range, and add an upper-boundary case to the port-validation table (`turn_port: 15599` must be rejected — it has no streamer case today).

```go
return (port >= 6520 && port <= 6620) || (port >= 15550 && port <= 15599) || (port >= turnRelayMin && port <= turnRelayMax)
```

---

kirkbrauer (changes requested):
Two blockers inline: the `health.py` IPv6 liveness bug carried up from #1072 (needs fixing at its origin or the rebase drops it), and an incomplete streamer port reservation in `cuttlefish.go`.

Separately: most open CodeRabbit comments here look stale — init-container ordering, exec+`webrtc_turn`, `_delete_stale` dedup, `client.py` `reuse_port`, the `cvdcli.py` non-dict guard and the gocyclo count all appear fixed in later commits, and `boolPtr` is pre-existing. Worth resolving them to clear the noise. The #1082 `chown httpcvd` issue is fixed here too.

VSOCK isolation checked specifically and it holds: `pin(crosvm, "vhost_user_vsock", "true")` yields a Pod-private Unix-socket backend and `/dev/vhost-vsock` is not mounted, so the reused CID 3 cannot collide. Since that rests on one `pin`, a test asserting it stays pinned would be cheap insurance. FWIW upstream `podcvd` still mounts `/dev/vhost-vsock`, so it is not a better model here.

---

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
```diff
@@ -0,0 +1,1091 @@
...
+		if vtc.Spec.Scheduling.Tolerations != nil {
+			pod.Spec.Tolerations = append([]corev1.Toleration(nil), vtc.Spec.Scheduling.Tolerations...)
+		}
+	}
+	return pod, nil
+}
+
+func podMeta(exporterSet *virtualtargetv1alpha1.ExporterSet, exporter *jumpstarterdevv1alpha1.Exporter) metav1.ObjectMeta {
+	meta := metav1.ObjectMeta{
+		Namespace:   exporterSet.Namespace,
```
kirkbrauer, line 582: **Placement is not gated, and the README example omits it.**

`NodeSelector`/`Tolerations` are copied only if the VirtualTargetClass sets them, there is no default, and the README's example sets neither — so a privileged Pod needing KVM/vhost-net/tun can land anywhere. hostPath does not help: `HostPathCharDev` is checked by the kubelet at mount time, not by the scheduler, so the replica is already bound and sits in Pending with `FailedMount` rather than rescheduling.

The runtime image *and* the fetched CVD build target (`aosp_cf_arm64_*` vs `aosp_cf_x86_64_*`) must both match the node. There is no `kubernetes.io/arch` default and no cross-check, so a mixed amd64/arm64 pool silently yields replicas that cannot work.

QEMU already solves this — `controller/config/samples/v1alpha1_virtualtargetclass.yaml`:

```yaml
scheduling:
  nodeSelector: {kubernetes.io/arch: arm64}
  tolerations: [{key: jumpstarter.dev/kvm, operator: Exists, effect: NoSchedule}]
  resources: {limits: {devices.kubevirt.io/kvm: "1"}}
```

This already works here: `runtimeResources` (988-989) DeepCopies `Scheduling.Resources` and only fills in cpu/memory, so a device-plugin limit survives onto the runtime container. **No code change is needed to gate placement.**

**Two asks:**
1. Add that `scheduling` block to the README example, and document the arch selector plus prepared-node taint as required rather than optional.
2. Make `deviceMounts` (618-622) conditional — skip the hostPath binds when the devices come from a plugin. They are unconditional today, and `cuttlefish-host-resources` chowns `/dev/kvm` on the *host* inode, which can lock out other KVM consumers on the node.

---

kirkbrauer (commented):
The README builds everything in `cuttlefish-lab` (namespace, PSA label, ServiceAccount, SCC grant, both example CRs), but `exporter-set-controller/main.go` pins its cache to `NAMESPACE`, a fieldRef to the controller's own namespace — the comment there explains it holds only a namespaced Role. An ExporterSet in `cuttlefish-lab` is never reconciled: no Pod, no event, no condition, nothing in the log. Shipped QEMU manifests use the install namespace. Doc fix, unless a real tenant namespace is wanted.

**Which means the PSA relaxation lands on the control plane.** `enforce=privileged` has to go on the namespace hosting jumpstarter-controller, routers and telemetry — all deliberately restricted-compatible today. Suggest recommending a dedicated Jumpstarter install for cuttlefish pools.

Smaller:
- Admission rejections and the ~10 new validation errors return before the status/condition block (`reconciler.go:276-278` vs `:284-294`), so `kubectl describe exporterset` shows nothing.
- Guest launcher logs are file-only in the disposable emptyDir, and the Pod is deleted on recycle, so boot-failure evidence dies with it.
- `DefaultTurnImage` is the only `docker.io` default and is not reachable via typed `spec.images` overrides.

Checked and clear: host TAP/cvd-account contention (all per-netns), ephemeral storage bounds, terminal-Pod recycling, the startup gate vs lease timeout, the docs symlink, and the absent SCC object (correct — the operator cannot delegate an SCC it has no `use` on).

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
```diff
@@ -125,29 +138,249 @@ def _wait_for_operation(self, op_name: str, timeout: float = 300) -> dict:
...
-        return result
+        group: str | None = None,
+        name: str | None = None,
+    ):
+        """Run a lifecycle operation on this driver's CVD, or on ``group``/``name`` if given."""
+        if op not in TARGET_STATE:
+            raise CuttlefishError(f"unknown operation {op!r}")
+        group = group or self._cvd_group or self.group
+        name = name or self._cvd_name or self.name
+        if not self.managed:
```
kirkbrauer: `self._cvd_group`/`_cvd_name` are read at 348-349 **outside** `_operation_lock`, but written inside it at 369-370 after a successful create.

Thread A resolves its target implicitly (the normal path for `stop`/`powerwash`/`restart`) and blocks at 352; thread B completes `create` and rebinds the identity; A then operates on the pre-create values.

The `except Exception` handler latches `state = "failed"` permanently, so `health.py` fails every later probe and liveness terminates an exporter whose guest is fine.

We could resolve this inside the lock. The unmanaged branch needs its own resolution since it returns first:

```python
if not self.managed:
    return self._backend.operate(
        op, group or self._cvd_group or self.group, name or self._cvd_name or self.name, data, timeout
    )
with self._operation_lock:
    group = group or self._cvd_group or self.group
    name = name or self._cvd_name or self.name
```

`test_managed_serializes_creation` will not catch this — no barrier forces overlap and it only tests create-vs-create.

---

kirkbrauer (commented):
One more from the correctness pass that I'd held back: a lock-scope race in the managed lifecycle path (`driver.py:348-352`). Inline detail attached.

It's narrow — it needs a genuine concurrent create — but the failure mode is disproportionate: the health state latches to `failed` permanently, so the liveness probe ends up terminating an exporter whose guest is actually healthy. Cheap to fix by moving two lines inside the existing lock.

---

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/health.py
bennyz: Fixed at the originating commit in #1072

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: runtimePortReserved now reserves the complete 15550–15599 range

controller/internal/exporterset/provisioners/cuttlefish/cuttlefish.go
bennyz: The README now requires explicit architecture/prepared-node selection and includes the selector, taint toleration, and device-plugin resource limit. configureDeviceResources removes each plugin-supplied device’s hostPath volume and mounts after validating the corresponding extended-resource limit

python/packages/jumpstarter-driver-cuttlefish/jumpstarter_driver_cuttlefish/driver.py
bennyz: Managed target identity is now resolved inside _operation_lock; unmanaged behavior remains outside the lock. A barrier-based regression test forces create and stop/restart/powerwash overlap and verifies the operation uses the newly created identity

PR #1085, opened 2026-09-10
fix(operator): grant exporter-set controllers exporters/finalizers

bennyz:
The exporter-set controller makes each Exporter the controller-owner of its config Secret and of its Pod via SetControllerReference, which sets blockOwnerDeletion on the ownerReference. The API server rejects such an ownerReference unless the writer may update the owner's finalizers, so with only the exporters rule the controller fails every reconcile with

  secrets "exporter-config-<exporter>" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on

and neither the Secret nor the Pod is ever created. The ExporterSet reports a degraded replica with no failed pods, which gives no hint at the cause.

It does not surface in the kind-based flows, which deploy the controller with broader permissions than the operator's namespaced Role.

The operator's own ClusterRole already grants jumpstarter.dev exporters/finalizers, so no additional privileges are needed to delegate it and the generated manifests are unchanged.

### DIFF
--- a/controller/deploy/operator/internal/controller/jumpstarter/exporterset.go
+++ b/controller/deploy/operator/internal/controller/jumpstarter/exporterset.go
@@ -513,6 +513,16 @@ func exporterSetPolicyRules() []rbacv1.PolicyRule {
 			Verbs:     []string{"get", "list", "watch", "create", "update", "patch", "delete"},
 		},
 		{
+			// The per-exporter config Secret and the exporter Pod are both owned by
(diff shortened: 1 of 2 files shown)


bennyz:
I've hit this testing cuttlefish on CRC, on kind it also happens, but only when I've enabled admission plugins

mangelajo:
Thanks! :)

mangelajo (approved):

Issue #1097, opened 2026-09-17
Expose exporter spec.enabled flag reliably in ListExporters protobuf response


mangelajo:
## Problem

Currently the `Exporter` message in `jumpstarter.client.v1` has an `enabled` field (`optional bool`), but its population is inconsistent. The `ExporterStatus` enum intentionally does not include a disabled state, because an exporter can be simultaneously online/available and disabled — meaning it won't accept new leases but is still running.

This creates a gap: CLI tools that want to show a quick status icon (e.g. 🚫 for disabled) need a reliable way to know whether `spec.enabled` is `false`, independent of the exporter's runtime status.

## Proposal

- Ensure the controller always populates the `enabled` field in `ListExporters` responses based on the exporter CR's `spec.enabled` value.
- Document that `enabled` and `status` are orthogonal: `enabled=false` means "no new leases will be assigned" while `status` reflects the exporter's runtime state (offline, available, running hooks, etc.).

## Motivation

The `jmp get exporters` CLI is adding a status icon column. Without a reliably-populated `enabled` field, we cannot show a 🚫 icon for disabled exporters. The `enabled` flag should not be a new `ExporterStatus` enum value because the two concepts are orthogonal — an exporter can be online and ready but simultaneously disabled.

## Context

The `enabled` boolean already exists in the protobuf `Exporter` message as an optional field. The issue is ensuring the controller consistently sets it from the CR's `spec.enabled` so clients can rely on it.

PR #1098, opened 2026-09-17
feat(cli): add status icon column to jmp get exporters

mangelajo:
## Summary

Adds a status icon as the leftmost column in the default table output of `jmp get exporters`, giving users an at-a-glance view of each exporter's runtime status.

### Icon mapping

| Status | Icon | Meaning |
|---|---|---|
| `AVAILABLE` | 🟢 | Online and ready for lease |
| `OFFLINE` | 🔴 | Exporter is offline |
| `BEFORE_LEASE_HOOK` | ⚙️ | Running before-lease hook |
| `AFTER_LEASE_HOOK` | ⚙️ | Running after-lease hook |
| `LEASE_READY` | ⏳ | Leased and serving a client |
| `BEFORE_LEASE_HOOK_FAILED` | ❗ | Before-lease hook failed |
| `AFTER_LEASE_HOOK_FAILED` | ❗ | After-lease hook failed |
| `UNSPECIFIED` / `None` | ❓ | Status unknown |

### Example output

```
   NAME          LABELS
🟢 my-exporter   env=prod,role=worker
🔴 old-exporter  env=staging
⚙️  hooking-exp   env=dev
⏳ busy-exp      env=test
```

### Details

- The icon column only appears in table output (default), not in `-o json`, `-o yaml`, or `-o name`
- Icons are derived purely from `ExporterStatus` — the `enabled` flag is not used for icon mapping (see #1097 for future work on reliably exposing `spec.enabled`)
- Added `Exporter.status_icon()` method with full test coverage for all status values
- Updated all existing column count/header assertions to account for the new column

### Related

- #1097 — Expose exporter `spec.enabled` flag reliably in ListExporters protobuf response

### Tests

- `pkg-test-jumpstarter`: 883 passed
- `pkg-test-jumpstarter-cli`: 267 passed


kirkbrauer:
@mangelajo I think this looks good, but are we respecting the `NO_COLOR` and `TERM=dumb` env variables to disable rich terminal output?

mangelajo:
This is how it would look:

image

Let me check the NO_COLOR and TERM=DUMB

btw, I think those UTF8 colors can be represented without color? but not sure if NO_COLOR requires "don't send me pesky UTF8", investigating.

mangelajo:
so good point, NO_COLOR and TERM=dumb ... I am adding a patch on top to change those icons into simple ascii when those variables are detected.

mangelajo:
image
the plaintext version :)

bkhizgiy (commented):
Looks good overall! Just a couple of UX nits.
Should we add a short explanation of the status icons to the help text? right now the column has no header, so especially with ASCII/NO_COLOR output, it’s not very obvious what +, -, *, ~, !, etc. mean.

also more of a suggestion, with `--with status` we show both the icon and the full status column, which feels a bit redundant once the user explicitly asks for the textual status, maybe in that case we could replace the icon with the status column instead?

kirkbrauer:
@mangelajo @bkhizgiy I wonder if a 🔒/`~` emoji might be a better description for a `LEASE_READY` exporter? Also, perhaps a ❌/`x` is better for `OFFLINE` status?

mangelajo:
Thanks for the feedback:

@bkhizgiy : good idea about the help text. I didn't want to add any column header to avoid making the output wider. But some help is deserved.

and about --status ... You're right, also I think it could make sense to put the icon after the name... like status.

@kirkbrauer :

about 🔒, it sounds like it could be locked or disabled by admin?, but at the time I think it's better than the clock ⏳ which could mean that we are performing some activity that needs time...

So even if not perfect, 🔒 sounds good, claude also suggests: 👤, 🎫, ...

The ❌/x sounds good for offline

For online/ready I am considering a change to ⚪ , because lots of green dots are a little bit overwhelming to the eye.

mangelajo:
image


hmm, for the online, I think that I prefer the green dot... let me try again

mangelajo:
image

mangelajo:
```
majopela@majopela-na:~/jumpstarter/jumpstarter/python$ NO_COLOR=1 uv run jmp get exporters --help
Usage: jmp get exporters [OPTIONS]

  Display one or many exporters

Options:
  --client-config PATH           Path to client config
  --client TEXT                  Alias of client config
  -l, --selector TEXT            Selector (label query) to filter on, supports
                                 '=', '==', and '!=' (e.g. -l
                                 key1=value1,key2=value2). Matching objects
                                 must satisfy all of the specified label
                                 constraints. Can be specified multiple times.
  -o, --output [json|yaml|name]  Output mode. Use "-o name" for shorter output
                                 (resource/name).
  --with TEXT                    Include fields: leases, online, status
                                 (comma-separated or repeated)
  --allow-disabled               Include disabled exporters in the listing
  --show-hidden-labels           Show labels hidden by controller config
  --page-size INTEGER RANGE      Number of results per page for pagination
                                 [x>=1]
  --help                         Show this message and exit.

  Status icons: +  available, x  offline, *  hook running, ~  leased, !  hook
  failed, ?  unknown
majopela@majopela-na:~/jumpstarter/jumpstarter/python$ uv run jmp get exporters --help
Usage: jmp get exporters [OPTIONS]

  Display one or many exporters

Options:
  --client-config PATH           Path to client config
  --client TEXT                  Alias of client config
  -l, --selector TEXT            Selector (label query) to filter on, supports
                                 '=', '==', and '!=' (e.g. -l
                                 key1=value1,key2=value2). Matching objects
                                 must satisfy all of the specified label
                                 constraints. Can be specified multiple times.
  -o, --output [json|yaml|name]  Output mode. Use "-o name" for shorter output
                                 (resource/name).
  --with TEXT                    Include fields: leases, online, status
                                 (comma-separated or repeated)
  --allow-disabled               Include disabled exporters in the listing
  --show-hidden-labels           Show labels hidden by controller config
  --page-size INTEGER RANGE      Number of results per page for pagination
                                 [x>=1]
  --help                         Show this message and exit.

  Status icons: 🟢  available, ❌  offline, ⚙️  hook running, 🔒  leased, ❗  hook
  failed, ❓  unknown
  ```

kirkbrauer:
@mangelajo The only reason I suggested the lock was to make it cohesive with the VS Code plugin I'm experimenting with, but we could change either of them to match the other if desired.

mangelajo:
sounds good, I have a request from @raballew , I could do as a follow up or here, adding something to completely avoid UTF8 icons, perhaps we should call it NO_ICONS? env var?

mangelajo:
My gentlemen agreement with @raballew is that I will add another PR that obeys a NO_ICONS env var.

raballew (approved):

Issue #926, opened 2026-07-28
Preserve signal information in jumpstarter-exec Exit protocol


mangelajo:
## Summary

When a child process is killed by a signal, `jumpstarter-exec` currently reports exit code `1`, which is indistinguishable from a normal failure. Callers cannot tell \"SIGTERM during shutdown\" from \"command failed\".

Deferred from PR #910 because the QEMU driver only branches on zero vs non-zero today.

## Background

From [PR #910](https://github.com/jumpstarter-dev/jumpstarter/pull/910#discussion_r3659197338):

> We can extend the protocol with `signal` field in `Exit` and map to `128+signum` in a follow-up if needed.

Original review: [signal-killed children reported as exit 1](https://github.com/jumpstarter-dev/jumpstarter/pull/910#discussion_r3655152963).

## Proposed approach

1. Extend `ServerMessage::Exit` with something like `signal: Option<i32>` (populated via `ExitStatusExt::signal()`)
2. Map client-side reported status to `128 + signum` when signal-terminated (shell convention), or expose signal explicitly to Python callers
3. Update protocol tests and any QEMU driver consumers that should distinguish shutdown vs failure

## Acceptance criteria

- [ ] Protocol carries signal information for signal-terminated children
- [ ] Client reports a distinguishable status (e.g. `128+signum`) or exposes `signal`
- [ ] Tests cover signal-terminated vs normal non-zero exit
- [ ] Document the behavior for driver authors

## Tracking

- Jira: https://redhat.atlassian.net/browse/PITCREW-503
- PR: https://github.com/jumpstarter-dev/jumpstarter/pull/910
- Code: `rust/jumpstarter-exec/src/{protocol,server,client}.rs`

PR #1100, opened 2026-09-17
fix(cli): propagate the command exit code from jmp shell

### LINKED ISSUE #1099: bug: jmp shell exits 0 when the command fails in local exporter mode
### Description

`jmp shell -- COMMAND` always exits `0` in local exporter mode, whatever COMMAND did. CI cannot tell a passing run from a failing one.

### Steps to Reproduce

```
jmp shell --exporter-config ./exporter.yaml -- sh -c 'exit 42'
jmp shell --exporter-config ./exporter.yaml -- this-does-not-exist
jmp shell --exporter-config ./exporter.yaml -- sh -c 'kill -9 $$'
```

### Error

All three exit `0`:

| command | expected | actual |
| --- | --- | --- |
| `sh -c 'exit 42'` | 42 | 0 |
| `this-does-not-exist` | 127 | 0 |
| `sh -c 'kill -9 $$'` | 137 | 0 |

The second prints `Error: command not found: this-does-not-exist` and still exits `0`.

### Root Cause

`shell()` has three terminal branches. `--tls-grpc` and client/lease both do `exit_code = anyio.run(...)` then `sys.exit(exit_code)`. The local exporter branch calls `launch_shell()` and discards the result, so the callback
returns `None` and click exits `0`.

Separately, `Popen.wait()` reports a signal death as `-N`, which `sys.exit()` turns into `256-N`, so SIGKILL gives `247` instead of `137`.

### Fix

Three commits:

1. Log the child's exit status in `_run_process`.
2. Normalise signal deaths to `128+signum`, and name the signal in the log. `137` on its own cannot be told from a command exiting `137`.
3. Capture `launch_shell()`'s return value and `sys.exit()` it, matching the other two branches.

### Testing

New coverage for signal normalisation and for exit-code propagation through the click command.

### Related

#926, which proposes the same `128 + signum` mapping for the Rust `jumpstarter-exec` Exit protocol. This does not close it.

lh-sag:
## Summary

`jmp shell -- COMMAND` exits `0` in local exporter mode whatever COMMAND did, so CI cannot tell a passing run from a failing one. Reproducer, expected codes and root cause in #1099.

## Fix

Three commits:

1. Log the child's exit status in `_run_process`, the one point all five `launch_shell` return paths funnel through. No behaviour change.
2. Normalise signal deaths to `128+signum`, and name the signal in the log. `137` on its own cannot be told from a command exiting `137`.
3. Capture `launch_shell()`'s return value and `sys.exit()` it, matching the `--tls-grpc` and client/lease branches.

| case | before | after |
| --- | --- | --- |
| `-- sh -c 'exit 42'` | 0 | 42 |
| `-- this-does-not-exist` | 0 | 127 |
| not executable | 0 | 126 |
| `-- sh -c 'kill -9 $$'` | 0 | 137 |
| `-- sh -c 'kill -15 $$'` | 0 | 143 |
| interactive shell, `exit 7` | 0 | 7 |

## Test plan

- [x] New coverage for signal normalisation and for exit-code propagation through the click command, which was untested
- [x] `make pkg-test-jumpstarter`, `pkg-test-jumpstarter-cli` and `pkg-test-jumpstarter-cli-common` green at each of the three commits
- [x] `make lint` and `make pkg-ty-jumpstarter-cli`
- [x] Every row of the table verified against a real local exporter

## Notes

This composes with #1084. That PR records the shell exit code from inside the worker thread, and the value it records comes from `_run_process`, so a signal-killed shell on the lease path now reports `137` instead of `247`.

Refs #926, which proposes the same `128 + signum` mapping for the Rust `jumpstarter-exec` Exit protocol. This does not close it.

Closes #1099


mangelajo (approved):
Thank you so much for addressing this bug and providing excellent test coverage.

PR #1102, opened 2026-09-18
fix(mcp): isolate a connection's background failure from its siblings

### LINKED ISSUE #1101: bug: one connection's background failure kills the whole jmp mcp serve session
## Summary

`ConnectionManager.running()` (`jumpstarter_mcp/connections.py:352`) opens exactly one `anyio` task group for the whole `jmp mcp serve` session, and every `connect()` call schedules that connection's entire background lifecycle into that same group (`connect()` -> `self._task_group.start(_run_connection)`).

With `anyio`/`trio`-style structured concurrency, an unhandled exception in any one of those sibling tasks cancels every other task in the group and then propagates out of `running()`'s `async with` block, which `run_server()` wraps around the whole stdio session.

So a transport hiccup on one connection (closed socket, gRPC error, anything past the point the connection reported itself ready) tears down every other live connection and ends the whole MCP process, not just the one that failed.

## Repro

Minimal repro against the real `ConnectionManager`, stubbing only `client_from_path`/lease so no live exporter is needed:

```python
import anyio
from contextlib import asynccontextmanager
from jumpstarter_mcp.connections import ConnectionManager

@asynccontextmanager
async def fake_client_from_path(path, portal, stack, allow, unsafe):
    yield object()
    if path.endswith("a.sock"):
        raise RuntimeError("simulated teardown error on connection a")

class FakeLease:
    def __init__(self, name):
        self.name, self.exporter_name = name, name
        self.allow, self.unsafe = [], True
        self.lease_transferred = self.lease_ended = False
        self.lease_ending_callback = None
    @asynccontextmanager
    async def serve_unix_async(self):
        yield f"/tmp/{self.name}.sock"
    @asynccontextmanager
    async def monitor_async(self):
        yield

class FakeConfig:
    def __init__(self, lease):
        self.lease = lease
    @asynccontextmanager
    async def lease_async(self, **kw):
        yield self.lease

async def main():
    import jumpstarter_mcp.connections as m
    m.client_from_path = fake_client_from_path
    manager = ConnectionManager()
    async with manager.running():
        a = await manager.connect(FakeConfig(FakeLease("a")), lease_name="a")
        b = await manager.connect(FakeConfig(FakeLease("b")), lease_name="b")
        await manager.disconnect(a.id)
        await anyio.sleep(0.5)
        print("b still connected:", b.id in manager.connections)

anyio.run(main)
```

On current `main` (`3a9dafb5`) this raises an unhandled `ExceptionGroup` out of `async with manager.running():` instead of printing anything, and connection `b` is cancelled along with `a`.

## Environment

- `jumpstarter-mcp` at `main` (`3a9dafb5`)
- `anyio` 4.15.1

AmirF194:
`connect()` schedules each connection's whole lifecycle into `self._task_group`, the single `anyio` task group `running()` opens for the entire session. Once an unhandled exception reached one connection's background task, the task group cancelled every sibling task and the exception propagated out of `running()`, ending the whole `jmp mcp serve` process (#1101).

The fix tracks whether `task_status.started()` has already fired for a connection. A failure before that point still raises, exactly as before, so `connect()` keeps reporting setup errors to its caller. A failure after that point is now caught, logged, and reported over MCP for that connection only; the task returns instead of raising into the shared group, so the other connections are untouched.

Added `connections_test.py` since the package had no direct test of `connect()`/`running()` yet: it drives the real `ConnectionManager` with a stubbed lease/client (no live exporter needed) and confirms a second connection survives when the first one's teardown raises. Without the fix it errors out of `async with manager.running():` with an unhandled `ExceptionGroup` instead.

Ran `uv run --isolated --directory packages/jumpstarter-mcp pytest`: 37 passed, up from 36. `ruff check` and `ty check` are clean on both files.

Fixes #1101


python/packages/jumpstarter-mcp/jumpstarter_mcp/connections.py
```diff
@@ -163,11 +176,17 @@ async def _run_connection(task_status=anyio.TASK_STATUS_IGNORED):
...
-            except BaseException:
+            except BaseException as exc:
                 _check_lease_error(lease_ref)
+                if tracker.called:
+                    # Isolate a post-startup failure to this connection; do not cancel siblings.
+                    unwrapped = _unwrap_exception(exc)
+                    logger.exception("Connection %s failed", connection_id)
+                    await self._send_log("error", f"Connection {connection_id} failed: {unwrapped}")
+                    return
                 raise
```
mangelajo: Agreeing with coderabbitai here: 

`_check_lease_error(lease_ref)` runs unconditionally before the `tracker.called` check. `Lease._notify_lease_ending()` sets `lease_ended = True` while the connection is fully active (post-startup), so if a teardown error hits the `except` block at that point, `_check_lease_error` raises a `ConnectionError` that escapes the task — bypassing the isolation this PR adds and cancelling siblings anyway.

Swap the order so lease-state checks only run for pre-startup failures:

```python
except BaseException as exc:
    if not tracker.called:
        _check_lease_error(lease_ref)
        raise

    # Isolate a post-startup failure to this connection; do not cancel siblings.
    unwrapped = _unwrap_exception(exc)
    logger.exception("Connection %s failed", connection_id)
    await self._send_log("error", f"Connection {connection_id} failed: {unwrapped}")
    return
```

---

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections.py
```diff
@@ -163,11 +176,17 @@ async def _run_connection(task_status=anyio.TASK_STATUS_IGNORED):
                     ) as lease:
                         lease_ref = lease
                         conn = await self._setup_connection(
-                            config, lease, portal, connection_id, event, task_status,
+                            config, lease, portal, connection_id, event, tracker,
                         )
                         logger.info("Connection %s tearing down (%s)", connection_id, conn.exporter_name)
-            except BaseException:
+            except BaseException as exc:
```
mangelajo, line 182: `except BaseException` also catches `Cancelled` (anyio's cancellation exception). After startup, if the task group's cancel scope cancels this connection during normal shutdown (e.g. `running()` cleanup), the cancellation will be caught, logged as a "failure", and swallowed via `return` instead of re-raised. This can prevent clean shutdown of the task group.

Consider re-raising cancellation unconditionally:

```python
except BaseException as exc:
    if isinstance(exc, anyio.get_cancelled_exc_class()):
        raise
    ...
```

---

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+import anyio
+import pytest
+
+from jumpstarter_mcp.connections import ConnectionManager
+
+# Sockets whose fake client should raise on teardown, simulating a transport
+# error (a closed gRPC channel, a vanished unix socket, ...) surfacing after
+# the connection already reported itself as ready.
+_RAISE_ON_TEARDOWN: set[str] = set()
```
mangelajo: Nit: `_RAISE_ON_TEARDOWN` is module-level mutable state shared across tests. It's fine with a single test, but if more tests are added (or tests run in parallel), there's a risk of cross-test interference. Consider turning this into a fixture or passing it through a closure to keep test isolation clean.

---

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections_test.py
```diff
@@ -0,0 +1,81 @@
...
+        conn_b = await manager.connect(config_b, lease_name="lease-b")  # ty: ignore[invalid-argument-type]
+
+        _RAISE_ON_TEARDOWN.add(conn_a.socket_path)
+        await manager.disconnect(conn_a.id)
+
+        # Give connection A's background task room to run its teardown (where
+        # the fake client raises) and, on unfixed code, for that exception to
+        # cancel the rest of the shared task group.
+        for _ in range(50):
+            await anyio.sleep(0.01)
```
mangelajo: Nit: the 500 ms sleep loop is a time-based race-condition workaround — it could be flaky on slow CI runners, or wasteful on fast ones. An event- or signal-based approach would be more deterministic, but given the difficulty of observing the internal task lifecycle externally this is acceptable for now.

---

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections_test.py
```diff
@@ -0,0 +1,81 @@
...
+
+        assert conn_b.id in manager.connections, "connection B was cancelled by connection A's unrelated failure"
```
mangelajo, line 88: Nice test for the post-startup path. It would be valuable to also add a negative test confirming that a **pre-startup** failure (e.g. `client_from_path` raises before `started()`) still propagates correctly as a `ConnectionError` to the `connect()` caller. This would guard against accidentally swallowing setup errors in future refactors.

---

mangelajo (changes requested):
Just some comments Amir but thank you so much for identifying and fixing this issue!

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections.py
AmirF194: Moved `_check_lease_error(lease_ref)` inside the `not tracker.called` branch, so a lease_ended flip that races a post-startup teardown can no longer raise ahead of the isolation logic. Added `test_lease_ended_race_during_teardown_does_not_kill_other_connections`: it fails against the old unconditional ordering and passes now.

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections.py
AmirF194: Added an `isinstance(exc, anyio.get_cancelled_exc_class())` check at the top of the except block that always re-raises, ahead of both the lease-error check and the isolation branch. That ordering also stops a pre-startup cancellation from getting replaced by a `_check_lease_error` ConnectionError. `test_cancellation_after_startup_is_not_reported_as_a_failure` cancels the manager's own task group after startup and asserts nothing gets logged as a failure. Against the old code it logs "Connection ... failed: Cancelled by cancel scope ..."; it passes now.

python/packages/jumpstarter-mcp/jumpstarter_mcp/connections_test.py
AmirF194: Added `test_pre_startup_failure_propagates_to_connect_caller`: `client_from_path` raises before `started()` is called, and `connect()` still raises `ConnectionError` to the caller.

mangelajo (approved):

mangelajo (approved):

AmirF194:
Thanks for the careful review, the lease_ended race and the swallowed cancellation were both real gaps in my first pass.

PR #1108, opened 2026-09-21
feat(nanokvm-usb): embedded RFB/VNC with HID and optional TCP bind

mparram:
This adds an embedded RFB 3.8 server to the NanoKVM-USB driver so HDMI capture and HID keyboard/mouse can be used from noVNC or a native VNC client. Video and VNC share a single FramePump so the UVC device is opened once; the VNC child follows the QEMU pattern (Unix socket + NovncAdapter) and is reachable over the Jumpstarter tunnel with j nanokvm-usb vnc session or forward-tcp.


python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,596 @@
...
+
+
+def _des_ecb_encrypt(block: bytes, key: bytes) -> bytes:
+    ip = _permute(int.from_bytes(block, "big"), _IP, 64)
+    left, right = ip >> 32, ip & 0xFFFFFFFF
+    for subkey in _des_subkeys(key):
+        left, right = right, left ^ _des_f(right, subkey)
+    preout = ((right & 0xFFFFFFFF) << 32) | (left & 0xFFFFFFFF)
+    return _permute(preout, _FP, 64).to_bytes(8, "big")
+
```
mangelajo: The inline DES implementation (~120 lines of tables + functions) exists only for VncAuth challenge-response (16 bytes). Consider adding a one-line comment at the top of this block explaining why it's inlined ("Pure-Python DES for VncAuth — avoids a C dependency for a 16-byte encrypt") so future readers don't try to refactor it into a library import.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,596 @@
...
+        name = b"NanoKVM-USB"
+        conn.sendall(
+            struct.pack("!HH", self._width, self._height)
+            + _pack_pixel_format(_default_pixel_format())
+            + struct.pack("!I", len(name))
+            + name
+        )
+
+    def _serve(self, conn: socket.socket) -> None:
+        pf = _default_pixel_format()
```
mangelajo: `_send_frame` sends the entire framebuffer as RAW encoding on every update request. For a 1920×1080×32bpp frame that's ~8 MB per frame. This will be very bandwidth-heavy over Jumpstarter tunnels.

Consider:
1. Only sending dirty rectangles (diff against previous frame)
2. Supporting Tight or ZRLE encoding (compressed)
3. At minimum, adding a `zlib.compress()` around the raw pixels would help significantly

This is fine for an initial implementation, but worth a `# ponytail:` or `# TODO:` note about the bandwidth ceiling.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,596 @@
...
+        if not client_ver.startswith(b"RFB "):
+            raise ConnectionError(f"invalid RFB version: {client_ver!r}")
+        if self._password:
+            conn.sendall(bytes([1, SEC_VNCAUTH]))
+            chosen = _recvexact(conn, 1)[0]
+            if chosen != SEC_VNCAUTH:
+                conn.sendall(struct.pack("!I", 1))
+                raise ConnectionError("client did not select VncAuth")
+            challenge = os.urandom(16)
+            conn.sendall(challenge)
```
mangelajo: After setting `conn.settimeout(0)` for the main serve loop, `select.select` with a 50ms timeout is used to poll for client messages. This is a busy-wait pattern — if there's no update to send and no client message, the loop spins at ~20 Hz doing nothing.

This is acceptable for a dedicated VNC session thread, but worth noting that each connected VNC client burns a thread spinning at this rate even when idle.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/driver.py
```diff
@@ -275,16 +318,56 @@ def __post_init__(self):
...
             ),
         }
-        for child in self.children.values():
-            child._owns_device = False
+        if self.vnc_enabled:
+            self._vnc_dir = tempfile.mkdtemp(prefix="nanokvm-usb-vnc-")
+            vnc_path = str(Path(self._vnc_dir) / "vnc.sock")
+            self._vnc_server = RfbServer(
+                vnc_path,
+                pump=lambda: self._shared_device._pump,
```
mangelajo: Accessing `self._shared_device._pump` (a private attribute) via a lambda crosses encapsulation boundaries. Consider adding a public property or method to `NanoKVMUSBDevice` like `@property def pump(self) -> FramePump | None` instead of reaching into the private `_pump`.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/driver.py
```diff
@@ -108,19 +117,31 @@ async def stream(self):
         send_stream, receive_stream = anyio.create_memory_object_stream(max_buffer_size=buffer_size)
```
mangelajo: The `_next_frame` closure also accesses `device._pump` directly via `getattr(device, "_pump", None)`. This should use the same public accessor suggested for `RfbServer` (e.g., `device.pump`).

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/device.py
```diff
@@ -92,15 +95,22 @@ def connect(self) -> InfoPacket | None:
                     )
 
                 self._connected = True
+                if self._video_device is not None:
+                    if self._pump is None:
```
mangelajo: The `FramePump` is started inside `connect()` under the `_connect_lock`, but `close()` also calls `self._pump.stop()` under the same lock. Since `FramePump.stop()` calls `self._thread.join(timeout=2)`, this is fine as long as the pump's capture callback doesn't try to acquire `_connect_lock` — which it shouldn't since `capture_frame_jpeg` doesn't use it. Just flagging the potential for future lock-ordering issues if the capture path changes.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,596 @@
...
+    b = b * pf["blue_max"] // 255
+    pix = (r << pf["red_shift"]) | (g << pf["green_shift"]) | (b << pf["blue_shift"])
+    dtype = ">u4" if pf["big_endian"] else "<u4"
+    if bpp == 16:
+        dtype = ">u2" if pf["big_endian"] else "<u2"
+    return pix.astype(dtype).tobytes()
+
+
+class RfbServer:
+    """Listen on a Unix socket (and optional TCP) and serve RFB sessions."""
```
mangelajo: The `RfbServer` spawns a new `threading.Thread` per client connection (`_session`). There's no limit on concurrent sessions — a misbehaving or reconnecting VNC client could spawn many threads. Consider capping concurrent connections (e.g., `if len(self._clients) >= MAX_CLIENTS: conn.close(); continue`).

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/client.py
```diff
@@ -161,6 +171,51 @@ def scroll(dx, dy):
...
+        with NovncAdapter(client=self, method="connect", encrypt=use_encrypt) as url:
+            yield url
+
+    def cli(self):
+        base = super().cli()
+
+        @base.command()
+        @click.option("--browser/--no-browser", default=True, help="Open the session in a web browser.")
+        @click.option(
+            "--encrypt",
```
mangelajo: The `session` CLI command blocks with `anyio.sleep_forever` until Ctrl+C. If the VNC session encounters an error on the server side (e.g., device disconnect), the client will hang forever with no feedback. Consider adding a health check or timeout mechanism.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,596 @@
...
+        "blue_max",
+        "red_shift",
+        "green_shift",
+        "blue_shift",
+    )
+    return dict(zip(keys, values, strict=True))
+
+
+def pack_rgb_frame(image: Image.Image, pf: dict[str, int]) -> bytes:
+    """Pack an RGB image into RFB raw pixels for ``pf``."""
```
mangelajo: The `pack_rgb_frame` function decodes the JPEG, converts to RGB numpy arrays, and repackages as raw RFB pixels on every frame. For 1080p at 30 fps this is CPU-intensive. Since the default pixel format is BGRA 32-bit (which PIL can produce directly), you could skip the numpy channel math for the common case:

```python
if pf == _default_pixel_format():
    return image.tobytes("raw", "BGRX")
```

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/driver.py
```diff
@@ -275,16 +318,56 @@ def __post_init__(self):
                 screen_height=self.screen_height,
             ),
         }
-        for child in self.children.values():
-            child._owns_device = False
+        if self.vnc_enabled:
```
mangelajo: The `vnc_enabled` default is `True` meaning every NanoKVM-USB exporter will start a VNC server even if not needed. For a feature that adds threads, a Unix socket, and imports numpy/PIL for frame conversion, `False` might be a safer default — users who want VNC can opt in. The QEMU driver doesn't start VNC by default either (it's always available because QEMU runs it, but it doesn't spin up extra infrastructure).

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_keymap.py
```diff
@@ -0,0 +1,179 @@
...
+LAYOUTS: dict[str, dict[int, CharCombo]] = {
+    "us": _us_layout(),
+    "es": _es_layout(),
+}
+
+
+def char_combo(keysym: int, layout: str = "us") -> CharCombo | None:
+    """Return (HID key, modifiers) to type the printable keysym on ``layout``."""
+    table = LAYOUTS[normalize_layout(layout)]
+    return table.get(normalize_keysym(keysym))
```
mangelajo: Only `us` and `es` layouts are supported. The `normalize_layout` function raises `ValueError` for anything else, which will crash the exporter at startup if misconfigured. This is fine behavior, but it would be good to document which layouts are planned in the README (or add a note that contributions for additional layouts are welcome).

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/driver.py
mangelajo: Yes let's disable by default.

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/driver.py
```diff
@@ -275,16 +319,58 @@ def __post_init__(self):
...
+            )
+        for name in ("video", "hid"):
+            self.children[name]._owns_device = False
 
         super().__post_init__()
+        if self._vnc_server is not None:
+            if self.vnc_tcp_port is not None and not is_loopback_bind(self.vnc_tcp_bind):
+                self.logger.warning(
+                    "RFB TCP bind %s:%s is reachable on the network; "
+                    "vnc_password (VncAuth) is not transport encryption",
```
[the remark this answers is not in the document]
mangelajo: it's ok for now, this is just test devices. Not a production server.

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/client.py
mangelajo: This is acceptable for now — the VNC session lifecycle matches the QEMU pattern. We can add health monitoring in a follow-up if needed.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,511 @@
...
+        got = pump.wait_jpeg(timeout=0.05, after_generation=last_gen if last_gen >= 0 else None)
+        if got is None:
+            return None
+        jpeg, gen = got
+        try:
+            image = Image.open(BytesIO(jpeg))
+            image.load()
+        except Exception:
+            logger.debug("failed to decode JPEG for RFB", exc_info=True)
+            return gen, last_pixels
```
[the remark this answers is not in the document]
mangelajo: Agreed, this should be fixed. On JPEG decode failure (line 496), returning `(gen, last_pixels)` tells `_serve` the update was sent, so the client hangs until it sends another `FramebufferUpdateRequest`. Should return `None` so the loop retries on the next valid frame.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/vnc_server.py
```diff
@@ -0,0 +1,511 @@
...
+        if image.size != (self._width, self._height):
+            image = image.resize((self._width, self._height))
+        pixels = pack_rgb_frame(image, pf)
+        conn.settimeout(_CLIENT_IO_TIMEOUT)
+        try:
+            if last_pixels is not None and pixels == last_pixels:
+                conn.sendall(struct.pack("!BxH", 0, 0))
```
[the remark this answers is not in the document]
mangelajo: Agreed. Per RFC 6143 §7.5.3, non-incremental requests mean "I lost my framebuffer, send everything." The incremental flag (first byte of the 9-byte payload at line 426) is currently discarded. It should be parsed and passed through so `_send_frame` skips the `pixels == last_pixels` dedup for non-incremental requests.

---

python/packages/jumpstarter-driver-nanokvm-usb/jumpstarter_driver_nanokvm_usb/keyboard.py
```diff
@@ -332,14 +351,34 @@ def reset(self) -> list[int]:
...
+        modifier = self._modifier & ~shift_altgr
+        for mod in combo_mods:
+            bit = MODIFIER_BITS.get(mod)
+            if bit is None:
+                raise ValueError(f"Unknown modifier: {mod!r}")
+            modifier |= bit
+        return self._build_report(modifier)
+
+    def printable_up(self, key: str) -> list[int]:
+        self._pressed.pop(key, None)
```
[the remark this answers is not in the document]
mangelajo: Valid and simple to fix (~10 lines). Track `combo_mods` per pressed printable key in a parallel dict `_printable_mods`. In `printable_up`, after removing the released key, union the remaining keys' combo_mods into the modifier byte. This preserves Shift/AltGr for still-held printable keys.

---

PR #1114, opened 2026-09-23
WIP: controller: add qemu-ssh provisioner library (SSH, quadlets, host pool)

bkhizgiy:
Adds the foundational packages for the qemu-ssh.jumpstarter.dev off-cluster provisioner: SSH/SFTP host management with context-aware commands and file reconciliation, quadlet .container file generation for runtime + exporter containers, and slot-based host pool selection with SSH config resolution. (JEP-0014 Phase 3)


controller/internal/exporterset/provisioners/qemu-ssh/host_pool.go
```diff
@@ -0,0 +1,204 @@
...
+	AnnotationHost = "qemu-ssh.jumpstarter.dev/host"
+)
+
+// HostConfig describes a single remote lab host parsed from
+// VirtualTargetClass/ExporterSet merged parameters.
+type HostConfig struct {
+	// Name is the FQDN or IP of the remote host.
+	Name string `json:"name"`
+
+	// Arch is the host CPU architecture (e.g. "aarch64", "x86_64").
```
mangelajo: do we really need/use arch / slots for anything?, if not I'd remove them for now.

controller/internal/exporterset/provisioners/qemu-ssh/host_pool.go
```diff
@@ -0,0 +1,204 @@
...
+// parameters.
+type SSHConfig struct {
+	User string `json:"user,omitempty"`
+	Port int    `json:"port,omitempty"`
+}
+
+// ParseHosts extracts the host list from merged parameters.
+// Expected structure: parameters.hosts: [{name, arch, slots, ...}]
+func ParseHosts(mergedParameters map[string]any) ([]HostConfig, error) {
+	hostsRaw, ok := mergedParameters["hosts"]
```
mangelajo:
```suggestion
	hostRaw, ok := mergedParameters["host"]
```

to simplify things every exporterset with this provider only should manage a single host. If the admin wants more hosts, would create additional exportersets under the same class.

This will make the controller simpler (you can use the number of replicas/etc just control one host, then you don't need to know about the available slots in it)

mangelajo (changes requested):
just one comment to simplify things.

PR #1115, opened 2026-09-23
WIP: controller: wire qemu-ssh provisioner into reconciler

bkhizgiy:
Wires the qemu-ssh provisioner into the exporter-set controller. Introduces the Deployer interface for off-cluster provisioners, implements Deploy/Cleanup/IsDeployed via SSH + Podman quadlets, adds driver enrichment for off-cluster paths, and fixes ExitAndReplace recycling for off-cluster exporters that have no Pods. (JEP-0014 Phase 3)


bkhizgiy:
depends on https://github.com/jumpstarter-dev/jumpstarter/pull/1114

controller/internal/exporterset/provisioners/qemu-ssh/enrich.go
```diff
@@ -0,0 +1,182 @@
...
+			"OVMF_VARS.fd": "/usr/share/AAVMF/AAVMF_VARS.fd",
+		}
+	default:
+		return map[string]string{
+			"OVMF_CODE.fd": "/usr/share/edk2/ovmf/OVMF_CODE.fd",
+			"OVMF_VARS.fd": "/usr/share/edk2/ovmf/OVMF_VARS.fd",
+		}
+	}
+}
+
```
mangelajo: there is probably some degree of duplication with the qemu provisioner in this area, we should probably create some common code between both.

controller/internal/exporterset/provisioners/qemu-ssh/host_pool.go
```diff
@@ -0,0 +1,204 @@
...
+	// Name is the FQDN or IP of the remote host.
+	Name string `json:"name"`
+
+	// Arch is the host CPU architecture (e.g. "aarch64", "x86_64").
+	// Informational; not used for scheduling in v1.
+	Arch string `json:"arch,omitempty"`
+
+	// Slots is the maximum number of concurrent exporter instances
+	// this host can run.
+	Slots int `json:"slots"`
```
mangelajo:
```suggestion
```

same as in the other review, based on the JEP, one exporterset should handle only one host, and one architecture to keep things simpler. (now I am not 100% if this was captured on the JEP, but it's something I remember discussing with @kirkbrauer )

mangelajo (changes requested):

controller/internal/exporterset/provisioners/qemu-ssh/enrich.go
bkhizgiy: yep make sense, will extract all the common code.

controller/internal/exporterset/provisioners/qemu-ssh/host_pool.go
bkhizgiy: ack sounds good, thanks for clarifying, will change to match.

